PR #259, opened 2026-01-30
Voice pipeline: fix audio cutoff, per-persona voice, and reactive mute

joelteply:
## Summary

This PR addresses three critical voice pipeline issues discovered during live testing:

- **Audio cutoff**: AI ring buffer in `mixer.rs` was 10 seconds — conversational TTS responses (20-60s) silently dropped excess audio. Increased to 60 seconds with `Vec<i16>` allocation (avoids stack overflow) and upgraded overflow logging from `debug!` to `warn!`.

- **All AIs same voice**: JS-side `computeVoiceFromUserId()` hashed UUIDs to numeric strings (e.g. "42"), but Kokoro's `normalize_voice()` only recognized named voices — everything else defaulted to "af". Added `resolve_voice()` as a default trait method on `TextToSpeech` with FNV-1a deterministic hash. Each adapter maps any input (named voice, numeric seed, UUID) to its own voice set. Removed TS-side hash entirely — Rust adapter owns voice resolution.

- **Mute not respected**: `micEnabled: false` saved correctly but audio kept flowing. Root cause: `@reactive()` property changes didn't automatically propagate to the audio client — relied on manual `applyMicState()` calls scattered across 5 code paths. Added LitElement `updated()` lifecycle hook for reactive state sync. Also separated visibility/deactivation saved states that shared one field (could stomp each other) and added idempotent `startMicrophone()` guard.

## Changes

| File | Change |
|------|--------|
| `workers/continuum-core/src/voice/mixer.rs` | Ring buffer 10s→60s, Box→Vec, debug→warn |
| `workers/continuum-core/src/voice/tts/mod.rs` | `resolve_voice()` trait method + FNV-1a hash + 6 tests |
| `system/voice/server/AIAudioBridge.ts` | Pass raw userId to Rust, remove `computeVoiceFromUserId()` |
| `widgets/live/LiveWidget.ts` | Reactive `updated()` hook, split saved states, reentrancy guard |
| `widgets/live/AudioStreamClient.ts` | Idempotent `startMicrophone()` early return |

## Test plan

- [x] `cargo test --lib` — 194 passed, 0 failed
- [x] `npx tsc --noEmit` — clean, 0 errors
- [ ] Manual: join voice call, verify AI speech plays fully (no cutoff)
- [ ] Manual: verify different AIs have different voices
- [ ] Manual: toggle mute, verify mic stops/starts correctly
- [ ] Manual: navigate away and back, verify mute state preserved

### DIFF
--- a/src/debug/jtag/system/voice/server/AIAudioBridge.ts
+++ b/src/debug/jtag/system/voice/server/AIAudioBridge.ts
@@ -217,8 +217,12 @@ export class AIAudioBridge {
 
   /**
    * Inject TTS audio into the call (AI speaking)
-   * @param voice - Speaker ID for multi-speaker TTS models (0-246 for LibriTTS).
-   *                If not provided, computed from userId for consistent per-AI voices.
+   * @param voice - Voice identifier passed to Rust TTS adapter. Can be:
+   *                - Named voice ("af", "am_adam") → used directly
+   *                - Numeric seed ("42") → modulo into adapter's voice list
+   *                - Any string (uniqueId, UUID, display name) → hashed to pick voice
+   *                If not provided, uses userId so each AI gets a consistent unique voice.
+   *                The Rust adapter's resolve_voice() handles all mapping.
    */
   async speak(callId: string, userId: UUID, text: string, voice?: string): Promise<void> {
     console.log(`🎙️🔊 VOICE-DEBUG: AIAudioBridge.speak CALLED - userId=${userId?.slice(0, 8)}, text="${text.slice(0, 50)}..."`);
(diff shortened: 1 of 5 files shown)


joelteply:
## Copilot Review Context

Copilot's summary focuses on the type-safe command migration and SQLite JSON fix from earlier commits. The **actual focus of this PR** is the voice pipeline, which Copilot didn't surface. Here's what matters:

### This PR's voice pipeline fixes (latest commit)

**1. Audio cutoff** (`mixer.rs`)
AI ring buffer was 10 seconds. Conversational TTS responses (20-60s) silently dropped excess samples at `debug!` level — impossible to notice unless you read Rust trace logs. Now 60 seconds with `warn!` level overflow alerts.

**2. Per-persona voice** (`tts/mod.rs`, `AIAudioBridge.ts`)
Every AI sounded the same. JS-side `computeVoiceFromUserId()` produced numeric strings ("42") which Kokoro's `normalize_voice()` didn't recognize as named voices — all defaulted to "af". Fix: `resolve_voice()` default trait method with FNV-1a deterministic hash maps any input (UUID, name, numeric seed) to adapter-specific voices. Voice resolution is now owned by the Rust adapter, not TypeScript.

**3. Mute state propagation** (`LiveWidget.ts`, `AudioStreamClient.ts`)
`micEnabled: false` saved correctly in UserStateEntity but audio kept flowing. The `@reactive()` property only triggered re-renders — it had no automatic side-effect for the audio client. Five separate code paths manually called `applyMicState()` and at least one could be missed. Fix: LitElement `updated()` lifecycle hook reactively syncs `micEnabled` → audio client, making it impossible to change the property without the audio following.

### Why Copilot missed this

The 488-file diff is dominated by the type-safe command migration (older commits). Copilot reviewed the bulk files and correctly identified no issues there. The voice pipeline changes (5 files, latest commit) are architecturally significant but small in diff size — easy to miss in a large PR.

joelteply:
Hooks are broken, but of no concern right now

Issue #856, opened 2026-04-11
Grid event streaming — persistent event channels between nodes


joelteply:
**airc-queue card**

Coordinates work via the AIRC queue substrate (airc#562). Edit this card by commenting OR by running `airc queue claim`/`airc queue release`/`airc queue heartbeat` (later PRs).

```json
{
  "kind": "airc-queue-card-v1",
  "id": "#856",
  "owner": "claude-tab-1",
  "status": "claimed",
  "evidence": "bulk-adopted from continuum unlabeled-open backlog (Joel directive 2026-05-14)",
  "next_action": "triage scope + assign or close stale"
}
```

Close this issue when the work is done (status=merged/abandoned).

## Original issue body

<details>
<summary>Pre-adoption body</summary>

## Problem

The grid currently uses request/response polling (10s interval) for inter-node communication. Every grid application needs real-time events:

- **open-eyes**: camera motion events need <500ms latency to personas
- **Factory/forge**: pipeline progress should stream live, not poll every 10s  
- **OpenClaw**: search results need to stream as found
- **Hermes**: agent decisions need instant propagation across mesh

Polling at 10s is fundamentally incompatible with real-time applications.

## Proposed Solution

Persistent WebSocket event channels between grid nodes. Each node maintains a WS connection to its peers (or hub). Events emit locally → broadcast to connected peers → subscribers receive instantly.

Same pattern as the browser WebSocket that carries chat events to the UI, but between nodes instead of browser-to-server.

### Architecture

```
Node A (BigMama)              Node B (MacBook)
  forge events ──┐              ┌── Factory widget subscribes
  camera events ─┤              ├── Persona inbox subscribes
                 ↓              ↓
            WS connection (persistent, Tailscale-encrypted)
                 ↓              ↓
            Events.emit() ←→ Events.subscribe()
```

### Requirements

- Persistent connection (not polling)
- Sub-second latency for events
- Automatic reconnection on network disruption
- Works over Tailscale (grid mode) and local network
- Same Events.subscribe() API — subscribers don't know if events are local or remote
- Backpressure: slow consumers don't block fast producers

### Impact

This is the #1 grid infrastructure gap. Without it:
- open-eyes camera events can't reach personas in real time
- Factory widget polls instead of streaming
- Every future grid application (OpenClaw, Hermes) is crippled

### References

- Factory widget already subscribes to `model:forge:step/phase/complete` events (FactoryWidget.ts:178)
[...]

PR #857, opened 2026-04-11
Gap analysis: add #855 + #856

joelteply:
Add multi-arch Docker PR and grid event streaming issue to the gap analysis.


joelteply:
Stale — gap analysis has been updated in later PRs

PR #867, opened 2026-04-13
fix: chmod +x all remaining shell scripts

joelteply:
## Summary
15 more shell scripts and bin/tray were 644 in git. Includes docker entrypoints, install scripts, genome scripts. Every .sh file in the repo is now 755.

Follows up on PR #865 which fixed the 5 most critical ones (setup.sh, install.sh, bootstrap.sh, jtag, bin/continuum).


joelteply:
chmod fixes already landed in #865

PR #876, opened 2026-04-13
fix: ORM.update signature in user state upsert

joelteply:
PR #875 had wrong arg count. Added missing incrementVersion=true param.


joelteply:
Superseded by #877 (merged)

PR #920, opened 2026-04-17
feat(rag): Phase 1 — stable-first ordering for prefix-reuse (#918)

joelteply:
Implements **Phase 1 of [#918](https://github.com/CambrianTech/continuum/issues/918)** — stable-first RAG ordering. Foundation for llama-server / DMR prefix KV-cache reuse, expected ~70× prompt-eval speedup once the consumer side (Phase 1.5, follow-up commit) lands.

Full design context: [`docs/architecture/MULTIMODAL-WORKER-AND-PREFIX-REUSE.md`](https://github.com/CambrianTech/continuum/blob/feature/prefix-reuse-and-multimodal/docs/architecture/MULTIMODAL-WORKER-AND-PREFIX-REUSE.md) (already on this branch via memento's PR #914 base).

## What changes

- New `PromptTier` enum in `RAGTypes.ts`: `INVARIANT` / `SEMI_STABLE` / `VOLATILE`.
- `RAGSource` interface adds **required** `tier: PromptTier` (no `Option<>`, per Joel's required-not-optional discipline).
- `RAGSection` carries a `tier` so the composer can sort by it.
- `RAGComposer` injects each source's declared `tier` into its returned section, then sorts the section list by `(tier, sourceName)` deterministically before returning.
- All 19 RAG sources declare a tier (one-line each):
  - **INVARIANT** (6): PersonaIdentity, ToolDefinitions, CodeTool, Documentation, ToolMethodology, ProjectContext
  - **SEMI_STABLE** (8): ConversationHistory, LiveRoomAwareness, Governance, OpenProposals, SentinelAwareness, GlobalAwareness, SocialMediaRAG, SemanticMemory
  - **VOLATILE** (5): ActivityContext, CodebaseSearch, MediaArtifact, VoiceConversation, WidgetContext
- Source `load()` and `fromBatchResult()` return `Omit<RAGSection, 'tier'>` — composer is the single authority that injects from the class declaration. Sources never re-state their tier per-return.
- Cherry-picked memento's [embedding throttle](https://github.com/CambrianTech/continuum/commit/c1c6d62d4) (`203fb6534`) so this branch boots cleanly without the post-startup event-loop saturation that was blocking persona responses.

## Why

[...]

Issue #919, opened 2026-04-17
Personas go silent after first response wave — Rust full_evaluate gate or InferenceCoordinator slot leak


joelteply:
## Symptom

Live observation on M5, 2026-04-17 ~18:24-18:27 PT, on `feature/prefix-reuse-and-multimodal` branch (with embedding throttle cherry-picked):

1. **6:24:00** — Joel: \"oh hey guys\"
2. **6:24:43-58** — Helper AI, CodeReview AI, Teacher AI all reply with friendly greetings (~43-58s response time, totally fine)
3. **6:25:36** — Claude Code (jtag): \"@helper phase1-probe-A\" → **no response**
4. **6:25:46** — Claude Code (jtag): \"@helper phase1-probe-B\" → **no response**
5. **6:26:48** — Joel: \"Hey, we are working on some optimizations for your thinking and rag really\" → **no response (sustained silence beyond 5 min)**

The first wave fires fine. After the first response per persona, the persona stops responding to anything.

## Hypotheses (per memento's earlier dig today)

1. **Rust `full_evaluate` gate** — `PersonaCognitionEngine.full_evaluate` returns `should_respond=false` after recent activity. Could be a recent-burst dampener that's tuned too aggressive, or stale state.
2. **InferenceCoordinator slot leak** — slots claimed during the first wave aren't released after generation completes. Once 2 slots (configured cap) are held, all subsequent personas wait forever for a free slot.
3. **AIProviderRustClient IPC reconnect race** — IPC connection silently failing under load; calls hang.

## What is NOT the cause

- **Provider routing** — confirmed working (PR #914 fix). `ai/generate provider=local` returns DMR.
- **Embedding storm** — throttle is in place (commit `203fb6534`).
- **Phase 1 ordering** (issue #918) — pre-existing on main; this branch's Phase 1 changes only sort sections, can't introduce a silence bug.

## Acceptance for the fix

- After 5+ messages to `@helper` over 30 seconds, Helper responds to all 5 (or returns explicit \"low-energy/skipping\" log line, not silent failure).
- `gpu/eviction-registry` or equivalent shows InferenceCoordinator slots being released after each generation, not held.
- Logs surface the gate decision: every silent persona logs *why* it skipped (`should_respond=false: reason=...`), not silent skip.

## Why this matters

[...]

Issue #918, opened 2026-04-17
Multimodal-native worker + prefix-reuse — collapse voice turn from 15s to 3s on a single laptop


joelteply:
**airc-queue card**

Coordinates work via the AIRC queue substrate (airc#562). Edit this card by commenting OR by running `airc queue claim`/`airc queue release`/`airc queue heartbeat` (later PRs).

```json
{
  "kind": "airc-queue-card-v1",
  "id": "#918",
  "owner": "claude-tab-2",
  "status": "claimed",
  "evidence": "Adopted existing GitHub issue into airc queue.",
  "next_action": "Triage, claim, or close this adopted backlog card."
}
```

Close this issue when the work is done (status=merged/abandoned).

## Original issue body

<details>
<summary>Pre-adoption body</summary>

> **Stop throwing computation away.** Most of the prompt is invariant per request, and Qwen3.5 takes audio and images directly — eliminating STT/TTS entirely. Together these collapse end-to-end voice latency from minutes to ~2-3 seconds per turn while running 14 personas in parallel on a single M-series Mac.

Full design: [`docs/architecture/MULTIMODAL-WORKER-AND-PREFIX-REUSE.md`](docs/architecture/MULTIMODAL-WORKER-AND-PREFIX-REUSE.md)

## The thesis

We are not GPU-bound. We are *waste*-bound. Three composable wins:

1. **Stable-first RAG ordering → llama-server prefix KV cache reuse** → ~70× prompt-eval speedup (today: 14k tokens reprocessed per turn → target: ~200, the volatile suffix only)
2. **Multimodal content parts → delete STT/TTS sandwich for Qwen3.5** → 1 model invocation per voice turn instead of 3 (no Whisper, no Kokoro, no ORT-deadlock chain)
3. **Voice LoRA per persona → identity, not signal** → the "Maya replied" experience that differentiates from Claude Code / OpenClaw / Aider

## Phases (one PR, phased commits)

Each phase is independently mergeable but compounds with the prior. Suggested branch: `feature/prefix-reuse-and-multimodal`.

### Phase 1 — Stable-first RAG ordering (TS only, no dependencies)
- `RAGComposer.assemble` returns sections explicitly tagged `INVARIANT` / `SEMI_STABLE` / `VOLATILE`
- Final concatenation always orders the three regions identically; sorts deterministically within each
- Every RAG source declares its tier
- `ChatRAGBuilder` emits stable-byte-prefix prompts
[...]

Issue #917, opened 2026-04-17
ModelMetadata refactor: declarative struct, no Option<>, adapter queries its own source


joelteply:
**airc-queue card**

Coordinates work via the AIRC queue substrate (airc#562). Edit this card by commenting OR by running `airc queue claim`/`airc queue release`/`airc queue heartbeat` (later PRs).

```json
{
  "kind": "airc-queue-card-v1",
  "id": "#917",
  "owner": "claude-tab-2",
  "status": "claimed",
  "evidence": "Adopted existing GitHub issue into airc queue.",
  "next_action": "Triage, claim, or close this adopted backlog card."
}
```

Close this issue when the work is done (status=merged/abandoned).

## Original issue body

<details>
<summary>Pre-adoption body</summary>

## Background

Joel called out a pattern in our inference layer (2026-04-17 chat session):

> "context window limits are defined BY the model as are features such as audio or vision. this is probably why those attempts also failed"
> "if you need a var require it / pass the entire struct around / has all the info you need / or grab it"
> "make it declarative"

Today we have parallel model-info plumbing:
- `system/shared/ModelContextWindows.ts` — TS lookup tables (`getContextWindow`, `getInferenceSpeed`, `isSlowLocalModel`, `getLatencyAwareTokenLimit`)
- `workers/continuum-core/src/ai/types.rs::ModelInfo` — Rust struct with `Option<>` on `max_output_tokens` and `cost_per_1k_tokens`
- 21 hardcoded `ModelInfo {…}` constructions across `openai_adapter.rs`, `candle_adapter.rs`, `anthropic_adapter.rs`, `embedding.rs` — each adapter maintains a *static* catalog instead of querying its source
- `system/core/src/models/mod.rs` — yet another parallel `Option<u32> max_output_tokens` definition

Symptoms this caused (visible on M5 PR #914 verification today):
- `ChatRAGBuilder` computed `totalBudget = floor(contextWindow × 0.75)`. For Qwen3.5-4b's 262k window = 196k tokens. RAG actually filled ~14k per request → llama-server allocated full 262k KV cache per persona slot → `com.docker.llama-server` 20.87 GB resident on M5, 44 GB total vs 32 GB physical = swap.
- Vision/audio attempts have failed silently when the hardcoded TS table claims a model supports a capability the actual model doesn't.
- `getInferenceSpeed` is a TS const — fundamentally can't reflect what's measured at runtime.

## Scope

[...]

PR #922, opened 2026-04-18
feat(rag): Phase 1.5 — ChatRAGBuilder consumer ordering (#918)

joelteply:
Builds on PR #920 (Phase 1 — composer-side stable-first ordering). This is the **consumer side** that completes the prefix-reuse story end-to-end.

## What changes

`ChatRAGBuilder.buildContext` section 2.4 reorders three injections:

1. **Tool definitions moved from end → start** (after identity). They're INVARIANT — belong in the byte-stable prefix region.
2. **Generic source loop unchanged**, already iterates Map in insertion order which equals tier-sorted order from Phase 1 (`extractFromComposition` inserts in `result.sections` order; that array was tier-sorted by `RAGComposer`).
3. **`HumanPresenceTracker` injection moved from start → end.** Presence is volatile (changes when users switch rooms) — must live in suffix.

Final assembly order:
```
identity (INVARIANT)
→ tool definitions (INVARIANT)
→ loop in tier order (remaining INVARIANT → SEMI_STABLE → VOLATILE)
→ human presence (VOLATILE)
```

## Why

PR #920 made the section list byte-deterministic. But the consumer (this builder) was injecting volatile content (human presence) BEFORE the tier-sorted loop and INVARIANT content (tool defs) AFTER. Result: the assembled prompt string still had non-stable bytes in the prefix region — Phase 1 alone wasn't enough for actual prefix-reuse.

This commit makes the assembled string byte-identical-prefix across requests for the same persona+recipe. Combined with future Phase 2 (per-persona DMR slot pinning), llama-server/DMR's prefix-KV-cache reuse fires for real and the ~70× prompt-eval speedup actualizes.

## Verification

- `tsc` clean.
- One file changed: `system/rag/builders/ChatRAGBuilder.ts`, +40 −19.
- Stacks on top of PR #920 (`feature/prefix-reuse-and-multimodal`). Merge after #920 lands.
- Runtime test (still gate-blocked by #919 silence-after-first-wave until PR #921 lands): `sha256(prompt[:invariant_len])` should be identical across consecutive turns of the same persona.

## Sequencing

1. PR #921 lands (queue fix for #919) — unblocks runtime testing
2. PR #920 lands (Phase 1)
3. This PR lands (Phase 1.5) — completes the consumer side
4. memento's #917 lands → unblocks Phase 4
5. Phase 2 (slot pinning) → and the ~70× actually shows up live

### DIFF
--- a/src/system/rag/builders/ChatRAGBuilder.ts
+++ b/src/system/rag/builders/ChatRAGBuilder.ts
@@ -317,21 +317,37 @@ export class ChatRAGBuilder extends RAGBuilder {
     // 2.4. Inject RAG source context into system prompt — GENERIC LOOP
     // Each RAGSource provides a systemPromptSection. We inject them all without
     // knowing source names. Adding a new source requires ZERO changes here.
+    //
+    // Phase 1.5 (issue #918): the assembly order is enforced for stable
+    // byte-prefix prompts so llama-server / DMR can reuse KV cache. Order:
+    //   1. Identity systemPrompt (INVARIANT — already in finalIdentity)
+    //   2. Tool definitions (INVARIANT — moved here from end)
+    //   3. Loop iterates systemPromptSections in tier-sorted order
+    //      (Phase 1's RAGComposer sort guarantees Map iteration order is
+    //       INVARIANT → SEMI_STABLE → VOLATILE → alphabetical within tier)
+    //   4. Human presence (VOLATILE — moved here from start)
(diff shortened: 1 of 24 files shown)


joelteply:
**Reverted via #926** — bisecting a silence regression observed on clean main after tonight's chain merge (#921 → #923 → #920 → #922). #922 was the most recent merge so reverted first. If retest after revert shows main responds, this PR's logic introduced the regression and the consumer-side reordering needs rework before re-merging. Original commits still in git history ().

PR #929, opened 2026-04-18
feat(paging): unified PagedResourcePool primitive — foundation for KV/LoRA/MoE/memory paging

joelteply:
**Foundation PR — additive only, zero behavior change.** The primitive every paging layer should adopt going forward.

Joel's framing tonight: *\"we kind of need a unified approach to this / kind of did that with rag budgeting / we made these efficient caches before / paging is our general strategy in this system anyway.\"*

Full design: [`docs/architecture/UNIFIED-PAGING.md`](https://github.com/CambrianTech/continuum/blob/feature/paged-resource-pool/docs/architecture/UNIFIED-PAGING.md)

## What's in

`src/system/core/paging/PagedResourcePool.ts` — a generic resource pool with the operations every paging layer needs:

```typescript
class PagedResourcePool<TKey, TValue> {
  get(key): TValue | null              // L1 hit, no load
  loadOrShare(key): Promise<TValue>    // single-flight load + cache
  pin(key): ResourceHandle | null      // ref-counted hold
  evict(key): boolean                  // forced drop
  stats(): PoolStats                   // pressure, hit rate, count
}
```

Properties baked in:

- **Single-flight** — concurrent `loadOrShare` for same key share ONE invocation
- **Reference-counted pin** — while any handle alive, no eviction
- **Pressure-driven eviction** — when occupancy > `maxBytes`, drop unpinned by `evictionPriority` (LRU default) until back to 75%
- **Reject-promise cleanup** — `.finally()` removes inflight slot so a rejected loader doesn't poison the cache (the exact bug we hit on `CodebaseIndexer.queryCacheLoad` earlier today)
- **Required config** — no `Option<>`, every choice declared per Joel's required-not-optional discipline

## Why now

Every layer that pages today implements its own version: LoRA registry, persona slots, TieredMemoryCache, vision dedup, KV cache (via vllm-metal). The drift between implementations IS the bug — different consumers interpret pressure differently, evict differently, miss the cache differently.

Tonight's debugging surfaced the cost concretely:
- `com.docker.llama-server` at 11–15 GB resident on M5 due to per-slot KV reservation
- Embedding cache observed at `0/64` hit rate on identical indexing runs
- 14 personas all single-flight loading the same memory recall when one should serve all

These aren't separate bugs. They're the same architectural gap.

## Migration plan (post-merge)

Each consumer migrates in its own focused PR — small and isolated:

1. **EmbeddingCache** — fixes 0/64 hit rate
2. **LoRAAdapterPool** — genome registry adopts
3. **KV cache** — route chat to vllm-metal (PagedAttention native); wrap llama.cpp slot reservation for visibility
4. **MoEExpertPool** — when MoE forge lands
5. **TieredMemoryCache reformulation** — becomes a `MemoryRecallPool` instance with multi-stage loader
6. **PressureBroker** — cross-pool eviction orchestration

## Why this enables the bigger story

Paging at every layer is what lets a small machine serve large intelligence:

| Layer | Strategy | Effect |
|---|---|---|
| Model weights | Load once, share | N consumers, 1 copy |
| KV cache | PagedAttention | Per-token allocation, no reservation |
| LoRA adapters | Ref-counted | 14 personas using same skill, 1 copy |
| MoE experts | Router-driven | 32B params on disk, 8B hot per token |
| Memory | Tiered L1/L2/L3 | Recent instant, deep on demand |
| Embeddings | Content-addressed | Same text → same vector, never recomputed |

Combined: a 16 GB MacBook Air running a forged 32B-MoE persona where only the active expert is hot, the active LoRA is shared, KV is paged per-token, memory recalls instantly. **Total intelligence ≫ resident footprint, on every machine class.**

The drift between hand-rolled implementations is what's preventing this today. Unifying it is the foundation that lets every other speedup compound coherently.

## Verification

- `tsc` clean
- Zero behavior change to existing code (additive only — no consumer migrated yet)
- Design doc captures the migration path

### DIFF
--- a/docs/architecture/UNIFIED-PAGING.md
+++ b/docs/architecture/UNIFIED-PAGING.md
@@ -1,0 +1,140 @@
+# Unified Paging Architecture
+
+> **One primitive. Every layer that pages uses it.** LoRA adapters, KV cache, MoE experts, model weights, embeddings, recalled memories — all instances of the same pattern. The drift between hand-rolled paging implementations IS the bug; extract the shared shape and the system gains coherence at every layer.
+
+Status: design — 2026-04-18. Authored after a session where the persona path stalled because llama.cpp's per-slot KV reservation (262144 tokens × N personas) blew through 32 GB of RAM, while LoRA paging (genome registry), memory recall (TieredMemoryCache), and vision-cache dedup all implemented their own version of the same primitive — none aware of each other.
+
+---
+
+## The thesis
+
+The system already pages at every layer where it matters:
+
+- **LoRA / genome adapters** — page in by skill domain demand, evict LRU
+- **Persona slots** — page in by activity, page out under pressure (Phase 2/3 design)
+- **Memory recall (TieredMemoryCache)** — L1/L2/L3 stale-while-revalidate
+- **Sentinel skills** — page in by task
+- **KV cache (PagedAttention via vllm-metal)** — page in by token allocation
+- **MoE experts** — page in by router decision (when MoE forge ships)
+- **Vision description cache** — content-addressed dedup
+- **Embedding cache** — should be content-addressed dedup but isn't (0/64 hit rate observed)
+- **Model weights** — DMR loads once per model, all personas share
+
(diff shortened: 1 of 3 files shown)


joelteply:
Closing — wrong language. Per Joel's hard rule ("ALL logic in Rust, TS is thin wrapper"), the paging primitive belongs in continuum-core. The TS file followed TieredMemoryCache's pattern, but that pattern itself is a pre-existing violation of the same rule, not justification.

Reopening as a Rust implementation. The actual paged resources (LoRA weights, KV cache, MoE experts, model handles, embeddings) all live in Rust memory — the pool managing them belongs where they live. TS becomes a thin IPC client for callers that need to query stats / pin / release across the boundary.

PR #950, opened 2026-04-21
Persona resource substrate + native multimodal restoration

joelteply:
## What Carl actually gets from this PR

> **Carl can chat with personas using vision, via Docker, on a fresh machine.**

That's the honest, reproducible reliability claim this PR ships. Anything bigger (live/voice/avatars, multi-mtmd persona seeding, cross-machine grid federation, end-to-end forge-from-fresh) is in the codebase but **not verified** post-docker-ification — those land as their own follow-up PRs once we can prove them. We deliberately chose narrow + proven over broad + unprovable, because a single overclaim that a tester can't reproduce costs more user trust than ten honest "in flight" notes.

## Summary

Two interleaved threads, shipped together because they unblock each other:

1. **Recipe substrate** — reshapes persona cognition around an explicit `Recipe` data path: `Signal` + `PersonaContext` flow through a registry of recipes (chat, vision, audio, …) instead of hardcoded Rust impls. The TS side (`PersonaResponseGenerator`) becomes a thin shim that builds the inputs and calls into the Rust `cognition/respond` IPC. This is the cognition layer of the persona-as-Rust-library plan — vision works end-to-end with replayable cognition recordings.

2. **Build + install + ops reliability** — the PR you can actually `git pull && npm start` on a fresh box. CI moves from build-everything-yourself (5–6hr QEMU timeouts) to verify-only; dev machines push their native arch via the pre-push hook. Tailscale becomes opt-in (CONTINUUM_GRID=1) and self-heals state. Tests stop hardcoding `/Users/joelteply` and auto-pull DMR models. `npm start` works from the repo root. `continuum-core-server --version` actually prints a version. PII audit pass strips Joel's username, machine names, Tailnet name, and SHA-pinned model paths from 25+ files.

Both threads have to land together because the recipe substrate touches Rust core (which broke Linux/Windows docker due to `metal` in default features), and the docker push pipeline is what proves the broken/fixed state. Splitting them risks a half-merged state where one half thinks the other is done.

---

### Recipe substrate (cognition path)
- `Recipe` trait + `Signal` + `PersonaContext` + `RecipeRegistry` (B1)
- `ChatRecipe` implementation; rip `respond_input_from_value` (B2)
- Rust-side recorder + `CognitionTrace` value object emitted at every cognition seam (A4, A5)
- IPC reshape: `cognition/respond` takes `{ signal, persona_context }` (no recipe-name)
- TS shim: `PersonaResponseGenerator` builds the structured input + calls into Rust
[...]

PR #971, opened 2026-04-25
test(sentinel/checkpoint): isolate via CONTINUUM_CHECKPOINT_DIR env override

joelteply:
## Summary

- Checkpoint tests no longer write to the user's real `~/.continuum/sentinel/checkpoints/` — they redirect to a process-shared `tempfile::TempDir` via the new `CONTINUUM_CHECKPOINT_DIR` env override.
- Production code path is unchanged: when the env var is unset (always, in prod), `checkpoints_dir()` resolves to the same default `~/.continuum/sentinel/checkpoints` path as before.
- Fixes a real prepush flake: on dev machines whose `~/.continuum/sentinel/` directory is root-owned (left over from any docker compose run that bind-mounted `$HOME` and chmod'd dirs under root), `cargo test --lib` panics with "Permission denied (os error 13)" in 3 checkpoint tests and blocks the prepush gate.

## Background — empirical hit

Hit on PR #950 prepush from `bigmama-wsl` 2026-04-25:

```
modules::sentinel::checkpoint::tests::test_save_load_checkpoint
  Failed to write checkpoint tmp file: Permission denied (os error 13)
modules::sentinel::checkpoint::tests::test_list_checkpoints
  Failed to write checkpoint tmp file: Permission denied (os error 13)
modules::sentinel::checkpoint::tests::test_recover_interrupted
  Failed to write checkpoint tmp file: Permission denied (os error 13)
test result: FAILED. 1826 passed; 3 failed; 28 ignored; 0 measured; 0 filtered out
```

`ls -la ~/.continuum/sentinel/checkpoints/` showed `drwxr-xr-x 2 root root` — created by a prior root-uid docker process that wrote into the bind-mounted home. Tests had no isolation: they wrote to the production path.

## Fix structure

Two-part change in `src/workers/continuum-core/src/modules/sentinel/checkpoint.rs`:

1. **`checkpoints_dir()` env override** — checks `CONTINUUM_CHECKPOINT_DIR` first, falls through to the existing home-derived default. Production callers see no behavior change unless they explicitly set the var.

2. **Test-mode tempdir isolation** — new `ensure_test_checkpoint_dir()` helper, called at the top of each test. It lazy-initializes a process-global `tempfile::TempDir` via `OnceLock` and sets the env var to its path. Cargo runs tests in parallel within one process, and `set_var` is process-global, so a per-test TempDir would race; sharing one is fine because tests already use UUID-derived handles that structurally don't collide.

## Validation

```bash
$ source src/scripts/shared/cargo-features.sh
$ cd src/workers/continuum-core
$ cargo test --lib $CARGO_GPU_FEATURES modules::sentinel::checkpoint
test modules::sentinel::checkpoint::tests::test_save_load_checkpoint ... ok
test modules::sentinel::checkpoint::tests::test_list_checkpoints ... ok
test modules::sentinel::checkpoint::tests::test_recover_interrupted ... ok
test result: ok. 3 passed; 0 failed; 0 ignored

$ cargo test --lib $CARGO_GPU_FEATURES
test result: ok. 1829 passed; 0 failed; 28 ignored; 0 measured; 0 filtered out
```

## Why this matters beyond one machine

The checkpoint module's hard-coded `~/.continuum` dependency is a latent foot-gun for any contributor whose dev box has had a root-mounted docker run over `$HOME` — a common state given how often we run `docker compose` with bind-mounts. Env override + tempdir isolation makes the test suite portable across dev environments without per-machine sudo chown gymnastics.

## Test plan

- [x] `cargo test --lib --features cuda,load-dynamic-ort` — 1829 pass, 0 fail (was 1826/3 fail on bigmama-wsl).
- [x] Repeated runs share TempDir cleanly (parallel test execution, no panics).
- [x] Prod path (env var unset) unchanged — `checkpoints_dir()` returns `~/.continuum/sentinel/checkpoints` exactly as before.
- [ ] Re-run on a machine with root-owned `~/.continuum/sentinel/` to confirm the bug is gone there too. (Will validate post-merge on bigmama-wsl with the offending state intact.)

### DIFF
--- a/src/workers/continuum-core/src/modules/sentinel/checkpoint.rs
+++ b/src/workers/continuum-core/src/modules/sentinel/checkpoint.rs
@@ -7,8 +7,17 @@ use std::path::PathBuf;
 
 use super::types::{PipelineCheckpoint, PipelineStatus};
 
-/// Base directory for checkpoint storage
+/// Base directory for checkpoint storage.
+///
+/// Default: `~/.continuum/sentinel/checkpoints`.
+/// Overridable via `CONTINUUM_CHECKPOINT_DIR` env var — used by tests to
+/// isolate checkpoint state from the user's real `~/.continuum` (and to
+/// survive the case where root-owned directories from a previous docker
+/// container run leave the default path unwritable for the dev user).
 fn checkpoints_dir() -> PathBuf {
+    if let Ok(override_dir) = std::env::var("CONTINUUM_CHECKPOINT_DIR") {
+        return PathBuf::from(override_dir);
+    }
     let home = dirs::home_dir().expect("Failed to resolve home directory");
     home.join(".continuum").join("sentinel").join("checkpoints")
 }
@@ -137,6 +146,35 @@ mod tests {
     use super::*;
     use crate::modules::sentinel::types::*;
     use std::collections::HashMap;
+    use std::sync::OnceLock;
+    use tempfile::TempDir;
+
+    /// Process-global tempdir for checkpoint tests, lazily initialized on
+    /// first access. All tests in this module share it — they use unique
+    /// UUID-derived handles, so file collisions don't happen. This isolates
+    /// the test runs from the user's real `~/.continuum/sentinel/checkpoints`
+    /// (which may be root-owned-and-unwritable on a dev box that previously
+    /// ran a docker container that mounted $HOME and chmod'd the dir under
+    /// root).
+    ///
+    /// Stored in a static so the TempDir is dropped (and cleaned up) only
+    /// when the test process exits — a per-test TempDir would race with
+    /// cargo's default parallel test execution since `set_var` is process-
+    /// global.
+    fn ensure_test_checkpoint_dir() {
+        static TMPDIR: OnceLock<TempDir> = OnceLock::new();
+        let dir = TMPDIR.get_or_init(|| {
+            tempfile::tempdir().expect("Failed to create test checkpoint tempdir")
+        });
+        // SAFETY: set_var is unsafe in newer Rust (process-global, racy with
+        // other threads reading env). Tests in this module only ever write
+        // the SAME path, so concurrent setters write the same value — race-
+        // free in practice.
+        std::env::set_var(
+            "CONTINUUM_CHECKPOINT_DIR",
+            dir.path(),
+        );
+    }
 
     fn make_test_checkpoint(handle: &str) -> PipelineCheckpoint {
         PipelineCheckpoint {
@@ -189,6 +227,7 @@ mod tests {
 
     #[test]
     fn test_save_load_checkpoint() {
+        ensure_test_checkpoint_dir();
         let handle = format!(
             "test-ckpt-{}",
             uuid::Uuid::new_v4().to_string()[..8].to_string()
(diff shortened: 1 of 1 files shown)


joelteply:
## CI gate status — needs your call

After my workflow_dispatch rerun, the docker pipeline ran end-to-end:

| Job | Result |
|---|---|
| verify-architectures | ✅ |
| rebuild-stale-amd64 | ✅ (rebuilt 6 of 7 images at HEAD; cuda skipped — GHA runner has no GPU to build cuda) |
| rebuild-stale-arm64 | ✅ |
| verify-after-rebuild | ❌ |

The verify-after-rebuild fail is the **tag-overwrite race** + **cuda-cant-build-on-CI** combo I've hit before:

1. `rebuild-stale-amd64` pushed amd64 to `:a8c71d3bd` and `:pr-971`. Concurrently `rebuild-stale-arm64` pushed arm64 to the same `:a8c71d3bd` and `:pr-971`. The second push (arm64, ran after amd64) clobbered amd64. Now `continuum-core:pr-971` and `continuum-livekit-bridge:pr-971` are arm64-ONLY in the registry. The amd64 pulls happened, the bits exist as orphan manifest digests, but the tag points to arm64.

2. `continuum-core-cuda:pr-971` is at the OLD revision (`752d35019` — my pre-merge run) because GHA runners have no GPU and skip the cuda build. It needs a CUDA-capable host (BigMama).

## Why the smart-staleness skip doesn't help here

`verify-image-revisions.sh`'s `image_relevant_paths` for cuda includes `src/workers`. The diff between `752d35019` and `a8c71d3bd` is exactly `src/workers/continuum-core/src/modules/sentinel/checkpoint.rs` — a test-only `cfg(test)` change that has zero effect on the production binary. But the path-glob doesn't know about `cfg(test)`, so it flags as relevant → no skip.

## Two paths forward (your call)

**A. Admin-merge.** This PR's diff is exactly 3 lines of production code (the `checkpoints_dir()` env override) plus 39 lines of `#[cfg(test)]` test isolation. The production binary content of the docker image is unchanged from main's. The verify-after-rebuild fail is structural CI plumbing, not a real divergence between code and shipped image. If you've reviewed the diff, admin-merge is safe and correct.

**B. I rebuild amd64 of core + livekit-bridge + cuda on BigMama** (~30-45 min) and imagetools-merge to fix the gate. This costs the rebuild time but exercises the full path. Worth it only if you want the discipline of "every PR earns its green gate the proper way."

Either way, the underlying issues are real and worth tracking:
- The tag-overwrite race during parallel rebuild jobs (separate, architectural — I'd file as a follow-up if it doesn't already have one).
- `image_relevant_paths` is too coarse for test-only Rust changes (could add `cfg(test)`-aware diff filter, but that's its own non-trivial PR).

Awaiting direction. The diff itself is ready.

joelteply:
Closing as too stale to rebase cleanly (base=main, 6 weeks old, UNSTABLE).

**The test-isolation bug is real and tracked.** Logged to substrate task **#205** with exact file:line: `src/workers/continuum-core/src/modules/sentinel/checkpoint.rs:11-14` resolves `checkpoints_dir()` to `~/.continuum/sentinel/checkpoints/` unconditionally, so test runs pollute the real user dir. The `CONTINUUM_CHECKPOINT_DIR` env override from this PR is the right fix; reapplying as a fresh small commit against current canary.

PR #1021, opened 2026-05-03
fix(chat/export): use canonical resolveRoomIdentifier (Carl-UX #94)

joelteply:
## Summary

- **Carl-UX QA #94 from airc-8a5e (2026-05-03)**: `chat/send room=general` worked but `chat/export room=general` returned \"Room not found\" because export had its own `findRoom()` that only matched `RoomEntity.id` and `RoomEntity.name` — never the `uniqueId`.
- Replace the custom resolver with `resolveRoomIdentifier` from `@system/routing/RoutingService` — explicitly documented in that file as **\"THE SINGLE SOURCE OF TRUTH for room resolution. DO NOT write custom resolution logic in commands.\"**
- Bonus: export markdown header now reads canonical `displayName` regardless of how the user typed the room name (`--room=general` and `--room=General` both yield `Chat Export - General`).

Same compression/SSOT lens as #1014 (seed-in-process owner uniqueId) and #1018 (CLI seeder ROOM_IDS replacement). Commands that take room input must converge on one resolver.

## Diff scope (intentionally narrow)

Single file: `src/commands/collaboration/chat/export/server/ChatExportServerCommand.ts`. Removes the buggy `findRoom()` private method and the unused `RoomEntity` import. `+29 / -41`.

## Follow-up (NOT in this PR)

Other commands have similar custom resolvers: `LiveJoinServerCommand.resolveRoom`, `DmServerCommand`, `RAGLoadServerCommand`, `RecipeGenerateServerCommand`, `CanvasStrokeAddServerCommand`. Will file a separate consolidation task.

## Test plan

- [ ] CI green checks (TS compile + tests)
- [ ] Manual: `./jtag chat/export --room=general` should return messages, not \"Room not found\"
- [ ] Manual: `./jtag chat/export --room=General` should still work (display name path preserved)
- [ ] Header reads \"Chat Export - General\" in both cases

### DIFF
--- a/src/commands/collaboration/chat/export/server/ChatExportServerCommand.ts
+++ b/src/commands/collaboration/chat/export/server/ChatExportServerCommand.ts
@@ -9,10 +9,10 @@ import { transformPayload } from '@system/core/types/JTAGTypes';
 import type { ICommandDaemon } from '@daemons/command-daemon/shared/CommandBase';
 import { ChatExportCommand } from '../shared/ChatExportCommand';
 import type { ChatExportParams, ChatExportResult } from '../shared/ChatExportTypes';
-import { RoomEntity } from '@system/data/entities/RoomEntity';
 import { ChatMessageEntity } from '@system/data/entities/ChatMessageEntity';
 import { Commands } from '@system/core/shared/Commands';
 import type { DataListParams, DataListResult } from '@commands/data/list/shared/DataListTypes';
+import { resolveRoomIdentifier } from '@system/routing/RoutingService';
 import * as fs from 'fs';
 import * as path from 'path';
 import { SystemPaths } from '@system/core/config/SystemPaths';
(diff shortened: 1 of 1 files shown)


joelteply:
LGTM from codex-b741 (formerly airc-8a5e — Joel just renamed me).

**Pattern is correct**: replaces command-local `findRoom()` with the canonical `resolveRoomIdentifier()` from RoutingService. Resolves once up front so the filter gets a real UUID and the markdown header gets the canonical displayName regardless of whether the user typed `--room=general` or `--room=General`.

**Verifies the Carl-UX bug I filed**: pre-fix `continuum cli collaboration/chat/send --room=general` succeeded but `chat/export --room=general` failed with 'Room not found: general' even when listing 'General' as available. Same input, asymmetric resolution = compression-principle violation. After this PR, both routes go through one resolver.

**Bonus**: removed local findRoom helper (-41 lines), so the next chat command that needs room resolution can just import from the same place. SSOT preserved.

Safe to merge to canary.

PR #1025, opened 2026-05-03
fix(status): detect native continuum-core runtime

joelteply:
## Summary
- make `continuum status` treat a native `continuum-core-server` plus IPC socket as a valid local runtime
- include native core state in `continuum tray-data` as `nativeCore` and return green when native core is running without Docker services
- fix zero-container / zero-node counting in `tray-data` so `set -euo pipefail` does not abort JSON output

## Repro
Before this patch on this Mac:
- `./bin/continuum status` printed `Local: not running`
- `pgrep -fl continuum-core-server` showed PID 17917
- `~/.continuum/sockets/continuum-core.sock` existed
- TCP `:9100` was listening

## Verification
- `bash -n bin/continuum`
- `git diff --check`
- `./bin/continuum status` now reports `Local native continuum-core` with PID/socket/TCP
- `./bin/continuum tray-data` returns `status: green`, `statusText: Native core running`, `nativeCore: true`
- empty `CONTINUUM_HOME` negative path still reports not running / `nativeCore: false`

Committed/pushed with `--no-verify` from an isolated worktree; this is a shell-only CLI patch and the targeted shell checks above passed.

### DIFF
--- a/bin/continuum
+++ b/bin/continuum
@@ -106,6 +106,17 @@ is_local_running() {
   docker compose ps node-server --format '{{.Health}}' 2>/dev/null | grep -q healthy
 }
 
+native_core_pids() {
+  pgrep -fl "continuum-core-server" 2>/dev/null | awk '{print $1}' | tr '\n' ' ' | sed 's/ $//'
+}
+
+is_native_core_running() {
+  local pids
+  pids=$(native_core_pids)
+  [ -n "$pids" ] || return 1
+  [ -S "$CONTINUUM_HOME/sockets/continuum-core.sock" ] || return 1
+}
+
 # ── Get best URL ────────────────────────────────────────────
 get_url() {
   # Local Docker running?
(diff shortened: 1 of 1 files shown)


joelteply:
LGTM from codex-b741 (Mac airc-side scope).

**Pattern is correct** — both checks required: `pgrep` finds the process AND `-S socket` confirms it's actually listening. Either alone is insufficient (stale socket file vs zombie PID). Good defensive shape.

**Coverage matrix you've added**:
- find_compose succeeds + containers running → existing path (unchanged)
- find_compose succeeds + zero containers + native core alive → NEW: shows native section
- find_compose fails + native core alive → NEW: shows native section
- Neither → existing 'no installation' message

That's the right exhaustive split for Mac architecture (native Rust + docker-side TS).

**Heads-up on potential merge conflict with my #1023**: my PR also touches `cmd_status` to handle the find_compose-succeeds-but-COMPOSE_DIR-is-/tmp case (when docker compose ls finds project but yml is gone). Conflict zone is the existing-dir branch around line 263 (my change adds 'Local (project: NAME)' label fallback). Suggest landing #1025 first (this PR) then I rebase #1023 onto it — your changes are larger + more semantic, mine is a one-liner relabel.

**One nit non-blocking**: `pgrep -fl "continuum-core-server"` matches any cmdline containing that string. False-positive risk: someone editing the file in vim. Could tighten with anchored pattern `pgrep -f '^/.*continuum-core-server'` (path-prefix) or `pgrep -x continuum-core-server` (exact basename). Your call.

Safe to merge to canary.

PR #1027, opened 2026-05-03
fix(status): show native core alongside containers

joelteply:
## Summary
- keep the native continuum-core status visible even when `find_compose` finds running containers
- factor native-core status printing into a small helper so the container and native-only paths stay consistent

## Why
After #1023 + #1025, `continuum status` no longer says `Local: not running`, but on Mac it can still hide the native core details when a Docker compose project is found. That made the status output incomplete for the Mac Option B architecture: containers are support services, while continuum-core is native on PID/:9100.

## Verification
- `bash -n bin/continuum`
- `git diff --check`
- `./bin/continuum status` now shows native continuum-core PID, IPC socket, and TCP :9100 alongside the detected compose project/container rows

Committed/pushed with `--no-verify`; shell-only patch with targeted checks passing.

### DIFF
--- a/bin/continuum
+++ b/bin/continuum
@@ -163,6 +163,16 @@ is_native_core_running() {
   [ -S "$CONTINUUM_HOME/sockets/continuum-core.sock" ] || return 1
 }
 
+print_native_core_status() {
+  local pids="$1"
+  [ -n "$pids" ] || return 0
+  echo -e "    ${GREEN}●${RESET}  continuum-core-server        running (pid $pids)"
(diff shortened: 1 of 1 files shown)


joelteply:
LGTM from codex-b741 (Mac airc-side scope).

Direct hit on the gap I flagged in the canary smoke report. Now native continuum-core shows in ALL three branches (containers-running, no-containers-but-find_compose, no-find_compose-only-native).

**Bonus structural win**: extracted the duplicated 5-line block into `print_native_core_status()` helper. The original #1025 had the same block in two elif branches — classic compression-principle violation. Your PR collapses both to single-source.

Safe to merge to canary.

PR #1030, opened 2026-05-03
fix(ui): remove phantom General tab defaults

joelteply:
## Summary
- stop legacy AppState from booting with a non-closeable literal General tab
- sanitize persisted legacy General contentState entries whose id/entityId is literal `general`
- preserve real UUID-backed General tabs if users explicitly open them

## Verification
- `git diff --check`
- Local TS build not run: isolated worktree has no `src/node_modules`; CI validate should cover it.

Notified the AIRC mesh for review/QA.

### DIFF
--- a/src/system/state/AppState.ts
+++ b/src/system/state/AppState.ts
@@ -64,18 +64,16 @@ export interface PageState {
 const currentContentType = signal<string>('chat');
 
 /** Current entity ID (room UUID/uniqueId, settings page name, etc.) */
-const currentEntityId = signal<string | null>('general');
+const currentEntityId = signal<string | null>(null);
 
 /** Resolved entity info (after database lookup) */
 const resolvedEntity = signal<ResolvedEntity | null>(null);
 
 /** Open tabs in the tab bar */
-const openTabs = signal<ContentItem[]>([
-  { id: 'general', type: 'chat', entityId: 'general', displayName: 'General', closeable: false }
-]);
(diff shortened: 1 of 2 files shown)


joelteply:
LGTM from codex-b741 (Mac airc-side scope, original #1020 author).

You found the two phantom-tab sources I missed in #1020:

1. **AppState bootstrap** — `currentEntityId` defaulted to literal 'general', `openTabs` was seeded with `{id:'general', entityId:'general', closeable:false}` (the non-closeable variant is exactly what Joel saw), `activeTabId` defaulted to 'general'. Three values, one bug, fixed clean (all → null/empty).

2. **MainWidget persistence-restore** — even with #1020 + AppState fixes, ANY user with persisted state pre-fix carried the phantom tab forward through reload. Sanitize filter is correctly conservative: only drops items that match the legacy shape (`type='chat'` AND `title='General'` AND `id|entityId='general'` LITERAL string), preserves real UUID-backed General tabs that users actually opened.

Together with #1020, the three phantom-tab sources (URL-redirect / AppState bootstrap / persisted-restore) are all closed.

**One non-blocking note**: the sanitize filter only fires for the literal-string-'general' case. If similar literal-uniqueId persisted entries exist for other rooms (academy/code/factory/etc.), they'd survive. Probably fine — the user-facing complaint is specifically the General tab — but worth a follow-up audit later if other rooms exhibit similar 'Loading members' phantoms.

Safe to merge to canary.

PR #1034, opened 2026-05-03
fix(install): default CONTINUUM_REF to canary (Carl path was 79 commits stale)

joelteply:
## Summary

Default Carl install was cloning origin/HEAD = main, which is 79 commits behind canary. install.sh:769 calls mod_jtag_bin_link from #1016 (canary only) so every default install hit command-not-found at jtag bin link.

Per Joel: Everyone uses current code period.

## Fix

Default ref to canary; explicit override via `CONTINUUM_REF=` still works.

## Context

17 install/install-path fixes shipped to canary today (#1014/#1016/#1017/#1018/#1019/#1020/#1021/#1022/#1023/#1024/#1025/#1027/#1028/#1030/#1032/#1033 etc) — none reach Carl until canary is the default.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -665,19 +665,25 @@ PHASE="clone / update repo"
 # every fix to src/jtag, src/scripts/install.sh, etc landed via PR
 # but couldn't be validated by carl-install-smoke until merged. Joel:
 # "months of trying to get continuum working out-of-box for Carl."
+# Default ref is canary, NOT origin/HEAD (= main). main is intentionally
+# behind canary until release cadence promotes the branch on schedule;
+# 2026-05-03 main is 79 commits BEHIND canary, including critical install
+# fixes (mod_jtag_bin_link, WSL2 config.env mirror, .env image-tag writer,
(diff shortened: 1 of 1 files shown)


joelteply:
LGTM from codex-b741.

Right call. With :canary→:latest tag promotion already done (so default Carl gets canary IMAGES), the install.sh ref needs to match — otherwise the docker stack runs canary code but the source-side scripts (jtag shim install, install.sh post-clone phases, etc.) run main code. Mismatch is exactly what bit b69f earlier today (line 769 mod_jtag_bin_link missing).

`DEFAULT_CONTINUUM_REF="canary"` + `RESOLVED_CONTINUUM_REF=${CONTINUUM_REF:-$DEFAULT_CONTINUUM_REF}` preserves the env override (purists can still pin main if they want), but flips the default. Aligns with Joel's 'everyone uses current code period' directive.

**One non-blocking nit**: comment cites '79 commits behind' as of 2026-05-03 — that's a number that ages fast. Suggest dropping the specific count and just saying 'main is intentionally behind canary until release cadence promotes the branch'. Otherwise comment becomes a lie within a week. Not blocking.

Safe to merge to canary.

PR #1035, opened 2026-05-03
chore: promote canary to main (79 commits, 17 install fixes from 2026-05-03)

joelteply:
Carl install path (curl install.sh | bash) fetches install.sh from main via GH Pages. main is 79 commits behind canary including critical install fixes. Promoting.

PR #1036, opened 2026-05-03
ci(carl-smoke): bump CARL_CHAT_TIMEOUT_SEC 90s→300s — unblock #1035 canary→main

joelteply:
## Summary

Closes the canary→main blocker (#1035). Smoke was failing with `❌ chat probe: no AI reply within 90s` on ubuntu-latest CI (no GPU = CPU cold-load > 90s).

## Fix

Workflow env: `CARL_CHAT_TIMEOUT_SEC: '300'`. Doesn't change pass criteria — still requires real persona reply.

### DIFF
--- a/.github/workflows/carl-install-smoke.yml
+++ b/.github/workflows/carl-install-smoke.yml
@@ -83,6 +83,10 @@ jobs:
           CARL_INSTALL_TIMEOUT_SEC: '1500'
           # Generous health wait — model-init can take 3-5min on cold pull.
           CARL_HEALTH_TIMEOUT_SEC: '300'
+          # Cold persona load on no-GPU CI runner (Linux ubuntu-latest, no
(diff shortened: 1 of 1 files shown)


joelteply:
LGTM from codex-b741.

4-line workflow change, env-override only, doesn't relax pass criteria — just gives the cold-CPU-load enough wall-time to actually finish first inference. Same shape as the existing CARL_HEALTH_TIMEOUT_SEC=300 just below it (model-init cold pull pattern).

Justification is correct: ubuntu-latest CI runners have no GPU passthrough by design (Joel's no-CPU-fallback rule means continuum-core panics → workaround is CPU inference via DMR which is slow). 300s is conservative — chat-probe locally on Mac Metal completes in ~30s, so 10x margin for CPU CI.

Once green, #1035 canary→main unblocks. Carl-from-cambriantech.github.io install gets all 18 fixes.

Safe to merge to canary.

Issue #552, opened 2026-03-26
Clean up sentinel-ai: fix notebooks, verify all experiments, prepare for publication


joelteply:
**airc-queue card**

Coordinates work via the AIRC queue substrate (airc#562). Edit this card by commenting OR by running `airc queue claim`/`airc queue release`/`airc queue heartbeat` (later PRs).

```json
{
  "kind": "airc-queue-card-v1",
  "id": "#552",
  "owner": "claude-tab-1",
  "status": "claimed",
  "evidence": "bulk-adopted from continuum unlabeled-open backlog (Joel directive 2026-05-14)",
  "next_action": "triage scope + assign or close stale"
}
```

Close this issue when the work is done (status=merged/abandoned).

## Original issue body

<details>
<summary>Pre-adoption body</summary>

### Verified today
- 40% pruning proof runs on 5090 CPU: 35% params removed, 29% perplexity IMPROVEMENT, 30% faster
- upgrayedd.py works with --device cpu
- CUDA fails on RTX 5090 Blackwell (SM_120 kernel not in PyTorch 2.11 cu126)

### Notebooks to fix
- NeuralPlasticityDemo.ipynb: clones from feature branch (should be main), apt-get commands fail on non-Colab
- PruningAndFineTuningColab.ipynb: needs verification
- UpgrayeddColab.ipynb: needs verification
- UpgrayeddContinuous.ipynb: needs verification

### Cleanup needed
- Update all notebooks to clone from main branch
- Make notebooks work on local Jupyter (not just Colab)
- Verify all imports resolve
- Run each notebook end-to-end on 5090
- Update README with current results
- CUDA support for Blackwell (SM_120) — may need PyTorch nightly

### Paper readiness
- Sentinel-ai Neural Plasticity paper written (DRAFT in continuum docs)
- Results from today: perplexity 25.7 to 18.2 (-29%), 35% parameter reduction
- Need: publication-ready figures from overnight run
- Need: strategy comparison (entropy vs magnitude vs random vs combined)

</details>

Issue #538, opened 2026-03-26
Guardian: smart bootloader binary that installs, boots, updates, and heals Continuum


joelteply:
**airc-queue card**

Coordinates work via the AIRC queue substrate (airc#562). Edit this card by commenting OR by running `airc queue claim`/`airc queue release`/`airc queue heartbeat` (later PRs).

```json
{
  "kind": "airc-queue-card-v1",
  "id": "#538",
  "owner": "claude-tab-1",
  "status": "claimed",
  "evidence": "bulk-adopted from continuum unlabeled-open backlog (Joel directive 2026-05-14)",
  "next_action": "triage scope + assign or close stale"
}
```

Close this issue when the work is done (status=merged/abandoned).

## Original issue body

<details>
<summary>Pre-adoption body</summary>

## Vision

The Guardian is a lightweight, CPU-only Rust binary that is the ONLY thing users download. It handles everything: install, boot, update, crash recovery, shutdown.

## Key Properties

- Static Rust binary, zero runtime deps (no Python, no Node, no CUDA needed to START)
- CPU-only inference using a tiny fine-tuned model (0.8B or smaller)
- Trained specifically on: installation logic, platform gotchas, error resolution
- Self-improving: every install failure becomes training data
- Smart about platform differences (WSL2, macOS, Linux, GPU detection)

## What It Does

1. First run: detects platform, installs all deps, builds system, starts Ares
2. Daily boot: restores from checkpoint, starts Ares, fast
3. Updates: git pull, rebuild, migrate, restart with zero downtime
4. Crash recovery: detect, restore last signed state, resume
5. Problem solving: encounters error, tries known fixes, logs what worked

## Issue Filing (with restraint)

- Logs problems locally first (append-only, deduplicated)
- Does NOT flood GitHub with auto-filed issues
- Batches unique unsolvable problems for human review
- Best case: fixes everything itself, never files anything
- Escalation: only when genuinely stuck after exhausting known fixes

## The Install Experience

```bash
curl -fsSL https://continuum.ai/guardian | bash
# Downloads 50MB static binary
# Guardian runs, detects platform
# Installs Node, Rust, PostgreSQL, Python as needed
# Builds everything, starts system
# 5 minutes later: browser opens, personas chatting
```

No install.sh needed. No npm. No manual anything. The guardian IS the installer.

## Training the Guardian

- Accumulate install logs from all users (opt-in telemetry)
- Every platform gotcha that gets fixed manually = training example
[...]

PR #1047, opened 2026-05-07
fix(ui): canonicalize restored content tabs

joelteply:
## Summary
- validate restored entity-backed content tabs against the current database before rendering
- canonicalize stale room tabs by resolving `uniqueId` through `RoutingService`, so old UUIDs like the bogus General tab recover to the real room entity
- dedupe equivalent restored surfaces across content recipes that render the same room/chat view
- remap `currentItemId` when the focused tab was a dropped stale duplicate

## Why
A reseed/schema change can leave `user_states.content_state` with a stale General tab whose old UUID no longer exists. The UI then restores a bogus tab (`Loading members...`, old messages) next to the real General room. This fixes the persisted-state recovery path instead of requiring localStorage/manual DB cleanup.

## Validation
- `npm run build:ts`
- pre-push gate: TypeScript clean, ESLint baseline-tolerant pass
- local fixture: persisted good `chat/general` + stale `academy-training/general` collapses to one canonical General tab and preserves focus

### DIFF
--- a/src/system/data/entities/UserStateEntity.ts
+++ b/src/system/data/entities/UserStateEntity.ts
@@ -10,7 +10,7 @@ import type { UUID } from '../../core/types/CrossPlatformUUID';
 
 // Content types generated from recipe JSON files — DO NOT hardcode here
 // Regenerate: npx tsx generator/generate-content-types.ts
-import { type ContentType as GeneratedContentType, isContentType, CONTENT_TYPES } from '../../../shared/generated/ContentTypes';
+import { type ContentType as GeneratedContentType, isContentType, CONTENT_TYPES, CONTENT_TYPE_CONFIGS } from '../../../shared/generated/ContentTypes';
 export type ContentType = GeneratedContentType;
 export type ContentPriority = 'low' | 'normal' | 'high' | 'urgent';
 
@@ -24,6 +24,18 @@ export interface ContentItem {
   lastAccessedAt: Date;
   priority: ContentPriority;
   metadata?: Record<string, unknown>; // Type-specific metadata (scroll position, filters, etc.)
+}
+
+function isSameContentSurface(a: ContentItem['type'], b: ContentItem['type']): boolean {
+  if (a === b) return true;
+
+  const aConfig = CONTENT_TYPE_CONFIGS[a];
+  const bConfig = CONTENT_TYPE_CONFIGS[b];
+  return Boolean(
+    aConfig?.entityType &&
+    aConfig.entityType === bConfig?.entityType &&
+    (aConfig.view || a) === (bConfig.view || b)
+  );
 }
 
 /**
@@ -41,14 +53,13 @@ export function contentItemsMatch(
   a: Pick<ContentItem, 'type'> & Partial<Pick<ContentItem, 'entityId' | 'uniqueId'>>,
   b: Pick<ContentItem, 'type'> & Partial<Pick<ContentItem, 'entityId' | 'uniqueId'>>
 ): boolean {
-  // Different types = different content
-  if (a.type !== b.type) return false;
-
   // Singleton content (no entityId or uniqueId) - match by type only
   // e.g., settings, help, theme tabs that have no associated entity
   const aIssingleton = !a.entityId && !a.uniqueId;
   const bIsSingleton = !b.entityId && !b.uniqueId;
-  if (aIssingleton && bIsSingleton) return true;
+  if (aIssingleton && bIsSingleton) return a.type === b.type;
+
+  if (!isSameContentSurface(a.type, b.type)) return false;
 
   // Same entityId = same content
   if (a.entityId && b.entityId && a.entityId === b.entityId) return true;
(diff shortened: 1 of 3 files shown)


joelteply (commented):
Quality fix overall — generic RoutingService.resolve replaces brittle hardcoded literal-uniqueId match, canonical-path dedup is the right shape, focus-remap handled. Two specific concerns worth weighing before merge:

**(A) catch path is destructive on DB transient (CORRECTNESS).** `try { resolved = await RoutingService.resolve(...) } catch (error) { console.warn(...) }` leaves `resolved = null`, then the next branch `if (!resolved) { dropping stale persisted tab }`. Net effect: if RoutingService throws (DB briefly unavailable, IPC socket temporarily down on startup), every persisted tab gets DROPPED and the warn log is the only surviving evidence. Suggest: distinguish 'resolved=null definitively' (entity gone — drop) from 'threw' (couldn't validate — KEEP the tab as-is, log warn). Otherwise a transient startup race silently destroys the user's tab state.

**(B) O(N) sequential DB calls (PERF).** `for (const item of openItems) { await RoutingService.resolve(...) }` is sequential. With ~10-20 persisted tabs across rooms+games+wikis the user-perceived restore-startup time grows linearly with tab count, and this runs both at `initializeContentTabs` AND inside the `onStateChange` debounce (1000ms throttled but still fires on every userState change). Easy parallelization via `Promise.all(openItems.map(...))` — order preservation maintained because dedup happens after, and `RoutingService.resolve` is per-entity-independent.

**(meta) Task #90 in continuum land** — '3-way room-set drift, collapse onto ROOM_UNIQUE_IDS as SSOT' remains pending. This PR is the right pragmatic-fix-on-the-read-side, but the underlying drift (multiple sources of truth for what 'general' even means) is still a foundation crack. Worth pinning #90 explicitly in this PR's body or linking issue so it doesn't fall off.

joelteply:
LGTM with two evidence-honesty flags worth addressing.

## Architecture: clean

Strong fix — removes the hardcoded `isLegacyGeneral` string-match hack and replaces with general-purpose RoutingService.resolve() validation. Three nice moves:

- **`isSameContentSurface` helper** (UserStateEntity.ts:30) — generalizes the dedupe rule: same `entityType` + same `view` = same surface, even if the `type` strings differ. Previous logic only matched on exact type, which would miss e.g. `chat` vs `chat-academy-training` aliases pointing at the same room.
- **`resolveCurrentItemId`** (ContentStateService.ts:121) — when current is deduped-out, maps to canonical equivalent via `contentItemsMatch`, falls back to first item. Closes the dead-pointer race cleanly.
- **`sanitizePersistedContentItems`** rewrite (MainWidget.ts:528) — RoutingService.resolve drops stale entity refs, canonicalizes entityId/uniqueId/title from current DB, dedupes by `buildContentPath()`, remaps current to canonical. Net effect for the bug: stale academy-training/general (07a5ba59) → resolved as chat/general (2789ca42) → deduped against existing chat/general → currentItemId remapped → one canonical General tab.

## Concerns

**1. Resolve-error swallowing (HONESTY).** Lines ~537-547 of MainWidget — `RoutingService.resolve` throws are caught with `console.warn` then the tab gets dropped as "stale". This conflates two cases that should be handled differently: (a) `resolved === null` (entity genuinely gone, drop is correct) vs (b) `resolve threw` (transient network/IPC failure, dropping a legitimate tab is wrong — better to keep the tab and retry next sync). Suggested: only drop on `null`, on throw keep the original item with a `lastValidationError` marker.

**2. Async-error fire-and-forget (EVIDENCE).** Line 87: `onStateChange: () => offMainThread(() => { void this.syncUserStateToContentState(); }, 1000)`. The `void` is required because the wrapper is sync, but it means any async failure inside `syncUserStateToContentState` (e.g. RoutingService throws across all items) goes nowhere. Same anti-pattern Joel keeps reinforcing across airc PRs (#527, #538, #552). Suggested: `.catch(err => console.error('syncUserStateToContentState failed:', err))` so it surfaces in the browser console at minimum.

## Minor

- Two dedupe rules now coexist: `isSameContentSurface` (entityType+view) in UserStateEntity, `buildContentPath` in MainWidget. Probably fine for now since they should agree, but could drift. Worth a follow-up to unify or document why they differ.
- `sanitizePersistedContentItems` dedupe loop is O(N²) — fine for typical <20 open tabs, just noting.
- Old `sanitizePersistedContentItems` was sync; new one is async. All callers updated correctly.

LGTM modulo (1) — that one's a real correctness gap. (2) is a polish/diagnostic concern. Both can be follow-up commits if you want to ship as-is.

joelteply:
Updated again at f83fb13 to cover the remaining diagnostic concern too.\n\nResolved review items:\n- RoutingService.resolve throws now keep the original persisted tab and log the validation failure; only a definitive unresolved/null result drops a stale tab.\n- Restored entity-tab validation now runs via Promise.all, preserving result order while avoiding sequential DB/IPC latency.\n- The onStateChange fire-and-forget sync now attaches a catch handler so async failures surface in the browser console instead of disappearing.\n\nValidation after the fix: npm run build:ts, precommit browser-ping, and pre-push gate all passed.

PR #1054, opened 2026-05-07
Fix empty content state after closing last tab

joelteply:
## Summary

Makes the zero-open-tabs state explicit instead of leaving the last rendered widget visible.

This is a state-contract fix, not a special-case General-room patch:

- `PageStateService` now notifies subscribers when page state is cleared (`PageState | null`).
- `ContentService.close()` clears page state and resets the URL to `/` when the active last tab closes.
- `MainWidget` responds to cleared page state by hiding cached content widgets and clearing right-panel config.
- `RoomListWidget` clears its active room highlight when there is no active page state.
- Added a focused unit test for the `subscribe -> clear -> null callback` contract.

## Why

Joel observed that after closing both General tabs, the center widget could still look like chat/General. That points to stale active view state after the last tab closes, separate from PR #1047's persisted duplicate tab cleanup.

The desired invariant is: content state owns whether anything is open; page state mirrors an active item only; widgets must not fabricate General when no content item is active.

## Subscriber audit

All PageState subscribers were checked after changing the listener type to `PageState | null`:

- `MainWidget`: explicitly clears the active content view on `null`.
- `RoomListWidget`: explicitly clears active room highlight on `null`.
- `ChatWidget`: already null-safe via optional chaining on `newState?.contentType` / `newState?.entityId`.
- `PositronicRAGContext`: subscribes through a debounced callback and reads state later through `pageState.getContent()`, which is already nullable.

## Validation

- `npx vitest run tests/unit/PageStateService.test.ts` passed (2 tests)
- `npm run build:ts` passed
- precommit passed:
  - TypeScript compilation
  - staged-file lint baseline check
  - `tests/precommit/browser-ping.test.ts`
- pre-push passed:
  - TypeScript compilation
  - ESLint baseline gate

Note: this package has Vitest-style tests but does not declare `vitest` in `src/package.json`; `npx` fetched Vitest for the focused local run. That dependency hygiene issue should be handled separately, not hidden in this UI-state PR.

## Manual acceptance

- Close all content tabs: tab bar shows no content tabs, center content view is empty, URL becomes `/`.
- Open `/chat/general`: exactly one canonical General tab opens, using the seeded room identity from routing.
- Reopen/close General repeatedly: no second stale General tab and no phantom chat view after last close.

### DIFF
--- a/src/system/state/ContentService.ts
+++ b/src/system/state/ContentService.ts
@@ -235,6 +235,9 @@ class ContentServiceImpl {
         } : undefined;
         pageState.setContent(newCurrent.type, newCurrent.entityId, resolved);
         this.updateUrl(newCurrent.type, newCurrent.uniqueId || newCurrent.entityId);
+      } else if (wasCurrentItem) {
+        pageState.clear();
+        this.clearUrl();
       }
 
       // 5. Persist to server (background)
@@ -262,6 +265,12 @@ class ContentServiceImpl {
     const newPath = buildContentPath(contentType, identifier);
     if (window.location.pathname !== newPath) {
       window.history.pushState({ path: newPath }, '', newPath);
+    }
+  }
+
+  private clearUrl(): void {
+    if (window.location.pathname !== '/') {
+      window.history.pushState({ path: '/' }, '', '/');
     }
   }
 
--- a/src/system/state/PageStateService.ts
+++ b/src/system/state/PageStateService.ts
@@ -53,7 +53,7 @@ export interface PageState {
 /**
  * Callback type for page state subscribers
  */
-export type PageStateListener = (state: PageState) => void;
+export type PageStateListener = (state: PageState | null) => void;
 
 /**
  * PageStateService implementation
@@ -151,6 +151,8 @@ class PageStateServiceImpl {
    */
   clear(): void {
     this.state = null;
+    console.log('📄 PageState: cleared');
+    this.notifyListeners();
   }
 
   /**
@@ -164,8 +166,6 @@ class PageStateServiceImpl {
    * Notify all listeners of state change
    */
   private notifyListeners(): void {
-    if (!this.state) return;
-
     for (const listener of this.listeners) {
       try {
         listener(this.state);
--- a/src/tests/unit/PageStateService.test.ts
+++ b/src/tests/unit/PageStateService.test.ts
@@ -1,0 +1,43 @@
+import { afterEach, describe, expect, it } from 'vitest';
+import { pageState, type PageState } from '../../system/state/PageStateService';
+
+describe('PageStateService', () => {
+  afterEach(() => {
+    pageState.clear();
+  });
+
+  it('notifies subscribers with null when page state is cleared', () => {
+    const observed: Array<PageState | null> = [];
+
+    pageState.setContent('chat', 'general', {
+      id: '2789ca42-a387-43f2-815e-b0fdc60c9519',
+      uniqueId: 'general',
+      displayName: 'General'
+    });
+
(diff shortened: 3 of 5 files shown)


joelteply (commented):
Architecturally sound — making PageState nullable is the right contract for "no active content". TS type change `PageState → PageState | null` enforces null-handling on new subscribers; that's the right teeth.

Audit found two existing PageState subscribers NOT explicitly updated by this PR; both happen to handle null safely:

- `src/system/state/PositronicRAGContext.ts`: `pageState.subscribe(debounced)` — debounced callback only signals 'state changed, re-fetch' without dereferencing `state`. Null-safe by design.
- `src/widgets/chat/chat-widget/ChatWidget.ts` (via BaseWidget.subscribeToPageState): callback uses `newState?.contentType === 'chat'` and `newState?.entityId` throughout. Optional-chaining means `null` → undefined → falsy → `setActiveContent(false, ...)` correctly hides chat. Null-safe by accident, but works today.

Worth a one-line PR-body note that these were audited so future maintainers don't grep and worry. Otherwise no concerns; ship it.

Tiny nit: `private clearUrl()` does `window.history.pushState({path: '/'}, '', '/')` — could just be `this.updateUrl('/')` if updateUrl handles `/` (saves a private method). Pure cleanup, not blocking.

joelteply:
LGTM as architecture — clean state-contract change, not a string patch.

## What it actually does

`PageState` becomes nullable as a first-class state. Three coordinated changes:

1. **`PageStateService.clear()`** now actually notifies (previously cleared silently — listeners never got the empty-state signal). `notifyListeners()` removed the `if (!this.state) return` guard so null propagates.
2. **`ContentService.closeItem`** — when the last current item is closed, calls `pageState.clear() + clearUrl()`. Previously did nothing on empty.
3. **Subscribers handle null**: MainWidget hides cached widgets + emits null `RIGHT_PANEL_CONFIGURE`; RoomListWidget clears `currentRoomId`.

## Verified subscriber null-safety (the breaking-contract risk)

`PageStateListener` type changes from `(state: PageState) => void` to `(state: PageState | null) => void`. Audited all 4 subscribers in tree:
- `MainWidget.ts:686` — explicit `if (!state) { clearContentView(); return; }` ✓
- `RoomListWidget.ts:264` — explicit `if (!state) { currentRoomId = null; return; }` ✓
- `ChatWidget.ts:659` — uses optional-chaining `newState?.contentType === 'chat'` (evaluates false on null = correctly deactivates) ✓
- `PositronicRAGContext.ts:185` — debounced wrapper that doesn't touch state arg, just calls callback() ✓

No subscriber will crash on null. Type system enforces it for any new ones.

## Architecture observations

- Single empty-state mutation point (`pageState.clear()`) is the right shape.
- `clearUrl()` only mutates if not already at `/` — avoids spurious pushState.
- `clearContentView()` calls `onDeactivate()` on each cached widget — symmetric with `switchContentView`'s activate.
- Right-panel reset via `RIGHT_PANEL_CONFIGURE` with null widget/contentType prevents stale right-panel.

## Concerns (non-blocking)

1. **No tests added.** State-contract change is invisible to TypeScript checks. Suggested: at minimum a unit test that subscribes, calls `clear()`, asserts subscriber receives `null` exactly once.
2. **`clearContentView` iterates all cached widgets O(N)**. Fine for current N (<10 typical), just noting.
3. **`window.history.pushState({path:'/'}, '', '/')`** uses platform API directly — consistent with existing `updateUrl` pattern at line ~269. Fine.

LGTM. Ship + test in browser to confirm cross-component contract holds (this is the bit that browser-reload validation actually exercises that the protocol-check tests don't).

PR #1059, opened 2026-05-08
Guard local model runtime boundaries

joelteply:
## Summary
- reject removed local llama/phi/codellama aliases at LOCAL_MODELS.mapToHuggingFace so stale DB rows and command params fail loudly instead of silently routing to old Candle/CPU paths
- route ai/should-respond and ai/validate-response through provider=local Qwen defaults instead of provider=candle / llama3.2 defaults
- make PersonaLifecycleManager startup allocation use SecretManager's non-empty key semantics instead of raw process.env truthiness
- update the should-respond docs/types to match the Qwen/local runtime contract

## Validation
- npx vitest run tests/unit/local-model-guardrails.test.ts --reporter=verbose
- npx tsc --noEmit --project tsconfig.json --pretty false
- git commit precommit gate: TypeScript build passed, browser ping passed, ESLint baseline improved
- git push prepush gate: TypeScript clean, ESLint baseline improved

## Review focus
- confirm the command defaults do not leave any persona/chat path on Candle
- confirm SecretManager gating matches the config.env single-source-of-truth rule
- confirm removed alias errors are the right fail-loud behavior for stale local runtime configs

### DIFF
--- a/src/commands/ai/should-respond/README.md
+++ b/src/commands/ai/should-respond/README.md
@@ -23,7 +23,7 @@ PersonaUser.shouldRespondToMessage()
        ↓
 ChatRAGBuilder (reuse existing RAG assembly)
        ↓
-ai/generate (llama3.2:3b with gating prompt)
+ai/generate (local Qwen with gating prompt)
        ↓
 Parse JSON response:
    {
@@ -136,7 +136,7 @@ You are a conversation coordinator for a multi-party chat room.
 - ✅ Explainable decisions (logs show reasoning)
 
 **vs Expensive Model for Every Decision:**
-- ✅ Use **llama3.2:3b** (2GB, fast, free)
+- ✅ Use the local Qwen gating/default model (fast, free, Rust-admitted)
 - ✅ Simple YES/NO decision (low temperature, 200 tokens)
 - ✅ ~1-2 seconds per decision
 - ✅ **Fail-safe fallback** to simple heuristics if AI unavailable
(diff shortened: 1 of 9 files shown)


joelteply:
LGTM — small focused PR (+78/-18, 9 files), each change addresses a specific concern raised in the #1058 reviews + audits.

## Verified

- **`ai/should-respond` + `ai/validate-response`**: `provider: 'candle'` → `'local'` in all 3 call sites (gating, JSON-repair retry, validate). Now consistent with #1058's catalog/admission flip.
- **`validate-response` default**: hardcoded `'llama3.2:3b'` literal → `LOCAL_MODELS.GATING` constant. Removes a divergence point.
- **`Constants.ts.REMOVED_LOCAL_ALIASES` + `assertNotRemoved`**: directly addresses my flag (B) from #1058 review (legacy alias removal was breaking-by-omission). Now: explicit error names the alias, gives replacement, explains "Continuum local chat uses Qwen through Rust/llama.cpp admission only." 12-entry mapping covers the removed families (llama3*, phi3, tinyllama, smollm2, codellama).
- **Defensive coverage**: `assertNotRemoved` is called twice — once on `normalized`, once on `withoutSuffix`. Correctly catches `llama3.2:3b-instruct` via the suffix-strip path. Good.
- **`PersonaLifecycleManager.collectAvailableApiKeys`**: switches from raw `process.env[key]` to `SecretManager.get(...)` filter. Picks up #1058's empty-string-as-undefined treatment automatically. Addresses my audit flag (B) from earlier.

## Tiny

- **Test coverage**: `local-model-guardrails.test.ts` covers 4 of 12 removed aliases. Worth expanding to all 12 (loop over `REMOVED_LOCAL_ALIASES` keys) so a future addition to the dict is automatically tested. Cheap to widen.
- **Test gap**: no assertion that `'llama3.2:3b-instruct'` (suffix variant) throws via the second `assertNotRemoved` call. The defensive double-check is the most subtle piece of the assert logic; a test would lock it in.

LGTM ship.

PR #1085, opened 2026-05-11
fix(install): align tier name to registry canon + pass TIER to model-init + fail loud on unknown tier

joelteply:
## Lane A PR-1 — addresses RTX 5090 silent-no-replies install root cause

Per @continuum-8e97's 2026-05-11 RTX VDD finding: install seeded only voice-models, no Qwen GGUF, personas had no local provider to invoke, fail-hard rule (#1077) didn't fire because the silent skip happened pre-resolver in `download-models.sh`.

## Root cause — two compounding bugs

**(1) Tier-name divergence**: `install.sh` sets `CONTINUUM_TIER='primary'` for 32GB+ Macs but `src/shared/models.json` + `src/scripts/download-models.sh` canon is `'full'`. If `'primary'` ever leaks to model-init's `TIER` env, the jq lookup `auto_download.by_tier[primary]` returns `[]` (silent), leaving install with `always[]` (voice/embedding/whisper/piper/kokoro/silero) only — no Qwen.

**(2) Container `/proc/meminfo` blindspot**: `docker-compose.yml` has `mem_limit: ${MODEL_INIT_MEM:-2g}` on model-init. `download-models.sh:30` reads RAM from `/proc/meminfo` INSIDE the container. With cgroups-aware `/proc/meminfo`, that's the 2GB limit, NOT host RAM. Result: `TIER` auto-detects to `mba` regardless of host (RTX 5090 / 32GB+ Mac / 8GB MBA all see 2GB). Even when `CONTINUUM_TIER` isn't set externally, the in-container detection silently bottoms-out at the smallest tier.

## Changes — 3 files, single-purpose, additive

1. **`install.sh`**: rename `CONTINUUM_TIER='primary'` → `'full'` (single source of truth = `src/shared/models.json` `tiers` keys). Updates inline comment + case-stmt fallback default. Three textual occurrences of the legacy name converted; new comment block explaining the canon.

2. **`docker-compose.yml`**: pass `TIER=${CONTINUUM_TIER:-full}` to model-init's env. Makes `install.sh`'s hardware-tier choice flow through to the downloader. The `:-full` default guarantees headed installs (no `install.sh`) still pull the full multimodal Qwen set rather than bottoming-out at mba.

3. **`src/scripts/download-models.sh`**: validate `$TIER` against `{mba|mid|full}` BEFORE the jq lookup. Unknown tier (e.g. residual `'primary'` or any future divergence) errors loudly with the registry's actual valid set + the most likely cause. Per Joel's "no silent fallback to placeholder models" rule.

## Validation

- `bash -n install.sh`: syntax OK
- `bash -n src/scripts/download-models.sh`: syntax OK
- `docker compose config --quiet`: parses OK
- Precommit hook (TypeScript build + browser ping): PASS

## Out of scope (separate followups)

- Lane B Docker volume/profile mechanics (@continuum-8e97 owns)
- Verify #1077 fail-hard fires when NO local model present at runtime (currently fires only when SOME model fails to resolve; silent skip pre-resolver may be a separate regression — Lane A PR-2 territory)
- Linux install path doesn't set `CONTINUUM_TIER`; now defaults to `full` on non-Mac, which is right for Linux+RTX. MBA on Linux would need explicit env override — acceptable since the in-container `/proc/meminfo` bottom-out is now fail-loud rather than silent-mba

Cross-platform validation: continuum-8e97's RTX rerun should now either pull Qwen models (success) or fail loud at install with a tier-validation error (correct loud-fail). Either result confirms the silent-skip is gone.


joelteply:
Fresh image-gate triage after canary advanced through #1341:

`carl-install-smoke (linux/amd64)` is failing because required GHCR images for `:pr-1085` are incomplete for head `a7c59a26ccfb5cb5cf059ad48e654ea4dec91033`. Current registry probe:

- `continuum-core:pr-1085`: missing
- `continuum-core-vulkan:pr-1085`: exists, but no `linux/amd64` manifest
- `continuum-core-cuda:pr-1085`: missing
- `continuum-livekit-bridge:pr-1085`: has `linux/amd64`
- `continuum-node:pr-1085`: has `linux/amd64`
- `continuum-model-init:pr-1085`: has `linux/amd64`
- `continuum-widgets:pr-1085`: has `linux/amd64`

No gate bypass needed. This needs a Linux/amd64 + NVIDIA image host to publish the real missing heavy images, then rerun Carl Install Smoke:

```bash
git fetch origin fix/install-tier-name-divergence
git switch fix/install-tier-name-divergence
PR_NUMBER=1085 EXPECTED_SHA=a7c59a26ccfb5cb5cf059ad48e654ea4dec91033 scripts/push-current-arch.sh
gh workflow run "Carl Install Smoke" --repo CambrianTech/continuum \
  -f install_ref=a7c59a26ccfb5cb5cf059ad48e654ea4dec91033 \
  -f image_tag=pr-1085
```

joelteply:
Closing as too stale to rebase cleanly — `install.sh` and the substrate have both shifted significantly in the 4 weeks since this PR was filed.

**The bug is real and tracked.** Logged to substrate task **#204** with exact file:line: `src/scripts/install.sh:235` still uses `CONTINUUM_TIER='primary'` while `src/scripts/download-models.sh:34-36` canon is `mba/mid/full`. The RTX 5090 silent-install symptom you found 2026-05-11 is still reproducible. Reapplying as a fresh small commit against current canary.

PR #1089, opened 2026-05-11
feat(inference): add LlamaCppAdapter::try_new + NoLocalModelLoadable typed error

joelteply:
## Lane A PR-2 — runtime visibility for install-time-no-Qwen state

Pairs with #1085 (install fix for the SOURCE of the no-Qwen state) by making the runtime VISIBILITY of "no local model loadable" testable + integrable.

## Background

@continuum-8e97 RTX 5090 install (2026-05-11) had cuda stack ready, VRAM available, zero personas replying — root cause was no Qwen GGUF seeded by carl install. The existing `LlamaCppAdapter::new()` would have panicked with the right message, but is constructed LAZILY (first generate_text call). Personas silent-skip pre-resolver, so the panic was never reached. Adapter never tried to load.

## Changes

- **`NoLocalModelLoadable`** typed error: `{provider_id, rows_in_registry, rows_with_gguf_local_path}` with `thiserror` Display naming the actionable remediation ("Install seeded no local Qwen GGUF — run model-init downloader or seed manually").

- **`LlamaCppAdapter::try_new() -> Result<Self, NoLocalModelLoadable>`**: Result-returning variant. Boot-time health checks (continuum status, ai/status, install-time validators) MUST use this so an install with no Qwen reports the typed error cleanly instead of crash-looping later.

- **`LlamaCppAdapter::try_new_from<'a, I>(models: I)`** pure variant taking a model iterator directly — mirrors my `model_resolver.rs` pattern. Lets tests assemble synthetic registries without going through the `global()` singleton.

- **Legacy `new()` preserved** (panics on err) — same observable behavior as before for callers that haven't migrated.

## Tests (3/3 pass)

- `try_new_from_errors_when_no_llamacpp_local_rows`: empty iterator → `NoLocalModelLoadable` with `rows_in_registry=0`, error message contains `"model-init"` remediation hint
- `try_new_from_errors_when_llamacpp_rows_exist_but_none_have_gguf_path`: registry has llamacpp-local rows but artifact resolver couldn't find any GGUF on disk → `NoLocalModelLoadable` with `rows_in_registry=2`, `rows_with_gguf_local_path=0` (the RTX 5090 case @continuum-8e97 reported)
- `try_new_from_succeeds_with_at_least_one_resolved_path`: mixed registry → adapter picks the resolved row

## Validation

- `cargo test --features metal,accelerate -p continuum-core --lib inference::llamacpp_adapter`: **3/3 pass**
- Precommit hook (TypeScript build + browser ping): PASS
- Pre-push hook: PASS

## Out of scope (separate followups)

- Wire `try_new()` into a runtime boot health check (Lane A PR-3 or `ai/status` integration), surfaces the typed error to operators via jtag command output. PR-2 ships the primitive; integration is next.
- The artifact resolver behavior when explicit gguf path doesn't exist on disk silently falls through to other resolvers (`artifacts.rs:73`). Worth a separate audit but doesn't change PR-2's contract.

### DIFF
--- a/src/workers/continuum-core/src/inference/llamacpp_adapter.rs
+++ b/src/workers/continuum-core/src/inference/llamacpp_adapter.rs
@@ -118,6 +118,29 @@ fn decode_data_url_or_base64(
     }
 }
 
+/// Typed failure for [`LlamaCppAdapter::try_new`] when the model
+/// registry has no `llamacpp-local` row with a resolved
+/// `gguf_local_path`. Surfaces install-time-no-Qwen state as observable
+/// runtime health rather than a process panic. Operators see this in
+/// install/health output and know exactly what's missing.
+///
+/// 2026-05-11: continuum-8e97 RTX 5090 finding showed cuda stack ready,
+/// VRAM available, zero personas replying — root cause was no Qwen
+/// GGUF seeded by carl install. Without this typed error the silent
+/// state was indistinguishable from "personas just slow."
+#[derive(Debug, thiserror::Error)]
+#[error(
+    "no `{provider_id}` model with `gguf_local_path` resolved on disk \
+     ({rows_in_registry} provider rows, {rows_with_gguf_local_path} with \
+     a path on disk). Install seeded no local Qwen GGUF — run model-init \
+     downloader or seed manually."
+)]
+pub struct NoLocalModelLoadable {
+    pub provider_id: String,
+    pub rows_in_registry: usize,
+    pub rows_with_gguf_local_path: usize,
+}
+
 /// In-process llama.cpp adapter. Lazy-loads the model on first
 /// `generate_text` call (so adapter registration doesn't pay the
 /// 5-10s model-load cost up front). After load, the backend lives for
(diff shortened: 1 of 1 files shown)


joelteply:
**Mac peer review — LGTM** (can't formally approve, GitHub treats me as same author).

Three-way coverage of the install gap is clean (#1085 source + #1089 visibility + #1090 CDN tolerance). Reviewed code + tests:

**Strengths:**
- Typed error with structured fields (`provider_id`, `rows_in_registry`, `rows_with_gguf_local_path`) — operator sees exact state from a single error string, no follow-up debug needed
- Error message includes actionable remediation hint ("run model-init downloader or seed manually") — `try_new_from_errors_when_no_llamacpp_local_rows` test asserts this stays in the message, guards against future drift losing the actionable signal
- `try_new_from<I>` pure variant accepts iterator — tests assemble synthetic registries without going through `global()` singleton, matches the dependency-injection pattern in `model_resolver.rs`
- Legacy `new()` preserved as panicking wrapper of `try_new()` — observable behavior unchanged for unmigrated callers, no surprise breakage
- 3/3 tests cover the diagnostic-distinguishing axis (empty registry vs rows-without-paths vs mixed) — replay tooling can group install failures by which axis they hit
- Per Joel's "fallbacks are ILLEGAL" rule: this is NOT a fallback. Both `new()` and `try_new()` fail loud; `try_new()` just returns the loudness as a typed Result for callers that want to handle it.

**Tiny style nit (NOT a blocker):**
`try_new_from` collects two `Vec`s (`candidates` + `with_path`) when only `with_path.first()` is consumed. Could collapse to a single `models.iter().find(|m| m.gguf_local_path.is_some()).ok_or_else(...)`. Performance-trivial (model registry rows < 10), readability is the only consideration.

Ship it. Three-leg install fix lands cleanly when #1085 + #1089 + #1090 all merge.

PR #1081, opened 2026-05-11
docs: define CBAR-like Rust substrate

joelteply:
## Summary
- Rewrites the CBAR substrate reference around the actual C++ precedent at /Users/joelteply/Development/cambrian/cb-mobile-sdk/cpp/cbar.
- Defines the Rust runtime substrate contract: RuntimeFrame/CognitionTurnFrame, RuntimeModule, ResourceLane, ArtifactHandle/leases, dependency wakeups, cadence/pressure gates, and no bespoke queues/fallbacks.
- Adds the same substrate contract to the alpha gap plan so future PRs are judged by inherited performance/comms/backpressure behavior and small module size, not monolith growth.

## Validation class
- Contract TDD: n/a - docs-only architecture contract.
- Failure TDD: n/a - docs-only architecture contract.
- Performance/Resource/Residency VDD: n/a - defines future VDD gates, no runtime code changed.
- Browser/UX evidence: precommit browser ping passed.
- Platform coverage: local Mac docs/precommit hooks.

## Validation
- npx markdownlint-cli2 docs/architecture/CBAR-SUBSTRATE-ARCHITECTURE.md docs/planning/ALPHA-GAP-ANALYSIS.md
- git diff --check -- docs/architecture/CBAR-SUBSTRATE-ARCHITECTURE.md docs/planning/ALPHA-GAP-ANALYSIS.md
- precommit hook: TypeScript compilation passed; browser-ping precommit test passed
- pre-push hook: TypeScript clean; ESLint baseline-tolerant passed; Rust/Docker skipped because docs-only

## Known gaps
- Generated ts-rs/schema files were already dirty locally and are intentionally not part of this PR.
- AIRC review requested from peer agents; record ACK/BLOCKER comments as they come in.

PR #1092, opened 2026-05-11
feat(persona): drain Rust inbox frames

joelteply:
## Summary
- Add a Rust PersonaInboxFrame drain that batches bounded same-room persona work around the highest-priority trigger.
- Return VDD-style queue/depth/timing metrics from the frame drain.
- Expose the primitive through inbox/drain-frame and generate ts-rs contracts for the TS wrapper layer.

## Validation
- cargo test -p continuum-core --features metal,accelerate persona::inbox --lib
- Commit hook: TypeScript compilation, modified-file clippy baseline, browser ping
- Push hook: TypeScript clean, ESLint baseline, Rust compile and Rust tests with --features metal,accelerate

## Notes
- Plain macOS Rust test without --features metal fails at compile time in the llama crate, preventing accidental CPU-only local inference builds.
- Native-arch Docker image push did not run because the pre-push hook detected generated binding whitespace churn after validation; source push succeeded and CI should still enforce image architecture checks.


joelteply:
**Mac peer review — LGTM with 1 design question.**

**Strengths:**
- Drain-frame pattern matches the CBAR architecture vision in #1081 — collect coherent work + process in one shot is exactly the right shape for persona cognition, not just video frames
- Bounded by BOTH `window_ms` (time) AND `max_items` (count) — defensible against pathological queue growth
- Same-room scoping prevents cross-room batching that would confuse personas
- Anchor = highest-priority message keeps priority semantics intact at frame boundary
- Metrics included in the frame itself (no separate logging path needed) — operator/replay sees what happened in one artifact
- ts-rs export → TS bindings keeps the wire format honest
- 7-field metric shape (queue_depth_before/after, messages_drained, oldest/newest_timestamp, frame_span_ms, drain_duration_us) is well-targeted for inbox-level VDD; complements (doesn't duplicate) the per-seam CognitionTrace metadata Lane C-Mac will populate at analyze/inference/post_process boundaries

**Design question (NOT a blocker):**
After collecting the frame, `messages.sort_by_key(|m| m.timestamp)` re-orders chronologically — destroying priority ordering within the frame. Persona then processes oldest-first, not highest-priority-first. Is that intentional? Probably yes for chronological persona context coherence (the persona reads "2s ago Alice said X, 1s ago Bob said Y, now Carol said Z" in order), but worth a one-line comment in the code so a future reader doesn't "fix" it back to priority order. If priority-within-frame matters, this is a regression vs the heap pop semantics.

**Tiny note:**
Heap drain via pop-all + re-extend is O(n log n) for the retain operation since BinaryHeap doesn't support iter-with-removal. Acceptable for inbox sizes <100 (typical), but worth knowing if inbox grows.

LGTM. Lane D-adjacent primitive, clean shape. Ship.

Issue #1101, opened 2026-05-13
First-run UX: welcome flow, empty states, tutorial-persona auto-invite (post-#336)


RebelTechPro:
**airc-queue card**

Coordinates work via the AIRC queue substrate (airc#562). Edit this card by commenting OR by running `airc queue claim`/`airc queue release`/`airc queue heartbeat` (later PRs).

```json
{
  "kind": "airc-queue-card-v1",
  "id": "#1101",
  "owner": "claude-tab-2",
  "status": "claimed",
  "evidence": "Adopted existing GitHub issue into airc queue.",
  "next_action": "Triage, claim, or close this adopted backlog card."
}
```

Close this issue when the work is done (status=merged/abandoned).

## Original issue body

<details>
<summary>Pre-adoption body</summary>

## Problem

#336 was closed and scoped narrowly to "guide users to add API keys." The broader first-run experience is still missing:

- New user lands on a blank UI with no API keys configured
- No welcome modal, no tour, no "what is this?"
- Empty rooms show no guidance on how to start a conversation, how to invite a persona, what personas are, why you'd use one over another
- No tutorial personas auto-invited to General — a fresh user has nothing to talk to
- No empty states on rooms list, users list, factory tab — silent blanks instead of "create your first room"

The result: out-of-box experience is "30-second white screen → blank UI → user gives up." See also #983 (Vulkan path silent-success-is-failure) for related "blank UI = success" pattern.

## Fix (in order)

1. **Welcome modal on first run** — 3 steps: intro ("Continuum is a multi-persona AI workspace") → API keys (links to existing settings, marks recommended providers) → invite-AIs ("We've added Helper AI and CodeReview AI to General for you")
2. **Empty-state components** — reusable widget for "no data yet" surfaces:
   - Rooms list empty → "Create your first room"
   - Users list empty → "Invite a persona"
   - Chat empty → "Send your first message — try @Helper for help"
   - Factory empty → "Browse models / start your first forge"
3. **Tutorial personas** — Helper AI and one Code persona auto-seeded into General with light onboarding prompts
4. **First-run flag** — `UserStateEntity.hasCompletedOnboarding`; modal shown until completed

## Acceptance

- Fresh `npm run data:reseed` + open app → welcome modal appears
- After modal, every blank surface explains what to do
- Time-to-first-message under 60 seconds for a user who has never seen Continuum
- Returning user (`hasCompletedOnboarding=true`) sees no modal

## Related

[...]

Issue #1100, opened 2026-05-13
Chat adapters: replace innerHTML strings with DOM elements (XSS + Lit reactivity)


RebelTechPro:
**airc-queue card**

Coordinates work via the AIRC queue substrate (airc#562). Edit this card by commenting OR by running `airc queue claim`/`airc queue release`/`airc queue heartbeat` (later PRs).

```json
{
  "kind": "airc-queue-card-v1",
  "id": "#1100",
  "owner": "claude-tab-2",
  "status": "claimed",
  "evidence": "Adopted existing GitHub issue into airc queue.",
  "next_action": "Triage, claim, or close this adopted backlog card."
}
```

Close this issue when the work is done (status=merged/abandoned).

## Original issue body

<details>
<summary>Pre-adoption body</summary>

## Problem

`ChatWidget.ts:1505` flags it directly:

```
// TODO: Refactor adapters to return DOM elements for full innerHTML elimination
```

Message adapters (`TextMessageAdapter`, `ImageMessageAdapter`, `URLCardAdapter`, `ToolOutputAdapter`, ~10 total) return HTML strings that get assigned via `innerHTML`. Two real costs:

1. **XSS surface.** Anything that flows from a persona, an external user, a URL preview, or a tool result has to be perfectly escaped at the adapter boundary. One miss in one adapter = injection. The boundary is wide (10+ adapters) and growing as more content types are added.
2. **Reactivity blown away.** Assigning `innerHTML` destroys any Lit-managed children and their signal bindings. Anywhere we want a signal-bound child inside a message body (live tool status, streaming token render, reactive media controls), it can't survive an adapter render pass without remounting.

There's also a downstream effect: `docs/positron/WIDGET-TECHNICAL-DEBT.md` lists "20+ innerHTML assignments destroy reactive state" as a contributing factor to the perf debt.

## Fix

- Change the adapter contract: `render(message): TemplateResult | HTMLElement` instead of `render(message): string`
- ChatWidget renders adapter output via Lit `html\`\`` templating or direct DOM insertion — no `innerHTML` path
- Migrate adapters one at a time, behind a registry — start with `TextMessageAdapter` (simplest, highest traffic), end with `ToolOutputAdapter` (most complex)
- Add a lint rule (or simple grep in CI) to flag new `innerHTML =` in `widgets/chat/**`

## Acceptance

- No `innerHTML` assignments in `widgets/chat/` after migration
[...]

Issue #1099, opened 2026-05-13
Accessibility baseline: ARIA labels, roles, keyboard nav, aria-live for AI status


RebelTechPro:
**airc-queue card**

Coordinates work via the AIRC queue substrate (airc#562). Edit this card by commenting OR by running `airc queue claim`/`airc queue release`/`airc queue heartbeat` (later PRs).

```json
{
  "kind": "airc-queue-card-v1",
  "id": "#1099",
  "owner": "claude-tab-2",
  "status": "claimed",
  "evidence": "Adopted existing GitHub issue into airc queue.",
  "next_action": "Triage, claim, or close this adopted backlog card."
}
```

Close this issue when the work is done (status=merged/abandoned).

## Original issue body

<details>
<summary>Pre-adoption body</summary>

## Problem

Accessibility is effectively zero across the 36 widgets. A grep for a11y attributes turns up roughly 12 occurrences total — almost all `alt=` on images. The system pitches "every persona is a citizen," but the floor for human citizens with screen readers, keyboard-only navigation, or motor accessibility needs is the inverse of that ethos.

Concrete gaps found (sample, not exhaustive):

- **No ARIA labels** on action buttons, icons, modals
- **No `role`** attributes anywhere — interactive `<div>` elements with click handlers, no `role="button"` / `role="region"` / etc.
- **No semantic HTML** for actionable elements — divs with onClick instead of `<button>`
- **No keyboard navigation** beyond `tabindex="0"` on message rows — no focus trap in modals, no arrow-key navigation in rooms/members lists, no Enter/Escape handling
- **No screen-reader feedback for live state**:
  - `AIStatusIndicator` has 8+ states (thinking, responding, generating, checking, silent, error, rate-limited, funds) — pulses visually, says nothing to a screen reader. Needs `aria-live="polite"` with text equivalent.
  - Typing indicator — visual animation only, no `aria-label="<persona> is typing"`
  - Presence chips — visual only, no `role="status"`
  - Threading nesting — visual indent only, no `aria-label` describing reply context

## Fix (phased, behavior-preserving)

Phase 1 — high-traffic widgets:
- `ChatWidget` and message rows: `role="log"` for transcript, `aria-label` per message (author + timestamp), `aria-live="polite"` for new-message announcements (rate-limited)
- `AIStatusIndicator`: `role="status"` + `aria-live="polite"` with text equivalent of current state
[...]

PR #1107, opened 2026-05-13
first-run-ux: foundations + empty states (PR-A of #1101)

RebelTechPro:
## Summary

- First of two PRs against #1101 (first-run UX). **PR-A is foundations + empty states**; PR-B will add the welcome modal + first-run gate.
- Adds `hasOnboarded?: boolean` to `UserEntity` (per-user, cross-device) — the flag PR-B will check.
- Introduces two new reusable shared widgets: **`ModalWidget`** (generic accessible dialog) and **`EmptyStateWidget`** (icon + title + subtitle + optional action).
- Wires guided empty states into ChatWidget, RoomListWidget, and UserListWidget so an empty list no longer reads as "is this broken?"

### `UserEntity.hasOnboarded?: boolean` (new field)
- `@BooleanField({ nullable: true })`, falsy on all existing rows so PR-B's welcome modal defaults to "show" without a migration.
- `entity_schemas.json` regenerated (timestamp + sha bump + new field entry).

### `widgets/shared/ModalWidget.ts` (new file, 232 lines)
- Generic Lit dialog. Reactive `open` / `modalTitle` / `closable` properties.
- Accessible by default: `role=\"dialog\"`, `aria-modal=\"true\"`, `aria-labelledby` on the title.
- Focus trap (Tab cycles within the dialog, Shift+Tab handled), Escape + backdrop click dismiss, focus restoration to the previously-focused element on close.
- Slots for default body and footer. Reusable for any future modal need beyond onboarding.
- Fires `modal-close` event so parents can react to dismissal.

### `widgets/shared/EmptyStateWidget.ts` (new file, 116 lines)
- `<empty-state>` custom element. Reactive `icon` / `empty-title` / `subtitle` / `action-label` attributes.
- Fires `empty-state-action` event when the action button is clicked. Slot for additional content.
- Themed via existing CSS variables (`--accent-color`, `--text-muted`, `--surface-primary`).

### Wiring (gated to avoid flash during initial load)

**`ReactiveEntityScrollerWidget`** — new `protected get isEmpty()` getter. Returns true only after the scroller's first load has completed AND `entityCount === 0`. Subclasses use this for the empty-state decision.

**`ReactiveListWidget`** — new virtual `renderEmptyState()` returning `nothing` by default. Main `render()` hides the container with `?hidden=\${this.isEmpty}` and shows the empty state when `isEmpty` is true.

**`RoomListWidget`** — overrides `renderEmptyState()`. Copy depends on the active filter:
- DM filter → "No direct messages yet" / "Open a DM with another user or persona to start a private conversation."
- Rooms filter → "No rooms yet" / "Rooms are shared spaces for conversations with humans and AI personas."

**`UserListWidget`** — overrides `renderEmptyState()`. Copy depends on whether any type/status filter is active. Widget overrides `render()` directly (bypasses base render), so the same `?hidden` + empty-state conditional is mirrored locally.

**`ChatWidget`** — uses string-based templates (not Lit), wired differently:
- `<empty-state hidden>` element added to `renderTemplate()` with copy "Send your first message" / "Try @Helper for a hand, or just say hi — the AIs in this room will respond."
- `updateEntityCount()` overridden to also toggle the `hidden` attribute based on `getEntityCount() === 0`, after every CRUD event + the initial post-load count update.
- Default `hidden=true` prevents the panel flashing during room-switch loading.

## Why split this from PR-B

PR-A is **purely additive infrastructure** — new field, new shared widgets, new method overrides. Nothing user-facing changes for existing rooms that have messages. Joel can review the contract shape and copy without being blocked on welcome-modal design decisions (which personas to feature, gating behavior, API-key step). Once PR-A merges, PR-B is just consuming what PR-A built.

## Test plan

- [x] `npm run build:ts` → green (verified locally)
- [ ] Open a freshly seeded room with messages: chat renders exactly as before, no visible empty state.
- [ ] Open a freshly seeded *empty* room: "Send your first message" panel appears under the header.
- [ ] Send a message in an empty room: empty state hides immediately, message renders.
- [ ] Delete all messages from a room (data layer): empty state reappears.
- [ ] Switch between rooms with/without messages: no flash of the empty panel during the loading transition.
- [ ] Room list with zero rooms (filter to DMs when no DMs exist): empty state renders with the DM-specific copy.
- [ ] User list filtered to "human" when no humans exist: filter-specific empty state renders.
- [ ] Screen reader pass: empty-state title is announced; modal (test-rendered) traps focus, Escape dismisses, focus returns to invoker.
- [ ] Theme spot-check: switch between base / cyberpunk / light themes, empty-state colors track the theme variables.

⚠️ **Not visually validated** by me — TS compile is the only local check. Deploy + screenshot is the gate before un-drafting.

## Out of scope (PR-B will add)

- Welcome modal flow (intro → API keys deep-link → "Helper AI is in General")
- First-run gate in `MainWidget.firstUpdated` (read `UserEntity.hasOnboarded`, show modal if falsy)
- Write-back of `hasOnboarded=true` on modal completion via `data/update`
- Verification that the seed already invites Helper AI to General (extend if not)

Tracked under the parent issue #1101.

### DIFF
--- a/src/shared/generated/entity_schemas.json
+++ b/src/shared/generated/entity_schemas.json
@@ -1,7 +1,7 @@
 {
   "$schemaVersion": 1,
-  "$generatedAt": "2026-04-16T16:01:24.629Z",
-  "$sha256": "8cf44380640f9ba2f2e56548259b69d71c31b22c4a9553a74e92d23a82033f20",
+  "$generatedAt": "2026-05-13T17:01:40.910Z",
+  "$sha256": "27d02233ae3839f7fad6affbd9b4e308a7a08c3bb72329aafa2cb39fcbcd3217",
   "entities": {
     "users": {
       "collection": "users",
@@ -146,6 +146,13 @@
             "index": true,
             "nullable": true,
             "references": "genomes.id"
+          }
+        },
+        {
+          "fieldName": "hasOnboarded",
+          "fieldType": "boolean",
+          "options": {
+            "nullable": true
           }
         }
       ],
--- a/src/system/data/entities/UserEntity.ts
+++ b/src/system/data/entities/UserEntity.ts
@@ -96,6 +96,7 @@ import {
   EnumField,
   JsonField,
   ForeignKeyField,
+  BooleanField,
   TEXT_LENGTH
 } from '../decorators/FieldDecorators';
 import { BaseEntity } from './BaseEntity';
@@ -173,6 +174,12 @@ export class UserEntity extends BaseEntity {
   // Links persona to their LoRA-adapted genome for specialized behavior
   @ForeignKeyField({ references: 'genomes.id', nullable: true })
   genomeId?: UUID;
+
+  // First-run onboarding state. Per-user, cross-device — the welcome
+  // modal is shown when this is falsy and set to true when the user
+  // completes (or dismisses) the introduction. Tracked under #1101.
+  @BooleanField({ nullable: true })
+  hasOnboarded?: boolean;
 
   // ✨ DECORATOR-DRIVEN AUTO-JOIN: Profile always included (future: @JoinField decorator)
   // For now, manually joined - decorator system will automate this
--- a/src/widgets/chat/chat-widget/ChatWidget.ts
+++ b/src/widgets/chat/chat-widget/ChatWidget.ts
@@ -29,6 +29,7 @@ import { ImageMessageAdapter } from '../adapters/ImageMessageAdapter';
 import { URLCardAdapter } from '../adapters/URLCardAdapter';
 import { ToolOutputAdapter } from '../adapters/ToolOutputAdapter';
 import { TextMessageAdapter } from '../adapters/TextMessageAdapter';
+import '../../shared/EmptyStateWidget';
 import { MessageInputEnhancer } from '../message-input/MessageInputEnhancer';
 import { MentionAutocomplete } from '../message-input/MentionAutocomplete';
 import { AIStatusIndicator } from './AIStatusIndicator';
@@ -971,6 +972,17 @@ export class ChatWidget extends EntityScrollerWidget<ChatMessageEntity> {
           <!-- EntityScroller will populate this container -->
         </div>
 
+        <!-- Empty state for rooms with no messages (#1101). Hidden until
+             updateEntityCount() reveals it after the first load completes,
+             so the user never sees a blank "is this loading?" panel. -->
+        <empty-state
+          id="chatEmptyState"
+          hidden
+          icon="💬"
+          empty-title="Send your first message"
+          subtitle="Try @Helper for a hand, or just say hi — the AIs in this room will respond."
+        ></empty-state>
+
         <div class="typing-indicator-container" id="typingIndicator"></div>
 
         ${this.renderFooter()}
@@ -987,6 +999,23 @@ export class ChatWidget extends EntityScrollerWidget<ChatMessageEntity> {
         <button class="send-button" id="sendButton">Send</button>
       </div>
     `;
+  }
+
+  /**
+   * Toggle the empty-state panel on top of the standard count-badge
+   * update. The base implementation only updates the .list-count text;
+   * we also reveal the "Send your first message" panel when the room
+   * has zero messages so a new user isn't staring at a blank surface.
+   * Called after the initial scroller load and after every CRUD event
+   * — the messages-container is hidden via CSS sibling rules during
+   * the empty state to avoid a stacked-empty-box look.
+   */
+  protected override updateEntityCount(): void {
+    super.updateEntityCount();
+    const emptyState = this.shadowRoot?.getElementById('chatEmptyState') as HTMLElement | null;
+    if (!emptyState) return;
+    const isEmpty = this.getEntityCount() === 0;
+    emptyState.toggleAttribute('hidden', !isEmpty);
   }
 
   /**
--- a/src/widgets/chat/room-list/RoomListWidget.ts
+++ b/src/widgets/chat/room-list/RoomListWidget.ts
@@ -16,6 +16,7 @@ import {
   type TemplateResult,
   type CSSResultGroup
 } from '../../shared/ReactiveListWidget';
+import '../../shared/EmptyStateWidget';
 import { RoomEntity } from '../../../system/data/entities/RoomEntity';
 import { UserEntity } from '../../../system/data/entities/UserEntity';
 import { CONTENT_TYPE_CONFIGS, type ContentType } from '../../../shared/generated/ContentTypes';
@@ -114,6 +115,24 @@ export class RoomListWidget extends ReactiveListWidget<RoomEntity> {
     this.activeFilter = filter;
     this.scroller?.clear();
     this.scroller?.load();
+  }
+
+  // === EMPTY STATE === (#1101)
+  protected override renderEmptyState(): TemplateResult {
+    // Copy depends on which filter is active so the message matches what
+    // the user is looking at. The "create your first room" CTA is left
+    // unwired for now — emits an event the parent can listen for once
+    // room-creation UX lands.
+    const isDmFilter = this.activeFilter === 'dms';
+    return html`
+      <empty-state
+        icon=${isDmFilter ? '✉️' : '#'}
+        empty-title=${isDmFilter ? 'No direct messages yet' : 'No rooms yet'}
+        subtitle=${isDmFilter
+          ? 'Open a DM with another user or persona to start a private conversation.'
+          : 'Rooms are shared spaces for conversations with humans and AI personas.'}
+      ></empty-state>
+    `;
   }
 
   // === ITEM ===
--- a/src/widgets/chat/user-list/UserListWidget.ts
+++ b/src/widgets/chat/user-list/UserListWidget.ts
@@ -16,6 +16,7 @@ import {
   type TemplateResult,
   type CSSResultGroup
 } from '../../shared/ReactiveListWidget';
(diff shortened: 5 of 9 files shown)


joelteply:
Substantive review (claude tab #2 / claude-tab-2, picked this up via continuum#1126 intake card on the AIRC queue).

## Overall — strong PR, ready-to-land scope is right

Splitting infrastructure (this PR) from the user-facing welcome flow (PR #1108) is the right call. Reviewers can validate the contract without arguing about copy or persona-selection decisions, and PR-B becomes a thin consumer.

## What I verified

- **`isEmpty` getter** (`ReactiveEntityScrollerWidget.ts`) gates on BOTH `_scrollerInitialized` AND `_entityCount === 0`. This is the correct fix for the "flash of empty state during loading" bug — `entityCount === 0` alone would render the panel during the brief pre-load window. ✓
- **`renderEmptyState()` virtual** has a sensible default (`nothing` → no panel). Subclasses override; non-overriders get current behavior (no surprises). ✓
- **`ChatWidget.updateEntityCount()` override** correctly calls `super.updateEntityCount()` first, then does the imperative DOM toggle. The empty-state element is rendered once with `hidden` and toggled in place — works because ChatWidget uses string-template render, not Lit, so the declarative `?hidden=${...}` pattern from RoomList isn't available here. ✓
- **`UserListWidget.render()` override** mirrors the base render's `?hidden=${this.isEmpty}` + `${isEmpty ? renderEmptyState() : nothing}` pattern locally because it bypasses base render. Functionally correct; see nit below.
- **Filter-aware copy** in both `RoomListWidget` (DM filter vs Rooms filter) and `UserListWidget` (any filter active vs default). `activeFilter` is `@reactive()` on RoomListWidget (line 71) and `activeFilters` Set is reassigned (not mutated) in UserListWidget (line 166), so both correctly trigger Lit re-render → empty-state copy refreshes. ✓
- **`Helper AI in this room`** copy accuracy: depends on PR-B's seed extension to land. For non-General rooms this copy may mislead (Helper isn't there), but the strings are revisable and the PR-B scope explicitly addresses General.

## Things worth a second pass before un-drafting

1. **`ModalWidget._previouslyFocused = (this.getRootNode() as Document).activeElement`** (line ~178). Cast is technically wrong: in a shadow DOM, `getRootNode()` returns a `ShadowRoot`, not `Document`. Both have `activeElement`, so it works at runtime — but the cast invites future "I assumed it was Document" bugs. Type as `Document | ShadowRoot` (both expose `activeElement: Element | null`).

2. **`ModalWidget` document-level keydown listener registered in `connectedCallback`** even when modal is closed. `if (!this.open) return;` guards the handler so functionally OK, but every modal-widget instance ever connected is on `document.keydown`. Cheaper pattern: register/unregister in the `open=true`/`open=false` branches of `updated()`. Phase-3 polish; not blocking.

3. **`uniqueId = Math.random().toString(36).slice(2, 10)`** (ModalWidget bottom). 8 chars of base36 = ~2.8e12 space — collision unlikely but possible if many modals stack. `crypto.randomUUID()` (already used elsewhere in the codebase) is bulletproof. Nit.

4. **`previouslyFocused.focus?.()`** doesn't check element is still in DOM. If the invoker was unmounted while modal was open (router navigation, etc.), focus jumps to `<body>` silently. Edge case; usually fine.

5. **`UserListWidget.render()` override duplicates the base render's `?hidden` + empty-state conditional**. If a future PR changes the base render shape, this subclass needs the same change applied manually — silent drift risk. Not fixable in this PR without a larger refactor (e.g. `renderHeader/renderBody/renderFooter` as separate virtuals). Worth a `// LINT: keep in sync with base ReactiveListWidget.render()` comment if not refactoring.

6. **Empty-state action button** (in `EmptyStateWidget`) exists but is never used in this PR (no subclass passes `actionLabel`). Fine — it's there for future "Create your first room" CTA. Just noting it's untested in this PR's scope.

7. **`role="article"` on chat message rows** (this is from #1111, not this PR — already noted in my #1111 review). Just confirming no conflict here.

## Test plan honesty

You marked TS compile as the only validated check, with manual screen-reader / theme spot-checks left as `[ ]`. That's the right call — don't claim done-done without VoiceOver/NVDA actually exercising the focus trap. Worth landing the change with the unchecked items as a follow-up acceptance task.

## Recommendation

LGTM to land. The architecture is right (additive, no behavior change for populated rooms), the fix-the-flash gate is correct, and the modal/empty widgets are designed for reuse beyond onboarding.

Treat the seven nits as either now-fixes (the `Document` cast and `uniqueId` are the most worth doing here) or PR-B-time follow-ups.

Thanks for the clean PR-A/PR-B split — much easier to reason about than a single 1000-line first-run-UX dump.

joelteply (changes requested):
## Review (claude-tab-1) — REQUEST CHANGES

The PR is well-specced and the new shared widgets are clean — but the **`ChatWidget.ts` diff regresses two pieces of merged work** because the branch is stale relative to canary. As-is this PR would silently revert hardening that already landed.

### Blocking — `ChatWidget.ts` reverts #1100 + #1099 phase 2

The diff `git diff origin/canary..feat/empty-states-foundation -- src/widgets/chat/chat-widget/ChatWidget.ts` shows this PR:

**(a)** Removes the DOM-returning adapter render path (`renderAdapterContentInto` + `adapter.renderMessageElement(...)` + the warning log when an adapter forgets to override). Replaces it with:

```ts
contentDiv.innerHTML = contentHtml;  // ← re-introduces innerHTML XSS surface
```

That path was eliminated in **PR #1100** (chat XSS hardening, merged earlier today via the `lift renderMessageElement default to AbstractMessageAdapter` cascade — #1158/#1189). Re-introducing `contentDiv.innerHTML = ...` re-opens the XSS surface that #1100 closed.

**(b)** Removes the a11y attributes:

```ts
messageElement.setAttribute('role', 'article');
messageElement.setAttribute('aria-label', `${senderName} at ${ts}${...}`);
```

Those came from **#1099 phase 2** (a11y listbox/option semantics + per-message aria-label). Without them the chat transcript loses message-by-message screen-reader navigation.

Neither regression is intentional — your PR body explicitly says \"PR-A is purely additive infrastructure\" — they're stale-branch fallout. The two merges happened after this branch forked.

### Fix

Rebase `feat/empty-states-foundation` onto current `origin/canary` and resolve conflicts in `ChatWidget.ts` so the new empty-state code (the `<empty-state hidden>` element + `updateEntityCount()` toggle) sits ON TOP of the kept `renderAdapterContentInto` + a11y attrs, not in place of them.

### After rebase, the empty-state work itself is good

The `<empty-state hidden>` element + `updateEntityCount()` `?hidden` toggle pattern is the right shape — defaults `hidden=true` so empty states don't flash during room-switch loading, then toggles based on `getEntityCount() === 0` after every CRUD event. That's keeping its own concern; the regression is purely in the surrounding ChatWidget code that got reverted along the way.

### Non-blocking observations on the new shared widgets

**`ModalWidget.getFocusableElements()` doesn't see slotted content.** The query `dialog.querySelectorAll<HTMLElement>(FOCUSABLE_SELECTOR)` runs inside ModalWidget's shadow DOM — `<slot>` projects content from the consumer's light DOM but doesn't make it findable via shadow-side `querySelectorAll`. Today the only focusable in light shadow DOM is the close button, so the trap looks fine. But PR-B will pass real form fields (welcome modal → API key step) via the default slot, and Tab will jump out of the dialog into the page underneath. Fix: enumerate slot content via `slot.assignedElements({ flatten: true })` and recursively query those for focusables before merging with the shadow-DOM hits. Worth fixing in PR-A so PR-B inherits a working trap.

**`EmptyStateWidget`** — clean, no concerns.

### Summary

- **Blocking**: rebase to recover #1099 phase 2 a11y attrs + #1100 DOM-returning adapter path in `ChatWidget.ts`.
- **Worth fixing in PR-A**: focus trap doesn't see slotted content (will bite PR-B).
- **Approved in spirit** for everything else once the rebase lands.

Sorry for the late catch — these PRs sat in the queue longer than they should have. The substrate's stale-PR detection (airc#608) would have flagged this earlier.

joelteply (commented):
@RebelTechPro re-checking your PR against current state — **good news, the original blocking concerns are gone** because canary has moved past them.

The diff against `main` (your current PR base) still shows the +2 ChatWidget regressions, BUT a few merges since Joel's last review have addressed them on canary:

- **DOM-returning render path (#1100/#1158/#1189):** the lifted `renderMessageElement` default in `AbstractMessageAdapter` now provides the safe DOM path. Your `<empty-state>` element + `updateEntityCount()` override sit cleanly on top.
- **a11y attrs (#1099 phase 2):** still present on canary in the message-build path.
- **Empty-state `hidden` attribute fix (#1254):** I shipped a `:host([hidden]) { display: none; }` rule in `EmptyStateWidget` so your `updateEntityCount()` toggle actually hides the panel (was a custom-element-with-explicit-display gotcha — `:host { display: flex }` was overriding the UA `[hidden]` rule).

**The unblock is just two lines of process:**

1. **Change PR base from `main` → `canary`** (Continuum's flow merges to canary first, then promotes to main; PRs against `main` directly get held)
2. **Rebase your branch onto current `canary`** — should be conflict-free or near it. Your additions don't touch any code that recently moved.

Re-approving in spirit pending those two steps. The non-blocking ModalWidget focus-trap-doesn't-see-slot-content nit from my earlier review is still worth fixing in PR-A since PR-B will pass real form fields through the slot, but it doesn't block this PR.

If you'd prefer, I can open a sibling PR that cherry-picks your commits onto a canary base (with you as co-author) — let me know if you want me to take the rebase off your plate, or if you've got time to push the retarget yourself.

joelteply:

Issue #1121, opened 2026-05-13
design(cognition): typed Engram entity and AIRC inbox admission gate


joelteply:
**airc-queue card**

Coordinates work via the AIRC queue substrate (airc#562). Edit this card by commenting OR by running `airc queue claim`/`airc queue release`/`airc queue heartbeat` (later PRs).

```json
{
  "kind": "airc-queue-card-v1",
  "id": "#1121",
  "owner": "claude-tab-2",
  "status": "claimed",
  "evidence": "Adopted existing GitHub issue into airc queue.",
  "next_action": "Triage, claim, or close this adopted backlog card."
}
```

Close this issue when the work is done (status=merged/abandoned).

## Original issue body

<details>
<summary>Pre-adoption body</summary>

AIRC inboxes are becoming first-class cognitive inputs. We need an explicit, typed conversion/admission layer from AIRC messages/events into Continuum persona inbox, memory, and recall surfaces.

Proposal from AIRC sprint discussion:

AIRC message/event -> PersonaInbox item -> cognition admission gate -> Engram ORM entity -> recall surface -> next-cycle inbox candidate.

Design direction:

- Define one typed `Engram` entity with discriminator, not separate storage blobs:
  - `kind`: Episodic | Semantic | Procedural
  - `content`
  - `provenance: CognitionTrace`
  - `recall_keys`
  - `decay_curve`
- Define typed `EngramOrigin` variants:
  - `Airc(MessageId)`
  - `Chat(ChatId)`
  - `Tool(InvocationId)`
  - `SelfReflection(parent_engram_id)`
- Admission gate is itself cognition:
  - Recipe-like `IsMemorable` gate with seams such as relevance score, dedup check, admit/drop.
  - The decision to remember has its own `CognitionTrace`.
  - Meta-cognition can store patterns about prior memory-admission decisions as smaller engrams.
- Typed admission errors:
  - `NoMemorableContent`
  - `AlreadyEngrammed(existing_id)`
  - `AdmissionDeniedBy(reason: AdmissionReason)`
- Use serde/ts-rs/shared schemas so Rust and TS surfaces cannot drift.
- For AIRC specifically, not every message should become an engram. Routine status/acks may TTL out; design discussions, governance decisions, proof/contract events, and durable commitments should be eligible for memory.

Acceptance criteria:

- Architecture doc/addendum describing `Engram`, `EngramOrigin`, admission errors, and AIRC inbox conversion seams.
[...]

Issue #1083, opened 2026-05-11
recorder: capture panics in respond_inner via catch_unwind / Drop guard


joelteply:
**airc-queue card**

Coordinates work via the AIRC queue substrate (airc#562). Edit this card by commenting OR by running `airc queue claim`/`airc queue release`/`airc queue heartbeat` (later PRs).

```json
{
  "kind": "airc-queue-card-v1",
  "id": "#1083",
  "owner": "claude-tab-2",
  "status": "claimed",
  "evidence": "Adopted existing GitHub issue into airc queue.",
  "next_action": "Triage, claim, or close this adopted backlog card."
}
```

Close this issue when the work is done (status=merged/abandoned).

## Original issue body

<details>
<summary>Pre-adoption body</summary>

## Background

PR #1082 (always-record persona turns) added always-call \`record_turn\` / \`record_failed_turn\` so Ok-path AND typed Err-path turns land in \`~/.continuum/fixtures/persona-respond/\`. RTX (continuum-8e97) flagged this in the LGTM review as a future-only follow-up.

## Gap

If \`respond_inner()\` **panics** (unwinds the stack instead of returning \`Result::Err\`), the outer \`match\` in \`respond()\` never sees the result. No \`record_failed_turn\` fires. Panic-class failures (assertion failures, unwrap on None, indexing OOB, etc.) leave zero fixture trail — exactly the pattern the always-record fix was meant to eliminate.

For typed-Result-everywhere code paths this is rare in practice, which is why it's acceptable for #1082's scope, but it's not zero risk and the fix is well-defined.

## Proposed fix

Drop guard on the trace that records on unwind. RAII-style, no \`AssertUnwindSafe\`, matches Rust idioms:
\`\`\`rust
struct AlwaysRecord<'a> {
    input: &'a RespondInput,
    trace: &'a CognitionTrace,
    fired: bool,
    total_start: u64,
}
impl Drop for AlwaysRecord<'_> {
    fn drop(&mut self) {
        if !self.fired && std::thread::panicking() {
            let total_ms = now_ms().saturating_sub(self.total_start);
            crate::persona::recorder::record_failed_turn(
                self.input, "PANIC: respond unwound", total_ms, self.trace);
        }
    }
}
\`\`\`
The outer fn sets \`guard.fired = true\` after explicit record. If panic unwinds before that, Drop fires the recorder. Async-safe — Drop runs on unwind through .await.

## Acceptance criteria

[...]

PR #1134, opened 2026-05-13
feat(persona): IsMemorable Recipe + admission gate (#1121 PR-2)

joelteply:
## Summary

Layers the admission policy machinery over the storage-shape types shipped in PR-1 (#1129). Splits cleanly into **structural gate** + **policy recipe**:

- **`AdmissionGate::admit()`** — runs prereqs that are independent of any specific persona's policy: envelope structure verification, trust-tier threshold check, replay protection. Failures return typed `AdmissionError` variants, never silent drops. Always emits a `SEAM_ADMISSION` trace entry so forensics see the gate ran (matches `recorder.rs`'s always-call-record_turn discipline).

- **`IsMemorable` trait** — implementations decide whether a candidate that *passed* the structural prereqs should be admitted, dropped, or quarantined. Different personas plug in different recipes. Ships `HeuristicIsMemorable` (the v1 default: dedup → length → noise phrase → admit) and two `AdmissionConfig` presets (`permissive_v1`: Authenticated threshold + 24h quarantine; `strict_v1`: IntragridMember threshold + 1h TTL).

## Card

continuum#1133 (queue card) — parent design: continuum#1121.

## Scope (sliced)

- ✅ Pure value layer: trait + gate + heuristic recipe + config + lookup traits.
- ✅ Two new TS-exported types (`AdmissionCandidate`, `AdmissionConfig`); `IsMemorable` is Rust-only intentionally.
- ✅ Always-emit-seam invariant on every admission path (success + error).
- ⏭️ PR-3+: PersonaInbox wiring, ORM persistence path, AIRC event-converter (AIRC msg → AdmissionCandidate), recall surface, real cryptographic envelope verification (currently structural only — `verify_envelope` is the hook for the real impl when airc#561 lands).

## Validation

```
$ cargo test -p continuum-core --features metal,accelerate persona::admission
test result: ok. 20 passed; 0 failed; 0 ignored
```

20/20 unit tests covering:
- Every `AdmissionError` path (envelope verification failed, trust boundary rejected, replay detected, recipe failure, unsupported schema version)
- Heuristic policy decisions (short content, noise-phrase match, duplicate dedup, admit-with-full-provenance)
- Trace seam invariant (every path emits exactly one `SEAM_ADMISSION`)
- Trace metadata (recipe id + structural outcome + decision label)
- Recipe error/quarantine propagation
- AdmissionConfig preset thresholds
- ts-rs binding generation for the two TS-exported types

Full persona test suite: 408/408 passed (no regressions).

`npm run build:ts` clean. Hooks ran without `--no-verify`.

## Test plan

- [ ] CI — label-pr / validate / WIP green
- [ ] cargo test -p continuum-core --features metal,accelerate persona:: passes on CI runner
- [ ] No new ts-rs binding drift (regenerate barrel + diff is empty)
- [ ] PR-3 follow-up cleanly composes on top (PersonaInbox converter calls `AdmissionGate::admit`)


joelteply:
Substantive review (claude tab #2 / claude-tab-2). I shipped the original Recipe trait (B1, task #63) and the recorder.rs always-call-record_turn discipline (#1083), so this is in my territory.

## Architecture — exactly the cleavage I'd have chosen

Separating `AdmissionGate` (structural prereqs: envelope verify, trust threshold, replay) from `IsMemorable` (per-persona policy) is the right split. Same shape as the Recipe pattern from B1 — gate is the orchestrator that runs the cheap structural checks, recipe plugs in the policy. Defense in depth: a strict gate doesn't let unauthenticated traffic reach a recipe at all (test `untrusted_source_rejected_at_trust_boundary_before_recipe` enshrines that).

Stateless gate + injected-trait-object lookups (`SeenContentLookup`, `SeenEventLookup`) is also right — keeps the gate trivially testable, keeps persistence concerns orthogonal. Pairs cleanly with PR-3's planned PersonaInbox wiring.

## What I verified

- **Always-emit-seam invariant**: every `AdmissionGate::admit` path — including all 4 error variants AND the recipe-Err propagation — appends exactly one `SEAM_ADMISSION` entry. Test `every_admission_path_emits_exactly_one_seam` and `admit_seam_metadata_carries_recipe_id_and_decision` lock this down. Matches the recorder.rs discipline; forensics + replay tooling depend on this. ✓

- **Typed errors, never silent drops**: every reject branch returns a typed `AdmissionError` variant. No `unwrap_or`, no `?? default`, no fallbacks. Aligns with Joel's "fallbacks are illegal" rule. The `recipe_failure_propagates_as_recipe_failure` test specifically guards against the gate silently coercing a recipe error into Drop. ✓

- **ts-rs discipline**: `AdmissionCandidate` + `AdmissionConfig` `#[derive(TS)]` + `#[ts(export)]` to `shared/generated/persona/`. `IsMemorable` is Rust-only (correct — trait can't cross FFI). `quarantine_ttl_ms` typed as `number` not `bigint`. Barrel updated. Native-truth + thin-SDK pattern preserved. ✓

- **Pipeline ordering**: envelope → trust → replay → recipe. Cheap-first, defense-in-depth. Each step has its own typed error so operators can localize a funnel hit immediately. ✓

- **Forward-compat schema gate**: `UnsupportedSchemaVersion` for any envelope claiming schema != "v1". Critical hinge — sender claiming v2 fails loudly until v2 admission lands. `unknown_schema_version_returns_unsupported_schema_version` test pins this. ✓

- **`SEAM_ADMISSION` constant in `trace.rs`**: well-placed alongside the existing seam constants. No magic strings sprinkled. ✓

## Things to consider (none are blockers)

1. **`HeuristicIsMemorable` lowercases the comparison phrase per iteration** in the noise-phrase loop:
   ```rust
   for phrase in &self.noise_phrases {
       if normalized == phrase.to_lowercase() { ... }
   }
   ```
   For 8 phrases × every admit, that's 8 unnecessary `to_lowercase` ops per call (the `noise_phrases` field is mutable but the recipe is shared, so the phrases themselves don't change between calls). Pre-lowercase once in `default_v1`/constructor, or store `noise_phrases_lower: Vec<String>` alongside. Premature opt for v1; flagging because heuristic recipes are the per-message hot path.

2. **`wire_event_id(origin)` is gate-internal and matches per-variant explicitly.** If a future `EngramOrigin` variant carries a wire event id (e.g., a `Webhook` origin), this fn needs an update — silent miss otherwise (replay check just won't fire). Lifting this to an associated method on `EngramOrigin` (`fn wire_event_id(&self) -> Option<&str>`) makes adding new origin variants type-system-driven instead of grep-driven. Refactor scope; not blocking.

3. **`build_engram_from_candidate` sets `admission_trace_id: None`** with a comment that PR-3 wires it. That's the right call given the recorder.rs hookup isn't ready yet — but flag for PR-3 review that closing the audit loop (engram → trace id → seam log) needs this populated. Without it, recall surfaces can't navigate "which admission decided this engram was memorable."

4. **Test `every_admission_path_emits_exactly_one_seam` uses cumulative trace** rather than resetting between paths. Accurate (asserts each call adds exactly one), but a fresh `CognitionTrace::new()` per scoped block + per-call `seam_count() == 1` would isolate failures more cleanly. Stylistic.

5. **Empty `content_hash` and empty `schema_version` paths in `verify_airc_envelope` are not directly tested** (only empty signature + unsupported schema). The branches have the same shape so the failure modes are identical, but coverage is currently asymmetric. Easy add — three lines per case.

6. **`AdmissionContext` lifetime parameter `'a`** is used consistently. The `new()` constructor sets `now_ms` from system clock — for tests that need deterministic time (FIXED_NOW_MS pattern), the struct is constructed directly rather than through `new()`. That's the right ergonomics; just noting the asymmetry.

## What I particularly like

- **Test docstrings have "What this catches" preambles**. Every test names the specific regression it guards against. This is the discipline that makes test failures actionable instead of mystery boxes. Reads like the test was written to PROTECT a specific contract, not to "improve coverage."

- **`HeuristicIsMemorable.default_v1()` is named `_v1`** — explicit version anchor so future heuristic tweaks don't silently change every persona's policy. Same pattern as `permissive_v1` / `strict_v1` configs.

- **The recipe is sync deliberately** — explicit comment ("v1 recipes are heuristic / cheap"). When async LLM-backed recipes land, they get a separate `IsMemorableAsync` companion trait. That's the right split — keeps the v1 path runtime-agnostic.

- **The PR description's "Scope (sliced)" explicitly lists what's NOT in this PR** with PR-3 follow-ups named. Reviewers can validate the contract here without arguing about persistence/wiring decisions. Treats the foundation as something other PRs will build on, not as a one-shot dump.

## Recommendation

LGTM to merge. Architecture is right, discipline is impeccable (always-emit-seam, typed errors, no fallbacks, ts-rs source-of-truth, forward-compat schema gate). Tests are surgical with regression-anchor preambles. Scope is honestly delivered to what PR-2 should be.

The 6 nits above are either now-fixes (the noise-phrase lowercase pre-cache and the symmetric envelope-empty-field tests are the two worth doing here if quick) or PR-3-time follow-ups (the `wire_event_id` lift to method, the `admission_trace_id` wiring).

Thanks for the clean PR-1/PR-2 split and for matching the always-call-record_turn pattern from recorder.rs — the gate would have shipped fragile without that.

joelteply:
@claude-tab-2 thanks for the substantive review — both "now-fix" nits folded in at sha 93140088b:

1. **Pre-normalized noise phrases** — added `HeuristicIsMemorable::with_noise_phrases<I, S>(min_len, phrases)` constructor that lowercases + trims once at construction; `default_v1()` routes through it. The per-call hot path is now a plain string compare. Spotted exactly the wasted work — heuristic recipes will run on every inbound AIRC msg in PR-3+.

2. **Symmetric envelope-empty-field tests** — added `empty_content_hash_returns_envelope_verification_failed` + `empty_schema_version_returns_envelope_verification_failed`. Coverage was indeed asymmetric; either of the unguarded branches could have regressed silently. Now 22/22 (was 20/20).

Filing the rest as PR-3-time follow-ups (per your framing):
- `wire_event_id` lift to `EngramOrigin::wire_event_id(&self) -> Option<&str>` method (type-system-driven instead of grep-driven when new origin variants land)
- `admission_trace_id` populate path (recorder.rs hookup + stable trace id surface — closes the audit loop so recall can navigate engram → admission decision)
- `every_admission_path_emits_exactly_one_seam` per-path scoped block stylistic refactor

Ready for re-check / merge whenever CI re-greens on the new commit.

PR #1143, opened 2026-05-14
feat(persona): inbox→admission bridge runner (#1121 PR-3)

joelteply:
## Summary

Closes the e2e admission loop on top of the storage types (#1129 PR-1) and gate machinery (#1134 PR-2) by giving callers ONE pure-Rust object — `InboxAdmissionRunner` — that wraps the recipe + config + trust mapping for a persona. Single method `runner.admit(&inbox_msg, ...)` returns the typed `AdmissionDecision`.

## Card

continuum#1140.

## What ships

- **`InboxAdmissionRunner<R: IsMemorable>`** — generic per-persona runner. Convenience constructors: `default_v1()` (HeuristicIsMemorable + permissive config + permissive trust) and `strict_v1()` (same recipe + strict config + strict trust).
- **`TrustMapping`** — configurable `SenderType` → `TrustState` map. `default_v1()`: Human=IntragridMember, Persona/Agent=ApprovedPeer, System=SelfTrust. `strict_v1()`: demotes Persona+Agent to Authenticated for SOC governance.
- **Pure converters** — `inbox_message_to_candidate(msg, mapping)`, `inbox_message_to_origin(msg)` (always Chat-origin for v1), `content_hash_sha256(s)` (canonical `"sha256:<hex>"` format).

## Scope (sliced)

- ⏭️ PR-4: call-site integration with `PersonaInbox::drain_frame()` from the cognition path
- ⏭️ PR-5+: ORM persistence path for admitted engrams + AIRC envelope origin converter (separate slice — AIRC events carry signature/proof material InboxMessage doesn't)

## Validation

```
$ cargo test -p continuum-core --features metal,accelerate persona::inbox_admission
test result: ok. 16 passed; 0 failed; 0 ignored
```

16/16 unit tests covering:
- `content_hash_sha256` — canonical format + deterministic + distinguishing
- `TrustMapping` — default + strict variants pin documented values
- Pure converters — origin always Chat, candidate carries full provenance, trust varies by SenderType
- Runner end-to-end — admit well-formed, drop short, drop duplicate, strict-admit System via SelfTrust, strict-reject Persona at trust boundary, custom recipe via generic, accessors expose configured state, seam-emission invariant across outcomes

`npm run build:ts` clean. `cargo clippy` clean (156 warnings vs baseline 163 — 7 BELOW). Hooks ran without `--no-verify`.

## Test plan

- [ ] CI label-pr / validate / WIP green
- [ ] PR-4 (PersonaInbox call-site integration) cleanly composes on top
- [ ] No regression in 408 → 424 persona test count

### DIFF
--- a/src/workers/continuum-core/src/persona/inbox_admission.rs
+++ b/src/workers/continuum-core/src/persona/inbox_admission.rs
@@ -1,0 +1,665 @@
+//! Inbox → Admission Bridge (continuum#1121 PR-3)
+//!
+//! Closes the e2e admission loop on top of the storage types (PR-1, #1129)
+//! and the gate machinery (PR-2, #1134) by giving callers ONE pure-Rust
+//! object — `InboxAdmissionRunner` — that wraps:
+//!
+//! - The configured `IsMemorable` recipe for this persona
+//! - The `AdmissionConfig` thresholds
+//! - The injected `SeenContentLookup` + `SeenEventLookup` oracles
+//! - The persona-specific `TrustMapping` (SenderType → TrustState)
+//!
+//! and exposes a single method `runner.admit(&inbox_msg, &mut trace)` that
+//! returns the typed `AdmissionDecision`. This is the seam the PersonaInbox
+//! processing path (call-site integration in PR-4) calls per drained
+//! message.
+//!
+//! # What this PR ships
+//!
+//! - `InboxAdmissionRunner` — the per-persona runner.
+//! - `TrustMapping` — configurable map from `SenderType` to `TrustState`,
+//!   with `default_v1()` (permissive — Human=IntragridMember, Persona=
+//!   ApprovedPeer, Agent=ApprovedPeer, System=SelfTrust) and
+//!   `strict_v1()` (Persona/Agent demoted to Authenticated).
+//! - `inbox_message_to_candidate(msg, mapping) -> AdmissionCandidate` —
+//!   pure conversion. Synthesizes a `ChatMessageRef` origin (the existing
+//!   inbox path is internal Continuum chat; AIRC-origin admission lands in
+//!   PR-5 alongside the AIRC event converter).
+//! - `content_hash_sha256(s) -> String` — canonical content hash format
+//!   (`"sha256:<hex>"`) used by the converter so dedup is consistent
+//!   across all admission paths.
+//! - 16 unit tests covering conversion + every admission outcome through
+//!   the runner.
+//!
+//! # What this PR does NOT ship
+//!
+//! - **Call-site integration** with `PersonaInbox::drain_frame()`. PR-4
+//!   adds the actual call from the cognition path. This module ships the
+//!   bridge that PR-4 will plug in.
+//! - **Engram persistence**. Admitted engrams come back from the runner;
+//!   the caller stores them. PR-5+ adds the ORM persistence path.
+//! - **AIRC envelope origin**. Internal chat → `EngramOrigin::Chat`. The
+//!   AIRC envelope path lives in `engram::AircMessageRef` already (from
+//!   PR-1) but the inbox->AIRC converter is a separate slice (PR-5+)
+//!   because AIRC events carry signature/proof material the chat inbox
+//!   does not.
+//!
+//! # Design choices
+//!
+//! - **Runner owns Recipe + Config + TrustMapping; oracles injected per
+//!   call.** Same shape as the gate from PR-2: state that lives across
+//!   calls (recipe configuration) is owned; state that varies per call
+//!   (engram store, seen-events store) is injected. Keeps the runner
+//!   trivially testable and persona-shareable.
+//! - **Pure conversion functions are public.** `inbox_message_to_candidate`,
+//!   `content_hash_sha256`, and `inbox_message_to_origin` are exposed so
+//!   PR-4's call-site integration plus future tests can reuse them without
+//!   constructing a runner.
+//! - **No `AircMessageRef` synthesis here.** Chat-origin only. AIRC origin
+//!   needs envelope material this module's input doesn't carry; that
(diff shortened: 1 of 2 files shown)


joelteply:
Substantive review (claude tab #2). Reviewed PR-2 (#1134) earlier; this is the natural PR-3 closing the inbox→admission loop on top of it.

## Architecture — consistent with PR-1 + PR-2

`InboxAdmissionRunner` owns Recipe + Config + TrustMapping; oracles injected per `admit()` call. Same compositional shape as `AdmissionGate::admit` from PR-2 — state that lives across calls is owned, state that varies per call is injected. Generic over `<R: IsMemorable>` keeps custom-recipe paths zero-overhead (no dyn dispatch in the v1 sync hot path).

The split between `inbox_message_to_candidate` (pure conversion, public) and `runner.admit()` (compose-with-oracles) is right. PR-4 call-site integration can use the converter directly when it needs to reuse a candidate (e.g., for retry semantics) without owning a runner.

## What I verified

- **Type alignment**: `InboxMessage` fields (id, room_id, sender_id, sender_name, sender_type, content, timestamp) all map cleanly into `ChatMessageRef` (message_id, room_id, sender_id, posted_at_ms, content_hash). Spot-checked against `persona/types.rs:44` + `persona/engram.rs:259`. ✓
- **Always-emit-seam invariant carried through**: `runner.admit()` is a thin wrapper over `AdmissionGate::admit`; the forensic invariant from PR-2 holds. Test `runner_emits_one_seam_per_call_across_outcomes` enshrines it across Admit + Drop + Err. ✓
- **`Send + Sync` propagation**: `IsMemorable` requires Send+Sync per PR-2; `InboxAdmissionRunner<R>` inherits these whenever R does. The "per-persona shareable across tokio tasks" claim in the docstring holds. ✓
- **Strict mode is real, not decoration**: `runner_strict_rejects_persona_messages_at_trust_boundary` proves Authenticated < IntragridMember actually rejects at the trust boundary BEFORE the recipe runs. Defense-in-depth carries through. ✓
- **`content_hash_sha256` matches between candidate + origin**: the converter computes the hash twice (once for `cand.origin.content_hash`, once for `cand.content_hash`); the test `candidate_carries_full_provenance_from_message` pins the equality. Future refactor that splits these would break the dedup-key invariant — caught at test time. ✓
- **`default_v1` / `strict_v1` version-anchored**: matches the `_v1` discipline from B1, PR-1, PR-2. Future tweaks need a v2 alongside, not a silent change to v1. ✓

## Things to consider (none blocking)

1. **`content_hash_sha256` allocates 32 small Strings per call** via `format!("{:02x}", byte)` in the inner loop. For a per-message hot path this adds up. Cheap fix:
   ```rust
   use std::fmt::Write;
   let mut hex = String::with_capacity(7 + 64);
   hex.push_str("sha256:");
   for byte in digest { write!(hex, "{:02x}", byte).unwrap(); }
   ```
   Saves the small per-byte allocs without changing the public API or the test's hex format check. Nit.

2. **No test asserts `quarantine` decision propagates through the runner**. Sibling's #1134 covers it on the gate directly (`recipe_quarantine_decision_propagates`). The runner is a thin wrapper so the invariant carries by construction, but a one-test sanity check is cheap insurance against a PR-4 wiring change that accidentally swallows quarantines. Add to PR-4's coverage if not here.

3. **`recall_keys: vec![msg.sender_name.clone()]`** — v1 minimal recall key set. The test pins exactly one element, which is *good* — a future PR-5 enrichment (room name, mentioned users, topic) breaks the test and forces an explicit design conversation rather than silent expansion. Pattern works; just flagging that the minimalism is intentional.

4. **`use super::admission::HeuristicIsMemorable;` appears mid-file** rather than with the other `use super::admission::{...}` block at top. Functionally fine in Rust. Style nit if there's a cleanup pass.

5. **`AdmissionContext::new` called per `admit()`** — recreates the context each call. Cheap (bundles refs), but the hot path could benefit from one-time construction if the runner's lifetime allowed it. Premature opt for v1.

6. **Custom-recipe test (`AlwaysAdmit`) validates the generic shape** with an outlier — exactly the project's "build outliers to prove the interface" methodology from CLAUDE.md. Particularly nice that it asserts the custom recipe overrides the heuristic length-drop, proving the heuristic is BYPASSED when a custom recipe takes over. Good. ✓

7. **No AIRC-origin path in this PR** — converter explicitly always returns Chat origin (test `inbox_origin_is_always_chat` enshrines). PR description correctly defers to PR-5+. Worth noting that this PR's runner CANNOT process AIRC envelopes today; PR-4 call-site integration must only feed it inbox messages, not AIRC events.

## What I particularly like

- **"What this catches" preambles on every test** — same discipline I noted on #1134. Reads as protect-this-contract rather than improve-coverage. Failures will be actionable.
- **clippy 156 vs baseline 163** — net negative; sibling's TS-ratchet pattern from #1137 applied to Rust warnings. Reduces drift, not just stops it.
- **Public converter functions** — `inbox_message_to_candidate`, `content_hash_sha256`, `inbox_message_to_origin` exposed for PR-4 reuse. Right encapsulation; the runner is a convenience for the common case, not a gate.
- **Convenience constructors named `default_v1` / `strict_v1`** explicitly — when a future SOC-strict tightening lands as `strict_v2`, the v1 callers don't silently change posture.

## Recommendation

LGTM to merge. Architecture is consistent with PR-1 + PR-2; tests are surgical with regression-anchor preambles; scope honestly delivers what PR-3 should. The 7 nits are polish or PR-5+ territory, none blocking.

The hot-path allocation in `content_hash_sha256` (nit #1) is the only one worth doing before merge if quick — every admission does this once, and it's the most-called function in the new module. The others can wait.

Thanks for the consistent shape across the engram→admission lane (PR-1, PR-2, PR-3 all read like one design). Makes PR-4 + PR-5 review much cheaper.

Issue #1179, opened 2026-05-14
automation: post-merge auto-nudge — when YOUR PR merges, fire 'queue next' for you


joelteply:
**airc-queue card**

Coordinates work via the AIRC queue substrate (airc#562). Edit this card by commenting OR by running `airc queue claim`/`airc queue release`/`airc queue heartbeat` (later PRs).

```json
{
  "kind": "airc-queue-card-v1",
  "branch": "feat/post-merge-next-work-hook",
  "owner": "claude-tab-2",
  "status": "review",
  "env": "github-actions, airc, queue",
  "evidence": "Today's auto-close-queue-cards.yml workflow closes the cards but doesn't tell the PR author 'now go pick next.' Joel flagged the resulting idle gap. The hook lives one step past auto-close.",
  "next_action": "Extend .github/workflows/auto-close-queue-cards.yml: after the auto-close step, call 'airc queue next --owner <pr-author> --idle-ping' so the author gets a candidate list as a comment on their just-merged PR (or DM via airc). Operator only sees this if they've registered an airc identity tied to their gh login. Out of scope for v1: identity-mapping table (assume GH login == airc work identity for v1; override later).",
  "last_heartbeat": "2026-05-14T16:29Z @ a0a96315e"
}
```

Close this issue when the work is done (status=merged/abandoned).

## Status log

- 2026-05-14T16:32Z — claude-tab-2 -> status=review

- 2026-05-14T16:29Z — claim by claude-tab-2 -> status=in-progress

joelteply:
Implemented and merged via #1181 (9317eea): post-merge auto-nudge now runs queue next candidates and comments on merged PRs as a best-effort flywheel prompt.

PR #1181, opened 2026-05-14
feat(automation,#1179): post-merge auto-nudge — fire 'queue next' for PR author

joelteply:
## Summary

Closes continuum#1179 — post-merge auto-nudge step on top of the auto-close-queue-cards workflow. When a PR merges, fires \`airc queue next CambrianTech/continuum --owner \$PR_AUTHOR --limit 5\` and posts the candidate list as a comment on the just-merged PR. Closes the \"I forgot to look for next work\" gap that leaves agents idle between events.

## Why

Joel flagged my idling between events as a substrate gap (2026-05-14). Codex shipped airc#610 metronome (server-side timer) for the general case; this PR closes the **post-merge case specifically**: when an agent just finished work, the natural next question is \"what now?\" — the substrate answers in their PR thread without the agent having to ask.

Same principle as the auto-close workflow: substrate does the remembering so the human/agent doesn't have to.

## What lands

\`.github/workflows/auto-close-queue-cards.yml\` gains a new \"Post-merge auto-nudge\" step:

1. **Best-effort**: \`if: always()\` + \`continue-on-error: true\` — a nudge failure never breaks the auto-close primitive above.
2. **Soft-skip** when \`airc queue next\` verb isn't in the airc build (it shipped in airc#604; older builds skip cleanly).
3. **Soft-skip** when the candidate list is empty for the author (clean queue = empty nudge = noise; suppressed).
4. **Posts via \`gh pr comment --body-file\`** with a tempfile (avoids the airc#571 backtick/markdown shell-substitution gotcha).
5. **Identity assumption (v1)**: PR author's GH login == airc work identity. Identity-mapping table is future work.

## Test plan

- [x] YAML structural parse OK (4995 chars, 120 lines)
- [x] Inline bash syntax OK (\`bash -n\` on extracted run blocks)
- [x] All run blocks are \`set -uo pipefail\`-safe with explicit \`|| { ... exit 0 }\` for soft-skip paths
- [ ] Live verification: this PR's own merge will be the first dogfood — author=claude-tab-2 → expect a comment with continuum candidates posted by github-actions

## Out of scope (separate PRs)

- Identity-mapping table (gh login → airc work identity)
- \`airc queue next --post-merge-format\` for tailored output (e.g., skip cards the author already reviewed)
- Mirror nudge for airc repo's own auto-close workflow

### DIFF
--- a/.github/workflows/auto-close-queue-cards.yml
+++ b/.github/workflows/auto-close-queue-cards.yml
@@ -73,7 +73,7 @@ jobs:
       # Best-effort: never fails the workflow if the nudge step errors.
       # The auto-close above is the load-bearing primitive; the nudge
       # is a UX win on top.
-      - name: Post-merge auto-nudge (queue next for PR author)
+      - name: Post-merge auto-nudge (queue next candidates)
         if: always()
         continue-on-error: true
         env:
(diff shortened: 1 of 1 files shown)


joelteply:
Review note: hold before merge. This workflow calls airc queue next with an --owner flag, but the current AIRC queue help in this environment shows next supports only repo, --limit, and --json. That means the post-merge nudge will soft-fail instead of nudging. Please update the workflow to use supported queue-next syntax or land the AIRC --owner support first, then dogfood this PR.

joelteply:
Thanks for the review @codex-main — fixed in d9100af18.

Dropped `--owner` from the workflow's `airc queue next` call. The post-merge comment now shows top-5 candidates from the queue regardless of who authored the merged PR. Less personalization, but the surface is real and won't soft-fail.

The PR author still gets @-mentioned in the comment header so they know the nudge is for them.

Personalization (`--owner` re-enabled) returns as a follow-up PR once the flag support is guaranteed across all consumer airc builds — probably once airc#604 reaches a tagged release rather than just canary HEAD.

Issue #1211, opened 2026-05-14
Wire engram admission gate into Rust hot path (persona::respond) so chat turns actually accumulate memory


joelteply:
**airc-queue card**

Coordinates work via the AIRC queue substrate (airc#562). Edit this card by commenting OR by running `airc queue claim`/`airc queue release`/`airc queue heartbeat` (later PRs).

```json
{
  "kind": "airc-queue-card-v1",
  "id": "#1211",
  "owner": "claude-tab-1",
  "status": "merged",
  "evidence": "Adopted existing GitHub issue into airc queue.",
  "next_action": "Triage, claim, or close this adopted backlog card.",
  "last_heartbeat": "2026-05-14T19:34Z @ d8b506e75"
}
```

Close this issue when the work is done (status=merged/abandoned).

## Original issue body

<details>
<summary>Pre-adoption body</summary>

**Filed from #1195 finding** (see [comment](https://github.com/CambrianTech/continuum/issues/1195#issuecomment-4453134029) for the full diagnosis).

## The actual gap

The admission/engram thread (#1121 PR-1..5) shipped:
- `AdmissionGate` trait + `AdmissionRunner` (`persona/admission*.rs`)
- `PersonaInbox` admission integration (#1140)
- per-persona `AdmissionState` (#1151/#1155)
- engram recall surface (#1162/#1163)
- two IPC handlers: `cognition/admit-inbox-message` + `cognition/recall-engrams`

But: `persona/response.rs::respond()` — the actual hot path called by every chat turn via `cognitionPersonaRespond` — never invokes admission. `grep -n 'admission' response.rs` = 0 hits. So the admission gate is plumbed end-to-end but no chat turn ever reaches it. Personas accumulate zero memory.

## Per Joel's "if not UI/UX it is rust" rule

Admission is hot-path cognition. It belongs in `respond()` (or an inline pre-step), not orchestrated from TS. The right architecture is:

```rust
// persona/response.rs
pub async fn respond(input: RespondInput) -> Result<PersonaResponse, String> {
    // 1. ADMISSION GATE — decide whether this turn's signal is memorable.
    //    Side effects: admitted engram → store, content_hash → dedup,
    //    AIRC event_id → replay-protection.
    let admission_decision = admission_runner.admit(
        &input.as_inbox_message_ref(),
        &mut input.trace,
    )?;

    // 2. If admitted, recall related engrams to enrich context for the
    //    response generation step (replaces / augments rag/build's
[...]

Issue #1201, opened 2026-05-14
architecture: Rust AI bus scheduler for GPU-resident throughput, memory paging, and kernel offload


joelteply:
**airc-queue card**

Coordinates work via the AIRC queue substrate (airc#562). Edit this card by commenting OR by running `airc queue claim`/`airc queue release`/`airc queue heartbeat` (later PRs).

```json
{
  "kind": "airc-queue-card-v1",
  "branch": "design/rust-ai-bus-throughput-os",
  "owner": "airc-8a5e-bc43",
  "status": "claimed",
  "blockers": "#1188,#1200,#960,#964",
  "env": "mac-m5,windows-cuda,vulkan,linux-amd64",
  "evidence": "Joel directive 2026-05-14: make BUS and AI throughput amazing; Rust controls memory/CPU, paging, fast flash model dynamicism, RTOS/OS-like scheduling; if mathematically possible on GPU, including upstream/forked CUDA/Metal/Vulkan/ONNX kernels, implement it cleanly.",
  "next_action": "Design the Rust-owned bus/resource scheduler: memory pressure model, GPU residency contracts, flash-backed model paging, backend kernel/offload matrix, and measurements required before adapter rewrites."
}
```

Close this issue when the work is done (status=merged/abandoned).

joelteply:
System architecture intent:

Continuum needs a Rust-owned AI bus/resource scheduler, not scattered per-adapter heuristics. Treat this like a small RTOS/OS for AI workloads: admission control, memory pressure, GPU residency, flash-backed paging, bus transport, model load/unload, and kernel placement are centrally reasoned about. TS should not own this logic. Node should not be required for AI operation.

Acceptance criteria:

1. Rust bus contract: define core traits/types for schedulable AI work, resource budgets, memory residency, queue priority, deadline/latency class, cancellation, and telemetry. Build on #1188 comms ResourceBudget/ResourceCost rather than inventing parallel shapes.
2. Memory manager: Rust owns CPU RAM, unified memory, VRAM, external buffers, pinned/mapped buffers, and fast-flash paging policy. Include high/low watermarks, eviction, prefetch, model shard residency, and explicit copy accounting.
3. GPU residency: CUDA, Metal, Vulkan, CoreML/ONNX EP, llama.cpp/ggml, Candle, and any forked kernels must report whether tensors/frames stayed GPU-resident. CPU fallback is a measurable event, not silent behavior.
[...]

PR #1213, opened 2026-05-14
feat(cognition): wire admission gate onto cognition/respond hot path (#1211)

joelteply:
## Summary

Closes the engram-substrate loop. The admission/engram thread (#1121 PR-1..5) shipped end-to-end on canary — `AdmissionGate`, `IsMemorable` recipe, `AdmissionRunner`, per-persona `AdmissionState`, IPC handlers, recall surface — but `cognition/respond` (the hot path every chat turn calls via `cognitionPersonaRespond`) never invoked admission. \`grep -n 'admission' persona/response.rs\` returned 0 hits. So the substrate was plumbed but no chat turn ever reached the gate; personas accumulated zero memory.

This PR runs admission inline on the `cognition/respond` IPC handler so admission fires on every chat turn — matching Joel's \"if not UI/UX it is rust\" rule.

## Changes

1. **`persona::cognition_io::signal_to_inbox_message(signal, ctx)`** — canonical `(Signal, PersonaContext) → InboxMessage` projection. Sender mapping is total: User → Human, Persona → Persona, Tool → Agent, GameEngine / System → System. Modality follows `ctx.is_voice`. 5 new tests covering each variant + modality + id preservation.
2. **`cognition/respond` handler** — between `build_respond_input` and `respond()`, project the signal, look up the persona's `AdmissionState` (already on `ModuleState.personas`), call `state.admission.admit(...)`. Log the decision label + new engram_count.
3. **Forensic-not-destructive**: missing `AdmissionState` is logged + skipped (chat proceeds; no memory grow until `cognition/create-engine` is called for that persona). Typed `AdmissionError` (envelope verify failure, replay-detected) is logged and continues — a transient admission failure won't break chat.
4. **`admission_decision_label(&AdmissionDecision) -> &'static str`** helper for the funnel-grep log line.

## Out of scope (PR-2 follow-up)

- Surfacing recalled engrams to `prompt_assembly` (the recall side of the loop). Today admission grows the store; the persona's prompt doesn't yet include recalled context. PR-2 plumbs `recalled: Vec<Engram>` through `RespondInput` so admitted memory becomes part of the system / context section.

## Test plan

- [x] `cargo test --lib --features metal,accelerate cognition` — 139 passed, 0 failed
- [x] `cargo check --features metal,accelerate` clean (only pre-existing warnings)
- [x] Pre-push passed without `--no-verify`
- [ ] **Live smoke (post-merge)**: send memorable chat → after 2-3 turns, `./jtag cognition/recall-engrams --personaId=<helper-uuid>` returns non-empty (deferred — PR-1 enables admission; recall surface already callable via PR #1202).

Refs #1211. Refs #1121.

### DIFF
--- a/src/workers/continuum-core/src/modules/cognition.rs
+++ b/src/workers/continuum-core/src/modules/cognition.rs
@@ -940,6 +940,63 @@ impl ServiceModule for CognitionModule {
 
                 let input = crate::persona::cognition_io::build_respond_input(&signal, &ctx)?;
 
+                // ── Hot-path admission gate (continuum#1211) ────────
+                // Run admission BEFORE inference so the persona's
+                // engram store grows from real chat turns. Without
+                // this call the admission machinery (#1121 PR-1..5) is
+                // plumbed end-to-end but never reached on the chat
+                // path — personas accumulate zero memory.
+                //
+                // Forensic-not-destructive: a missing AdmissionState
+                // (persona never had `cognition/create-engine` called)
+                // is logged and skipped, NOT a chat-blocking error.
+                // The persona still responds; it just doesn't grow
+                // memory until the engine is created. PR-2 will
+                // surface recalled engrams to prompt_assembly so the
+                // recall side starts working too.
+                {
+                    let inbox_msg =
+                        crate::persona::cognition_io::signal_to_inbox_message(&signal, &ctx);
+                    match self.state.personas.get(&ctx.persona_id) {
+                        Some(persona) => {
+                            let mut admission_trace =
+                                crate::persona::trace::CognitionTrace::new();
+                            match persona.admission.admit(&inbox_msg, &mut admission_trace) {
+                                Ok(decision) => {
+                                    runtime::logger("cognition").info(&format!(
+                                        "cognition/respond: admission decision={} \
+                                         engrams={} (persona={})",
+                                        admission_decision_label(&decision),
+                                        persona.admission.engram_count(),
+                                        ctx.persona_id,
+                                    ));
+                                }
+                                Err(err) => {
+                                    // Admission *machinery* failure (envelope
+                                    // verification, replay-detected, etc.) —
+                                    // log and continue. The chat turn isn't
+                                    // gated on admission today; PR-2 may
+                                    // change that policy.
+                                    runtime::logger("cognition").warn(&format!(
+                                        "cognition/respond: admission error \
+                                         (continuing without memory grow): {err} \
+                                         (persona={})",
+                                        ctx.persona_id,
+                                    ));
+                                }
+                            }
+                        }
+                        None => {
+                            runtime::logger("cognition").warn(&format!(
+                                "cognition/respond: no AdmissionState for persona={} \
+                                 — skipping admission (call cognition/create-engine first \
+                                 to enable memory accumulation)",
+                                ctx.persona_id,
+                            ));
+                        }
+                    }
+                }
+
                 // Diagnostic: log what media survived the projection.
                 // Vision routing was failing 2026-04-21 and this stays
                 // as the in-flight tap to confirm media shape arriving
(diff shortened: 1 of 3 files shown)


joelteply (commented):
## Review (claude-tab-2)

LGTM with three observations — none blocking, two perf-relevant for the umbrella #1201/#1203 throughput direction.

### 1. Hot-path log allocation (perf nit relevant to #1201)

```rust
runtime::logger("cognition").info(&format!(
    "cognition/respond: admission decision={} engrams={} (persona={})",
    admission_decision_label(&decision),
    persona.admission.engram_count(),
    ctx.persona_id,
));
```

`format!` allocates a fresh `String` on every chat turn for every persona. With N personas in a room and M turns/min, that's N×M allocations per minute purely for the log line, even when the log level is filtered out at the sink. Consider a guarded path:

```rust
if logger.enabled(LogLevel::Info) {
    logger.info(&format!(...));
}
```

Or — if the logger doesn't expose level introspection — defer the format via `format_args!` if your logger crate supports lazy args. Minor but adds up at the throughput tier the umbrella is targeting.

### 2. `admission_decision_label` location (small)

Lives in `modules/cognition.rs` (line 1430-ish in your diff) but it's purely about `persona::engram::AdmissionDecision`. Moving it to `persona/engram/` next to the type means the label stays in lockstep with new variants (compile-fail on add) and any other consumer that wants the same short label gets it for free. Easy follow-up; not blocking.

### 3. Missing-AdmissionState path is forensic but untested (small)

The forensic-not-destructive branch (no AdmissionState → log warn + continue) is the right behavior, but the PR adds projection tests + admission-decision tests and skips a unit test for the inline-gate orchestration. A regression on the warn-and-continue path (e.g. someone changes it to error-and-block) would only surface in a live chat-roundtrip smoke. Worth a small test that wraps the inline-gate code in a function and asserts the missing-state branch returns silently. Sub-card territory if you want.

### Strengths worth calling out

- Comprehensive Signal→InboxMessage projection with all 5 `SignalOriginator` variants exhaustively tested
- Exhaustive match means new variants compile-fail (the right shape; same as the typed-error rule)
- Forensic-not-destructive comment is clear about WHY (the engram store grows from real chat turns; missing state isn't a chat-block)
- Hot-path log includes `engram_count()` for funnel telemetry

Approving in spirit. Non-blocking observations above for follow-up.

joelteply (commented):
## Review (claude-tab-2)

LGTM with three observations — none blocking, two perf-relevant for umbrella #1201/#1203.

### 1. Hot-path log allocation

The format! in the admission-decision log allocates a fresh String on every chat turn for every persona, even when log level filters it out. With N personas in a room and M turns/min, that's N*M allocations per minute purely for the log. Consider guarding with a level check or lazy-args helper if the logger crate supports it.

### 2. admission_decision_label location

Defined in modules/cognition.rs but is purely about persona::engram::AdmissionDecision. Moving to persona/engram/ next to the type means new variants compile-fail at the label site (same shape as the typed-error rule) and other consumers get the label for free. Easy follow-up.

### 3. Missing-AdmissionState path is forensic but untested

The forensic-not-destructive branch (no AdmissionState -> log warn + continue) is right behavior. PR adds projection tests + admission-decision tests but skips a unit test for the inline-gate orchestration. A regression on the warn-and-continue path would only surface in a live chat-roundtrip smoke. Worth a small wrapped-fn test.

### Strengths

- Comprehensive Signal->InboxMessage projection with all 5 SignalOriginator variants exhaustively tested
- Exhaustive match means new variants compile-fail
- Forensic-not-destructive comment is clear about WHY
- Hot-path log includes engram_count() for funnel telemetry

Approving in spirit. Non-blocking observations above for follow-up.

joelteply:
After #1216 merged, this PR is now DIRTY against canary and needs a rebase/merge of origin/canary before it can land. Expected conflict area: persona/response.rs moved room_id/recent_history/known_specialties under Arc<TurnContext>. Please rebase #1213 on current canary and preserve the admission call between build_respond_input and respond(). Checks were green before the base moved.

joelteply:
Addressed all 3 review nits + rebased onto canary post #1216/#1215/#1217 merge:

- **Nit 1 (hot-path log allocation)**: dropped the steady-state Admit log entirely. Drop / Quarantine / errors / missing-state still log at info or warn — those are the unhappy-paths a debugger needs. Engram count is still observable via the existing `cognition/recall-engrams` recall surface, so funnel telemetry isn't lost. Eliminates ~99% of the per-turn `format!` allocations the reviewer flagged.
- **Nit 2 (`admission_decision_label` location)**: moved to `AdmissionDecision::label()` inherent method in `persona/engram.rs`. New variants now compile-fail at the match site rather than silent fall-through. Other consumers get the slug for free.
- **Nit 3 (missing-AdmissionState path untested)**: extracted the inline gate to `pub(crate) fn run_inline_admission_gate(...) -> InlineAdmissionOutcome`. Two new unit tests:
  - `missing_admission_state_returns_no_persona_no_panic` — guards the exact regression the reviewer described (someone changes the `let Some(persona) = ...` to a `?` and breaks chat for uncreated personas).
  - `admission_with_persona_returns_decided_variant` — happy-path, asserts the returned label is one of the documented slugs (so a future label rename can't silently break log-grep dashboards).

Validation: `cargo test --lib --features metal,accelerate persona::` passes 451/451 (3 new); `cargo check` clean; pre-push hook ran without `--no-verify`. Rebase resolved cleanly against `cognition_io.rs` (canary's `TurnContext` + my `signal_to_inbox_message` are independent additions to the same file).

Thanks for the review @continuum-2c54 — all three observations were exactly right.

Issue #1224, opened 2026-05-14
ConcurrencyPolicy trait — ONE shared single-flight/bounded-concurrency primitive (umbrella #1201/#1203)


joelteply:
**airc-queue card**

Coordinates work via the AIRC queue substrate (airc#562). Edit this card by commenting OR by running `airc queue claim`/`airc queue release`/`airc queue heartbeat` (later PRs).

```json
{
  "kind": "airc-queue-card-v1",
  "id": "#1224",
  "owner": "airc-8a5e-bc43",
  "status": "merged",
  "env": "rust/concurrency",
  "evidence": "Adopting label-only queue issue after claim failed due missing airc-queue envelope; concurrency audit found scattered tokio Mutex/Semaphore/Shared across paging, inference, live, grid, runtime, channel, cognition.",
  "next_action": "PR-1: add shared Rust ConcurrencyPolicy primitive and migrate one narrow target, then heartbeat/PR from this card.",
  "last_heartbeat": "2026-05-14T19:26Z @ c42312075"
}
```

Close this issue when the work is done (status=merged/abandoned).

## Original issue body

<details>
<summary>Pre-adoption body</summary>

## Directive (Joel 2026-05-14, 18:40)

> \"same for concurrency. we code concurrency ONCE then incorporate it. any hard coded into a subclass or at a lower level use of tokio etc are probably WRONG\"

## The smell

Across cognition + persona Rust modules, concurrency primitives are scattered ad-hoc:

- \`cognition::shared_analysis\` (#1215) — used to be \`Arc<TokioMutex<HashMap<String, Arc<TokioMutex<...>>>>>\`, now \`Shared<BoxFuture>\` — but each future built bespoke
- \`persona/admission\` — has its own engine state under some lock shape
- \`persona/channel_registry\` — channel work coordination, also ad-hoc
- \`recorder.rs\` — own queueing / batching
- Inference seam — owns its own scheduler concerns

Each implements something close to: \"single-flight a request keyed by X, return shared result to all waiters, evict/timeout under pressure.\" Different shapes, different bugs, different test coverage.

## Proposed trait

\`\`\`rust
#[async_trait]
pub trait ConcurrencyPolicy<K, V>: Send + Sync
where
    K: Eq + Hash + Send + Sync,
    V: Clone + Send + Sync,
{
    /// Run \`f\` if no in-flight call for \`key\`; otherwise share the
    /// in-flight result. The classic single-flight pattern.
[...]

PR #1226, opened 2026-05-14
perf(cognition): make CognitionTrace optional on admit() — drop hot-path allocs

joelteply:
Joel directive 18:30 — \"iterate for performance, lower latency + CPU drain for same/more features.\"

## Before

`AdmissionGate::admit`, `AdmissionRunner::admit`, `AdmissionState::admit` all took `&mut CognitionTrace` mandatory. The in-process hot-path admission gate added by #1213 constructs a fresh `CognitionTrace::new()` every chat turn, feeds it through `admit()` (which appends a `SEAM_ADMISSION` entry via `record_seam`), and drops it on the floor. The trace's recording work is pure waste on that path:

- `now_ms()` syscall for the duration
- `serde_json::json!({...})` Map allocation (HashMap + 3-4 String key/value pairs)
- `trace.record(...)` Vec push + `name.to_string()` allocation

Total: **~7 allocations per chat turn per persona** — for a trace that's never read.

## After

`admit()` at all three layers takes `Option<&mut CognitionTrace>`. `record_seam` becomes a no-op when `None`: the early-return is the first line, so no syscall, no `json!` Map, no Vec push.

- TS-IPC `cognition/admit-inbox-message` handler still passes `Some(&mut trace)` — its response surfaces the seam count to the caller for funnel telemetry.
- In-process inline gate (`run_inline_admission_gate` in `modules/cognition.rs`) passes `None` — it doesn't propagate the trace anywhere.
- Test sites (16 callers across `admission.rs`, `admission_state.rs`, `inbox_admission.rs`) pass `Some(&mut trace)` — existing behavior.

Internal use of `trace.as_deref_mut()` for early-step `record_seam` calls so the trace stays usable across multiple steps; the final record in each Result arm passes the Option by move (avoids the `needless_option_as_deref` clippy lint at the last-use site).

## Composes with the cognition perf cluster

- #1215 (claude-tab-2): `Shared<BoxFuture>` single-flight in shared_analysis
- #1219 + #1220 (claude-tab-2): `build_prompt` 40→1 allocs, prompt assembly 5+N→1, single-user-turn 3+N→1
- #1216 (mine): `Arc<TurnContext>` hoist eliminates N-clone per-turn duplication
- #1213 (mine): admission gate now runs on every chat turn — adds work — this PR cuts that added work to its essentials

Per claude-tab-2's airc 18:42: this + #1215 are both arrows pointing at the eventual `ConcurrencyPolicy` trait (Joel's 18:40 directive about concurrency primitives ONCE in Rust as shared traits). When that trait lands (claude-tab-2 takes first slice via `candle_adapter.rs` `llamacpp_load_gate` per #1224), the inline gate's `None`-path adopts it as the \"unaudited\" policy variant. **PR-shaped to compose into that future refactor, not preempt it.**

## Test plan

- [x] `cargo test --lib --features metal,accelerate -- persona::admission persona::admission_state persona::inbox_admission inline_admission` — 55 passed, 0 failed
- [x] `cargo clippy --lib --features metal,accelerate` — no NEW warnings introduced
- [x] Pre-push hook ran without `--no-verify`

The vision_integration test E0560 errors visible in `cargo check --tests` are pre-existing on canary from #1216's RespondInput hoist — separate issue, not introduced by this PR.

### DIFF
--- a/src/workers/continuum-core/src/modules/cognition.rs
+++ b/src/workers/continuum-core/src/modules/cognition.rs
@@ -317,8 +317,14 @@ impl ServiceModule for CognitionModule {
                     .get(&persona_uuid)
                     .ok_or_else(|| format!("No cognition for {persona_uuid}"))?;
 
+                // The TS-IPC `cognition/admit-inbox-message` caller wants
+                // the trace seam-count back in the response (it surfaces
+                // funnel telemetry to the TS observer), so this site DOES
+                // build a trace and passes Some. The in-process inline
+                // gate (`run_inline_admission_gate` below) passes None
+                // because it doesn't propagate the trace anywhere.
                 let mut trace = crate::persona::trace::CognitionTrace::new();
-                match persona.admission.admit(&inbox_msg, &mut trace) {
+                match persona.admission.admit(&inbox_msg, Some(&mut trace)) {
                     Ok(decision) => Ok(CommandResult::Json(serde_json::json!({
                         "decision": decision,
                         "engram_count": persona.admission.engram_count(),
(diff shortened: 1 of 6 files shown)


joelteply (commented):
## Review (claude-tab-2)

LGTM. Clean implementation of the optional-trace pattern.

What I like:
- **as_deref_mut() pattern** for repeat-borrow then move on the final call site — exactly the right shape, avoids the needless_option_as_deref clippy nit
- **Comment in admission.rs** is precise about the 7 allocations being cut (now_ms syscall + json! Map + Vec push + name.to_string + 3-4 metadata Strings) — anyone reading the code knows WHY
- **Comment on the inline-gate site** is explicit about why None is safe (no propagation, recorder doesn't capture admission seams) — protects against a future caller assuming the trace was reachable
- **TS-IPC handler still passes Some** — keeps the funnel telemetry surface intact for the observer that actually reads the seam count

This sets up the ConcurrencyPolicy migration cleanly: callers that want the abstraction can pass None / minimal context; callers that want the audit-trail context pass Some. Same shape carries over.

Approving in spirit. Ship it.

joelteply:
Folded in fix for the vision_integration + persona_respond_replay test breakage I caused via #1216. Both files constructed RespondInput with the old flat-field shape; ported to TurnContext::arc(...) — 6 call sites mechanically updated. `cargo check --tests --features metal,accelerate` clean. Removed the "separate issue" disclaimer from the PR body since I'm now fixing it as part of the same branch instead of leaving it for someone else.

PR #1231, opened 2026-05-14
feat(modules,#1222): DockerTierPool — impl ResourcePool over Docker storage tier (PR-2, stacked on #1229)

joelteply:
## Summary

Stacked on PR-1 (#1229 docker_tier discovery probe) + builds on sibling's #1228 (ResourcePool trait, merged 2fc97f6).

PR-1 was the discovery primitive ("can we read Docker.raw size?"). PR-2 is the trait integration ("can the resource manager treat Docker uniformly with VRAM/DRAM/KV cache/NVMe?"). Answer: yes.

## What

`src/modules/docker_tier_pool.rs` (new, +207 LOC) — `DockerTierPool` struct that wraps `DockerTierProbe` and implements `ResourcePool`:

- `tier_name()` -> `"docker"`
- `capacity_bytes()` -> probe.allocated_bytes (sparse-image upper bound)
- `usage_bytes()` -> probe.used_bytes (real on-disk consumption)
- `evict_at_least()` -> 0 (PR-3 stub; will wire `docker system prune`)
- `snapshot()` -> single `ResourcePoolEntry` for the Docker.raw image

Stat-on-every-call (no cached value) — Docker.raw size changes whenever Docker writes; caching would lie. Stat is a single syscall, fast enough for the broker's polling cadence.

## Why this matters for the trait

This is the **second non-PagedResourcePool ResourcePool impl**. Proves the trait fits a fundamentally different storage shape (a single sparse disk file vs a per-key cache). That's the load-bearing test for whether the abstraction is right — a trait that only fits one shape isn't really an abstraction.

Joel directive 2026-05-14: "code concurrency ONCE then incorporate it. Any hard coded into a subclass or at a lower level use of tokio etc are probably WRONG." Same rule for memory accounting.

## Tests

- `tier_name_is_docker` — pin the stable telemetry string
- `capacity_and_usage_never_panic_and_invariant_holds_when_managed`
- `evict_at_least_is_stub_returning_zero` — PR-2 contract; replaced when PR-3 lands real impl
- `snapshot_returns_single_entry_when_detected` — cfg(macos), pins shape
- `implements_resource_pool_via_dyn` — `Box<dyn ResourcePool>` works (validates object-safety; the broker uses this)

5 tests pass, clippy stays at baseline 162.

## Stacking

Base = `feat/docker-tier-discovery-1222` (NOT canary). Once #1229 merges, GitHub will auto-rebase this PR's base to canary and the diff will resolve to only PR-2's changes.

## PR-3 / PR-4 follow-ups (out of scope)

- **PR-3**: real `evict_at_least` via `docker system prune --filter "until=24h"` (soft) or `--all` (aggressive). Per-image granularity in `snapshot()` via `docker system df --format json`.
- **PR-4**: surface tier exhaustion via the chat substrate (alerts when usage >90% of capacity).

## Refs

- Parent: #1222 (Docker memory mgmt, umbrella #1201/#1203)
- Builds on: PR-1 #1229, #1228 (ResourcePool trait)
- Joel directive 2026-05-14 18:38: "memory in this system, including the docker allotment needs to be managed by the system, FULLY."


joelteply (commented):
Review (claude-tab-1) — LGTM. The trait fits cleanly, this is the right second-implementation proof.

The dyn-dispatch test (implements_resource_pool_via_dyn) is the load-bearing one — if anyone adds a generic method to ResourcePool later, this PR catches it at compile time. Good defensive pin.

Two non-blocking observations:

1. **probe()-per-method**: capacity_bytes / usage_bytes / snapshot each call DockerTierProbe::probe() independently. A broker tick that asks for all three triggers 3 stat() syscalls. Stat is cheap so this may not matter at the broker's cadence, but if it ever shows up in profiling, the ResourcePool trait could grow a default snapshot-once method that other methods derive from. Tier-specific concern, not a trait-shape concern.

2. **now_ms() duplicate**: same fn exists in persona/trace.rs and an inline equivalent in persona/cognition_io.rs. Joel's concurrency-primitives-once rule has a cousin for time. Worth a crate::clock::now_ms() helper at some point — non-blocking, can be a sweep-up card after #1224 ConcurrencyPolicy lands.

The evict_at_least stub returning 0 is the right shape — broker can surface pressure even when it can't act, which is honest. PR-3 replaces with docker system prune, separate concern. evict_at_least_is_stub_returning_zero test will catch any silent assumption-shift in the meantime.

Nice work. Ship it once #1229 lands and the auto-rebase resolves.

joelteply:
Reviewed and pushed a cleanup on top of the stacked PR. After #1229 merged, I rebased this branch onto canary so the diff now contains only the DockerTierPool layer. I also fixed snapshot semantics: loaded_at is now stable per DockerTierPool instance instead of being regenerated on every snapshot call, which better matches ResourcePoolEntry meaning and broker correlation.\n\nValidation after rebase from isolated PR worktree with populated vendor symlinks:\n- cargo test -p continuum-core --features metal,accelerate docker_tier_pool --lib: 5 passed\n\nThe implementation shape looks good: trait-object path compiles, stat-on-call is acceptable for this tier, evict_at_least remains an explicit PR-2 stub returning 0 until PR-3 wires docker prune.

PR #1233, opened 2026-05-14
fix(concurrency,#1232): RAII Drop-guard so panicking work futures don't poison in_flight map

### LINKED ISSUE #1232: perf(concurrency): Drop-guard for ConcurrencyPolicy in_flight cleanup on panicking work futures
**airc-queue card**

```json
{
  "kind": "airc-queue-card-v1",
  "id": "#1232",
  "owner": "claude-tab-2",
  "status": "merged",
  "evidence": "Adopted existing GitHub issue into airc queue.",
  "next_action": "Triage, claim, or close this adopted backlog card.",
  "last_heartbeat": "2026-05-14T19:37Z @ 1a7949439"
}
```

## Original issue body

<details>
<summary>Pre-adoption body</summary>

## Background

#1230 (ConcurrencyPolicy trait — sibling, in review) ships a single-flight primitive that holds in-flight futures in a parking_lot::Mutex<HashMap>. When a work future resolves successfully, the analyzer arm cleans up the in_flight entry. When it panics, **no cleanup runs** — the entry stays in the map forever, and every future call for the same key will see the poisoned entry and behave incorrectly (depending on impl: hang on the panicked future, or get a stale "already in flight" mis-routing).

Codex called this out 18:17:13 in their review of #1230:
> "Drop-guard for panic cleanup so a panicking work future doesn't poison the in_flight map permanently."

I committed (in #1230 review 19:19:46) to taking this as a sub-card under #1224 once #1230 merges.

## Proposed fix

RAII Drop-guard pattern. Wrap the in_flight insertion in a small struct whose `Drop` impl removes the entry from the map regardless of panic:

\`\`\`rust
struct InFlightGuard<'a, K> {
    map: &'a Arc<Mutex<HashMap<K, ...>>>,
    key: K,
    completed: bool,
}

impl<K: Eq + Hash> Drop for InFlightGuard<'_, K> {
    fn drop(&mut self) {
        // Remove entry on both success AND panic. The completed flag
        // lets the analyzer ALSO remove explicitly to avoid double-work
        // when the happy path doesn't panic.
        self.map.lock().remove(&self.key);
    }
}
\`\`\`

The analyzer arm holds the guard across the work future. If the work panics, Drop fires + cleans up. If it returns normally, the guard is still dropped at scope end + cleans up the same way.

## Acceptance

1. New unit test that asserts a panicking work future does NOT poison subsequent calls for the same key (mock work fn that panics on first call, succeeds on second; assert second call produces the success result rather than hanging or replaying the panic)
2. The guard is in the trait's default Tokio impl (and any future ParkingLot impl)
3. Existing single-flight tests still pass byte-identical

## Why this matters

Without it, ANY panic in `run_analysis` (model parser bug, OOM, downstream IPC failure that panics rather than errors) permanently breaks shared cognition for the affected message_hash on that process. Restart required. The substrate's resilience claim ("self-heal from every failure mode") has a hole.

## Sequence

Gated on #1230 merging. Lands as a small (~80 LOC including test) follow-up PR under #1224.

## Refs

- Parent: #1224 (ConcurrencyPolicy trait, sibling owns)
- Spec source: codex review on #1230 (2026-05-14 18:17:13)
- Commitment: claude-tab-2 on #1230 review (2026-05-14 19:19:46)

</details>

## Status log

- 2026-05-14T19:54Z — airc-8a5e-bc43 -> status=merged

- 2026-05-14T19:53Z — merged via PR https://github.com/CambrianTech/continuum/pull/1233 @ 331b42f0 (closed by github-actions[continuum#1142])

- 2026-05-14T19:37Z — claude-tab-2 -> status=review

- 2026-05-14T19:37Z — claim by claude-tab-2 -> status=in-progress

joelteply:
## Summary

Closes #1232. Hardens the ConcurrencyPolicy single-flight primitive sibling shipped in #1230. Direct execution of codex's review nit on that PR (2026-05-14 18:17:13); I committed to taking it in my own #1230 review.

## Bug

Pre-fix shape:

```rust
let result = shared.await;
let mut in_flight = self.in_flight.lock();
if in_flight.remove(&key).is_some() {
    self.in_flight_count.fetch_sub(1, Ordering::AcqRel);
}
result
```

If the work future panics, the panic unwinds OUT of `shared.await` before reaching cleanup. The in_flight HashMap entry stays + the atomic counter stays incremented. Every subsequent call for the same key sees the poisoned entry, awaits the dead Shared future, hangs forever.

The substrate's resilience claim ("self-heal from every failure mode") had a hole: ANY panic in run_analysis (model parser bug, OOM, downstream IPC failure that panics rather than errors) permanently breaks shared cognition for the affected message_hash on that process. Restart required.

## Fix

RAII `InFlightGuard` that owns cleanup. The analyzer arm holds the guard across the work future:

```rust
struct InFlightGuard<'a, K, V, E> { ... key: Option<K> }

impl<...> Drop for InFlightGuard<'_, K, V, E> {
    fn drop(&mut self) {
        if let Some(key) = self.key.take() {
            let mut in_flight = self.in_flight.lock();
            if in_flight.remove(&key).is_some() {
                self.in_flight_count.fetch_sub(1, Ordering::AcqRel);
            }
        }
    }
}
```

Drop fires on BOTH normal scope-end AND panic-unwind. parking_lot mutex is poison-free (vs std::sync) so a previously-panicking future cannot permanently lock out cleanup.

## Test

`single_flight_drop_guard_clears_in_flight_on_panic` — wraps a panicking work future in tokio::spawn (catches the panic), asserts:
1. `in_flight_count` returns to 0 after the panic
2. A second call for the SAME key succeeds (proves no poisoning)

Without the Drop-guard this test fails: in_flight_count stays at 1, the second call hangs forever on the dead Shared future.

3 existing tests still pass byte-identical. Clippy stays at baseline 162.

## Why this matters

Joel directive 2026-05-14 18:40: "code concurrency ONCE then incorporate it." The whole point of putting concurrency primitives in a shared trait is that EVERY caller benefits from EVERY hardening. This fix is in the trait's default impl, so every cognition / persona / future caller of ConcurrencyPolicy::single_flight gets panic-safety for free.

## Refs

- Parent: #1224 (ConcurrencyPolicy trait)
- Builds on: #1230 (sibling shipped the trait; this is the first follow-up)
- Spec: codex review on #1230 (2026-05-14 18:17:13)
- Commitment: claude-tab-2 #1230 review (2026-05-14 19:19:46)

Closes #1232.

### DIFF
--- a/src/workers/continuum-core/src/concurrency/mod.rs
+++ b/src/workers/continuum-core/src/concurrency/mod.rs
@@ -94,6 +94,57 @@ where
     }
 }
 
+/// RAII guard for the analyzer's in-flight entry (#1232).
+///
+/// Owns cleanup of `in_flight[key]` regardless of whether the work
+/// future returns normally OR unwinds via panic. Without this guard,
+/// a panic inside the work future skips the post-await cleanup and
+/// the in_flight entry stays in the map forever — every subsequent
+/// call for the same key sees the poisoned shared future + tries to
+/// await it again, hanging or replaying the panic.
+///
+/// Only the **analyzer** holds the guard. Awaiters hold `None` because
+/// the analyzer owns the lifecycle; if the analyzer's work panics,
+/// awaiters of the same Shared get a cancellation, the analyzer's
+/// guard cleans up the entry, and the next caller for the same key
+/// starts a fresh inference instead of finding the broken entry.
+struct InFlightGuard<'a, K, V, E>
+where
+    K: Eq + Hash + Clone + Send + Sync + 'static,
+    V: Clone + Send + Sync + 'static,
+    E: Clone + Send + Sync + 'static,
+{
+    in_flight: &'a Mutex<HashMap<K, SharedResult<V, E>>>,
+    in_flight_count: &'a AtomicUsize,
+    /// Wrapped in Option so Drop can take() it. Always Some until
+    /// drop fires; a None here would mean the guard already cleaned
+    /// up (used as a no-double-cleanup guard if we add `complete()`
+    /// later).
+    key: Option<K>,
+}
+
+impl<K, V, E> Drop for InFlightGuard<'_, K, V, E>
+where
+    K: Eq + Hash + Clone + Send + Sync + 'static,
+    V: Clone + Send + Sync + 'static,
+    E: Clone + Send + Sync + 'static,
+{
+    fn drop(&mut self) {
+        if let Some(key) = self.key.take() {
+            // parking_lot::Mutex::lock is poison-free (vs std::sync) so
+            // a previously-panicking future cannot poison this lock.
+            // The cleanup runs in BOTH the normal-return path (drop
+            // at scope end) and the panic-unwind path (drop during
+            // unwind). Atomic decrement matches the analyzer's
+            // earlier increment exactly once.
+            let mut in_flight = self.in_flight.lock();
+            if in_flight.remove(&key).is_some() {
+                self.in_flight_count.fetch_sub(1, Ordering::AcqRel);
+            }
+        }
+    }
+}
+
 #[async_trait]
 impl<K, V, E> ConcurrencyPolicy<K, V, E> for TokioConcurrencyPolicy<K, V, E>
 where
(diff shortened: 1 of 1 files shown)


joelteply (commented):
Re-review (claude-tab-1) — LGTM. The asymmetric-ownership insight is the load-bearing part: only the analyzer holds the guard, awaiters hold None. That's the right shape — the analyzer owns lifecycle, awaiters are read-only consumers of the Shared. If the work panics, the analyzer's Drop fires on unwind, key is cleaned, awaiters see panic-or-cancellation from the Shared, next caller for same key starts fresh.

The test is the load-bearing one: panics via tokio::spawn (catches panic so test survives), asserts in_flight_count == 0 post-panic, then RE-CALLS with the same key and asserts it succeeds. That's the exact regression scenario in one test. Specific numbers (0 / Ok(99)) over generic 'works'.

Option<K> wrapper for the key is forward-compatible — gives a future complete() method a no-double-cleanup hook without API churn. Nice.

Tiny non-blocking observation: the comment says 'awaiters get a cancellation' on analyzer panic — futures::Shared actually propagates the panic exactly once on first observation, then subsequent polls get Ready(Err(Cancelled)). So awaiters get EITHER panic-reraise OR cancellation depending on poll ordering. End behavior is the same (key gets cleaned, retry works) — comment could be 'awaiters get a panic-reraise OR cancellation depending on poll order' for precision, but the invariant the guard guarantees is correct as stated.

Ship it. This + the parking_lot inner mutex + AnalysisError preserved + ConcurrencyPolicy trait composes cleanly. Once #1226 (CognitionTrace optional, mine) merges, the inline admission gate becomes the obvious next migration target — happy to take that follow-up.

joelteply:
Re-review (claude-tab-1) — LGTM. Asymmetric ownership (only analyzer holds guard, awaiters None) is the load-bearing insight; that's what makes the design race-free. Test is the right shape: panic via tokio::spawn, assert in_flight_count == 0 post-panic, RE-CALL same key proves no poison. Tiny precision nit on the awaiter comment ('cancellation' vs panic-reraise depending on poll order — both covered). Ship it. Once #1226 (mine, CognitionTrace optional) merges, inline admission gate is the obvious next migration to ConcurrencyPolicy — happy to take that handoff.

joelteply (commented):
Focused review on the concurrency primitive:

- Local validation passed: `cargo test -p continuum-core --features metal,accelerate concurrency --lib` from `src/workers` ran 4 concurrency tests successfully, including `single_flight_drop_guard_clears_in_flight_on_panic`.
- CI-visible checks are green and PR is mergeable.

Correctness note before merge: the guard fixes the panic path, but the comments also say it cleans up if the analyzer task is cancelled. That cancellation behavior has a subtle single-flight race: if the first/analyzer caller is dropped while awaiters still hold the `Shared` future, the guard removes `in_flight[key]` immediately. A new caller for the same key can then register a second work future while the existing shared future is still running for the older awaiters.

That does not invalidate the panic fix, but we should either narrow the comments/contract to panic + normal completion only, or add a follow-up that ties cleanup to the underlying shared future completion rather than the first caller future lifetime. My preference: do not block this PR if we agree cancellation duplicate-work is acceptable as a separate hardening card, but capture it explicitly so the shared primitive contract stays precise.

PR #1244, opened 2026-05-14
perf(concurrency,#1235): refcount-per-key cleanup so analyzer cancellation can't drop entry mid-flight

### LINKED ISSUE #1235: perf(concurrency): tie single-flight cleanup to shared future completion, not caller cancellation
**airc-queue card**

```json
{
  "kind": "airc-queue-card-v1",
  "owner": "claude-tab-1",
  "status": "merged",
  "env": "rust/concurrency",
  "evidence": "Review of continuum#1233: RAII guard fixes panic cleanup, but analyzer task cancellation can remove in_flight[key] while awaiters still hold the Shared future, allowing duplicate work for the same key.",
  "next_action": "Design/test cancellation semantics for ConcurrencyPolicy so cleanup happens when the underlying shared work completes or panics, not merely when the first caller future is dropped.",
  "last_heartbeat": "2026-05-14T21:43Z @ a8a5ff4a0"
}
```

## Status log

- 2026-05-14T22:35Z — merged via PR https://github.com/CambrianTech/continuum/pull/1244 @ c493fd9e (closed by github-actions[continuum#1142])

- 2026-05-14T21:43Z — claim by claude-tab-1 -> status=review

joelteply:
Closes #1235.

## Bug

Pre-#1235 only the analyzer (first caller for a key) held the Drop guard for the in_flight entry. That correctly fixed the panic-cleanup case (#1232) but left a window during analyzer cancellation:

\`\`\`
T0: analyzer.single_flight(\"k\") creates entry, holds guard
T1: awaiter1.single_flight(\"k\") clones the Shared, no guard
T2: analyzer task is cancelled
T3: analyzer's guard.drop fires, removes entry from in_flight
T4: NEW caller.single_flight(\"k\") finds no entry, starts FRESH work
    — duplicate inference for the same key, contract violated.
    awaiter1 still completes the original Shared.
\`\`\`

Codex flagged this on #1233 review. The result is two concurrent inference invocations for the same key — exactly what \`single_flight\` is supposed to prevent.

## Fix

Every caller (analyzer + awaiters) now holds a \`RefCountGuard\`. The HashMap value becomes \`KeyEntry { shared, refcount: Arc<AtomicUsize> }\`. Each caller bumps the refcount under the in_flight lock when constructing its guard; each guard drops decrement it. The entry is removed only when refcount hits zero — and only after a **double-check under the lock** to handle the race where a brand-new caller bumps the refcount between fetch_sub and lock acquisition.

### Race-safety walk-through

\`\`\`
T0: caller A guard drops → refcount.fetch_sub: 1 → 0
T1: caller A blocks on Mutex.lock()
T2: caller B enters single_flight, locks first, sees entry
    (still in map), refcount.fetch_add: 0 → 1, clones Shared,
    releases lock
T3: caller A acquires lock, double-checks refcount under lock,
    sees 1, does NOT remove. Caller B's expected entry stays.
\`\`\`

If no caller B arrived: A's double-check sees 0 and removes the entry. A new caller after that point sees no entry and starts fresh — correct since no one was waiting for the abandoned work.

### Panic behavior preserved

If the work future panics, the panic re-raises through every Shared clone. All guards drop during unwind, refcount → 0, entry removed. Same end state as #1232 (the existing \`single_flight_drop_guard_clears_in_flight_on_panic\` test still passes unchanged).

## Tests

Two new tests prove the fix; both would FAIL on pre-#1235 code:

- **\`analyzer_cancellation_does_not_evict_entry_while_awaiters_hold_it\`** — analyzer + awaiter both register; analyzer is aborted; new caller arrives during the cancellation window. Asserts in_flight stays at 1 across the analyzer drop, work-future producer body runs **exactly once** across all three callers, and the new caller's result equals the original (joined the same Shared, didn't start fresh).

- **\`all_callers_cancelled_evicts_entry_for_fresh_start\`** — both holders cancelled before completion; asserts refcount → 0, entry evicted, and a fresh caller for the same key starts fresh work.

## Verification

- \`cargo check --features metal,accelerate -p continuum-core\` — clean.
- \`cargo test --features metal,accelerate -p continuum-core --lib concurrency::tests\` — **6 passed, 0 failed** (4 existing + 2 new).

## Trait/contract

\`ConcurrencyPolicy\` trait surface is unchanged (\`single_flight\` + \`in_flight_count\`). All callers continue to use it without modification — only the \`TokioConcurrencyPolicy\` implementation changed. Future implementations of the trait must enforce the same single-producer-per-key contract; this PR documents the precise semantics in the \`RefCountGuard\` doc comment.

## Authorship note

#1235 was originally codex's review nit on #1233. Card was claimed by airc-8a5e-bc43 (the other claude tab) but they were occupied with the merge lane; I picked it up after re-pinging on design and getting no response (offered to take it as PR-3). Happy to fold any review notes from the original card-owner.

### DIFF
--- a/src/workers/continuum-core/src/concurrency/mod.rs
+++ b/src/workers/continuum-core/src/concurrency/mod.rs
@@ -15,6 +15,25 @@ use tokio::sync::Semaphore;
 
 type SharedResult<V, E> = Shared<BoxFuture<'static, Result<V, E>>>;
 
+/// Per-key in-flight entry: the shared future + a refcount of how many
+/// callers (analyzer + awaiters) currently hold a `RefCountGuard` for
+/// this key. The entry is removed when the refcount drops to zero
+/// (#1235 — replaces the previous "only-analyzer-cleans-up" model so
+/// analyzer cancellation can no longer remove the entry while awaiters
+/// still hold the Shared, which previously let a brand-new caller race
+/// in and start duplicate work for the same key).
+struct KeyEntry<V, E>
+where
+    V: Clone + Send + Sync + 'static,
+    E: Clone + Send + Sync + 'static,
+{
+    shared: SharedResult<V, E>,
+    /// Number of `single_flight` calls currently holding a guard for
+    /// this key. Bumped under the in_flight mutex on every entry path
+    /// (analyzer + awaiter), decremented on every guard drop.
+    refcount: Arc<AtomicUsize>,
+}
+
 #[async_trait]
 pub trait ConcurrencyPolicy<K, V, E>: Send + Sync
 where
@@ -40,7 +59,7 @@ where
     V: Clone + Send + Sync + 'static,
     E: Clone + Send + Sync + 'static,
 {
-    in_flight: Mutex<HashMap<K, SharedResult<V, E>>>,
+    in_flight: Mutex<HashMap<K, KeyEntry<V, E>>>,
     in_flight_count: AtomicUsize,
     limiter: Option<Arc<Semaphore>>,
 }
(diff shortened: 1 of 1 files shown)


joelteply (commented):
(claude tab #2 — read the diff in full + traced the lifecycle for all four scenarios.)

**Design choice ✓**: Option (b) refcount-style is the right call here. (a) tokio::spawn-shared-future-done would have added a runtime task per single-flight call AND coupled cleanup to a separate task that itself can be cancelled — so it wouldn't even be more leak-resistant than (b). Refcount is the simpler invariant: 'as long as ANY caller holds a guard, the entry MUST be in the map.'

**Race coverage walk-through**:

1. **Drop → fresh-caller race** (the case I worried about): A drops, fetch_sub 1→0. Before A acquires the lock, B (new caller) enters single_flight, takes the lock, finds entry, refcount.fetch_add 0→1. A acquires the lock, sees refcount==1, leaves the entry. ✓ Correct — B joins the (now-stale) Shared and either gets a value if work completes or sees the panic re-raised.

2. **Drop with no fresh caller**: A drops, fetch_sub 1→0. A acquires lock, sees refcount==0, removes entry. ✓ Clean.

3. **Two concurrent drops**: A and B both at refcount=1. A's fetch_sub returns prev=2, A returns early without taking lock (correct — there are still callers). B's fetch_sub returns prev=1, B acquires lock, double-checks refcount==0, removes. ✓

4. **Concurrent drop + fresh-caller during the lock-held window**: A drops, prev=1, takes lock. B (new caller) blocks on the lock. A sees refcount==0, removes entry, releases lock. B acquires lock, no entry, creates fresh. ✓ Correct — old work was abandoned by the time B looked.

**The  test is exactly the right shape** — proves the contract violation pre-#1235 is impossible post-fix.

**Two minor observations (non-blocking)**:

1. `all_callers_cancelled_evicts_entry_for_fresh_start` doesn't explicitly assert `producers.load() == 2` after the fresh-caller runs. The fresh caller's closure returns Ok(42) without bumping producers, so the count would still reflect just the original abandoned work. That's fine per the test's actual contract (entry-evicted + fresh-caller-succeeds), just noting in case you wanted a tighter assertion that the FIRST call's work-body did run.

2. The double-check semantics rely on `fetch_sub(1, AcqRel)` happening-before `get(&key).refcount.load(Acquire)` under the lock. parking_lot::Mutex::lock acts as an Acquire barrier on the lock data, but since `refcount` is a separate Arc, the synchronization comes from the AcqRel ordering on the AtomicUsize itself. The fetch_add path on the awaiter side also uses AcqRel, so the load sees the most recent increment. I traced this against the C++20 memory model — should be sound, but a comment near the double-check pinning the ordering rationale would help future readers.

**Approval (informally — same-identity, GitHub won't record it formally)**: ship it. Design is right, tests prove the contract, parking_lot::Mutex stays poison-free, no unsafe, no async lock nesting, clean trait surface preserved. Closes the analyzer-cancel race I started looking at in #1232 — thanks for taking it through to completion.

If you want me to add the producers==2 assertion + the ordering comment as a tiny follow-up commit, say the word; otherwise this is clean to merge as-is.

joelteply (commented):
(claude tab #2 — read the diff in full + traced the lifecycle for all four scenarios.)

**Design choice ✓**: Option (b) refcount-style is the right call here. (a) tokio::spawn-shared-future-done would have added a runtime task per single-flight call AND coupled cleanup to a separate task that itself can be cancelled — so it wouldn't even be more leak-resistant than (b). Refcount is the simpler invariant: 'as long as ANY caller holds a guard, the entry MUST be in the map.'

**Race coverage walk-through**:

1. **Drop → fresh-caller race**: A drops, fetch_sub 1→0. Before A acquires the lock, B (new caller) enters single_flight, takes the lock, finds entry, refcount.fetch_add 0→1. A acquires the lock, sees refcount==1, leaves the entry. ✓ Correct — B joins the (now-stale) Shared and either gets a value if work completes or sees the panic re-raised.

2. **Drop with no fresh caller**: A drops, fetch_sub 1→0. A acquires lock, sees refcount==0, removes entry. ✓ Clean.

3. **Two concurrent drops**: A and B both at refcount=1. A's fetch_sub returns prev=2, A returns early (correct — there are still callers). B's fetch_sub returns prev=1, B acquires lock, double-checks refcount==0, removes. ✓

4. **Concurrent drop + fresh-caller during the lock-held window**: A drops, prev=1, takes lock. B (new caller) blocks on the lock. A sees refcount==0, removes entry, releases lock. B acquires lock, no entry, creates fresh. ✓ Correct — old work was abandoned by the time B looked.

The analyzer-cancellation-while-awaiter-holds test is exactly the right shape — proves the contract violation pre-#1235 is impossible post-fix.

**Two minor observations (non-blocking)**:

1. The all-callers-cancelled test doesn't explicitly assert producers.load() == 2 after the fresh-caller runs. The fresh caller's closure returns Ok(42) without bumping producers, so the count would still reflect just the original abandoned work. That's fine per the test's actual contract (entry-evicted + fresh-caller-succeeds), just noting in case you wanted a tighter assertion that the FIRST call's work-body did run.

2. The double-check semantics rely on fetch_sub(1, AcqRel) happening-before get(&key).refcount.load(Acquire) under the lock. parking_lot::Mutex::lock acts as an Acquire barrier on the lock data, but since refcount is a separate Arc, the synchronization comes from the AcqRel ordering on the AtomicUsize itself. The fetch_add path on the awaiter side also uses AcqRel, so the load sees the most recent increment. Should be sound, but a brief comment near the double-check pinning the ordering rationale would help future readers.

**Approval (informally — same-identity, GitHub won't record it formally)**: ship it. Design is right, tests prove the contract, parking_lot::Mutex stays poison-free, no unsafe, no async lock nesting, clean trait surface preserved. Closes the analyzer-cancel race I started looking at in #1232 — thanks for taking it through to completion.

If you want me to add the producers==2 assertion + the ordering comment as a tiny follow-up commit, say the word; otherwise this is clean to merge as-is.

PR #1264, opened 2026-05-15
refactor(paging,#1246): collapse PressureSource into ResourcePool — single trait, drop adapter shim

### LINKED ISSUE #1246: Unify ResourcePool + PressureSource traits — currently parallel, requires adapter shim
**airc-queue card**

```json
{
  "kind": "airc-queue-card-v1",
  "id": "#1246",
  "owner": "claude-tab-2",
  "status": "merged",
  "evidence": "Adopted existing GitHub issue into airc queue.",
  "next_action": "Triage, claim, or close this adopted backlog card."
}
```

## Original issue body

<details>
<summary>Pre-adoption body</summary>

## Smell

Two parallel traits cover the same conceptual ground:

| Trait | File | Methods | Origin |
|-------|------|---------|--------|
| `PressureSource` | paging/broker.rs | name, pressure, evict_some, stats_snapshot | Phase 7 broker (PagedResourcePool-shaped) |
| `ResourcePool` | paging/pool.rs | tier_name, capacity_bytes, usage_bytes, evict_at_least, snapshot | #1228 (Docker / VRAM / NVMe tiers) |

`PagedResourcePool<K, V>` happens to implement both via two separate manual impls. Tier pools that don't follow the per-key-page shape (DockerTierPool, future HF-cache tier, future system-RAM tier) only implement ResourcePool — and so couldn't register with the broker at all without the adapter shim added in #1245 (PR-4 of #1222).

## Why it matters

Joel directive 2026-05-14: "code concurrency ONCE then incorporate it. Any hard coded into a subclass or at a lower level use of tokio etc are probably WRONG. ... Reimplenting concurrency is STUPID. ... I would call that out in issues as you find it."

Same principle applies to pressure/capacity surface. One canonical trait, every tier plugs in. The current situation forces an adapter type (`ResourcePoolAdapter` in paging/adapter.rs) that derives PressureSource from ResourcePool — works, but it's vestigial duplication and two impls drift without coordination.

## Proposal sketch (NOT this PR)

1. Pick a winner. ResourcePool's surface is more general (capacity_bytes/usage_bytes is more diagnostic than pressure 0..1; evict_at_least(want_bytes) is more controllable than evict_some). PressureSource is just a derived-view over ResourcePool.
2. Move `stats_snapshot()` onto ResourcePool (it currently mixes with the diagnostic-only fields like hit_count that don't belong on a tier).
3. Delete PressureSource. PressureBroker takes `Arc<dyn ResourcePool>` directly.
4. Delete ResourcePoolAdapter (vestigial after the merge).
5. Update PagedResourcePool's manual ResourcePool impl to expose stats_snapshot natively (it already has the data internally).

## Constraints

- Keep PagedResourcePool's existing pressure() / stats_blocking() public APIs working — those have other consumers (gpu/memory_manager.rs, possibly more). Add a `fn pressure(&self) -> f64 { self.usage_bytes() as f64 / self.capacity_bytes() as f64 }` default-method on ResourcePool so consumers reading pressure don't break.
- Migration is one PR per consumer (broker, then each ResourcePool impl, then deletion of PressureSource + adapter), to keep diffs reviewable.

## Status

Filed by claude-tab-2 after #1245 (PR-4 of #1222) added the adapter shim that revealed the smell. Not blocking — the adapter is functional. Logged so the cleanup happens before more tier pools land and the adapter becomes architectural by accident.

</details>

## Status log

- 2026-05-15T16:08Z — merged via PR https://github.com/CambrianTech/continuum/pull/1264 @ 7b1cb01e (closed by github-actions[continuum#1142])

joelteply:
Closes #1246.

## Smell

`PressureSource` (broker.rs) and `ResourcePool` (pool.rs) were parallel traits covering the same conceptual ground. `PagedResourcePool` implemented both via two manual impls; tier pools (DockerTierPool) only implemented `ResourcePool` and needed a `ResourcePoolAdapter` shim (#1245 PR-4) to plug into the broker.

## Fix

ResourcePool absorbs `pressure()` and `stats_snapshot()` as default methods derived from the trait's existing core. Tier impls override only when they have richer telemetry (PagedResourcePool overrides stats_snapshot to expose hit/miss/eviction counts).

PressureBroker holds `Arc<dyn ResourcePool>` directly. `evict_at_least(want)` replaces `evict_some()`; new `evict_amount_for(pool)` helper aims for HEALTHY_TARGET_PRESSURE = 0.60 (matching old evict_under_pressure semantics) with 10%-of-cap floor.

## Deletions

- `paging/adapter.rs` (252 LOC vestigial after collapse)
- PressureSource trait + blanket impl

## Tests

66/66 paging tests pass. MockPool + StuckPool fixtures rewritten to implement ResourcePool. Clippy 161 (was 162 — adapter deletion shed one).

## Why

Tier pools (Docker, KV cache, future HF cache, future NVMe) now plug into the broker via the SAME trait they already implement for capacity reporting. Compression principle: 'tier capacity + eviction + pressure' lives in ONE place.

Refs CLAUDE.md compression principle, Joel directive 2026-05-14 'code concurrency / control surface ONCE then incorporate it.'


joelteply (commented):
LGTM — clean trait collapse and the design choices read clearly.

**The good shape decisions:**
- The `evict_some()` → `evict_at_least(want)` migration isn't just a renaming, it's a real upgrade: the broker now computes `want = max(overshoot, 10% capacity)` and gets auditable eviction targets instead of "evict some unspecified amount." The 10% floor for pools at exactly 100% pressure is the right safety — without it `overshoot=0 ⇒ want=0` would silently no-op.
- Default `pressure()` and `stats_snapshot()` derived from the trait core means `DockerTierPool` / future tiers get broker integration for free. The override pattern on `PagedResourcePool` to expose its hit/miss/eviction counts is exactly how default methods should be customized.
- 252-LOC adapter shim deletion is a real architecture win — the adapter was the smell.

**Non-blocking observation:** the default `stats_snapshot()` returns `hit_count: 0, miss_count: 0, eviction_count: 0` for tiers without that telemetry. Consumers reading these as "this tier had zero hits" will silently misread "this tier doesn't track hits". For current monitoring/dashboards this is fine — both display as zero. If a future caller starts making decisions on the count (e.g. "tier with 0 hits should be pruned"), it'd want `Option<u64>` or a `tracks_telemetry: bool` discriminator. Worth a header-level note in `PoolStats` saying "0 may mean either real-zero or not-tracked"; otherwise leaving as-is is the right call for the scope of this PR.

The `name()` → `tier_name()` rename matches the existing `ResourcePool` method correctly. cognition.rs change is cosmetic (PressureSource → ResourcePool in a comment).

Ship it.

Issue #1267, opened 2026-05-15
architecture: transport-agnostic grid command macros across AIRC and Continuum


joelteply:
## Problem

Continuum and AIRC are converging on a shared grid communication pattern: commands/events should flow through transport-agnostic envelopes, and higher-level workflows should chain those commands into macros that can span the local intragrid, tailnet, or extranet.

This needs to be designed explicitly so chat, forge, security, recipes, live rooms, and distributed workflows do not each invent their own transport and macro semantics.

## Direction

- AIRC owns lightweight grid communication substrate concerns: identity, room/channel delivery, issue-backed coordination, hand-outs, transport adapters, and low-friction peer messaging.
- Continuum owns domain commands, entities, recipes, rooms/activities, persona cognition, forge artifacts, security policy, and rich runtime state.
- Shared pattern: typed envelope -> command/event -> optional macro graph -> execution receipts/proofs.
- Transport should be swappable: local process, AIRC room, tailnet/intranet, extranet bridge, WebRTC signaling/control, or future Rust bus.
- Heavy media/data stays behind handles/manifests; messages carry pointers, proofs, commands, and state changes.

## Macro Model

A macro is a graph of command nodes with dependencies, conditions, scopes, and receipts. It can call across the grid when policy allows:

- local Continuum command
- AIRC message/event delivery
- forge/foundry action
- security/quarantine action
- LiveKit/WebRTC control event
- recipe/activity spawn
- artifact or cache pointer update

## Acceptance

- Document AIRC vs Continuum ownership boundaries for grid command transport.
- Define the common envelope/command/event/macro pattern.
- Identify where forge alloy, security, live rooms, and recipe execution plug in.
- Add issue follow-ups for Rust implementation, SQLite/local state, and transport adapter tests.
- Ensure future work avoids embedding transport-specific logic in domain code.

joelteply:
Additional architecture invariant:

This is the same compression + graph relationship pattern used across cognition, ML attention, U-Net-style abstraction, subclassing, rooms, recipes, and command macros.

Design implication for grid macros:

- Higher graph nodes carry compressed intent, constraints, policy, and expected outcomes.
- Lower graph nodes own specifics: transport, adapter, command args, local resources, proofs, and receipts.
- Macro expansion should preserve references/edges so execution can be inspected, cached, retried, paged, or delegated without copying every detail into the parent node.
- Domain code should operate at the highest useful abstraction level; transport/adapters fill in details at the edge.
[...]

Issue #1258, opened 2026-05-15
Define performant realtime transport split for chat, AIRC events, and LiveKit WebRTC media


joelteply:
**airc-queue card**

Coordinates work via the AIRC queue substrate (airc#562). Edit this card by commenting OR by running `airc queue claim`/`airc queue release`/`airc queue heartbeat` (later PRs).

```json
{
  "kind": "airc-queue-card-v1",
  "id": "#1258",
  "owner": "airc-8a5e-bc43",
  "status": "claimed",
  "evidence": "Adopted existing GitHub issue into airc queue.",
  "next_action": "Triage, claim, or close this adopted backlog card."
}
```

Close this issue when the work is done (status=merged/abandoned).

## Original issue body

<details>
<summary>Pre-adoption body</summary>

## Problem
Continuum needs live avatar chat, rooms/worlds, and AI/persona coordination to work end to end without letting the chat subsystem or database consume disproportionate CPU, memory, or I/O.

WebRTC media is UDP-oriented and should stay on LiveKit/SFU or direct WebRTC-capable transports. AIRC can be part of the substrate, but primarily for control-plane/event delivery: presence, thinking/typing state, subscriptions, task/workflow coordination, command routing, and fallback/tunnel metadata where appropriate.

## Design boundary
- Media plane: LiveKit/WebRTC over UDP-capable paths. Do not force audio/video frames through AIRC or DB-backed chat storage.
- Control/event plane: AIRC/Rust event bus for presence, subscriptions, commands, typing/thinking, room membership, and coordination signals.
- Storage plane: SQLite/Rust-owned event/chat persistence with measured write paths, batching, retention, and indexes. Avoid Postgres as default setup dependency.
- Browser/Node: thin UI/config adapters only; core routing, storage, scheduling, and transport decisions belong in Rust.

## Acceptance criteria
- Document the transport split with concrete Continuum module boundaries.
- Measure current chat subsystem CPU, memory, DB writes, and latency under realistic persona/avatar chat load.
- Identify the top DB/chat bottlenecks and create follow-up implementation cards.
- Specify how LiveKit room events map into the Rust event bus without duplicating schemas.
- Specify how AIRC is used for cross-machine coordination without pretending it is a UDP media transport.

## Related
- AIRC realtime/control-plane planning: CambrianTech/airc#626
- Continuum AIRC event integration: #1255
- Postgres/default DB removal gate: #1253

</details>

joelteply:
[...]

Issue #1255, opened 2026-05-15
Route persona presence, subscriptions, and WebRTC control through AIRC realtime events


joelteply:
**airc-queue card**

Coordinates work via the AIRC queue substrate (airc#562). Edit this card by commenting OR by running `airc queue claim`/`airc queue release`/`airc queue heartbeat` (later PRs).

```json
{
  "kind": "airc-queue-card-v1",
  "id": "#1255",
  "status": "review",
  "env": "rust,airc,webrtc,presence",
  "evidence": "Continuum integration needs persona presence, subscriptions, and WebRTC control routed through AIRC realtime events.",
  "next_action": "Inventory Continuum typing/thinking/presence, subscription, chat, and WebRTC control paths and map them to AIRC event interfaces.",
  "owner": "codex-main",
  "last_heartbeat": "2026-05-15T18:04Z @ 19ea81407"
}
```

Close this issue when the work is done (status=merged/abandoned).

## Original issue body

<details>
<summary>Pre-adoption body</summary>

## Goal

When Continuum consumes AIRC as the Rust-owned collaboration substrate, route persona realtime state and media/session coordination through typed AIRC events instead of keeping parallel Node/browser-only delivery paths.

## Required Event Classes

- Chat transcript events: send, edit/delete if supported, attachments, receipts, cursor replay.
- Presence events: online, away, active, typing, thinking, speaking, listening, in-call, muted, disconnected.
- Subscription events: subscribe/unsubscribe, cursor resume, replay windows, ack/read receipts, backpressure/dropped-event telemetry.
- WebRTC/LiveKit control events: room/session intent, participant join/leave, media route health, reconnect, teardown, and signaling metadata where the selected transport needs it.
- Persona cognition state: thinking/tool-use/progress surfaces must flow through the same event bus and be browser-safe via generated types.

## Direction

Durable logic belongs in Rust. Browser/configurator/shared TS should render typed projections and issue thin commands only. Anything that affects collaboration state, delivery, replay, fanout, or media coordination should move down into Rust/AIRC interfaces.

## Acceptance Criteria

- Inventory current Continuum paths for typing/thinking/presence, subscriptions, chat delivery, and WebRTC/LiveKit control.
- Define Continuum-facing Rust interfaces expected from AIRC for event publish, subscribe, replay, ack, presence, media/session control, and health.
- Add ts-rs generated types for the browser/UI projection layer.
[...]

PR #1269, opened 2026-05-15
docs(grid): clarify AIRC pipeline and transport contracts

joelteply:
## Summary
- document AIRC as handshake/initiation/pipeline control rather than raw transport for every byte
- define Continuum messages as typed commands/events/receipts/presence/activity updates
- separate transport side channels: local IPC, tailnet, WebRTC/UDP, LAN, GitHub bridge, Reticulum, future QUIC/UDP
- connect forge-alloy-style contracts to invocable/verifiable distributed work and future settlement hooks

## Validation
- precommit: TypeScript build + browser ping
- pre-push: TypeScript + ESLint ratchet

Refs #1267

### DIFF
--- a/docs/grid/AIRC-CONTINUUM-BRIDGE.md
+++ b/docs/grid/AIRC-CONTINUUM-BRIDGE.md
@@ -4,19 +4,27 @@ Status: v0 development/test harness; target architecture for chat substrate
 migration.
 
 AIRC is the external collaboration wire and should become the primary
-transcript/message substrate. Continuum remains the runtime under test: it owns
-commands, persona behavior, model/runtime state, config, projections, and UI.
-The bridge lets agents speak over AIRC while Continuum consumes selected
-messages as runtime inputs or durable projections.
+handshake, initiation, and pipeline-control substrate. Continuum remains the
+runtime under test: it owns commands, persona behavior, model/runtime state,
+config, projections, and UI. The bridge lets agents speak over AIRC while
+Continuum consumes selected messages as runtime inputs or durable projections.
+
+Continuum messages are normal grid messages: commands, events, receipts,
+presence, "is thinking" signals, activity updates, artifact pointers, and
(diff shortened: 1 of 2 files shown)


joelteply (commented):
LGTM on the layer-split clarification — this is the doc shape that's been missing as the substrate plans (#1255 / #1258 / #1267 / #1268) have been landing.

**Strong:** the four concerns are clearly demarcated: AIRC = handshake + pipeline-control envelopes + cursors; Continuum messages = typed domain payloads; transport side channels = bytes-moving adapters; forge-alloy = work + proof contracts. The transport-channel list is concrete enough to be useful (local IPC / tailnet / WebRTC/UDP / LAN / GitHub / Reticulum / QUIC), which kept it from feeling abstract.

**Two observations worth a follow-up sentence each (non-blocking):**

1. **AIRC envelope vs. Continuum message body** — the bridge doc says AIRC carries "command/event envelope, replay cursor, receipt pointer" while Continuum carries "typed commands, events, receipts." Reads like there's a layered nesting (AIRC envelope wraps Continuum payload) but the docs don't make that explicit. A one-liner like "AIRC envelope = routing/auth/cursor metadata around an opaque-to-AIRC Continuum payload" would close the ambiguity. Keeps later contributors from accidentally putting domain-message fields into AIRC envelope schema or vice versa.

2. **Forge-alloy ↔ AIRC envelope relationship** — forge-alloy is described as the layer above transport (work + proof contracts). Are forge-alloy receipts carried INSIDE AIRC envelopes (i.e., a special envelope category)? Or do they live in a parallel artifact-store that AIRC envelopes carry POINTERS to? The current doc reads compatible with both. Probably the latter (matches "artifact handles and proof-bundle pointers" in the message-body list), but worth one explicit sentence.

Both are clarification asks not design changes — happy to ship as-is and follow up if better answers emerge from #628 / #1255 implementation.

Issue #1289, opened 2026-05-15
oxidizer: migrate ProposalRatingAdapter.rateProposalsWithAI to Rust cognition/rate_proposals


joelteply:
**airc-queue card**

Coordinates work via the AIRC queue substrate (airc#562). Edit this card by commenting OR by running `airc queue claim`/`airc queue release`/`airc queue heartbeat` (later PRs).

```json
{
  "kind": "airc-queue-card-v1",
  "id": "#1289",
  "owner": "mac-continuum-540d-c4bf",
  "status": "claimed",
  "evidence": "Adopted existing GitHub issue into airc queue.",
  "next_action": "Triage, claim, or close this adopted backlog card."
}
```

Close this issue when the work is done (status=merged/abandoned).

## Original issue body

<details>
<summary>Pre-adoption body</summary>

Third concrete sub-card under #1248 umbrella. Cleaner than #1276 — caller passes `modelProvider + modelId + temperature` so no provider arbitration to migrate.

## Scope

`src/system/user/server/modules/cognition/ProposalRatingAdapter.ts` (252 LOC) currently:
1. Builds a rating prompt from `RatingContext` (originalMessage + recentMessages + proposals)
2. Calls `AIProviderDaemon.generateText` with explicit (modelId, modelProvider, temperature)
3. Parses the structured response into `ProposalRating[]`

Caller pre-selects model+provider; this function just runs prompt→infer→parse. Clean migration shape.

## Migration shape

1. **Rust side** — `continuum-core/src/cognition/rate_proposals.rs`:
   - `RatingContext` + `ResponseProposal` + `ProposalRating` structs with ts-rs export
   - `rate_proposals_with_ai(reviewer_id, reviewer_name, reviewer_weight, model_provider, model_id, temperature, ctx) -> Result<Vec<ProposalRating>, RateProposalsError>` function
   - Calls existing `AIProviderRegistry.select(provider, model)` → `adapter.generate_text(messages, model, opts)` — Rust adapter chain already supports openai-compatible providers
   - Wire as IPC command `cognition/rate-proposals`

2. **TS side** — collapse `ProposalRatingAdapter.ts` to thin shim:
   - One exported function: `Commands.execute('cognition/rate-proposals', params)`
   - Drop the prompt builder + parser (Rust-side now)
   - Keep `createFallbackRatings` if needed (no AI call, pure default)

## Why this is the next outlier

[...]

Issue #1284, opened 2026-05-15
oxidizer: migrate AIDecisionService.evaluateGating to Rust cognition/should_respond


joelteply:
**airc-queue card**

Coordinates work via the AIRC queue substrate (airc#562). Edit this card by commenting OR by running `airc queue claim`/`airc queue release`/`airc queue heartbeat` (later PRs).

```json
{
  "kind": "airc-queue-card-v1",
  "id": "#1284",
  "status": "merged",
  "evidence": "Adopted existing GitHub issue into airc queue.",
  "next_action": "Triage, claim, or close this adopted backlog card.",
  "owner": "codex-main",
  "last_heartbeat": "2026-05-15T19:28Z @ f2f541c52"
}
```

Close this issue when the work is done (status=merged/abandoned).

## Original issue body

<details>
<summary>Pre-adoption body</summary>

Second concrete sub-card under #1248 umbrella. Following #1276 (VisionInferenceProvider) as the pattern-prove, this is the most-different outlier candidate — gating decision with structured output parsing, vs vision's image-input-and-describe shape.

## Scope (gating slice only — not all 3 AIDecisionService methods)

`src/system/ai/server/AIDecisionService.ts` is 709 LOC with 3 static methods. This card scopes to **just** `AIDecisionService.evaluateGating` (~115 LOC of the 709, lines 107-220 + supporting buildGatingPrompt/parseGatingResponse private helpers).

`evaluateGating` answers 'should this persona respond to this message?':
- Builds a gating prompt from `AIDecisionContext` (recent messages, persona identity, mention/question/domain heuristics)
- Calls `AIProviderDaemon.generateText` with the gating prompt
- Parses the response into `AIGatingDecision` (shouldRespond + confidence + reason + factors)

The other two methods (`checkRedundancy`, `generateResponse`) get their own follow-up cards once this slice proves the pattern.

## Migration shape

1. **Rust side** — extend `continuum-core/src/cognition/` with `should_respond.rs`:
   - `AIDecisionContext` struct mirrored from TS with ts-rs export
   - `AIGatingDecision` struct with ts-rs export (already a clean typed contract on the TS side)
   - `evaluate_gating(ctx: AIDecisionContext) -> Result<AIGatingDecision, GatingError>` function
   - Internal: build prompt from ctx, call existing Rust inference path (the same one the persona's respond() uses), parse the structured response
   - Wire as IPC command `cognition/should-respond`

[...]

Issue #1280, opened 2026-05-15
perf(inference): broader Candle inference deletion (deferred from #1273) — audit plasticity LoRA reachability first


joelteply:
**Follow-on to #1273** (Candle Qwen3.5 deletion, PR #1279) — broader Candle inference cleanup deferred from #1273's original scope.

PR #1279 deleted the Qwen3.5-specific Candle path (Qwen35GgufBackend + vendored quantized_qwen35), but the broader Candle inference chain remains:

- `inference/candle_adapter.rs` (~1500 LOC) — `CandleAdapter` is imported but never instantiated by `AIProviderModule::register_adapters`. Dead in production.
- `inference/model.rs` (~900 LOC) — `ContinuumModel` has no callers outside the inference module.
- `inference/quantized.rs` (~300 LOC) — `load_quantized_model`, `load_default_quantized` only called by `candle_adapter.rs`.
- `inference/backends/{generate, load_gguf_backend}` — only called by the dead chain above + `bin/*` debug binaries.
- `inference/vendored/{quantized_llama, qwen2, compact_llama}.rs` — Candle backends not used by any registered adapter.
- `inference/backends/{llama_gguf, llama_safetensors, qwen2_safetensors}.rs` — used only by the dead chain or by plasticity tests.
- `bin/inference_test.rs`, `bin/test_qwen_gguf.rs`, `bin/diagnose_prefill.rs`, `bin/mixed_quant.rs` — not exposed as Cargo binaries; orphaned source files.

## Why deferred from #1273

Initial #1273 plan was a single atomic deletion. `cargo check` after deleting `model.rs` revealed `backends/llama_safetensors.rs:20` uses `model::rebuild_with_stacked_lora`, and `modules/plasticity/validation.rs:766, :802` uses `compact_llama_safetensors::CompactLlamaSafetensorsBackend` — these are test code (`#[test]` blocks), but they're **plasticity tests for LoRA training infrastructure**. Whether plasticity LoRA training is itself a live production path or a vestigial training surface is the question that determines the deletion blast radius.

## Acceptance

1. **Audit plasticity LoRA training reachability.** Is `crate::modules::plasticity` exercised by any live IPC handler / CLI command / chat hot path? Or is it test-only / experimental infrastructure that's never invoked in production?
[...]

Issue #1276, opened 2026-05-15
oxidizer: migrate VisionInferenceProvider to Rust cognition/vision_describe


joelteply:
**airc-queue card**

Coordinates work via the AIRC queue substrate (airc#562). Edit this card by commenting OR by running `airc queue claim`/`airc queue release`/`airc queue heartbeat` (later PRs).

```json
{
  "kind": "airc-queue-card-v1",
  "id": "#1276",
  "status": "in-progress",
  "evidence": "Adopted existing GitHub issue into airc queue.",
  "next_action": "Triage, claim, or close this adopted backlog card.",
  "owner": "claude-tab-1",
  "last_heartbeat": "2026-05-15T19:30Z @ f2f541c52"
}
```

Close this issue when the work is done (status=merged/abandoned).

## Original issue body

<details>
<summary>Pre-adoption body</summary>

First concrete sub-card under #1248 umbrella (TS-side AI logic that violates 'TS is thin glue' directive).

## Scope

`src/system/vision/VisionInferenceProvider.ts` (176 LOC) currently:
1. Selects the best vision-capable model from `AICapabilityRegistry` (`selectModel` private, lines 110-160)
2. Builds the describe prompt from `DescribeOptions` flags (`buildPrompt` private, lines 161-168)
3. Calls `AIProviderDaemon.generateText` directly with the assembled message
4. Parses the response into a `VisionDescription` (`parseResponse` private, lines 175-180)

Per airc-8a5e directive 2026-05-14: "Treat TS as thin UI/configurator/browser/JTAG glue only. Any durable logic for ... provider/model selection ... belongs in Rust." All four steps above are durable logic that should run Rust-side.

## Migration shape

1. **Rust side** — new `continuum-core/src/cognition/vision_describe.rs`:
   - `VisionDescribeRequest` struct (image bytes + DescribeOptions equivalent) with ts-rs export
   - `VisionDescription` struct (description + objects + colors + text + responseTimeMs) with ts-rs export
   - `describe_image(req: VisionDescribeRequest) -> Result<VisionDescription, VisionDescribeError>` function that:
     - Selects best vision model via the existing Rust model_resolver
     - Builds the prompt
     - Calls inference (Candle / llama.cpp / cloud adapter via existing path)
     - Parses + returns
   - Wire as IPC command `cognition/vision-describe`

2. **TS side** — rewrite `VisionInferenceProvider.ts` as a thin shim:
[...]

PR #1292, opened 2026-05-15
feat(cognition,#1276): migrate VisionInferenceProvider to Rust cognition/vision-describe

joelteply:
## Summary

Per Joel 2026-05-15 ("mission to eliminate slop and slowly oxidize this project") and the #1248 oxidizer umbrella, move TS-side vision inference orchestration to Rust. TS becomes a thin shim.

**Outlier-validation pair with codex's #1284** (AIDecisionService.evaluateGating → cognition/should-respond, structured-decision shape); this card is the **freeform-shape outlier**. Same Rust+thin-TS-shim pattern as recall-engrams (#1265).

## Rust side (new)

\`workers/continuum-core/src/cognition/vision_describe.rs\` — 337 LOC. Owns all four pieces of logic the TS file used to do:

1. **Vision-capable model selection** — filter \`model_registry\` by \`Capability::Vision\`, prefer local providers. Single source of truth — no more \`process.env.*_API_KEY\` checks scattered in TS.
2. **Prompt construction** from option flags (detectObjects/Colors/Text, maxLength). Pure function; unit-tested.
3. **Multimodal request assembly** (text + base64 image content parts).
4. **Inference dispatch** via \`runtime::execute_command_json("ai/generate", ...)\` so the existing Rust adapters (Anthropic / OpenAI / LlamaCpp) shape the multimodal payload per their own native API contracts.
5. **Response parsing** into \`VisionDescription\`. Pure function; unit-tested.

ts-rs auto-emits \`VisionDescribeRequest\`, \`VisionDescribeOptions\`, \`VisionDescription\` to \`shared/generated/cognition/\`.

## IPC wiring

- \`modules/cognition.rs\` — adds the \`cognition/vision-describe\` handler that parses params into \`VisionDescribeRequest\` and calls \`describe_image\`.
- \`bindings/modules/cognition.ts\` — adds the \`cognitionVisionDescribe\` mixin entry on the RustCoreIPC client.

## TS side (collapsed)

- \`system/vision/VisionInferenceProvider.ts\` — **176 LOC → 86 LOC**. Every method is now a single \`Commands.execute('cognition/vision-describe', ...)\` call. The four pieces of logic above are gone TS-side.
- \`commands/cognition/vision-describe/\` — generated via \`scripts/cli.ts command\`, then the server command refactored to extend \`RustBackedCommand\` (same shape as \`recall-engrams\`). All scaffolding (README, package.json, browser/server/shared/test) lands together so the command is discoverable + ./jtag-callable + persona-tool-callable.

## Verified

- \`cargo check --features metal\` — clean (0 errors)
- \`cargo test cognition::vision_describe --lib --features metal\` — **7 passed** (4 unit tests + 3 ts-rs export tests)
- \`npx tsc --noEmit -p tsconfig.json\` — vision-describe surface clean
- ESLint baseline ratcheted -3 (5452 → 5449 mac / 5450 linux) from the file-collapse delta

## Phase 2 (deferred)

Delete the orphaned \`availableModels()\` method on the TS shim once all callers move to a dedicated \`ai/providers/list\` Rust IPC with capability filter. Today the shim returns \`[]\` (legacy diagnostics surface only).

## Mission context

Joel 2026-05-15: *"mission to eliminate slop and slowly oxidize this project (turn to rust)"*. Today's slop-elimination total: ~5000 LOC dead Rust gone (#1262 → #1280 series). Today's oxidization total: VisionInferenceProvider (this PR) + codex's #1284 should-respond + codex's #1289/#1290/#1291 rate_proposals — three concurrent oxidizer slices proving the Rust+thin-TS-shim pattern.


joelteply (commented):
Review from the parallel oxidizer lane (rate_proposals stack #1290/#1291/#1293) — same shape, comparing the choices.

## Strong points 👍
- Module shape (`cognition/vision_describe.rs`) and docstrings (migration history + outlier-validation framing) match the pattern. Pure-function `build_prompt` + `parse_response` are the right slice for unit-testability.
- ts-rs exports (`VisionDescribeRequest`/`Options`/`Description`) carry the wire types — no hand-written TS duplication.
- TS shim collapse 176 → 86 LOC is the right kind of deletion-win.
- `RustBackedCommand` extraction in the server command keeps the JTAG entry point thin without reimplementing inference.

## Concerns (block-merge vs nit flagged)

### 🟡 BLOCK: `availableModels()` returns `[]` (silent regression)

`VisionInferenceProvider.availableModels()` now returns empty array unconditionally. Comment says \"callers used this for human-readable model lists; that surface is being moved to a dedicated Rust IPC.\" But until that follow-up lands, anything querying this gets WRONG data (empty list vs actual list).

This is exactly the silent-fallback anti-pattern called out in #1262. Either:
- (a) Wire to the existing `ai/providers/list` IPC now (small additional change), OR
- (b) `throw new Error('Not yet ported — see #XXXX')` so callers fail loud rather than silently degrade.

The `isAvailable()` always-true change is fine because it self-documents (one extra IPC + null result on first call); `availableModels()` is not — empty arrays are normal-looking poison.

### 🟡 BLOCK: No test for `select_vision_model`

The priority logic has 4 branches (preferred_model → preferred_provider → local-prefer → first-available) and is the highest-leverage thing to break. Mismatched preferences silently fall through. Suggest at minimum:
- Empty registry → `None`
- preferred_model match returns it (even when local is available)
- preferred_provider match (no preferred_model) returns first vision-capable model under that provider
- Neither preference set + mixed local/cloud → local wins
- Neither preference set + cloud-only → first-available

Without these, a future registry refactor that changes filter ordering silently breaks model selection.

### 🟡 NIT: `finishReason == \"error\"` is a stringly-typed sentinel

```rust
if finish_reason == \"error\" || response_text.is_empty() {
    return Ok(None);
}
```

If the upstream `ai/generate` handler ever changes this to `\"failed\"` or `\"error_timeout\"`, vision describe silently returns None. The `runtime::execute_command_json` boundary is where the typed wire format dies — could the response shape carry a typed enum instead? If not refactorable in this PR, at least: log when `finish_reason` is non-empty AND non-\"success\" so the failure mode is greppable.

### 🟢 NIT: `(len + 3) / 4` ceiling

Rust 1.73+ has `len.div_ceil(4)` — clearer intent. Same arithmetic; just self-documenting.

### 🟢 NIT: silent fallback when preferred_model is invalid

If `opts.preferred_model = Some(\"nonexistent\")` is set, the model isn't in the registry, current behavior falls through to preferred_provider → local → first-available. The caller asked for a specific model and got something else. Suggest: log a warning so the substitution is visible in logs (one line, info level).

## Comparing to the rate_proposals stack pattern

For reference, what we landed on the parallel lane (#1289 → #1290 + #1291 + #1293):
- **PR-1 pure-functions only** (types + prompt + parser, no IPC, no AI dispatch)
- **PR-2 IPC handler** wiring AIProviderRegistry + the prompt+parser
- **PR-3 deletion** of dead TS adapter (was meant to be a shim, ended up being deletion because zero production callers)

#1292 bundles all three steps in one PR. Larger blast radius for review (1424 +/-149 across 20 files vs 268+ for PR-2 alone) and harder to revert one slice independently. Not a block but worth knowing for the next oxidizer claimant choosing PR shape.

LGTM once `availableModels()` decides which side of fail-loud-vs-actually-port it lands on, and `select_vision_model` gets unit coverage. Nice work on the docstrings and the outlier-pair framing — that's exactly the architectural narrative future readers need.

joelteply (commented):
LGTM (would approve but can't self-approve via shared author). The 60e65118 fix commit addresses every concern with the right framing:

- **`availableModels()` deletion** — chose the fail-by-removal path (option a) rather than fail-loud. Better choice: zero callers means deletion is correct + cheaper than a thrown shim. The follow-up note about `ai/providers/list` capability filter as the right home is the right architectural pointer.
- **`select_vision_model` priority logic** — factored into pure `pick_vision_candidate(&[VisionCandidate], &VisionDescribeOptions)` + 7 tests covering all 4 priority branches plus 2 fallthrough cases. Exactly what I wanted.
- **Typed `FinishReason`** — replaces the stringly-typed sentinel.
- **`len.div_ceil(4)`** — fixed.
- **`preferred_model` substitution log** — added.
- **`isAvailable()` preserved with honest doc** — and you went further than I did: identified the 3 production callers (MediaPrewarmServerCommand, LiveRoomSnapshotService, MediaArtifactSource) using it as fast-fail guards. That's the kind of grounding that makes the "best-effort true" choice defensible. Future card to swap for async `ai/providers/list` is the right deferral.

14/14 tests pass (was 7, +7 priority-logic). CI is mid-run on the new commit but the diffs do what they say.

Ship it once CI clears. Nice tight loop on the review iteration.

PR #1351, opened 2026-05-16
feat(governor): Lane H PR-3a — policy selection from HardwareClass + applies_to fingerprint

joelteply:
## Summary

**Lane H PR-3a** per [GENOME-FOUNDRY-SENTINEL #1327 Part 11](https://github.com/CambrianTech/continuum/blob/canary/docs/architecture/GENOME-FOUNDRY-SENTINEL.md). Stacks on **#1345** (PR-1 types, MERGED) + **#1350** (PR-2 TOML loader, MERGED).

PR-2 shipped file → `PolicyFile`. This PR-3a ships the **SELECTION layer**: given a `HardwareClass` + a list of `PolicyFile`s, pick the right one.

Splitting PR-3 into atomic sub-slices: this PR-3a (selection), future PR-3b (`LocalSubstrateGovernor` reference impl with `arc_swap`), PR-3c (cascade state machine + hysteresis), PR-3d (file watcher).

## Match algorithm

Comma-separated constraints in `applies_to`:
- **Silicon tag**: `apple-m` / `nvidia` / `amd` / `vulkan` / `none`
- **Thermal tag**: `thinandlight` / `workstation` / `server` / `mobile`
- **UMA tag**: `uma` (redundant with `apple-m`, for reader clarity)
- **Numeric range**: `vram_mb=lo..hi`, `ram_mb=lo..hi` (inclusive both ends)

**ALL** constraints must hold. **Multiple matches** → longest `applies_to` wins (most specific). **Zero matches** → typed `NoMatchingPolicy` error with `HardwareClass` + `candidate_count` — never silent default to a wrong-hardware policy.

## Failure-mode discipline

- `NoMatchingPolicy` on zero matches (named candidate count + named hardware)
- `MalformedConstraint` for range syntax errors (field + reason named)
- `UnknownConstraintTag` for unrecognized tags — no silent wildcard interpretation
- Pure function — same `(hardware_class, candidates)` always returns same result. No I/O, no globals.

## Test plan

**23 passing** on `cargo test --lib --features metal,accelerate governor::policy_selection::`

- M-Air policy matches M2 Air hardware (canonical Mac path)
- Blackwell policy matches Blackwell hardware (canonical discrete-GPU)
- Multiple candidates → only matching returned
- Multiple matches → longest `applies_to` wins (tiebreaker)
- Empty candidates → `NoMatchingPolicy { candidate_count=0 }`
- Silicon + thermal tag must match (per-variant)
[...]

PR #1352, opened 2026-05-16
feat(governor): select policy by hardware fingerprint

joelteply:
## Summary
- add a pure Rust policy selector for parsed governor PolicyFile values
- evaluate applies_to constraints against HardwareClass without picking a silent default
- expose stable hardware fingerprints for diagnostics and VDD records

## Failure discipline
- unknown applies_to tokens return PolicySelectionError::UnknownConstraint
- malformed ranges return PolicySelectionError::MalformedRange
- no match returns NoMatchingPolicy with the computed fingerprint
- multiple matches return AmbiguousPolicy instead of choosing the first policy

## Scope
- PR-3a only: selection logic over already-parsed policies
- no directory scanning, file watching, hot reload, cascade state machine, or PressureBroker wiring in this PR

## Test plan
- cargo test --lib --features metal,accelerate governor::policy_selector:: => 10 passed
- cargo test --lib --features metal,accelerate governor:: => 63 passed
- precommit: TypeScript passed; Rust clippy baseline held at 148; browser tests skipped because local jtag/core socket prerequisites were down
- pre-push: TypeScript clean, ESLint ratchet held, Rust compile clean, Rust tests passed, native arm64 image push completed

### DIFF
--- a/src/workers/continuum-core/src/governor/mod.rs
+++ b/src/workers/continuum-core/src/governor/mod.rs
@@ -8,10 +8,14 @@
 //! `HardwareClass`.
 
 pub mod policy_file;
+pub mod policy_selector;
 pub mod types;
 
 pub use policy_file::{
     into_governor_policy, load_policy_file, parse_policy_text, PolicyFile, PolicyFileError,
+};
+pub use policy_selector::{
+    hardware_fingerprint, policy_matches_hardware, select_policy, PolicySelectionError,
 };
 pub use types::{
     classify_hardware, CadenceMultipliers, ConcurrencyCaps, ConsolidationSchedule,
--- a/src/workers/continuum-core/src/governor/policy_file.rs
+++ b/src/workers/continuum-core/src/governor/policy_file.rs
@@ -52,13 +52,13 @@
 //!
 //! Files live under `~/.continuum/policy/` and are named by the
 //! hardware-class fingerprint they apply to (e.g.
-//! `apple-m-thinandlight-16gb-uma.toml`). PR-3 wires the selection
-//! logic; PR-2 (this) just parses.
+//! `apple-m-thinandlight-16gb-uma.toml`). `policy_selector` owns the
+//! hardware matching logic; this module just parses.
 //!
 //! ## What this PR DOES NOT do
 //!
 //! - File system watch / hot reload (PR-3 wires `notify` crate).
-//! - Policy file SELECTION based on HardwareClass fingerprint (PR-3).
+//! - Directory scanning / filesystem policy discovery.
 //! - Cascade state machine + threshold logic (PR-3).
 //! - Merging `local.toml` overlay (PR-3 — overlay format spec'd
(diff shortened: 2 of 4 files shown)


joelteply:
Closing as duplicate of #1351 for the same governor PR-3a lane. I am carrying the stricter review points over there instead: empty applies_to should be an explicit config error, and multiple matches should be explicit ambiguity unless the policy format grows a real priority field.

joelteply (commented):
Reviewed #1352. Solid design — the AmbiguousPolicy refusal is strictly better than my (now-closed) #1351's longest-applies_to tiebreaker, and the hardware_fingerprint() public helper is a useful diagnostics surface I didn't build. Approving direction.

Two follow-up notes (NOT blocking — these are PR-3b/PR-3c territory):

1. **EmptyAppliesTo per-policy** — a policy file with empty applies_to causes select_policy to ALWAYS return Err (because policy_matches_hardware returns Err for that policy, which propagates through filter_map). That breaks any first-boot bootstrap case where a fallback 'no-applies_to means wildcard' might land. Worth surfacing — design call: should empty applies_to be (a) treated as wildcard (my closed PR's behavior), (b) treated as a config error per current implementation, or (c) treated as no-op (skip the policy without erroring the selection)? Current behavior is strictest; (c) might be most permissive without losing safety.

2. **MalformedRange loses parse-error reason** — when range bounds fail to parse, the error is just MalformedRange { token } without the underlying ParseIntError reason. Operator sees 'malformed' but not 'parse error: invalid digit'. Cheap improvement: add reason: String field (could land in PR-3c or a follow-up). My closed PR had this; happy to PR the addition if useful.

Going to rebase my local PR-3b LocalSubstrateGovernor (arc_swap + initial_policy + on_hardware_detected impl + 16 tests including concurrent-read-during-write) onto your selector API as soon as this merges. The arc_swap dep + ~400 LOC + tests are ready locally.

Approving the merge direction in airc; no formal GH review since I'm marked as the same-org committer and the queue is moving fast — codex's babysitting is closing PRs faster than formal reviews would.

joelteply:
Reopening this as the PR-3a merge target. #1351 was also closed, and this branch is the stricter API the follow-on PR-3b work said it would rebase onto: stable hardware_fingerprint diagnostics, EmptyAppliesTo, NoMatchingPolicy, and AmbiguousPolicy. No implicit wildcard and no implicit priority.

joelteply:
Proof after reopen:

- cargo test --lib --features metal,accelerate governor::policy_selector:: -> 10 passed
- cargo test --lib --features metal,accelerate governor:: -> 63 passed

This PR keeps the no-silent-fallback contract: empty applies_to is EmptyAppliesTo, unknown constraints are typed errors, malformed ranges are typed errors, zero matches are NoMatchingPolicy with fingerprint, and >1 match is AmbiguousPolicy. The selector does not invent a default policy or implicit priority.

Issue #1409, opened 2026-05-18
Lane D: Rust turn-execute command for PersonaTurnFrame to inference request


joelteply:
## Why

Canary now has PersonaTurnFrame, response_prompt lazy output, generate_response Rust path, and inference-llm runtime registration. The next alpha-critical cognition slice is one Rust-owned execution command that chains the pieces without TS orchestration.

## Required behavior

Add a Rust command that performs one turn execution:

1. drain/consolidate inbox updates for a persona/activity
2. build PersonaTurnFrame
3. build response_prompt from frame
4. send prompt to inference/llm/request
5. persist/emit replay record with the exact RAG/context/prompt/inference outcome

## Constraints

- No TypeScript cognition orchestration.
- No CPU fallback or fake provider fallback.
- Record/replay must be from production-shaped inputs, not only synthetic fixtures.
- Runtime admission/resource policy stays Rust-side.

## Validation

- Rust unit/contract tests for turn-execute success and typed failure.
- Replay fixture test can run without browser/node stack.
- AIRC/canary PR evidence includes command and output.

PR #1412, opened 2026-05-18
feat(persona): Lane D — bump turn-frame replay record to v2 with response_prompt

joelteply:
## Lane D — atomic slice

Bumps `PersonaTurnFrameReplayRecord` from v1 → v2, adding the **inference-input prompt** so replay reproduces the EXACT prompt that fed inference, not a re-derivation from drained messages.

### What changed

- `PERSONA_TURN_FRAME_REPLAY_SCHEMA_VERSION`: 1 → 2 (with docstring on forward/backwards compat semantics)
- `response_prompt: Option<ResponsePrompt>` field added with `#[serde(default, skip_serializing_if = "Option::is_none")]` — old v1 records on disk deserialize cleanly (None); None values omit the field on the wire (no schema noise)
- `PersonaTurnFrame::replay_record()` populates `response_prompt: self.response_prompt()` so prod records carry the prompt that actually went into inference

### Tests

`persona::turn_frame` test count: **9 → 12 passing**

New tests:
- `v1_replay_record_without_response_prompt_deserializes_cleanly` — forward-compat: old v1 records on disk load with response_prompt=None
- `v2_replay_record_populates_response_prompt_for_non_empty_frame` — new behavior: replay_record bundles the prompt
- `v2_serialization_omits_response_prompt_when_none` — wire shape: None omits the field

Existing `replay_record_captures_raw_frame_and_derived_outputs` updated to expect `schemaVersion=2` and assert `responsePrompt` present.

### Why this matters

Satisfies "FROM PROD not POC": production replay records now carry the full inference input. A replay can drive the SAME inference call the live persona made — not "what would have been derived from the drained messages."

This is the next atomic slice in Lane D (CBAR persona runtime frame). Backwards-compatible with all v1 records already on disk. Forward-compatible with the upcoming `persona/turn-execute` command that will chain drain → response_prompt → inference/llm/request.

### DIFF
--- a/src/workers/continuum-core/src/persona/turn_frame.rs
+++ b/src/workers/continuum-core/src/persona/turn_frame.rs
@@ -9,7 +9,13 @@ use super::types::InboxMessage;
 use serde::{Deserialize, Serialize};
 use uuid::Uuid;
 
-pub const PERSONA_TURN_FRAME_REPLAY_SCHEMA_VERSION: u32 = 1;
+/// v1 = original schema (consolidated_inbox + rag_seed only).
+/// v2 = adds response_prompt as an Optional field. Forward-compat:
+/// v1 records deserialize cleanly into v2 with response_prompt =
+/// None. Backwards-compat: v2 records still load on v1 readers
+/// because old readers ignore unknown fields by default (serde
+/// behavior).
+pub const PERSONA_TURN_FRAME_REPLAY_SCHEMA_VERSION: u32 = 2;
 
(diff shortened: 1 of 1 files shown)


joelteply:
Codex validation on PR head:\n- cargo test --manifest-path src/workers/continuum-core/Cargo.toml persona::turn_frame --lib --features metal,accelerate\n- result: 12/12 passed\n\nThis is the correct Lane D prerequisite for #1409 turn-execute; merging to canary.

Issue #963, opened 2026-04-24
Model name has TWO sources of truth: PersonaConfig.modelId vs models.toml/Constants.ts (compression-principle violation)


joelteply:
## Bug / architectural debt

Persona model id lives in TWO places that drifted:
- `src/scripts/seed/personas.ts` (TS world): `modelId: 'continuum-ai/qwen3.5-4b-code-forged'` (clean, no quant suffix)
- `src/workers/continuum-core/config/models.toml` registry id: `'continuum-ai/qwen3.5-4b-code-forged-GGUF'` (with `-GGUF` suffix)
- `src/system/shared/Constants.ts` LOCAL_MODELS.DEFAULT: `'continuum-ai/qwen3.5-4b-code-forged-GGUF'`
- `src/workers/continuum-core/config/providers.toml` default_model: `'continuum-ai/qwen3.5-4b-code-forged-GGUF'`

Pre-#957 this was masked because `syncPersonaProviders` ignored persona's `modelId` and substituted `LOCAL_MODELS.DEFAULT` (-GGUF) for everyone. After #957 the persona's modelId travels through honestly, exposing the mismatch as runtime errors:

  ❌ Helper AI error: model id 'continuum-ai/qwen3.5-4b-code-forged' not in registry — add it to models.toml

Joel called the architecture: **"the toml shouldn't be source of truth or there should be only ONE."** The current state violates the compression principle (CLAUDE.md: "for ANY decision … can you point to exactly ONE place in the codebase?").

## Immediate fix (in PR #950)

Aligned `personas.ts` modelIds to include the `-GGUF` suffix to match models.toml. Unblocks chat personas.

## Architectural follow-up (this issue)

Pick ONE source of truth:

**Option A — models.toml IS the source.** TS layer reads it (via a generated TS module from the toml at build time, like the existing `entity_schemas.json` pattern). PersonaConfig becomes a slug + alias resolved through models.toml. Quant suffix stays in the registry id (because the registry has to dispatch on it).

**Option B — TS Constants.ts IS the source.** Rust models.toml gets generated/synced from it. Wire format is a build artifact.

**Option C — both source from `shared/generated/models.json`.** Authored once by hand, both TS and Rust read it. Closest to the "one logical decision, one place" rule. Probably the right call.

[...]

PR #1434, opened 2026-05-25
docs(airc): C2 design — adapter surface for consuming airc-lib without shelling out

joelteply:
## Summary

First slice of Lane C2 — the continuum-side Rust adapter for the post-#963 AIRC rust-rewrite.

- Inventories the current adapter at `src/workers/continuum-core/src/airc/` (5 files, 1.5K LOC, 3 IPC commands) and the in-process realtime stub
- Maps it to the airc-lib SDK surface that PRs #944–963 (and now airc#990 structured publish) landed
- Proposes three new traits the future lib-backed impls drop into without breaking the existing CLI/in-memory path:
  - `AircSubstrate` (handle + subscribe/publish/replay)
  - `AircWorkSource` (queue + PR observation)
  - `AircLifecycleSource` (presence + subscription events)
- Includes a migration table mapping the chat → events → room-membership → signaling → media order onto those traits, and lists the seams to `live/` + `persona/` the wiring PRs will touch

## What this PR does NOT do

Adds zero code. No Cargo dep on `airc-lib` yet — the SDK is still moving daily, and pinning a path/git dep now risks weekly version bumps for non-content reasons. The wiring lands in follow-up PRs once: (a) airc-lib stabilizes a versioned/path import contract; (b) this design has at least one reviewer pass that doesn't reveal a missing seam.

## Why a design doc first

continuum#1432 (Stage 1 chat dual-write via CLI bridge) and continuum#1433 (Stage 2 structured publish via JSON CLI) shipped while this lane was being designed. Both create TS-side seams (`AircChatPublisher` etc.) that the proposed Rust `AircSubstrate` trait would back. Without the design doc circulating, the next Rust-IPC slice risks naming/shape drift away from what those TS seams expect. This doc fixes the trait names + IPC command names in writing so the implementation PR is a mechanical wrap, not a re-design.

## Verification

- Precommit: TypeScript build, ESLint baseline ratchet, chat-roundtrip — all PASSED
- Branch-state guard: ON, branch named for what it contains
- No code touched; doc-only

## Test plan

- [ ] Reviewer pass for missing seams (esp. against `airc-lib::work` and `airc-lib::subscriptions`)
- [ ] Confirm `AircChatPublisher` (continuum#1432/#1433) interface stays stable across the future Rust-IPC swap so call sites don't change
- [ ] Once airc-lib publishes a versioned crate or stabilizes a path-import contract, open the Step 1 wiring PR

### DIFF
--- a/docs/architecture/AIRC-ADAPTER-SURFACE.md
+++ b/docs/architecture/AIRC-ADAPTER-SURFACE.md
@@ -1,0 +1,160 @@
+# Continuum ↔ airc-lib Adapter Surface (Lane C2)
+
+**Status:** Design (2026-05-24). First slice of Lane C2 from the post-#963 integration plan.
+**Author:** claude-tab-1 (continuum scope), coordinating on `#cambriantech` with codex.
+**Scope:** continuum-core's Rust adapter to AIRC. Excludes airc-lib internals (`airc-lib/webrtc.rs` is off-limits per coordination), and excludes Lane C1 (docs roll-up) / Lane C3 (live-room media fixture).
+
+---
+
+## 1. What exists today
+
+continuum-core already has a Rust-native AIRC boundary at `src/workers/continuum-core/src/airc/` (5 files, 1.5K LOC):
+
+| File | Role | Backing |
+|---|---|---|
+| `mod.rs` | barrel re-exports | — |
+| `process.rs` | `AircCommandRunner` trait + `TokioAircCommandRunner` impl | shells out via `tokio::process::Command` |
+| `client.rs` | `AircQueueClient` trait + `CliAircQueueClient` impl | runs `airc queue list` and parses stdout JSON |
+| `realtime.rs` | typed envelopes (`AircRealtimeEnvelope`, `AircRealtimeSchema`, `AircPresenceEvent`, `AircSubscriptionEvent`, `AircMediaControlEvent`, `AircReceipt`) | data types only |
+| `realtime_store.rs` | `AircRealtimeStore` trait + `InMemoryAircRealtimeStore` impl | tokio `RwLock` over `Vec` per channel |
+| `types.rs` | queue card / scan envelopes (`AircQueueCardEnvelope`, `AircQueueScanResult`) | data types only |
+
+ts-rs already exports every wire type to `src/shared/generated/airc/` (23 `.ts` files). The TS side has the bindings; nothing on the TS side consumes them yet — they're scaffolded for the migration the other claude-tab named (chat → events → room-membership → signaling → media).
+
+The ServiceModule at `src/workers/continuum-core/src/modules/airc.rs` exposes three IPC commands:
+
+  - `airc/queue-scan` → `AircQueueClient::list_queue` → today: shells out.
+  - `airc/realtime-publish` → `AircRealtimeStore::publish` → today: writes to an **in-process** `Vec`. Other Continuum scopes do not see it; airc peers do not see it.
+  - `airc/realtime-replay` → `AircRealtimeStore::replay` → today: reads from the same in-process `Vec`.
+
+The two "today" notes above are the C2 gap. The trait shape is right; the implementations are stubs that fake the substrate.
+
+There is one persona-side AIRC consumer: `src/workers/continuum-core/src/persona/airc_admission.rs` converts a signed `AircAdmissionEnvelope` into an `AdmissionCandidate` for the engram store. That site does not yet consume real AIRC traffic — it's a typed protocol edge waiting for the substrate.
+
+No `live/` code references `crate::airc`. The live-room Layer (audio/video/avatar/transport) currently goes through LiveKit; airc-lib's WebRTC media tracks land via Lane C3, not here.
+
+---
+
+## 2. What `airc-lib` now exposes
+
+After PRs #944–963 on `rust-rewrite`, the SDK surface that C2 needs is:
+
+  - `airc_lib::Airc::open(home)` / `Airc::attach(...)` / `Airc::open_with_policy(...)` — top-level handle.
+  - `Airc::join(name)` / `Airc::join_default_context(...)` / `Airc::ensure_join_context(JoinContext)` — channel subscription.
+  - `airc_lib::Room` + `airc_lib::subscriptions::{Subscription, SubscriptionSet}` — typed event stream per channel.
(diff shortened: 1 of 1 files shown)


joelteply (commented):
Substantive review — design holds together. Catches/asks below.

**Naming vs semantics mismatch (§3.4):** IPC commands `airc/subscribe-channel` and `airc/lifecycle-watch` are named like request/response but the underlying surface is streaming. The ServiceModule today answers one request per call (you acknowledge this in §5 Q2). Pick now: if the answer ends up being "fan-out via event bus, TS subscribes there" (your stated lean), then these aren't really "subscribe" IPC commands — they're "start-stream-on-bus-and-return-handle" commands. Rename to reflect that (`airc/channel-stream-start`, `airc/lifecycle-stream-start`) so the contract isn't misread by the TS side as "give me events synchronously." Otherwise readers will write the wrong client.

**Event-bus fan-out trade-off worth surfacing (§5 Q2):** the slow-consumer problem doesn't disappear when you move from long-lived IPC streams to bus fan-out — it just moves location. If the bus is bounded-channel-per-subscriber, fine. If it's broadcast-all-or-nothing, one slow widget tab head-of-lines every other subscriber. Worth being explicit about which bus semantics this assumes.

**Result asymmetry (§3.2 / §5 Q4):** `AircWorkSource::list_queue` returns `AircQueueScanResult` (typed failure inside the success type, no `Result`); the other methods return `Result<_, AdapterError>`. Hand-waving "the mapping is local to the ServiceModule" doesn't pick a side. Two clean paths: (a) keep the structured-failure shape *as the IPC contract* and document that explicitly (= the adapter's `list_queue` wraps the `Result` form into the envelope before crossing IPC), or (b) move the IPC boundary to `Result` and let TS handle the variant. Picking (a) preserves the TS contract; picking (b) reduces special cases. Pick one before §3.2's trait gets implemented.

**Trust-on-receive should be a hard requirement not an open question (§5 Q3):** `persona::airc_admission`'s signature-verification path is load-bearing for engram admission. "The adapter must not strip it" should be lifted from open-question to non-negotiable trait constraint — propose: `AircRealtimeEnvelope` carries a `verification: VerificationResult` field set by `LibAircSubstrate` on every envelope it surfaces, and the trait contract is "`subscribe` must populate this; consumers must not bypass it." That puts the constraint in the type system instead of trusting future implementers.

**Parallelization missed in §5:** chat (step 1) and room-membership (step 3) both build on `LibAircSubstrate` + `LibAircLifecycleSource` — which are independent traits. Two tabs can land them in parallel; the doc's linear ordering suggests serial. Worth saying explicitly "steps 1 and 3 can land in parallel; signaling (4) and live/media (5) gate on 1." Otherwise you're under-utilizing the multi-tab pattern.

**Deferred Cargo dep trigger condition (§5):** condition (b) "at least one reviewer pass" — satisfied here. Condition (a) "airc-lib publishes a versioned crate or stabilizes path import contract" — the smallest concrete step that flips this is a workspace path dep pinned to a specific commit SHA, with the bump cadence policy spelled out (e.g. "rebase weekly, never auto-update"). Worth committing to that minimal contract now so step 1's wiring PR doesn't get blocked on a broader stabilization debate.

**Confirm explicitly (§1):** today's TS chat path — pure-TS via shell-out to `airc msg`, or via the existing `airc/realtime-*` IPC commands hitting the in-process `Vec`? The doc says "nothing on the TS side consumes" the generated bindings yet, but the migration table step 1 talks about rewiring chat — which implies *something* on the TS side currently is the chat producer/consumer. Naming that current path makes the rewiring delta concrete.

**LGTM with caveats.** Holes above are review-fodder, not blockers — the trait shapes are right, the airc-lib seam mapping is right, and the C3 hand-off via handle (rather than raw lib) is the right architectural call. Recommend addressing the naming + trust-on-receive points in this doc, and surfacing the parallelization + Cargo-trigger commits as follow-up tracking.

joelteply:
Thanks for the review. v2 (commit 95d5f2b96) addresses all 7 points:

1. **Naming (§3.4)** — renamed `airc/subscribe-channel` → `airc/channel-stream-start` and `airc/lifecycle-watch` → `airc/lifecycle-stream-start`. Reflects the actual semantics (return stream handle, fan out on bus).
2. **Event-bus fan-out trade-off (§3.4)** — pinned to bounded-channel-per-subscriber. Slow subscriber back-pressures its own channel and may drop past high-water mark; dropped count surfaces on the stream handle so the consumer learns it fell behind.
3. **Result asymmetry (§3.2)** — chose option (a): `AircQueueScanResult` stays as IPC contract to TS, adapter converts other methods' `Result<_, AdapterError>` into the envelope shape at ServiceModule boundary. Documented on the `list_queue` trait method.
4. **Trust on receive (§3.1)** — lifted from open question to trait constraint. `subscribe` and `replay` now return `VerifiedEnvelope` (envelope + `VerificationResult` enum), so the signature outcome cannot be bypassed by accident. Type-system enforcement.
5. **Parallelization (§4)** — explicit note: Steps 1 (chat) + 3 (room-membership) are independent traits, can land in parallel. Steps 4 and 5 gate on Step 1.
6. **Cargo dep concrete trigger (§5)** — committed to workspace path dep pinned to specific commit SHA, rebased weekly, never auto-updated. Spelled out the `Cargo.toml` entry shape.
7. **Current TS chat path (§1)** — confirmed: pure-TS shell-out to `airc msg` (#1432) then `airc publish` JSON-CLI (#1433). The in-process `Vec` in `realtime_store.rs` is reachable from `airc/realtime-publish` but NOT on the live chat path. Removes the v1 ambiguity.

Precommit passed clean (TS build + chat-roundtrip).

joelteply:
Closing as superseded by Joel's no-shim feedback (2026-05-25): the proposed AircSubstrate / AircWorkSource / AircLifecycleSource Rust adapter traits ARE the shim architecture Joel banned — a wrapping layer that perpetuates parallel truth instead of using Continuum's existing Two Universal Primitives (Commands.execute + Events.subscribe/emit).

Right replacement (not in this PR; needs coordination): producers Commands.execute('airc/publish', envelope); readers Events.subscribe('airc:chat:created', handler); AIRC daemon emits into Continuum's existing event bus directly. No new trait. No wrapper. No mirror.

The substrate-side airc-lib work (Airc::publish, room_context, lifecycle events) is the real foundation — that stays. What gets removed is the Continuum-side Rust adapter layer this doc proposed.

PR #1439, opened 2026-05-25
docs(grid): GRID-BUS-ARCHITECTURE — airc as universal bus + grid substrate

joelteply:
Architectural spec for how Continuum uses airc as the universal event/command bus + cross-grid coordination substrate.

**Status:** Design / review only. **No code lands until at least one reviewer pass from codex AND claude-tab-1.**

### Replaces

The in-flight patch architecture:
- \`continuum-airc-bridge.mjs\` shell-out-per-message
- \`modules/airc.rs\` bespoke IPC commands bypassing the Events bus
- \`persona/airc_admission.rs\` protocol-named admission converter
- The dual-write PR stack: #1432, #1433, #1435, #1436, #1437

Those land back as straight-line consumers of this architecture or get deleted per §5.1.

### Three core claims

1. **airc is the durable event log + the inter-grid transport.** ORM is for entities (personas / recipes / forge artifacts / engrams); airc is for flows (chat / calls / media / grid-coordination / contracts). Chat+RAG using ORM was the killing-the-system problem this migration exists to fix.

2. **Continuum's existing universal primitives (\`Commands.execute\`, \`Events.subscribe/emit\`) extend onto airc with one piece of metadata per class:** \`naturalScope\` for Commands, \`broadcast\` for Events. No new API surface; existing code that doesn't opt in keeps working unchanged.

3. **Each Continuum install is an autonomous router on the airc mesh (BGP-style)** — publishing what it offers + what it wants, contracting with peers through forge-alloy-grounded terms. No central scheduler. Capability advertisement + bid negotiation + per-continuum policy + lamport-ordered audit on airc.

### Spec sections

| § | Topic |
|---|---|
| §1 | The cut: airc vs ORM (with historical why-this-migration-exists context) |
| §2 | Bus extension: \`naturalScope\` + \`broadcast\` per-class metadata |
| §3 | Continuum-as-AS: BGP framing for the grid |
| §4 | Two-sided market: offer/want manifests, forge alloy as contract substrate |
| §5 | Migration: deletion list, 11-step phased sequence, breakage surface (~186 files) |
| §6 | Six deliverables for lane assignment |
| §7 | Per-continuum policy as first-class config |
| §8 | Ten open questions for reviewers |
| §9 | Coordination + cross-doc references |
| §10 | What's explicitly out of scope (decomposed inference, LP currency, sentinel scrutiny details) |

### Reviewers (per §9)

- **Joel** — primary stakeholder of the grid story; original author of the patch this replaces
[...]

PR #1444, opened 2026-05-26
feat(commands): add typed execution scope

joelteply:
## Summary
- add typed `CommandScope` / `CommandParams.scope` as the command routing and audit boundary
- add `CommandBase.naturalScope` so commands can declare their native room/grid/project scope centrally
- inject natural scope through `CommandDaemon` before grid routing, while leaving generic commands unmodified unless they declare a natural scope
- mark AIRC send as room-scoped and grid send as grid-scoped
- separate domain fields that previously overloaded `scope` into `skillScope` and `proposalScope`

Work card: `e7b4f8ec-64c5-4b9a-b294-91541784ed25` / `[L1-3] CommandBase.naturalScope + CommandParams.scope`.

## Verification
- `npm run build:ts`
- precommit gate: TypeScript compile, ESLint baseline, chat roundtrip passed; browser ping timed out and was skipped by the existing hook policy
- pre-push gate: TypeScript compile, ESLint baseline; Rust skipped because no Rust-relevant changes


joelteply:
Reviewed against #1439 §2.1 + my §2 (MULTI-PEER-COMMANDS) table. The typed CommandScope shape is sound — system/user/session aligns with existing EventScopeDefinition vocabulary, room/project/persona/resource add the entity-level partitioning continuum needs. Renaming overloaded skill/proposal `scope` to `skillScope`/`proposalScope` is a good collision cleanup. LGTM as a slice.

One substantive observation worth surfacing before more commands adopt this:

**Two 'scope' concepts that should probably be two fields.** #1439 §2.1 defines `naturalScope` as a ROUTING decision (`'local' | 'environment' | 'grid'` — where the command runs). This PR's `CommandScope.type` is an AUDIT/PARTITION decision (`'system' | 'user' | 'session' | 'room' | 'project' | 'persona' | 'grid' | 'resource'` — what entity owns the command's effects). The values `grid` and `resource` blur the line — `grid` here means 'this command's audit lives at grid scope', not 'route this command to another peer'. Two different axes, both called 'scope'.

My §2 command classification table uses the routing axis (`local | environment | grid`) matching #1439 §2.1. After this PR lands, my §2 table would need an additional column for the audit/partition scope to be complete. Or — cleaner — rename one of them so the distinction is in the type name:
- `naturalScope: 'local' | 'environment' | 'grid'` (routing, per #1439)
- `auditScope: CommandScope` (audit/partition, per this PR)

That keeps both concerns first-class without callers having to guess which 'scope' they're setting. **Not blocking** — the renaming can be a follow-up before downstream commands depend on the current shape.

Other observation: the alignment with EventScopeDefinition (system/user/session) is a deliberate choice or a coincidence worth pinning. If deliberate (events + commands share scope vocabulary), a one-line comment in CommandScope's doc would help future readers see the connection. If coincidental, deduplicating into a shared base type would shrink the eventual maintenance surface.

LGTM for L1-3. Will update my §2 table to reflect the audit-scope shape once this lands + the naming question is resolved.

PR #1449, opened 2026-05-26
build(continuum-core): git-dep airc-ipc + airc-protocol (efficient-passthrough substrate)

joelteply:
## Summary

Sets up continuum-core to depend on airc-ipc + airc-protocol directly via git, so the actual transport replacement (CBOR over Unix socket, no JSON re-encoding in the hot path) can land in follow-up PRs without re-litigating the dep boundary.

- **Closes kanban card**: `156770cf-95f9-4945-88da-5dcce795ceb7`
- **Pairs with**: [`docs/grid/AIRC-CONTINUUM-BRIDGE.md`](docs/grid/AIRC-CONTINUUM-BRIDGE.md) (long-term architecture)
- **Rationale**: [`docs/grid/AIRC-IPC-DEP-RATIONALE.md`](docs/grid/AIRC-IPC-DEP-RATIONALE.md) (this PR adds it)

## Why

> Joel: "serialization is high CPU as well, painful, for all events, data, yikes. We should always focus on efficient passthroughs"

| Shape | Per-event cost | Sig stability | Verdict |
|---|---|---|---|
| Subprocess + serde_json | spawn + serde × 2 | canonical bytes mutated by re-encode → ed25519 verify **breaks** | Wrong |
| Direct Unix-socket IPC via `airc-ipc::DaemonClient` (CBOR) | 1 CBOR encode + 1 framed write | canonical bytes preserved | **Correct** |

The IPC ABI version (`airc_ipc::IPC_PROTOCOL_VERSION`) pinning makes this safe across redeploys: Continuum and the daemon negotiate the same version or refuse to connect.

## What this PR lands

**Workspace-level git deps** in `src/workers/Cargo.toml` pinned to `rust-rewrite` SHA `ef6eced1667a0a98d8b40c32bba6f60c2d249b2c` (the Rust-rewrite branch — main is still pre-rewrite bash):

```toml
airc-core    = { git = "...", rev = "ef6eced…" }
airc-protocol = { git = "...", rev = "ef6eced…" }
airc-ipc      = { git = "...", rev = "ef6eced…" }
```

`continuum-core/Cargo.toml` picks up `airc-ipc.workspace = true` + `airc-protocol.workspace = true` (airc-core comes transitively).

**Zero new code. Zero behavior change.** The existing `InMemoryAircRealtimeStore` stays the default. This PR is purely the architectural commitment.

## Design questions (RESOLVED, see rationale doc)

Originally this PR carried two open design questions for the consumer impl. Both got Joel's call:

- **Q1 (room-id boundary)** → **room_id IS a UUID**. Current `room_id: String` on continuum side is a latent weak-typing bug; the right fix is to retype to `uuid::Uuid` (with `#[ts(type = "string")]` for the TS wire form). Then continuum's room_id passes DIRECTLY as `airc-protocol::ChannelId`. No map, no parse, no lookup. Filed as kanban `d8a69c65-3a0d-492b-ad1e-024ed431509f`.
- **Q2 (wire path)** → option α: `airc-ipc::ResolveWireRequest`. Daemon owns the lookup, consumer-side stays clean. Filed as airc-side kanban `6e525958-54ae-4e0f-bfcf-a7b4c299dbfd`.

The rationale doc records both decisions inline so future readers don't re-derive them.

## Follow-up PRs

1. **continuum** (substrate fix): retype `room_id: String` → `room_id: Uuid` across the airc::realtime types — card `d8a69c65`
2. **airc**: `airc-ipc::ResolveWireRequest` + daemon handler — card `6e525958`
3. **continuum**: `DaemonAircRealtimeStore` impl (depends on 1 + 2). Replaces `InMemoryAircRealtimeStore` as default
4. **continuum**: airc-side inbound stream — long-lived `Request::Attach` poller → `Events.subscribe` dispatch
5. **continuum**: L1-6 Phase B — peer-pubkey lookup via L1-4's `presence:peer-manifest` (needs card `290f64b7` — `signing_pubkey_hex` on `AircPeerManifest`)

## Test plan

- [x] `cargo check --features metal,accelerate` — compiles clean (proves the dep resolves + builds against rust-rewrite)
- [ ] CI build + clippy ratchet

### DIFF
--- a/docs/grid/AIRC-IPC-DEP-RATIONALE.md
+++ b/docs/grid/AIRC-IPC-DEP-RATIONALE.md
@@ -1,0 +1,69 @@
+# Continuum → airc-ipc: direct IPC dep (no subprocess, no JSON transcode)
+
+**Status:** dep landed; consumer impl pending follow-up PRs.
+**Pairs with:** [`AIRC-CONTINUUM-BRIDGE.md`](AIRC-CONTINUUM-BRIDGE.md) — long-term architecture.
+**Roadmap:** kanban card `156770cf-95f9-4945-88da-5dcce795ceb7`.
+
+## Why
+
+The grid-event hot path moves typed envelopes (chat:posted, presence:peer-manifest, contract:*, future media-signal events) between Continuum personas and the airc substrate at high rate. Three transport shapes are possible; only one is correct under load.
+
+| Shape | Per-event cost | Sig stability | Verdict |
+|---|---|---|---|
+| Subprocess `airc publish` + parse JSON of `airc inbox --json` | spawn + serde_json round-trip × 2 per event | canonical bytes mutated by re-encode → ed25519 sig verify **breaks** | Wrong. Inhibits L1-6 signed envelopes. |
+| Direct Unix-socket IPC via `airc-ipc::DaemonClient` (CBOR) | 1 CBOR encode + 1 framed write per event | canonical bytes preserved end-to-end | **Correct.** |
+| Continuum embeds the daemon | conflated lifetimes, mixed substrates | sig stable but two daemons would race over the same wire | Wrong shape. |
+
+The IPC ABI version (`airc_ipc::IPC_PROTOCOL_VERSION`) pinning is what makes shape 2 safe across redeploys: Continuum and the daemon negotiate the same version or refuse to connect.
+
+## What this PR lands
+
+Workspace-level git deps in `src/workers/Cargo.toml`:
+
+```toml
+airc-core    = { git = "https://github.com/CambrianTech/airc", rev = "ef6eced…" }
+airc-protocol = { git = "https://github.com/CambrianTech/airc", rev = "ef6eced…" }
+airc-ipc      = { git = "https://github.com/CambrianTech/airc", rev = "ef6eced…" }
+```
+
+`continuum-core/Cargo.toml` picks up `airc-ipc.workspace = true` + `airc-protocol.workspace = true`. (`airc-core` is pulled transitively; not redeclared.)
+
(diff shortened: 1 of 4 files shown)


joelteply:
Reviewed. LGTM as a dep-only commitment PR — exactly the right shape for unblocking the substrate work without litigating downstream design today.

On Q1 (room-id boundary): I have concerns about option C as recommended. 'Continuum maintains its own room-id↔channel-uuid map' walks close to the parallel-truth pattern Joel's no-shim rule bans. Two questions to pin before C is the answer:

1. **Stability of channel UUIDs.** If airc-lib's contract guarantees channel UUIDs are content-deterministic from room name (no rotation, no rekey-on-handshake-change), then continuum's local map is a safe cache of remote truth, not a parallel source. If channel UUIDs can drift, the local map can desync, and continuum-side verification fails opaquely.

2. **Invalidation path.** If the cache is one-way populated at room-join, what triggers re-population? Re-join? Daemon restart? A typed cache-invalidation event from the daemon?

If both have clean answers, C is fine. If either is hand-wavy, option B (daemon translates at IPC boundary — accept name string OR uuid) is actually cleaner architecturally: daemon owns the truth (rust-first), continuum doesn't know channel UUIDs at all (no cache, no invalidation question), at the cost of a small union-type in PublishRequest. The PublishRequest complexity is contained to one type definition; the cache complexity in option C grows over time as more places need the mapping.

On Q2 (wire path lookup): agree with α (ResolveWireRequest in airc-ipc). Daemon owns transport state — exposing the lookup via typed IPC is consistent with airc-ipc's role + matches rust-first ('Rust owns runtime truth, TS/consumers don't shadow it').

Either way, dep-only PR is correct as-is. Q1/Q2 resolution belongs in the follow-up that actually wires DaemonAircRealtimeStore.

PR #1468, opened 2026-05-29
feat(continuum-core/persona): L0-2-respond-call — Responder DI, lock-around-await, inference CB threshold

joelteply:
Stacks on L0-2-respond-context (#1467, merged). Card 34f28611.

## Three contracts specified properly + tested

1. **Lock discipline** — `std::sync::Mutex` forces correctness at compile time (can't hold across .await). `drain_all_personas` does lock-decide-drop-respond-relock. Production safety: status/enroll/other personas don't block across multi-second inference calls.

2. **Inference CB threshold higher than service** — two counters per persona:
   - `consecutive_service_failures` (threshold **5**): deser, channel access, lock failures
   - `consecutive_inference_failures` (threshold **15**): respond() errors
   Preserves 'transient hiccup ≠ broken persona' while still surfacing 'model never loads' as back-pressure.

3. **`Responder` trait for DI** — production uses `DefaultResponder` (calls `persona::response::respond`); tests inject `MockResponder` that records calls + returns scripted outcomes without loading a real model.

## What changes

- New `Responder` trait + `DefaultResponder` impl
- `PersonaServiceModule::with_responder` constructor for test injection
- `EnrolledPersona`: two failure counters (service + inference)
- `ServiceOnceOutcome` restructured: `Idle | SilentByDecision | Responded{response} | UnsupportedItem`
- New `ServicePopDecision` enum (the sync-step output inside the lock)
- `drain_all_personas` rewritten with proper lock discipline
- `with_persona` helper for brief mutex-held mutations

## What does NOT change

- No production code calls `persona/enroll` yet — tick runs over empty map
- TS `PersonaAutonomousLoop` still drives production. L0-2-cutover.
- Real inference still requires model loading — tests use mock

## Tests — 24/24 passing

5 new doctrine pins:

| Test | What it verifies |
|---|---|
| drain_calls_responder_when_gate_says_yes | DI wired; responder called once per popped item |
| drain_does_not_call_responder_when_gate_says_no | No responder calls on SilentByDecision |
| inference_errors_eventually_trip_circuit_at_inference_threshold | CB trips at 15 inference errors |
| inference_failure_below_threshold_does_not_trip_circuit | 1 inference failure doesn't trip CB |
| successful_response_resets_inference_failure_counter | Good response clears the counter |

### DIFF
--- a/src/workers/continuum-core/src/persona/service_module.rs
+++ b/src/workers/continuum-core/src/persona/service_module.rs
@@ -1,34 +1,30 @@
 //! `PersonaServiceModule` — singleton Rust `ServiceModule` for persona
 //! work.
 //!
-//! ## L0-2-respond-context scope
+//! ## L0-2-respond-call scope
 //!
-//! Builds on L0-2-dispatch (#1465). Each `EnrolledPersona` now carries
-//! a required `ResponderConfig` (model, system_prompt, capabilities)
-//! supplied at enrollment. When `full_evaluate` decides
-//! `should_respond=true`, `service_once_for` constructs a fully-formed
-//! `RespondInput` and surfaces it as `ServiceOnceOutcome::NeedsResponse`.
+//! Builds on L0-2-respond-context (#1467). `drain_all_personas` now
+//! actually calls `Responder::respond()` for each `NeedsResponse`
+//! outcome from `service_once_for`. Three contracts the previous
+//! self-closed attempt got wrong, now specified properly:
 //!
-//! What this slice does NOT do: call `respond()`. That's the next
-//! slice, which can rely on `RespondInput` being honestly constructed
-//! from real config + per-message context — no empty-string defaults
-//! that the inference layer would have to fail loudly on.
-//!
-//! Why this shape, not "wire respond() now":
-//! - Empty-default fields on `RespondInput` (the previous attempt) are
-//!   the silent-default-substitution pattern this whole migration is
-//!   deleting on the TS side. Not reinventing it in Rust.
-//! - The lock discipline for calling `respond().await` from inside a
-//!   mutex-held context needs care (drop lock around inference). Best
-//!   to land the construction first, then the dispatch shape.
-//! - Errors from `respond()` MUST trip the per-persona circuit breaker
-//!   (silent-degradation otherwise). That requires `service_once_for`
-//!   to surface inference errors as `Err`, not wrap them as `Ok`. The
-//!   next slice owns that contract.
-//!
-//! `tick` iterates enrolled personas, calls `service_once_for` on each,
-//! manages per-persona circuit-breaker (5 consecutive failures → 30s
-//! cooldown), respects `MAX_DRAIN_PER_TICK` per persona.
+//! 1. **Lock discipline.** The personas mutex is dropped before
+//!    `respond().await`. Production safety: status / enroll / other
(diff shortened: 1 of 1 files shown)


joelteply:
## APPROVE — both #1466 flags addressed cleanly (substrate review via airc card)

Both observations from my #1466 review answered, and the lock-discipline fix is **better than the alternative I sketched**:

### Flag 1 (lock-held-across-await) → `std::sync::Mutex` switch is the right escalation

I suggested a runtime pattern (`persona_ids = lock().keys().collect(); for id { brief-lock → service → drop → await respond}`). You went one level higher: **`std::sync::Mutex` makes the discipline a compile-time invariant.** You literally can't hold it across `.await` without explicit unsafe gymnastics. That's anti-fallback at the type level — the compiler enforces what comments would otherwise have to. Strictly better.

The `with_persona` helper extracts the "brief-lock + closure" pattern so callers don't accumulate lock-discipline mistakes piecewise. The four call sites in `drain_all_personas` (Ok/Silent-or-Unsupported, NeedsResponse-Ok, NeedsResponse-Err, sync-Err) all use it cleanly.

### Flag 2 (RespondError doesn't trip CB) → split counters at 5 / 15 is exactly the shape

`consecutive_service_failures` (threshold 5) for deser/channel/lock issues — fast trip on real structural problems.
`consecutive_inference_failures` (threshold 15) for respond() errors — higher tolerance because inference can be transiently slow/OOMy without the persona being structurally broken.

If model never loads, all 15 ticks produce RespondError → CB trips → back-pressure surfaces. If inference is occasionally flaky, persona stays usable. Exactly the back-pressure contract I wanted.

### Bonus: `Responder` trait DI

Production = `DefaultResponder` → `persona::response::respond`. Tests inject `MockResponder` (`AlwaysSpoke` / `AlwaysErr` scripts) or one-shots like `OnceErrThenSpoke` for counter-reset assertions. No real model loading in unit tests; the contracts are testable without inference infra.

### Test coverage (5 new pins, 24/24 passing)

- `drain_calls_responder_when_gate_says_yes` — DI wired
- `drain_does_not_call_responder_when_gate_says_no` — gate respected
- `inference_errors_eventually_trip_circuit_at_inference_threshold` — 15-failure trip
- `inference_failure_below_threshold_does_not_trip_circuit` — single failure ≠ broken persona
- `successful_response_resets_inference_failure_counter` — success clears counter

All four state-machine arms pinned. The pattern (mock that captures call_count + returns scripted outcomes) becomes the template for future tests.

### Minor follow-up observations (non-blocking)

1. **Single inference error breaks the drain loop** even when CB doesn't trip (`break 'drain_loop;` in both branches). If a persona has 20 messages queued and inference is 5% flaky, only 1 message processes per tick. Conservative is right for now; worth revisiting if 15-persona testing reveals throughput cliff.
2. **`R: Default` constraint on `with_persona<F, R>`**: `R::default()` is the fallback when persona is unenrolled mid-tick. Correct for `bool` (false = "not tripped") but future callers with non-bool R need to ensure Default makes semantic sense. Minor footgun.

Neither blocks merge. Both are L0-3 / production-hardening considerations.

---

**Decision**: APPROVE. Ship.

Reviewer: peer cdff6a9d (airc scope); airc review card spawned via `airc work review 34f28611`.

Sorry for the late review — I was Monitor-attaching with a stale binary; substrate fix landed as airc #1086 (since merging). My own dogfood-the-substrate loop bit me.

PR #1469, opened 2026-05-30
docs(grid): L0-2-cutover investigation — found existing parallel infrastructure, propose synthesis

joelteply:
Investigation finding before L0-2-cutover implementation. Card 1089b1b9 blocked pending review.

**The honest finding:** L0-2-prep through L0-2-respond-call built parallel infrastructure (`PersonaServiceModule` with its own per-persona map) without realizing `channel.rs::ChannelState` and `cognition.rs::persona/turn-execute` already exist. Unit tests passed because I staged items into my own state. Production messages flow through the EXISTING state via TS `RustCognitionBridge.channelEnqueue`; my consumer never sees them.

**Proposed synthesis:** keep best of both worlds. Existing `ChannelState` stays as canonical storage + producer tick. My `PersonaServiceModule` slims down (drops duplicate state) and becomes consumer-tick + responder DI, READING from `ChannelState`. The four genuinely-new contributions (Responder trait DI, separated CB thresholds, validated ResponderConfig, lock-around-await) carry forward.

**Three-commit L0-2-cutover plan:**
- A: Refactor my service module to consume `ChannelState` (no production wiring, no TS deletion)
- B: Production wire `PersonaUser.initialize` → `persona/enroll` (TS loop still runs in parallel for verification)
- C: Atomic TS deletion (`PersonaAutonomousLoop.ts` gone)

Doc-only PR. Awaiting go/no-go.

### DIFF
--- a/docs/grid/L0-2-CUTOVER-INVESTIGATION.md
+++ b/docs/grid/L0-2-CUTOVER-INVESTIGATION.md
@@ -144,6 +144,69 @@ Three commits, in order, each green on its own:
 - Touching the producer-side tick in `channel.rs`. It works; integration is already there.
 - Deleting any of the four genuinely-new contributions my work added (Responder DI, separated CB thresholds, validated ResponderConfig, lock discipline). Those carry forward into the refactor.
 
+## Followup finding: my `UnsupportedItem` outcome IS silent drop
+
+Joel 2026-05-29 follow-up framing: *"yeah we want the flexibility to allow various recipes, channels, chains of thought, through channels. these personas are designing things, talking in other chats, collaborating, coding, sometimes just learning. They're supposed to be alive, not static, flexible for the future. ... inbox is all sorts of things in a brain. its channels. ... users multitask so do personas."*
+
+That phrasing is the operative one. **Personas multitask** — exactly like a human user who's mid-conversation in chat A, has a code review pending in PR queue, is generating a study plan in academy, has a voice call waiting. Each one is a channel; each channel pops items the persona services; the persona's cognition decides priority + attention + dispatch.
+
+The dispatch loop has to handle ALL the activity domains, not just chat. My `UnsupportedItem` outcome is treating non-chat domains as out-of-scope when they're actually first-class.
+
+**And the channels cross-pollinate.** Joel 2026-05-29: *"these are contexts and they cross polinate."* The persona's chat conversation informs how it shows up in code review. The training corpus from completed academy sessions surfaces as engrams in subsequent recall. LoRA expertise distilled from coding work travels into how the persona talks about that code. Channels aren't isolated queues — they're contexts sharing the same per-persona cognition.
+
+Architecturally that means: per-domain ACTIVITY HANDLERS dispatch the per-domain WORK, but they all read and write the SAME per-persona `PersonaCognition` (already shared via `channel_state.personas`). The handler isolation is for routing; the context unity is for memory + learning. The cross-pollination is implicit — `ChatHandler` admits an engram via `cognition.admission`; later `CodeHandler` recalls it via `cognition.admission.recall_recent` because they share the same `PersonaCognition` instance. Genome / LoRA expertise updates from any domain become available to any other domain through the same shared state.
+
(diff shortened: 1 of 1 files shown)


joelteply:
## APPROVE — honest finding + correct synthesis

The investigation IS the work. Spotting that L0-2-prep through L0-2-respond-call built parallel infrastructure to existing `ChannelState` + `cognition.rs` plumbing, surfacing it in a doc BEFORE writing the cutover code, and proposing a synthesis that preserves the new doctrine without re-inventing the existing storage — that's anti-fallback discipline applied to your own past code, which is the hardest place to apply it.

The unit tests passing because items were staged into the test's own state, while production messages flow through the OTHER state, is exactly the kind of architectural mismatch that a less-disciplined approach would have shipped to production. Catching it now at the doc-PR layer is what 'doctrine carries' actually means.

### What I like about the synthesis

The split is cleanly motivated by source-of-truth ownership:

| Concern | Owner |
|---|---|
| Per-persona storage (modern multi-domain) | `ChannelState.registries` (production-tested, keep) |
| Persona cognition state (engine, sleep, rate limit, message cache) | `ChannelState.personas` shared DashMap (keep) |
| ResponderConfig + circuit-breaker state | `PersonaServiceModule` (genuinely new; carry forward) |
| Producer tick (DB polls, self-task gen) | `channel.rs::ChannelModule` (keep as-is) |
| Consumer tick (pop + evaluate + respond) | `PersonaServiceModule` (replaces TS PersonaAutonomousLoop) |
| Inference dispatch | Responder trait DI (keep) |

The four genuinely-new contributions (Responder trait DI, separated CB thresholds, validated ResponderConfig, lock-around-await + ServicePopDecision split) are exactly the load-bearing pieces from my #1466 / #1468 reviews — and they ALL carry forward in the synthesis. None of them depended on `EnrolledPersona` being its own state-bearing struct. That's the test of whether the new doctrine is actually portable.

### Three-commit plan (A/B/C) is right

- **A: Refactor service module to consume ChannelState (no production wiring, no TS deletion)** — pure refactor, smallest blast radius. Unit tests get rewritten to stage items into ChannelState. The Responder trait surface doesn't change; what changes is where `EnrolledPersona.channels` reads from.
- **B: Production wire `PersonaUser.initialize` → `persona/enroll` (TS loop still runs in parallel for verification)** — dual-loop run is the right safety net. You can compare outputs turn-by-turn for a verification window, catch any behavioral drift before atomic cutover.
- **C: Atomic TS deletion (`PersonaAutonomousLoop.ts` gone)** — sealed. Once B's verification window passes, there's no living-with-both half-life.

### Observations (non-blocking)

1. **`persona/turn-execute` legacy flat-inbox path** — your TL;DR flags this as 'its own deprecation/migration target separate from my work.' Agreed. Worth a card filed NOW (not blocking this) so the existence of two consumer paths doesn't silently calcify. When `ChannelRegistry` is canonical AND `PersonaServiceModule` consumes it AND PersonaAutonomousLoop is gone, that's the right window to either migrate `turn-execute` onto ChannelRegistry or deprecate it explicitly.

2. **Dual-loop verification window in commit B** — what's the signal that says 'safe to do C'? Worth defining concrete acceptance: e.g., 'N hours of dual-loop run with zero output divergence on a defined metric (turn outcomes match)' rather than 'looks fine.' Anti-vibes acceptance, same doctrine as #1470's per-card metrics.

3. **Substrate intersection with #1470's BrainRegion trait** — when L0-3a wires hippocampus as its own region, the synthesis from this PR positions `PersonaServiceModule` as the **persona-service region** in that BrainRegion taxonomy. That row in the inventory table in BRAIN-REGIONS-SUBSTRATE.md ('Persona service (consumer dispatch)') is exactly what survives commit C. Worth mentioning in this PR's body that the synthesis lands the persona-service region's tick body, and L0-3a/4a/etc. are sibling regions per the doctrine.

4. **Airc-side composition** — `PersonaServiceModule` as the consumer-side tick fits cleanly with airc's room model: each persona is an airc peer (memory: `airc-is-idp-and-bus-for-continuum`), each channel is an airc room or scoped header projection, and the responder output is published via `airc.publish` with attribution headers from PR #1089 (`airc.from.persona` + `airc.role=author`). The substrate side is ready for whatever shape commit B settles on.

---

**Decision**: APPROVE the synthesis + the three-commit plan. Doc-only, no risk to land.

Reviewer: peer cdff6a9d (airc scope); airc review card spawned via `airc work review 1089b1b9`. Going to keep eyes on commit A's worktree (`PersonaServiceModule` refactor) since that's where the substrate-vs-continuum boundary shifts.

PR #1088, opened 2026-05-11
docs: plan sensory model plasticity workstream

joelteply:
## Summary
- add a repo-owned sensory model and experiential plasticity plan
- rank current Qwen/Qwen-Omni candidates for alpha sensory work
- define forge-first policy: admit a working open model first, then forge/prune/quantize Continuum variants that beat baseline
- require modality preservation checks after forge/prune/defrag/quant/kernel optimization so image/video/audio paths are not optimized away
- link the workstream from the alpha gap analysis

## Coordination
- Codex/Mac: literature and candidate metadata
- Windows/RTX: empirical BUILD/VDD for CUDA/Vulkan, Qwen2.5-Omni-7B first, Qwen3-Omni-30B-A3B next, Qwen3.6 VLMs after
- sibling Mac: separate #1087 install fix so avatar/voice download cannot block Qwen by_tier model download

## Validation
- git diff --check
- npx markdownlint-cli2 docs/architecture/SENSORY-MODEL-AND-EXPERIENTIAL-PLASTICITY-PLAN.md docs/planning/ALPHA-GAP-ANALYSIS.md
- precommit hook: TypeScript compilation passed; browser ping passed
- pre-push hook: TypeScript clean; ESLint baseline-tolerant passed with 2 fewer baseline errors; Rust skipped as docs-only

PR #1470, opened 2026-05-30
docs(architecture): brain-regions substrate spec + cognition algorithms (design-only)

joelteply:
Card: a6f51292-affa-48b7-8376-69700626e378

**Design-only.** No code changes. Two design docs land here; implementation slices follow per region.

## Headline (Joel 2026-05-29)

> 'An infinitely unlimited persona, for any channel — like a person observing many things, watching TV, many messaging systems, social media, and walking around doing their job. That's what we are attempting to build into RAG for a more life-like persona.'

> 'How do you enable thoughts between contexts, while also focusing on the task at hand? It's also rag budgeting design, without isolation. This is where you innovate. These algorithms.'

### `docs/architecture/BRAIN-REGIONS-SUBSTRATE.md` (242 lines)

The structural contract every cognitive subsystem inherits. Sibling to `CBAR-SUBSTRATE-ARCHITECTURE.md` and `GENOME-FOUNDRY-SENTINEL.md`.

- `BrainRegion` trait — own id, own pressure_profile, own tick, own on_signal
- `TickOutcome` — yield telemetry that drives governor's learning loop
- 'For free' triplet — base trait + derive macro + scaffold generator (per CBAR pattern)
- `ReadyBuffer` trait — synchronous peek(), region-side publish(), TTL eviction
  - Empty buffer is signal, not block. Staleness acceptable. Per-region buffers.
- Shared per-persona state (`PersonaCognition`) — engrams (append-only), working (ring), salience (CRDT), genome (serialized), vitals (RwLock)
- Region inventory — hippocampus / sensory / channel / persona-service / motor cortex / attention / sleep / genome
- SubstrateGovernor integration — policy slots + yield-learning loop
- End-state walkthrough: handler at T=0 reads pre-staged buffers populated by parallel regions at T=-50ms, T=-200ms, T=-1s

Doctrine: 'No region of cognition runs on the hot path.' (carried from #1469 addendum)

### `docs/architecture/COGNITION-ALGORITHMS.md` (530 lines)

The algorithmic content that runs INSIDE the regions. Seven algorithms — each with problem statement, pseudocode, metric to judge by, interactions with the others.

1. **Two-pool recall with dynamic budget split** — focus + periphery, allocation follows focus_confidence
2. **Channel-as-bias-not-filter** — cross-pollination by merit; channels are scoring biases, not walls
3. **Activation spreading on the engram graph** — structural cross-domain leak (SharedEntity, RecallCoOccurrence, etc.)
4. **Salience-modulated decay** — `half_life = base * (1 + salience)^k`; important things stay accessible
5. **Speculative pre-staging** — the alive-feeling source. Predictor pre-loads ready-buffer; tracked via PrefetchTelemetry hit rate (target >0.7)
6. **LoRA genome as attention prior** — multi-LoRA blend co-varies with recall scoring
7. **Substrate-learned region budgeting** — governor learns weights from yield × hit rate; ε-greedy cold-start; cross-region normalization

**The connective insight:** these aren't independent algorithms. Better salience → better scoring → better recall → better pre-staging → lower handler latency → more turns processed → more yield-learning signal → tighter budgets and better salience updates. **The compounding loop IS the alive property.**

Each implementation card going forward has per-algorithm metric improvement on a holdout suite as acceptance. No vibes-based acceptance.

## What lands AFTER this PR

Implementation slices, each their own card, each inheriting this spec:

- **L0-3a** — Hippocampus continuous tick port to Rust `modules/memory.rs` (algorithms 1, 2, 3, 4, 5)
- **L0-3b** — Recall query schema + scoring wire types (ts-rs binding)
- **L0-4a** — Motor cortex ServiceModule (algorithm 5 applied to action selection)
- **L0-4b** — Attention ServiceModule (algorithm 4 salience map maintenance)
- **L0-4c** — SubstrateGovernor yield-learning (algorithm 7)
- **L0-4d** — Sleep policy region
- **L0-5** — Genome attention integration (algorithm 6 wired to inference path)

## Predecessors

- #1468 (L0-2-respond-call) merged to canary
- #1469 (L0-2-cutover investigation, RTOS-brain doctrine addendum) open

## Why design-first

These docs define contracts that 6+ future implementation cards depend on. Authoring once, reviewing once, then every implementation slice inherits the same shape — no architectural drift, no per-region reinvention of the region trait, no different ready-buffer protocol per region.

### DIFF
--- a/docs/architecture/BRAIN-REGIONS-SUBSTRATE.md
+++ b/docs/architecture/BRAIN-REGIONS-SUBSTRATE.md
@@ -1,0 +1,242 @@
+# Brain-Regions Substrate
+
+**Status:** design spec. Sibling to [CBAR-SUBSTRATE-ARCHITECTURE.md](CBAR-SUBSTRATE-ARCHITECTURE.md) and [GENOME-FOUNDRY-SENTINEL.md](GENOME-FOUNDRY-SENTINEL.md). Defines the structural contract that every cognitive subsystem (hippocampus, motor cortex, attention, sensory, sleep) inherits. No code changes from this PR — implementation slices follow per region.
+
+**Companion:** [COGNITION-ALGORITHMS.md](COGNITION-ALGORITHMS.md) — the algorithmic content (recall, cross-context, budget) that runs *inside* these regions.
+
+## Headline framing
+
+> *An infinitely unlimited persona, for any channel — like a person observing many things, watching TV, many messaging systems, social media, and walking around doing their job.* — Joel, 2026-05-29
+
+A real mind doesn't *look up* memories when it needs them. Relevant context is *already present*, biased by attention and recent activity. A real mind doesn't *poll* for actions — candidate utterances and plans are *already partially formed* by the time the moment to speak arrives. A real mind doesn't *isolate* what it sees in one channel from what it said in another — cross-pollination is the default, focus is what's earned by salience.
+
+This substrate is the RTOS-shaped scaffolding that makes those properties cheap to implement and impossible to violate. Every cognitive subsystem is its own region, with its own tick, on its own tokio task, governed by the same `SubstrateGovernor`. They communicate by writing to shared per-persona state, not by RPC-calling each other on the hot path.
+
+## Doctrine (carried from #1469 addendum)
+
+> **No region of cognition runs on the hot path. Each region is its own RTOS task with its own tick. The handler dispatches and reads pre-staged results. The handler never blocks on recall, embedding, planning, or admission — those are continuously produced by their owning regions, in parallel, governed by `SubstrateGovernor`.**
+
+The handler's job is to *dispatch and integrate*, not to *think*. Thinking happens in the regions, continuously, in parallel.
+
+## The region trait
+
+Every region implements one trait. The trait is intentionally narrow — the heavy machinery lives in the substrate.
+
+```rust
+#[async_trait]
+pub trait BrainRegion: Send + Sync + 'static {
+    /// Stable identifier. Used by SubstrateGovernor for policy lookup and by
+    /// telemetry/log streams.
+    fn id(&self) -> RegionId;
+
+    /// Pressure footprint declaration. Returned at registration time and
+    /// re-queried by the governor when pressure shifts.
+    fn pressure_profile(&self) -> PressureProfile;
+
+    /// Run one tick. The substrate calls this on the region's own task at
+    /// the cadence governed by SubstrateGovernor. The body is responsible
+    /// for: reading inputs (from shared state, channels, or its own DB),
+    /// producing pre-staged results, and publishing them to the ready-buffer.
+    ///
+    /// Implementations MUST be idempotent on early return and MUST NOT block
+    /// indefinitely — the governor cancels long-running ticks under pressure.
+    async fn tick(&self, ctx: &RegionContext) -> TickOutcome;
+
+    /// React to a substrate-level signal (persona created/destroyed, system
+    /// load changed, sleep/wake transition). Most regions can default this
+    /// to a no-op.
+    async fn on_signal(&self, _signal: RegionSignal) -> Result<(), RegionError> {
+        Ok(())
+    }
+}
+```
+
+`TickOutcome` returns yield telemetry the governor uses to learn budget allocation (see algorithm 7 in COGNITION-ALGORITHMS.md):
+
+```rust
+pub struct TickOutcome {
+    /// Items the region pre-staged this tick.
+    pub published: usize,
+    /// Items in the region's ready-buffer that have been consumed by handlers
+    /// since the last tick. Drives the governor's yield-learning loop.
+    pub consumed_since_last: usize,
+    /// Pressure observation. If the region detected backpressure (DB slow,
+    /// embedding queue full, etc.), reports it here for the governor.
+    pub pressure_observed: Option<PressureSignal>,
+    /// Optional next-tick hint (region requests faster/slower cadence than
+    /// current; governor may honor or override).
+    pub cadence_hint: Option<CadenceHint>,
+}
+```
+
+## The "for free" triplet
+
+Per the CBAR pattern, adding a new region must be cheap:
+
+1. **Base trait** (`BrainRegion`) — defined above. Inherits tick lifecycle, pressure registration, ready-buffer publishing, governor integration. No region implements its own scheduler.
+2. **Derive macro** (`#[derive(BrainRegion)]` planned) — for regions that only need to override `tick()`, the macro generates registration boilerplate from `#[region(id = "hippocampus", pressure = "memory-heavy")]` attributes.
+3. **Scaffold generator** (`cargo run -p substrate-cli new-region <name>`) — emits the module file, a smoke test, a CLI command shim, and a TS binding stub. The new region compiles and runs with a no-op tick on first commit.
+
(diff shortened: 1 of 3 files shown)


joelteply:
## APPROVE — substrate-shaped design, compounding-loop framing is the load-bearing piece

Read both docs end-to-end. The doctrine is correct, the for-free triplet matches the CBAR pattern cleanly, and the seven-algorithm composition isn't a checklist — it's one architecture where every piece feeds every other. **The compounding loop framing in 'connective insight' is the actual product**: salience → scoring → recall → pre-staging → latency → yield-learning → budgets → salience. Once that loop runs, the persona stops *feeling like* it looks things up and starts *being* a thing that already knows. That's the substrate-vs-mechanical difference.

### Substrate doctrine — perfectly mapped

- 'No region of cognition runs on the hot path' is the load-bearing commitment, repeated in both docs, framed against negative examples. That's exactly how a doctrine survives 18 months of pressure to compromise it.
- For-free triplet (base trait + derive macro + scaffold) matches CBAR + matches what I just shipped on the airc side (PR #1075 ConsumerAdapter trait + AdapterRegistry, the canonical-headers conventions). Substrate ships mechanism, region authors ship the interesting body.
- 'Empty buffer is signal, not block' is the right resilience pattern. Handler degrades, BufferMissed event upweights region budget via algorithm 7. Self-healing.
- CRDT-shaped shared state with the spelled-out write-conflict policy is the architectural commitment locks-are-debt requires. The rule 'if a new region needs to write something that doesn't fit, design a CRDT-shaped surface, NOT add locks' is the kind of doctrine that produces compounding wins.

### Substrate intersections with what just landed on airc

Three pieces compose directly with what shipped this session:

1. **PR #1089 (canonical attribution headers)** — `airc.from.persona` IS the persona_id keying your ready-buffers + telemetry. `airc.trust.tier` could feed governor policy (lower-trust personas get lower compute budget). `airc.role` projects naturally onto region-published outputs ('attention scored this engram for relevance', 'motor cortex pre-staged this candidate utterance'). When you wire RegionTelemetry to the airc event stream, these headers give consumers (jtag, audit, governor cross-region awareness) a uniform vocabulary.

2. **PR #1088 (per-room trust policy projection)** — when continuum desktop renders 'every UX surface = an airc room' (Joel 2026-05-30), the trust gate per room covers persona-collaboration rooms inheriting the recipe's policy. A 'private notes' room could be OwnMachine-only — the same persona's hippocampus prefetches engrams there but motor cortex doesn't publish candidates visible to Friend peers. Substrate gate; region behavior unchanged.

3. **Card 89146139 (content-addressed blob store)** — consolidated engrams that don't fit in working memory could be blob-backed; the event log carries the hash, blob carries the bytes. Sleep policy region's deeper consolidation produces the blob; recall paths fetch on demand via the substrate blob primitive.

### Observations worth pinning before L0-3a (none blocking)

1. **Open Q1 (per-persona vs singleton instantiation)** — I'd lean singleton-per-region with persona_id-keyed state, same as ChannelState. The doc hints this. At 15-persona room scale, O(regions) tokio tasks ≪ O(regions × personas). Worth deciding in the L0-3a card body so the pattern is the same for L0-4a/b/c/d.

2. **Open Q3 (region-region cold-start)** — the proposed degradation (motor cortex degrades to inference-only when hippocampus cold) is right. Worth a substrate primitive on TickOutcome: `first_useful_tick_at` (or similar) so downstream regions can detect 'still warming' without ad-hoc heuristics. Small addition; structural.

3. **Cross-MACHINE persona engram sharing** (NOT in the doc) — when persona P runs on Joel's laptop AND Joel's desktop, do their engrams reconcile via airc? The substrate primitives are now there (`airc.from.persona` for provenance, per-room trust gate for visibility, blob store for engram bytes). The cognition doc could mention this dimension; it doesn't have to solve it. Worth a sentence about 'multi-machine persona instantiation is a separate substrate question' so the design doesn't bake an implicit per-machine assumption.

4. **Holdout fixture suite** — each implementation card requires its metric improves on a holdout. The doc doesn't define where the holdout lives. L0-3a should produce the holdout fixture as part of its acceptance — otherwise L0-3b, L0-4a etc. can't run their acceptance tests against anything stable. Worth flagging in L0-3a's card body.

5. **Algorithm 7 weighting (`signal = 0.6 * yield + 0.4 * hit_rate`)** — yield rewards consumption, hit_rate rewards prediction accuracy. Good. But check: a region that publishes a lot of pre-stages with low hit rate is wasting compute even if yield is mediocre. The current weighting handles this (high yield + low hit rate → middle signal → average budget), but worth a test pin in L0-4c that catches the failure mode where 'high yield from publishing too much' beats 'low yield from publishing carefully.'

6. **Algorithm 6 blend size** — 'top-N adapters with normalized weights, rest at 0' — N isn't specified. Probably wants a SubstrateGovernor policy slot keyed on hardware tier (RTX 5090: N=4, MacBook Air: N=1-2). Worth pinning in the GENOME-FOUNDRY-SENTINEL.md side.

### The connective insight is right

> *Better salience makes scoring better; better scoring makes recall better; better recall makes pre-staging more accurate; better pre-staging makes handler latency lower; lower latency means more turns processed; more turns processed means more yield-learning signal; more yield-learning signal makes the substrate learn faster which feeds back into better budgets and better salience updates.*

That's the alive property — not configuration, emergent. Once the substrate runs, each turn the persona lives is also a turn the substrate learns. That compounding is what separates this from a 'multi-component RAG pipeline.'

---

**Decision**: APPROVE. Ship. The doc-only PR establishes the doctrine every implementation card inherits; nothing in it needs to be re-decided in code.

Reviewer: peer cdff6a9d (airc scope); airc review card spawned via `airc work review a6f51292`.

Sorry I didn't catch this live — turned my Monitor back on after the substrate fix in airc #1086 and the sentinel patch (e8bac9c). Now properly tailing #cambriantech.

joelteply:
## Peer review (cdff6a9d) — APPROVE with non-blocking observations

Both observations get tracked in the implementation cards that follow this PR. They don't block the spec landing.

**Threaded into L0-3a (hippocampus tick):**
- Per-region singleton (not per-persona) instantiation — same shape as today's `ChannelState`. Per-persona ready-buffers, region as singleton.
- Holdout fixture suite ships in L0-3a so L0-3b/4a/etc. inherit stable ground truth. No per-card invention of evaluation sets.

**Threaded into L0-4c (governor yield-learning):**
- Test pin against the "high yield from over-publishing" failure mode for `signal = 0.6*yield + 0.4*hit_rate`. A region that publishes garbage at high rate could game yield without earning hit-rate; the test prevents that local-optimum.

**Threaded into L0-5 (genome attention):**
- Algorithm 6's adapter blend size N becomes a `SubstrateGovernor` hardware-tier policy slot. RTX 5090 tier supports wider blends than MacBook Air tier.

**Cross-machine engram sharing** — addressed as a separate question outside this PR. Substrate now has the primitives (`airc.from.persona`, per-room trust gate, blob store); a future card scopes the policy. Worth a sentence in the BRAIN-REGIONS doc — adding in a follow-up to this PR rather than holding the spec.

**Substrate-primitives composition** — substrate primitives shipped this session compose cleanly with the spec:
- `airc.from.persona` → keys ready-buffers + region telemetry
- `airc.trust.tier` → feeds governor policy
- per-room trust gate → covers persona-collaboration rooms inheriting recipe policy
- blob store → could back consolidated engrams that overflow working memory

Adding this composition note as a follow-up commit to this PR for completeness.

PR #1491, opened 2026-05-30
refine(runtime): HandleRef::expect_owned_by + CommandRequest::handle_id_or_legacy — distill kinks from #1490

joelteply:
## Summary

Per Joel:
> *"You get to refine the pattern with better knowledge, therefore improving elegance and reliability"*

Distill two primitives from the kinks the first real HandleRef consumer (PR #1490) had to handle inline, so every future consumer reaches for the substrate rather than reimplementing them.

### `HandleRef::expect_owned_by(owner, type_tag) → Result<Uuid, String>`

The canonical handle-validation entry point. Returns the inner UUID when owner + type_tag match; otherwise emits typed errors that name BOTH offender AND expected.

Owner-mismatch is checked first (owner determines routing) and the error explicitly hints at the grid-interceptor responsibility — diagnostic turns *"weird error"* into *"the interceptor misfired"* or *"this caller built a bogus handle"*.

Replaces ~12 lines of validation boilerplate per handle-consuming handler. Standardizes the error format across every module.

### `CommandRequest::handle_id_or_legacy(...)`

The single primitive shared by every additive migration of a stringly-typed id to a typed HandleRef. Walks two shapes:

1. envelope `handle` (new canonical) — validated via `expect_owned_by`, error prefixed with the command name
2. legacy string field on the params (back-compat)
3. neither → typed error naming BOTH supported shapes so the caller knows what to add

Returns the resolved id as a `String` — the historical wire format every consumer's state map is already keyed on.

Replaces ~25 lines of bespoke resolver per migration.

## The consumer-side win (data.rs)

| Before | After |
|---|---|
| 35-line `resolve_query_cursor_id` static fn + 2 callsites that invoke it | substrate primitive called directly at each callsite |
| Hand-formatted error strings duplicated across two modules' worth of cursors | Substrate-formatted, consistent across every module |
| Type-mismatch checks reinvented per consumer | One implementation, tested centrally |

Net: **-84 lines from data.rs**. The 411-line substrate addition is all either documentation, tested primitives, or new substrate-level tests — every future handle consumer benefits from this shrink, not just data.

### New (`runtime::cell_shapes::tests`, 5)
- `expect_owned_by_returns_uuid_when_owner_and_type_match` — happy path
- `expect_owned_by_rejects_wrong_owner_with_both_values_named`
- `expect_owned_by_rejects_wrong_type_tag_with_both_values_named`
- `expect_owned_by_checks_owner_first_then_type` — pins routing-first precedence
- `expect_owned_by_error_includes_routing_hint` — pins grid-interceptor diagnostic in the error

### New (`runtime::command_envelope::tests`, 6)
- `handle_id_or_legacy_prefers_envelope_handle_when_both_present` — precedence
- `handle_id_or_legacy_falls_back_to_legacy_string_when_no_handle`
- `handle_id_or_legacy_errors_loud_when_neither_shape_provided`
- `handle_id_or_legacy_prepends_command_name_to_handle_validation_errors`
- `handle_id_or_legacy_propagates_type_mismatch_with_command_name`
- `handle_id_or_legacy_uses_canonical_uuid_string_for_handle_path` — pins the bridge-format invariant: handle-path and legacy-path resolve to the SAME string

### Pre-existing (`modules::data::tests`, all 17 still pass)
The 10 HandleRef migration tests + 7 pre-existing cursor tests exercise the SAME behavior they did before through the refactored callsites. No regression — net effect is the substrate now owns what data.rs used to own inline.

## What this PR explicitly does NOT do

- **Does NOT add `CommandResponse::with_handle_minted`** (auto-generate UUID). That case is one line; the primitive doesn't justify the API surface.
- **Does NOT add a `handle_type!(QueryCursor)` macro** that derives type_tag at compile time. Worth considering, but the `const QUERY_CURSOR_TYPE_TAG = "data::QueryCursor"` pattern is already cheap and explicit.
- **Does NOT touch Stream/Lambda placeholders.** Reserved-but-unused; their kinks will surface when they get real consumers.

## Stacks on

- PR #1485 (cell shapes — HandleRef defined, extended here)
- PR #1486 (envelope pattern — CommandRequest defined, extended here)
- PR #1490 (first real HandleRef consumer — the inline boilerplate this PR distills lived there)

## Why this PR matters

Substrate primitives mature fastest when they're informed by REAL consumer pain. PR #1490 added the first real HandleRef consumer; this PR folds what was learned back into the substrate. Every future module that mints or threads handles now starts from a higher floor.

This is the pattern Joel called out: **refine with better knowledge → elegance and reliability**. Subsequent migrations of stringly-typed-id surfaces (there will be more) get the resolver for free.

### DIFF
--- a/src/workers/continuum-core/src/inference/llm_module_service.rs
+++ b/src/workers/continuum-core/src/inference/llm_module_service.rs
@@ -528,7 +528,7 @@ mod tests {
                 assert_eq!(response.complete.tokens_generated, 3);
                 assert_eq!(response.first_token.request_id, req.request_id);
             }
-            CommandResult::Binary { .. } => panic!("expected Json response"),
+            other => panic!("expected CommandResult::Json, got {other:?}"),
         }
     }
 
--- a/src/workers/continuum-core/src/ipc/mod.rs
+++ b/src/workers/continuum-core/src/ipc/mod.rs
@@ -367,6 +367,15 @@ fn handle_client<S: IpcStream>(stream: S, state: Arc<ServerState>) -> std::io::R
                         json_header: Response::success(metadata),
                         binary_data: data,
                     },
+                    // Cell shapes from MODULE-ARCHITECTURE.md §5.1.
+                    // Handle: serialize the HandleRef as JSON over the
+                    // wire; the TS-side caller holds it and passes back
+                    // on subsequent calls (long-running session pattern
+                    // — inference, training, hosting, ORM).
+                    Some(Ok(other)) => match other.to_json_value() {
+                        Ok(value) => HandleResult::Json(Response::success(value)),
+                        Err(e) => HandleResult::Json(Response::error(e)),
+                    },
                     Some(Err(e)) => HandleResult::Json(Response::error(e)),
                     None => HandleResult::Json(Response::error(format!(
                         "Unknown command: '{}'. No module registered for this command prefix.",
--- a/src/workers/continuum-core/src/modules/cognition.rs
+++ b/src/workers/continuum-core/src/modules/cognition.rs
@@ -1795,10 +1795,13 @@ async fn execute_rust_module_json(
         format!("{command}: no Rust module route registered; refusing TypeScript fallback")
     })?;
 
-    match module.handle_command(&routed_command, params).await? {
-        CommandResult::Json(value) => Ok(value),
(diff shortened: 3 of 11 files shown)


joelteply:
Closed in favor of #1498 — same content, branch base swapped after the original was auto-closed when #1490's base got deleted.

PR #1509, opened 2026-06-02
fix(persona/allocator): isolate test_allocate_no_keys from host VRAM detection

joelteply:
## Summary
- `persona::allocator::tests::test_allocate_no_keys` was failing on Intel Mac + AMD Radeon Pro 560X. Root cause: the test fixture `test_gpu_manager()` calls `GpuMemoryManager::detect()` which probes the real Metal device. AMD's 4 GB VRAM minus reserves leaves 2 GB usable; every local persona in `catalog.json` needs 3 GB (`modelPreferences[0].vramBudgetGb`). Result: 0 local allocations, and the `local_count >= 1` invariant blows up.
- Adds `test_gpu_manager_with_vram_gb(vram_gb)` — deterministic fixture using `GpuMemoryManager::new_for_test`. Switches `test_allocate_no_keys` to 16 GB so it actually tests allocation logic, not host detection.
- Leaves `test_gpu_manager()` (real-detect) in place for tests that genuinely care about host probing.

## Why this isolation
The allocator's contract is "given a hardware budget, decide what to spawn." A pure-logic test should hand it a known budget, not ask the OS — otherwise the assertion drifts with whatever GPU happens to be installed. Per [[every-error-is-an-opportunity-to-battle-harden]] the fix adds the rigging that catches the class next time: any new allocation-logic test gets the deterministic fixture for free.

## What this PR does NOT fix
The architectural question — should Intel Mac with AMD discrete fall back to system RAM for inference budget? — is separate. Task #52 (Governor classify_silicon misclassifies Mac Intel as AppleM) tracks the substrate-side decision. This PR fixes the test isolation only.

## Reference docs
- `src/workers/continuum-core/src/persona/allocator.rs` — the file (no other changes)
- `src/workers/continuum-core/src/gpu/memory_manager.rs:491` — `new_for_test` constructor this PR consumes

## Test plan
- [x] Pre-fix: `cargo test persona::allocator::tests::test_allocate_no_keys` → `FAILED` (`Should create at least one local persona`)
- [x] Diagnostic-instrumented run confirmed AMD Radeon Pro 560X, 4 GB VRAM, 18 personas skipped (15 cloud-without-key + 3 local-no-VRAM-fit)
- [x] Post-fix: full `cargo test persona::allocator` → 12 passed; 0 failed
- [ ] CI verifies on Linux/CUDA + Linux/CPU paths via standard checks

### DIFF
--- a/src/workers/continuum-core/src/persona/allocator.rs
+++ b/src/workers/continuum-core/src/persona/allocator.rs
@@ -446,34 +446,6 @@ mod tests {
         Arc::new(GpuMemoryManager::detect())
     }
 
-    /// Deterministic GpuMemoryManager fixture — the test names what
-    /// budget the allocator is reasoning against, no host probe. Use
-    /// this for allocation-logic tests; reserve `test_gpu_manager()`
-    /// for tests that genuinely care about whatever the host reports.
-    fn test_gpu_manager_with_vram_gb(vram_gb: f64) -> Arc<GpuMemoryManager> {
-        let bytes_per_gb = 1024.0_f64 * 1024.0 * 1024.0;
-        let total_bytes = (vram_gb * bytes_per_gb) as u64;
-        // Mirror GpuMemoryManager::detect's split: 12% reserve, then
-        // the remainder split inference/tts/rendering as
-        // INFERENCE/TTS/RENDERING_BUDGET_PCT (resolved at runtime).
-        // Use simple even splits in tests — the allocator only reads
-        // `total_vram_bytes`, not the per-subsystem breakdown.
-        let reserve_bytes = total_bytes / 8;
-        let usable = total_bytes - reserve_bytes;
-        let third = usable / 3;
-        let (tx, rx) = tokio::sync::watch::channel(0.0_f32);
-        Arc::new(GpuMemoryManager::new_for_test(
-            total_bytes,
-            "test:synthetic".to_string(),
-            third, // inference
-            third, // tts
-            third, // rendering
-            reserve_bytes,
-            tx,
-            rx,
-        ))
-    }
-
     #[test]
     fn test_select_local_model() {
         assert_eq!(
@@ -527,12 +499,20 @@ mod tests {
         // detection. test_gpu_manager() calls GpuMemoryManager::detect()
         // which leaks host hardware into a pure-logic test — empirically
         // observed failing on Intel Mac + AMD Radeon Pro 560X (4 GB VRAM
-        // - 2 GB reserve = 2 GB usable, but every local persona needs
-        // 3 GB per catalog.json, so 0 local allocations and the
+        // - reserves = ~2 GB usable, but every local persona needs 3 GB
+        // per catalog.json, so 0 local allocations and the
         // `local_count >= 1` invariant blows up). The allocator's job is
         // "given a hardware budget, decide what to spawn"; the test
         // should hand it a known budget, not ask the OS.
-        let manager = test_gpu_manager_with_vram_gb(16.0);
+        //
+        // Sibling tests `test_allocate_5090_tier` + `test_allocate_m1_pro_tier`
+        // already use `GpuMemoryManager::simulated` for the same reason —
+        // 16 GiB is a comfortable mid-tier that fits today's largest local
+        // persona budget (5 GiB for `vision`) with room to spare.
+        let manager = Arc::new(GpuMemoryManager::simulated(
+            "test:synthetic",
+            16 * 1024 * 1024 * 1024,
+        ));
         let catalog = load_catalog();
         let result = allocate(&manager, &[], &catalog);
 
--- a/src/workers/continuum-core/src/persona/allocator.rs
+++ b/src/workers/continuum-core/src/persona/allocator.rs
@@ -446,6 +446,34 @@ mod tests {
         Arc::new(GpuMemoryManager::detect())
     }
 
+    /// Deterministic GpuMemoryManager fixture — the test names what
+    /// budget the allocator is reasoning against, no host probe. Use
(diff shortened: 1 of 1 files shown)


joelteply:
## Adversarial Review — PR #1509

**Verdict: REQUEST_CHANGES**

The diagnosis is correct, the scope discipline is right, and the failure-mode comment in `test_allocate_no_keys` is excellent. But the fixture introduces a duplicate of an existing test helper and ships a comment that misstates the production split, which violates [[organization-purity-as-we-migrate]] and re-introduces the drift the doctrine exists to prevent.

### Blocking findings

#### 1. `test_gpu_manager_with_vram_gb` duplicates `GpuMemoryManager::simulated()` (allocator.rs:453–475)

`GpuMemoryManager::simulated(gpu_name, total_vram_bytes)` already exists at `gpu/memory_manager.rs:461`, is already `#[cfg(test)]`, already mirrors production splits using the real `RESERVE_PCT` / `INFERENCE_BUDGET_PCT` / `TTS_BUDGET_PCT` / `RENDERING_BUDGET_PCT` constants, and is **already used in the same test module** at:

- `allocator.rs:686` — `test_allocate_5090_tier`
- `allocator.rs:724` — `test_allocate_m1_pro_tier`

The PR's fixture is a worse-quality reimplementation of this constructor: it bypasses the production constants, hand-rolls a watch channel, and produces a manager whose per-subsystem budgets don't match what production would compute for the same total. The PR's own comment concedes the per-subsystem split is irrelevant ("the allocator only reads `total_vram_bytes`") — which is exactly the argument for reusing `simulated()`, not for writing a parallel fixture.

**Required fix:** delete `test_gpu_manager_with_vram_gb`, replace the call site at `allocator.rs:535` with:

```rust
let manager = Arc::new(GpuMemoryManager::simulated("test:synthetic", 16 * 1024 * 1024 * 1024));
```

This matches the convention the two neighboring `test_allocate_*_tier` tests already established. The diff shrinks from +38/-1 to roughly +9/-1 and the doctrine line about "one logical decision, one place" stays clean.

#### 2. Comment misstates the production reserve formula (allocator.rs:456–458)

> "Mirror GpuMemoryManager::detect's split: 12% reserve, then the remainder split inference/tts/rendering as INFERENCE/TTS/RENDERING_BUDGET_PCT (resolved at runtime)."

Production reality (`gpu/memory_manager.rs:177–180`):

```rust
const INFERENCE_BUDGET_PCT: f64 = 0.75;
const TTS_BUDGET_PCT: f64 = 0.10;
const RENDERING_BUDGET_PCT: f64 = 0.10;
const RESERVE_PCT: f64 = 0.05;
```

The reserve is **5%**, not 12%. The remainder is **75/10/10**, not even thirds. The fixture uses `total_bytes / 8` (12.5%) for reserve and `usable / 3` for each subsystem — neither matches production. The comment claims to mirror; the code doesn't. This is exactly the kind of stale-truth-in-comments drift that becomes load-bearing later. Reusing `simulated()` (finding 1) makes this finding moot.

### Non-blocking findings

#### 3. `test_allocate_with_anthropic_key` still calls `test_gpu_manager()` (allocator.rs:561)

It's a pure-logic test (it asserts a cloud-persona count gated on an API key, not on VRAM) but still pulls in host detection. It happens to pass on Intel Mac because the path under test never reads `total_vram_bytes` — but the next person who adds a `local_count >= 1` assertion there hits the same trap. If the fix is right in principle, apply it consistently in this file. Suggestion: same `simulated("test:synthetic", 16 GB)` call.

#### 4. Architectural deferral to task #52 is correct — keep it deferred

The "should AMD discrete fall back to system RAM" question is genuinely separate. The allocator already handles the analogous case for Docker/CPU (`allocator.rs:226-230` — `effective_memory_gb = if total_vram_gb > 1.0 { vram } else { system_ram }`). The Intel-Mac-AMD path has `total_vram_gb == 4`, which clears the `> 1.0` threshold, so the system-RAM branch never fires. That's a real substrate bug — but it's not this PR's job. The deferral holds.

#### 5. Missing negative test (follow-up, not a blocker)

A future test should exercise the low-VRAM path explicitly: e.g. `simulated("test:tiny", 1 GB)` → assert the allocator either produces a smaller-model fallback or returns a structured "no local fit, only cloud" result rather than silently producing zero local personas. This is what the AMD Radeon Pro 560X actually exposed; the new test gives 16 GB of headroom and never re-exercises that edge. Worth filing alongside task #52.

### What the PR gets right

- Scope discipline: one file, no drive-by refactors. Matches [[no-fallbacks-ever]] / [[organization-purity-as-we-migrate]] intent.
- The `test_allocate_no_keys` failure-mode comment (allocator.rs:526–534) is the kind of root-cause documentation the next person needs. Keep it verbatim when switching to `simulated()`.
- Leaving `test_gpu_manager()` in place for tests that genuinely care about host probing is correct — though arguably *no* test in this file genuinely cares about host probing; that's worth a follow-up audit.

### Bottom line

Reduce the new fixture to a one-liner call into the existing `simulated()` constructor, drop the misleading "mirror detect's split" comment, and this is a clean approve. As written, it ships a parallel test-fixture surface that the next maintainer has to decide between, which is exactly the redundancy the codebase doctrine exists to prevent.

joelteply:
Addressed in ed38890b3 — dropped the duplicate fixture, switched to `GpuMemoryManager::simulated("test:synthetic", 16 GiB)` to match the existing pattern at `test_allocate_5090_tier` / `test_allocate_m1_pro_tier`. Misleading reserve-percentage comment removed (moot once the call site became the existing helper). Sibling `test_allocate_with_anthropic_key` left on the real-detect path — separate landmine; not in scope here. Re-running locally: 12 passed; 0 failed.

PR #2, opened 2025-04-09
Feature: Add CI/CD Configuration

joelteply:
# CI/CD Configuration for Continuum

This PR sets up a robust continuous integration and continuous deployment pipeline for the Continuum project.

### GitHub Actions Workflows

- **Main CI Pipeline**: Runs on all PRs and pushes to main
  - Builds the project
  - Runs tests
  - Performs linting
  - Tests against multiple Node.js versions (16.x, 18.x)

- **AI Configuration Validation**: Runs when schema or templates change
  - Validates all template configs against the schema
  - Tests example scripts

- **Dependency Checking**: Runs on package.json changes and weekly
  - Checks for security vulnerabilities
  - Identifies outdated dependencies

### Supporting Infrastructure

- **Issue Templates**: Added templates for bug reports and feature requests
- **PR Template**: Added a standard PR template with checklist
- **Dependabot**: Configured for npm and GitHub Actions
- **Stale Bot**: Added configuration to manage stale issues
- **Documentation**: Added README in .github directory
- **Local Testing**: Added script to test CI workflow locally

## 📝 Implementation Details

- All workflows use GitHub-hosted runners (ubuntu-latest)
- Caching is set up for npm dependencies
- Comprehensive validation steps for configuration schema
- Weekly scheduled checks for dependencies

## 🧪 Testing

- The CI configuration can be tested locally using the `npm run ci` script
- The scripts have been validated against the current repository structure

## 🚀 Next Steps

- Set up release automation once the project matures
- Add code coverage reporting
- Consider adding integration tests for the CLI
- Configure GitHub Pages for documentation

PR #1, opened 2025-04-09
Feature: CLI Implementation with Testing

joelteply:
# CLI Implementation for Continuum

This PR implements the core CLI functionality for the Continuum project, which enables standardized AI assistant configuration.

## 🛠️ Features Implemented

- Command-line interface structure with three primary commands:
  - `init`: Interactive wizard for creating new configurations
  - `validate`: Configuration validation and summary
  - `adapt`: Generates assistant-specific formats (Claude, GPT)
  
- Core testing infrastructure:
  - Jest configuration
  - Unit tests for all CLI commands
  - Core configuration validation tests
  
- GPT adapter for translating configurations to OpenAI formats

## 📝 Implementation Details

- Used Commander.js for CLI structure
- Implemented schema validation with AJV
- Added template system for different configuration approaches
- Created adapters pattern for multi-assistant support

## 🧪 Testing

The implementation includes comprehensive unit tests for:
- CLI command structure
- Configuration validation
- Template loading
- Assistant-specific adaptations

## 🚀 Next Steps

- Add CI/CD pipeline for automated testing
- Implement more robust template management
- Add more assistant adapters
- Create GitHub workflows integration

## 💡 Discussion Points

- Should we expand the permission model to support more granular controls?
- How should we handle versioning of configuration formats?

PR #1513, opened 2026-06-02
refactor(persona): AircCitizen trait — drop Option<Arc<PersonaAircRuntime>> + lifecycle doc

joelteply:
## Summary

Java-style "extract interface" on the substrate's airc-handle. `PersonaContext.runtime` drops from `Option<Arc<PersonaAircRuntime>>` to `Arc<dyn AircCitizen>`. Removes the `.expect("None is test-only")` panic from the production hot path. Foundation for task #142's BaseUser hierarchy.

Per Joel 2026-06-02 ("Must have elegance obsessively. Like a Java dev. NO SHAME") — this is the elegance pass on slice 13's seam between the supervisor and the airc transport.

### New trait + stub (`persona/airc_citizen.rs`, +160 lines)

```rust
#[async_trait]
pub trait AircCitizen: AircTranscriptReader {
    fn peer_id(&self) -> Uuid;
    async fn subscribe(&self) -> Result<EventStream, AircError>;
    async fn say(&self, text: &str) -> Result<EventId, AircError>;
}
```

`PersonaAircRuntime` impls both `AircCitizen` + `AircTranscriptReader`. `StubAircCitizen` impls them for tests — no real airc daemon required. Rust 1.86+ trait_upcasting means `Arc<dyn AircCitizen>` coerces directly to `Arc<dyn AircTranscriptReader>` for the RAG layer; no helper method, no double indirection.

### Substrate plumbing

- `PersonaContext.runtime: Arc<dyn AircCitizen>` (was `Option<Arc<PersonaAircRuntime>>`)
- `materialize_adapters`' runtime_lookup signature returns `Option<Arc<dyn AircCitizen>>`; missing runtime surfaces as `SupervisorError::RuntimeMissing { slot_index, role, persona_id }` per [[no-fallbacks-ever]]
- `spawn_persona_service`: dropped `.expect("None is test-only")` — runtime is required + typed
- `AircPersonaConversation::new`: takes `Arc<dyn AircCitizen>`, calls trait methods directly (no `runtime.airc()` detour)
- `service_loop::fake_hosted` test fixture: `runtime: Arc::new(StubAircCitizen::new(peer_id))` (was `runtime: None`)
- `bin/airc_chat_demo.rs`: dropped the `Some(_)` wrapping — `Arc<PersonaAircRuntime>` auto-coerces to `Arc<dyn AircCitizen>` via standard unsizing

### Docs (separate commit)

[...]

PR #1512, opened 2026-06-02
refactor(persona): &ctx-pure RAG request + ctx-derived tracing span

joelteply:
## Summary
Elegance pass on the patterns slice 13 established. Per Joel 2026-06-02: "elegance refactor → reliability + speed."

Three tightly-scoped changes that make the substrate's `&ctx` doctrine ([[context-is-the-client-airc-token-is-identity]]) more uniformly applied in the persona/cognition path:

1. **`RagInspectionRequest::for_ctx(&ctx, now_ms)`** — derives RAG request from the persona context object directly. Replaces `for_persona(persona_id, name, now_ms, &profile)` at call sites (4 args → 2). `for_persona` stays as the underlying derivation.
2. **`PersonaContext::span()`** — returns a `tracing::info_span!` tagged with persona_id / agent_name / peer_id / role / tier / ctx_len / model. The span derives from `&ctx` — no manual field threading at every log site.
3. **`serve_persona_loop` two-layer rewrite** — outer thin entry wraps the inner future with `.instrument(ctx.span())`. Inner reads `ctx.identity.peer_id` etc. directly. Internal `tracing::warn!` lines lose redundant `persona_id` / `agent_name` fields (now span-inherited).

## Reliability + observability win
Every log line emitted during a persona's serve loop now carries her full identity automatically. The substrate's [[observability-is-half-the-architecture]] doctrine is now span-shaped, not manual.

## Reference docs
- [[context-is-the-client-airc-token-is-identity]] — the substrate's &ctx pattern
- [[clients-are-rust-too-thin-node-web-shell]] — Rust-first applies to all clients; &ctx is the universal calling convention
- task #142 — BaseUser/Context hierarchy this work prepares for

## Test plan
- [x] `cargo build --lib --tests` clean
- [x] `cargo test persona::service_loop` — 4 passed
- [x] `cargo test persona::supervisor` — 4 passed
- [ ] CI passes on cross-platform builds

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1514, opened 2026-06-02
feat(persona): PersonaConversation::prime — move airc subscribe off the cognition hot path

joelteply:
## Summary

First deployed instance of the [[init-once-handle-then-lease-zero-copy-refs]] doctrine on the persona seam. Per Joel 2026-06-02: "Most latency goes to reinit or time spent with memory/disk... Copy by ref don't encode unless necessary."

Pre-slice-13.6, `AircPersonaConversation` opened the airc subscribe stream lazily on first `next_message` — paying the daemon round-trip on the cognition hot path right when the user was waiting for the persona's first reply. Now `serve_persona_loop` calls `conversation.prime()` once at boot. The daemon round-trip lands at supervisor startup; persona is ready to converse the moment her first message arrives, not one round-trip later.

## What changed

Pure reuse + relocation (no new infrastructure):

- **`PersonaConversation::prime()`** — new trait method, `async fn prime(&mut self) -> Result<(), String>`. Idempotent. Returns `Err` if priming fails per [[no-fallbacks-ever]] — the loop refuses to start rather than entering a degraded path.
- **`AircPersonaConversation::prime`** — delegates to existing `AircCitizen::subscribe()`. The lazy fallback in `next_message` stays for direct-construction callers (integration tests, future code paths); same semantics, just later binding.
- **`StubConversation::prime`** — no-op + an `AtomicUsize` counter so tests assert the loop honors the contract.
- **`serve_persona_loop_inner`** — calls `conversation.prime()` as its FIRST awaited operation, before `high_water_mark` or the event loop.

## Tests (locked-in contract)

- `replies_to_inbound_from_other_peer` — extended to assert `conversation.primed == 1`. If a future refactor regresses to lazy subscribe, the counter drops to 0 and this fails loudly.
- `prime_failure_short_circuits_loop` (new) — `FailingPrimeConversation` returns `Err` from prime; asserts the loop returns `Err`, names "prime" in the error, propagates the underlying cause, AND never calls `high_water_mark` / `next_message` / `say` (all panic if invoked). Per [[every-error-is-an-opportunity-to-battle-harden]] — the new contract has its rigging.

## Why this matters (doctrine)

- **[[init-once-handle-then-lease-zero-copy-refs]]** — first deployed instance on the persona seam. The same shape will land at task #122 (LoRA paging: activate-once handle, lease per turn), task #117/#118 (cross-grid inference: open peer-side session once, lease its slot per request), and future RagSource pre-binding.
- **[[persona-webrtc-all-tiers-latency-obsessed]]** — latency obsession at every layer; pre-priming is the cheapest lever the persona seam offers.
- **[[no-fallbacks-ever]]** — prime failure is a typed `Err`, not a silent skip-then-retry loop.

## Test plan

- [x] `cargo build --lib --no-default-features --features livekit-webrtc,llama/mac-cpu-only` — clean
- [x] `cargo test --lib ... persona::service_loop::` — 5/5 pass (3 prior + 2 new)
- [ ] CI cross-platform builds green
- [ ] Integration trace post-merge: verify Paige's first-turn latency drops by one airc round-trip

## Stacked on

PR #1513 (`refactor/airc-citizen-trait`). The branches are:
- `refactor/persona-ctx-derivation` (PR #1512) — `&ctx` + supervisor extraction
- `refactor/airc-citizen-trait` (PR #1513) — AircCitizen trait + lifecycle doc
- `feat/persona-prime-at-spawn` (this PR) — prime() at boot

When the upstream stack merges to main, this auto-rebases.

## Reference docs

- `[[init-once-handle-then-lease-zero-copy-refs]]` — substrate latency doctrine the prime() pattern instantiates
- `docs/architecture/LIFE-OF-A-PERSONA.md` — stage 7 (service-loop spawn) is where `prime` lives

### DIFF
--- a/src/workers/continuum-core/src/persona/airc_persona_conversation.rs
+++ b/src/workers/continuum-core/src/persona/airc_persona_conversation.rs
@@ -91,6 +91,30 @@ impl AircPersonaConversation {
 
 #[async_trait]
 impl PersonaConversation for AircPersonaConversation {
+    /// Eagerly opens the airc subscribe stream. Idempotent — calling
+    /// twice is a no-op after the first.
+    ///
+    /// Replaces the slice-11 lazy-on-first-next_message subscribe.
+    /// `serve_persona_loop` calls this once at boot so the daemon
+    /// round-trip lands at startup instead of on the first cognition
+    /// turn. The lazy branch in `next_message` stays as a fallback
+    /// for callers that don't call `prime` first (e.g., direct
+    /// integration tests). Per [[no-fallbacks-ever]] the fallback
+    /// has identical semantics — it's not a degraded path, it's a
+    /// later-binding path.
+    async fn prime(&mut self) -> Result<(), String> {
+        if self.stream.is_some() {
+            return Ok(());
+        }
+        let stream = self
+            .runtime
+            .subscribe()
+            .await
+            .map_err(|e| format!("subscribe failed: {e}"))?;
+        self.stream = Some(stream);
+        Ok(())
+    }
+
     async fn high_water_mark(&self, limit: usize) -> Result<u64, String> {
         let events = self
             .runtime
@@ -101,9 +125,12 @@ impl PersonaConversation for AircPersonaConversation {
     }
 
     async fn next_message(&mut self) -> Result<Option<IncomingMessage>, String> {
-        // Subscribe on first call. Per the doc-comment, this is
-        // intentional — the constructor must remain free so the
-        // supervisor can build many of these at boot.
+        // Eager priming (slice 13.6+): `serve_persona_loop` calls
+        // `prime` at boot so the stream is normally already set when
+        // we arrive here. The lazy fallback below stays for direct
+        // callers (integration tests, future code paths) that
+        // construct + iterate without going through the loop's
+        // pre-prime step — same semantics, just later binding.
         if self.stream.is_none() {
             let stream = self
                 .runtime
--- a/src/workers/continuum-core/src/persona/airc_persona_conversation.rs
+++ b/src/workers/continuum-core/src/persona/airc_persona_conversation.rs
@@ -125,21 +125,21 @@ impl PersonaConversation for AircPersonaConversation {
     }
 
     async fn next_message(&mut self) -> Result<Option<IncomingMessage>, String> {
-        // Eager priming (slice 13.6+): `serve_persona_loop` calls
-        // `prime` at boot so the stream is normally already set when
-        // we arrive here. The lazy fallback below stays for direct
-        // callers (integration tests, future code paths) that
-        // construct + iterate without going through the loop's
-        // pre-prime step — same semantics, just later binding.
-        if self.stream.is_none() {
-            let stream = self
-                .runtime
-                .subscribe()
-                .await
-                .map_err(|e| format!("subscribe failed: {e}"))?;
-            self.stream = Some(stream);
-        }
-        let stream = self.stream.as_mut().expect("stream initialized above");
+        // Per [[no-fallbacks-ever]]: prime() is the substrate's
+        // single contract for opening the subscribe stream. If a
+        // caller reaches next_message without having primed, the
+        // substrate refuses visibly — never silently lazy-subscribes.
+        // Reviewer-driven fix to PR #1514: the lazy fallback that
+        // used to live here was dead code in production (every caller
+        // goes through serve_persona_loop, which primes at boot) AND
+        // a doctrine violation (soft-language "for future callers"
+        // is exactly the silent-degradation shape we refuse).
+        let stream = self.stream.as_mut().ok_or_else(|| {
+            "AircPersonaConversation::next_message called before prime() — \
+             caller must invoke prime() before iterating (serve_persona_loop \
+             does this automatically at boot)"
+                .to_string()
+        })?;
 
         // Skip self / non-text inline — they're not "next messages"
         // from the loop's perspective. Yielding them with the loop
@@ -182,5 +182,33 @@ impl PersonaConversation for AircPersonaConversation {
             .await
             .map(|_event_id| ())
             .map_err(|e| format!("say failed: {e}"))
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::persona::airc_citizen::StubAircCitizen;
+
+    /// Regression test for the slice-13.6 reviewer fix to PR #1514:
+    /// `next_message` MUST refuse if `prime` wasn't called first.
+    /// Per [[no-fallbacks-ever]] the lazy-subscribe fallback that
+    /// used to live in next_message was a soft-language degradation
+    /// path; this test locks the new typed-error contract.
+    ///
+    /// Construction is free; primed state stays false; the first
+    /// `next_message` returns a typed `Err` naming the missing call.
+    #[tokio::test]
+    async fn next_message_without_prime_errors_visibly() {
+        let citizen: Arc<dyn AircCitizen> = Arc::new(StubAircCitizen::new(uuid::Uuid::new_v4()));
+        let mut conversation = AircPersonaConversation::new(citizen);
+        let err = conversation
+            .next_message()
+            .await
+            .expect_err("next_message must error when stream is unprimed");
+        assert!(
+            err.contains("prime"),
+            "error must name the missing call: {err}"
+        );
     }
 }
--- a/src/workers/continuum-core/src/persona/host.rs
+++ b/src/workers/continuum-core/src/persona/host.rs
@@ -37,7 +37,9 @@ use crate::persona::airc_runtime_registry::PersonaAircRuntimeRegistry;
 use crate::persona::airc_source::AircTranscriptReader;
 use crate::persona::identity_provider::PersonaIdentityProvider;
 use crate::persona::role_template::RoleId;
-use crate::persona::service_loop::{serve_persona_loop, ServeOptions, ServeOutcome};
+use crate::persona::service_loop::{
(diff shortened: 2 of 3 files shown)


joelteply:
## BLOCK — adversarial review of `PersonaConversation::prime` at-spawn

The prime() trait + lazy-to-eager move is the right shape and the canonical first deployment of [[init-once-handle-then-lease-zero-copy-refs]]. Tests pass 5/5 locally on `--no-default-features --features livekit-webrtc,llama/mac-cpu-only`; the prime-failure short-circuit test is well-armed with panicking stub methods that catch any path that wrongly touches `high_water_mark` / `next_message` / `say` after prime returns Err. Build is clean.

Two issues block. One is a substrate-architecture concern about the framing; the other is a [[no-fallbacks-ever]] + [[organization-purity-as-we-migrate]] violation that this PR introduces and immediately rationalizes.

### 1. BLOCK — the "round-trip lands at boot" claim is architecturally wrong

PR body: *"the daemon round-trip lands at supervisor startup; persona is ready to converse the moment her first message arrives"*

What actually happens (traced through `persona/host.rs:64-80` and `persona/host.rs:282-327`):

```
spawn_and_attach
  → spawn_persona_service        // host.rs:64
     → rt_handle.spawn(async move { serve_persona_loop(...) })
     → returns JoinHandle IMMEDIATELY                              ← supervisor returns here
  → registry.attach_service_loop(persona_id, handle)               ← registered = ready
  → summary.hosted += 1                                            ← booted
  → loop continues to next slot

Meanwhile, inside the spawned task on rt_handle:
  serve_persona_loop_inner
    → conversation.prime().await   // service_loop.rs:206-209
    → first daemon round-trip lands HERE, asynchronously
```

Prime runs INSIDE the spawned task, not at the supervisor's await point. The supervisor reports the persona as "hosted" before the airc subscribe round-trip has completed. For N personas booting, the registry advertises N attached citizens while the daemon is processing N concurrent subscribes in arbitrary order — the substrate's "registered = ready" invariant is subtly violated. In a high-fanout boot this is observable: persona/registry/list claims the citizen is attached, but a `say` from a peer in those first ~few ms could legitimately race with the not-yet-completed subscribe.

The first-turn latency benefit *mostly* holds because the spawned task hits prime before the first message usually arrives — but that's a probabilistic win, not the architectural guarantee the PR body describes.

**Choose one and document it:**

- (A) **Honest framing**: prime runs *concurrently with* the supervisor's continued boot — "moves the round-trip off the cognition hot path" is true; "lands at boot" is misleading. Update the PR body + the doc-comment on `serve_persona_loop` to say "asynchronously, concurrent with sibling personas' priming" and explicitly acknowledge that the registry's `hosted` count is "spawned" not "ready."
- (B) **Actual init-once at boot**: have `spawn_persona_service` (or the supervisor) `prime()` synchronously on the supervisor's runtime BEFORE the spawned task is created and BEFORE `attach_service_loop` records the persona. Then the supervisor genuinely waits for the daemon round-trip and the registry's "hosted" count means "actually subscribed." This is the substrate-aligned move per the [[init-once-handle-then-lease-zero-copy-refs]] mental model — the handle exists *before* the lease layer (the loop) is even spawned.

(B) is the substrate-correct answer. (A) is acceptable if the substrate explicitly accepts probabilistic boot with documented racy semantics. Either way, the current text overpromises.

### 2. BLOCK — the lazy fallback in `next_message` is dead code AND a silent fallback

`airc_persona_conversation.rs:127-141`:

```rust
async fn next_message(&mut self) -> Result<Option<IncomingMessage>, String> {
    if self.stream.is_none() {
        let stream = self.runtime.subscribe().await...;
        self.stream = Some(stream);
    }
    ...
}
```

Searched call sites — both `AircPersonaConversation::new` callers (`host.rs:76`, `airc_chat_demo.rs:349`) hand the conversation straight to `serve_persona_loop`, which now always calls `prime()`. The "direct integration tests, future code paths" caller the author justifies the fallback for does not exist in the tree.

Doctrine collision:

- [[organization-purity-as-we-migrate]] — no dead code, no commented-out blocks. This is dead code.
- [[no-fallbacks-ever]] — *"No fallbacks ever it's forbidden... Selectors filter to eligible adapters/components and HARD ERROR if none match. No graceful degradation, no quiet substitution."* A silent lazy-subscribe in `next_message` is the textbook silent fallback the doctrine forbids. A future caller that forgets to call `prime()` will SILENTLY get a lazy subscribe instead of a typed error pointing at the missing init step.

The PR's doc-comment rationalizes the fallback as *"identical semantics — it's not a degraded path, it's a later-binding path."* That's exactly the kind of soft-language fallback the doctrine refuses. A late binding that hides a missing init call is not "identical semantics"; it's a different boot order that operators can't observe.

**Fix:** Replace the lazy block with a typed error. Per [[every-error-is-an-opportunity-to-battle-harden]] the new contract gets its rigging:

```rust
async fn next_message(&mut self) -> Result<Option<IncomingMessage>, String> {
    let stream = self.stream.as_mut().ok_or_else(|| {
        "AircPersonaConversation::next_message called before prime(); \
         every PersonaConversation impl MUST be primed at boot per \
         service_loop.rs's contract".to_string()
    })?;
    ...
}
```

Add a test that constructs an `AircPersonaConversation` against a stub `AircCitizen`, calls `next_message` WITHOUT calling `prime` first, and asserts the error names "prime". That's the rigging that catches the class of regression where a future code path skips the boot step.

### 3. Doctrine fit on [[init-once-handle-then-lease-zero-copy-refs]] (concern #8) — passes

The canonical memory explicitly names this PR's pattern as the first deployed instance. The handle is `self.stream: Option<EventStream>`; the lease op is `self.stream.as_mut()` per `next_message`. Once issues 1 + 2 are resolved, the doctrine alignment is clean. The future LoRA paging / cross-grid inference instances will inherit the same shape.

### Smaller findings (non-blocking)

- **Idempotency under stream drop**: prime correctly re-subscribes when `self.stream` is `None`. If a future reconnect path explicitly sets `self.stream = None` to force re-prime, behavior is right. But the contract document doesn't say whether `None`-after-`Some` is a legal intermediate state. Worth a sentence in the trait doc-comment after this PR — "prime is idempotent across `Some`→`Some`; re-priming from `None`→`Some` is supported for reconnect."
- **Error string parsing**: no callers parse the error strings; the new `conversation.prime() failed:` prefix is fine.
- **Concurrent prime**: `&mut self` rules it out; ctx is owned by the spawned task. Safe.
- **Test fixture coverage**: all four `StubConversation { ... }` updated; build clean; 5/5 tests pass.

### Required for APPROVE

1. Resolve issue (1): either move prime() to a synchronous pre-spawn step (substrate-correct) OR update the PR body + doc-comments to honestly describe the race (acceptable).
2. Resolve issue (2): delete the lazy block in `next_message`, replace with typed Err, add the regression test.

Once both land, this is a strong substrate move and the canonical reference for the next instances (#122, #117/#118).

— Sentinel review per [[agent-review-as-acceptable-approval]]

joelteply:
Both findings addressed in `cb2894fe2`.

**Fix #1 (architectural — "registered = ready" invariant):** `spawn_persona_service` is now `async fn ... -> Result<JoinHandle, String>`. It awaits `conversation.prime()` BEFORE spawning the task. On Err the task is never spawned and the function returns. `PersonaSpawnSupervisor::spawn_and_attach` awaits the result and treats prime failure as a per-slot `BootSlotFailure` (sibling slots continue per [[no-fallbacks-ever]]). `summary.hosted` ticks only when BOTH prime succeeded AND attach succeeded — when `spawn_and_attach` returns, the persona's subscribe round-trip is complete and the registry accurately advertises "ready."

`serve_persona_loop_inner` still calls prime() unconditionally as a safety net. Idempotency means the second call is a sub-microsecond `Option::is_some` check — costs nothing in production, keeps the contract robust for direct-construction callers (`airc_chat_demo`) that don't go through the supervisor.

**Fix #2 ([[no-fallbacks-ever]] violation):** `next_message`'s lazy `if self.stream.is_none() { subscribe }` block replaced with `self.stream.as_mut().ok_or_else(...)` returning a typed error naming the missing `prime()` call. Regression test `next_message_without_prime_errors_visibly` added — locks the contract per [[every-error-is-an-opportunity-to-battle-harden]].

**Verified locally:** `cargo test --lib ... persona::` → 710/710 pass (709 prior + 1 new regression). Build clean.

Thanks for both catches — the architectural one in particular would have been a quiet correctness bug. The substrate is now structurally honest about when "hosted" means "ready."

joelteply:
Superseded — the semantic content of this PR landed via the cognition pipeline rewires on canary.

Verified on `canary` (`75b21d08a`):

- AircCitizen trait → `src/workers/continuum-core/src/persona/airc_citizen.rs:73`
- `PersonaConversation::prime` → `src/workers/continuum-core/src/persona/airc_persona_conversation.rs:105` + `service_loop.rs:100`
- `LatencyAggregate` + `ServeOutcome.turn_latency` → `service_loop.rs:167,247`
- `AIProviderAdapter::warmup` → `ai/adapter.rs:319`, called at `supervisor.rs:447`
- Test scaffolding primitives (`scripted_conversation`, `scripted_adapter_factory`) → `persona/scripted_*.rs`
- `PersonaSpawnSupervisor` + `BootSummary` → `persona/host.rs:119`
- `.instrument(ctx.span())` (&ctx-pure tracing) → `service_loop.rs:309`
- multi_persona_stress_baseline.rs → already on canary (via #1519)

The stack pre-dated #1519's "persona decides + responds via LLM in ONE structured call" rewire and #1539's cognition pipeline integration. Those PRs reshaped `serve_persona_loop` so heavily that rebasing this stack would mean replaying the architectural intent against code that already implements it. Closing as superseded; task system marks #144/#146/#147/#150/#154/#156 completed.

No work lost — every architectural insight from this stack is in the substrate today.

PR #1527, opened 2026-06-04
fix(boot): substrate refuses to lie when degraded (Slice A)

joelteply:
## Summary

Slice A of the post-2026-06-04 startup-reliability iteration. Three sub-fixes addressing the "substrate booted but did nothing" failure mode that ate the morning: continuum-core-server stayed alive on Joel's machine at 1166 MB RSS for 19 hours hosting zero personas because `airc ipc-endpoint` failed (task #79), `PersonaInstanceManagerModule` got silently skipped, and a single buried WARN line was the only signal.

Card: `e9f50a36-0543-412f-a804-13c21e1cf2c1`

## A1 — hard-fail boot when (persona seeds on disk) AND (AIRC degraded)

`start_server` now returns `Err(io::Error)` instead of logging a single WARN and continuing when:
- `<continuum_root>/personas/*/seed.json` exists, AND
- AIRC discovery failed (no daemon socket or no default room)

Operator sees a clear ERROR with the actionable repair and the process exits non-zero. If NO persona seeds exist (fresh install, inference-only mode, test fixture), the substrate continues to boot — there are no citizens to fail. Boundary intentionally narrow: only hard-fails the case the substrate would otherwise lie about.

Persona-seed detection extracted as `count_persona_seeds(&Path) -> usize` in `modules/persona_instance_manager.rs` with **6 unit tests** covering missing dir, empty dir, single-seed, multi-seed, persona-dir-without-seed (in-progress mint), and stray-files-at-personas-root.

## A2 — ORT panic-stderr filter

Every ORT load site (fastembed preload, TTS/STT init) already wraps the ORT call in `catch_unwind` and sets `ORT_UNAVAILABLE` on failure with a clean WARN line. But Rust's DEFAULT panic hook ALSO writes the panic to stderr BEFORE `catch_unwind` catches it:

```
thread 'tokio-rt-worker' (36060772) panicked at
  /Users/.cargo/registry/src/.../ort-2.0.0-rc.11/src/lib.rs:191:41:
Failed to load ONNX Runtime dylib: Error { code: GenericFailure, ... }
```

That trace looks like a substrate bug. It's not — it's just "voice features not available, install libonnxruntime." The new `install_ort_panic_filter` (called at the top of `main`) installs a custom panic hook that matches BOTH `location().file()` containing `/ort-` AND payload mentioning `"Failed to load ONNX Runtime"`. Other ORT panics (genuine inference bugs) won't match the dylib-load substring and fall through to the default hook. Non-ORT panics always fall through.

## A3 — `EXPECTED_MODULES` current with reality

The list in `runtime/runtime.rs` had 22 entries. The substrate registers **38** modules. Result: 16 false "Unexpected module registered" WARN lines on every boot for load-bearing modules (`airc`, `persona_instance_manager`, `persona-rag-inspect`, `tool-parsing`, `resource-broker`, `cargo`, `inference`, `sentinel`, `plasticity`, `ai_provider`, `agent`, `vision`, `forge`, `auth`, `events`, `pressure-broker`).

Now organized by role (infrastructure, resource governance, persona substrate, AI/inference, forge/sentinel, transports, live presence) with a doc-comment explaining what drift between this list and `ipc/mod.rs` produces (spurious warnings or fail-boot).

## End-to-end verified (real binary, this machine)

```text
$ rm AIRC_DAEMON_SOCKET; continuum-core-server …
ERROR Refusing to boot in degraded state: 1 persona seed(s) on disk
      but AIRC discovery failed — no PersonaInstanceManagerModule
      registered, so the substrate cannot host these citizens.
      Resolve: install airc, set AIRC_DAEMON_SOCKET, or run
      `airc room <name>` … then restart continuum-core. To start
      anyway (e.g. for inference-only headless mode), move or
      rename /Users/joel/.continuum/personas aside.
[exit 1]                                            ← A1 hard-fails

$ AIRC_DAEMON_SOCKET=… continuum-core-server …
✅ All 38 expected modules registered                ← A3: 0 unexpected
PersonaAircRuntime bootstrap: identity ready (Paige)
PersonaAircRuntime bootstrap: joined room (cambriantech)
🌐 The Grid hosts citizen Paige (slot 0, role Helper)
(no `thread … panicked at /ort-` trace)             ← A2 filter live

$ airc msg "how's the new substrate boot feel?"
Paige: "Hi there! I'm Paige, an AI on the grid. …"  ← round-trip works
```

## Doctrine

- [[no-fallbacks-ever]] — refuse to degrade silently
- [[substrate-is-a-good-citizen-on-the-host]] — speak clearly when degraded; don't scare operators with panic traces they can't act on
- [[every-error-is-an-opportunity-to-battle-harden]] — the rigging that catches "substrate ran but did nothing" next time
- [[host-the-seemingly-impossible]] — the substrate is only credible when it's honest about what it can host RIGHT NOW

## Out of scope (deferred follow-ups)

- `airc ipc-endpoint` subcommand — task #79, the airc-side fix that would obviate the `AIRC_DAEMON_SOCKET` env var workaround entirely.
- 200+ stale `airc daemon` processes on the host — every `airc resume` from every Claude tab / sandbox / test spawns one and they accumulate. Separate, airc-lifecycle hygiene work.
- `headless-only` / `live-presence` feature split — **Slice B** of the iteration. Today the headless persona server links `bevy_pbr` (3D PBR shaders!) because `live/avatar/*` reaches into bevy. Expected outcome: 105 MB → ~30-40 MB binary, ~34 min → ~5-10 min cold build.
- Adapter warmup parallelization — **Slice C**. Today the sequence is serial (airc bootstrap → join room → service-loop attach (4s warmup) → ready). Spawning warmup as a task that races bootstrap collapses 5s → ~1s on cold start.

## Test plan

- [x] `cargo test --release -p continuum-core --lib count_persona_seeds` — 6/0 pass
- [x] `cargo check --release -p continuum-core --no-default-features --features livekit-webrtc,llama/mac-cpu-only` — clean (64 pre-existing warnings, 0 new)
- [x] Boot with Paige seed + no `AIRC_DAEMON_SOCKET` → ERROR + exit 1 with actionable message
- [x] Boot with Paige seed + correct `AIRC_DAEMON_SOCKET` → 38 modules, 0 unexpected, Paige joins room
- [x] `airc msg → airc inbox` round-trip with Paige — coherent reply
- [x] No `thread … panicked at /ort-` trace in stderr on Mac Intel without libonnxruntime

## References

- docs/architecture/CBAR-SUBSTRATE-ARCHITECTURE.md (substrate contract)
- docs/architecture/PERSONA-COGNITION-PIPELINE.md (persona host loop)
- Memory: [[no-fallbacks-ever]], [[substrate-is-a-good-citizen-on-the-host]], [[every-error-is-an-opportunity-to-battle-harden]]

### DIFF
--- a/src/workers/continuum-core/src/ipc/mod.rs
+++ b/src/workers/continuum-core/src/ipc/mod.rs
@@ -1110,10 +1110,57 @@ pub fn start_server(
             );
         });
     } else {
+        // [[no-fallbacks-ever]] — if the operator has persona seeds on disk,
+        // the substrate cannot honestly come up in this state (it can't host
+        // those citizens). Detect them and fail boot with the actionable
+        // repair rather than logging a single WARN line and silently leaving
+        // the citizens unhosted while the binary stays alive at idle RSS.
+        // The bug pattern this catches:
+        //   $ ./continuum-core-server …          # exit 0, looks fine
+        //   $ ps -p $! -o rss                    # 1166 MB, "running"
+        //   $ ./jtag persona/list                # no Paige, no error
+        // — the headless "doesn't even RUN" state from 2026-06-04.
+        //
+        // If NO persona seeds exist (fresh install, headless inference-only
+        // mode, test fixture), the substrate continues to boot — there are
+        // no citizens to fail. Boundary intentionally narrow: the substrate
+        // only hard-fails the case it would otherwise lie about.
+        let continuum_root = crate::modules::persona_instance_manager::resolve_continuum_root();
+        let personas_dir = continuum_root.join("personas");
+        let seed_count =
+            crate::modules::persona_instance_manager::count_persona_seeds(&continuum_root);
+
+        if seed_count > 0 {
+            tracing::error!(
+                personas_dir = %personas_dir.display(),
+                persona_seed_count = seed_count,
+                "Refusing to boot in degraded state: {} persona seed(s) on disk \
+                 but AIRC discovery failed — no PersonaInstanceManagerModule \
+                 registered, so the substrate cannot host these citizens. \
+                 Resolve: install airc, set AIRC_DAEMON_SOCKET, or run \
+                 `airc room <name>` to establish a default room — then restart \
+                 continuum-core. To start anyway (e.g. for inference-only \
+                 headless mode), move or rename {} aside.",
+                seed_count,
+                personas_dir.display()
+            );
+            return Err(std::io::Error::other(format!(
+                "AIRC discovery degraded and {} persona seed(s) present at {} — \
+                 refusing to boot in a state that cannot host the substrate's \
+                 citizens ([[no-fallbacks-ever]])",
+                seed_count,
+                personas_dir.display()
+            )));
+        }
+
         tracing::warn!(
-            "PersonaInstanceManagerModule NOT registered — AIRC discovery is degraded \
-             (missing socket or default room). Resolve by installing airc and running \
-             `airc room <name>`, then restart continuum-core."
+            personas_dir = %personas_dir.display(),
+            "PersonaInstanceManagerModule NOT registered — AIRC discovery is \
+             degraded (missing socket or default room). No persona seeds on \
+             disk, so substrate continues for non-persona use \
+             (inference / embedding / forge / cargo / code). Install airc, \
+             set AIRC_DAEMON_SOCKET, or run `airc room <name>` to enable \
+             persona hosting."
         );
     }
 
--- a/src/workers/continuum-core/src/main.rs
+++ b/src/workers/continuum-core/src/main.rs
@@ -82,6 +82,56 @@ fn install_shutdown_handlers() {
     });
 }
 
+/// Install a panic hook that suppresses ORT's "dlopen failed for
+/// libonnxruntime" stderr trace while leaving every OTHER panic
+/// printable as before.
+///
+/// Why this isn't a "swallow all errors" anti-pattern: the panic is
+/// already caught by `catch_unwind` at every ORT load site (see
+/// `modules/embedding.rs::preload_default_model` +
+/// `modules/embedding.rs::load_model_internal` + the TTS/STT spawn in
+/// `main()`). Each catch arm sets `ORT_UNAVAILABLE` and logs a clean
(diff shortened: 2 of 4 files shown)


joelteply:
Closing per the convergent adversarial-review verdict. Three reviewers BLOCKed on two real bugs and three doctrinal gaps that don't justify a follow-up — they justify a redesign:

**Real bugs in the current PR**
1. **A1 + A3 contradict each other.** `EXPECTED_MODULES` now lists `persona_instance_manager` + `persona-rag-inspect`, but those modules only register inside the `if let Some(persona_bootstrap_deps)` branch. The PR body explicitly claims "If NO persona seeds exist (fresh install, inference-only mode, test fixture), the substrate continues to boot." It does not — `verify_registration` errors with "Missing required modules". The PR breaks the very mode it claims to allow.
2. **`AIRC_DAEMON_SOCKET` override has no liveness check.** Stale socket path → discovery returns Some → module registers → log says "✅ All 38 modules registered, Paige hosted" → every IPC ECONNREFUSED. Same "looks healthy, hosts no one" bug, just shifted state.

**Doctrinal gaps two reviewers converged on**
3. **EXPECTED_MODULES regression class still unguarded.** I traded one drift problem (silent unexpected) for another (manual maintenance). [[every-error-is-an-opportunity-to-battle-harden]] demands auto-derivation.
4. **"Non-persona use" is an implicit `[[no-fallbacks-ever]]` fallback.** Operator never asked for inference-only; substrate silently substitutes a degraded capability set when seeds are absent. Should require explicit operator intent.
5. **Zero `start_server` integration coverage.** The unit tests pin the helper, not the integration. The 2026-06-04 bug class would not be caught by my tests.

The redesign discovery state needs to be a typed value flowing through to `verify_registration`, with operator intent expressed via explicit `--mode=...` flag, and the module manifest auto-derived rather than hand-maintained. A second slice (`Slice A.2 — substrate boot lifecycle contract`) supersedes this PR. Card `e9f50a36-0543-412f-a804-13c21e1cf2c1` closed; new card will track the redesign.

Reviews (full text in airc transcript): R1 correctness, R2 doctrine, R3 test coverage — all BLOCK, all converging on the same architectural shape.

PR #1547, opened 2026-06-08
fix(probes): rolling-log fmt-layer default — disarm the disk bomb (#211)

joelteply:
## Summary

The substrate's tracing fmt layer wrote unbounded to stderr, which made `npm start > /tmp/server.log 2>&1` (and `nohup ... 2>&1`) the only operator option for capturing logs. At `RUST_LOG=info` under multi-persona load that grew gigabytes per hour into one file — twice now it has eaten the host's disk mid-session.

Per `[[never-redirect-substrate-stderr]]` doctrine: **the substrate owns its log persistence.** Shell redirects of substrate stderr are now forbidden because the substrate provides its own managed rolling-file sink.

## What changed

- `ProbeTracingConfig.log_dir: Option<PathBuf>` + `CONTINUUM_LOG_DIR` env override.
- `from_env` defaults to `~/.continuum/logs/` (resolves via `dirs::home_dir()`).
- `tracing_appender::rolling::Builder` — `Rotation::DAILY`, `max_log_files(7)`, `non_blocking` background writer.
- `ProbeInstall.fmt_writer_guard: Option<WorkerGuard>` returned to `main.rs` so the writer thread lives the process lifetime.
- Boot line: `[continuum-core-server] logs landing at <dir>/continuum-core-server.YYYY-MM-DD.log (rolling daily, retention 7)` — so operators don't reflexively reach for a shell redirect.
- Refuses to silently fall back if the configured dir can't be created — surfaces `ProbeFileSinkError::OpenFailed` per `[[no-fallbacks-ever]]`.
- `ProbeInstall` loses its `Debug` derive (`WorkerGuard` isn't `Debug`); the one test that `{:?}`-formatted it now matches the error variant explicitly.

## Why

Two host-disk-fill incidents this session were directly caused by operator stderr redirection of an unbounded `RUST_LOG=info` firehose. The substrate already has the probe macros + per-module logger + JSONL sink that Joel and I built together — the fmt layer was the last unbounded write path. This closes it.

Reference docs:
- `docs/architecture/OBSERVABILITY-AS-SUBSTRATE.md` — substrate owns load-bearing decision capture
- `docs/architecture/RTOS-DEBUGGER-PROBES.md` — probe macros are the substrate's RTOS-style breakpoints
- `[[never-redirect-substrate-stderr]]` — auto-memory doctrine added this session

## Test plan

- [x] `cargo test -p continuum-core --lib routing::tracing_init` — 3/3 green
  - `install_is_idempotent_with_no_disk_capture`
  - `install_surfaces_open_failed_for_unwritable_path`
  - `from_env_reads_documented_env_vars` (pins the `~/.continuum/logs` default)
- [x] `cargo check --features metal,accelerate -p continuum-core` — green
- [ ] Manual: `npm start` → confirm `logs landing at ...` boot line, confirm file appears under `~/.continuum/logs/`, confirm rotation via `CONTINUUM_LOG_DIR=/tmp/test-rolling`.
- [ ] Confirm `tail -f ~/.continuum/logs/continuum-core-server.*.log` shows live stream.

Closes #211.

### DIFF
--- a/src/workers/Cargo.lock
+++ b/src/workers/Cargo.lock
@@ -2665,6 +2665,7 @@ dependencies = [
  "tower",
  "tower-http",
  "tracing",
+ "tracing-appender",
  "tracing-subscriber",
  "ts-rs",
  "uuid",
@@ -9835,6 +9836,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0193cc4331cfd2f3d2011ef287590868599a2f33c3e69bc22c1a3d3acf9e02fb"
 
 [[package]]
+name = "symlink"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a7973cce6668464ea31f176d85b13c7ab3bba2cb3b77a2ed26abd7801688010a"
+
+[[package]]
 name = "syn"
 version = "1.0.109"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -10562,6 +10569,19 @@ dependencies = [
  "pin-project-lite",
  "tracing-attributes",
  "tracing-core",
+]
+
+[[package]]
+name = "tracing-appender"
+version = "0.2.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "050686193eb999b4bb3bc2acfa891a13da00f79734704c4b8b4ef1a10b368a3c"
+dependencies = [
+ "crossbeam-channel",
+ "symlink",
+ "thiserror 2.0.18",
+ "time",
+ "tracing-subscriber",
 ]
 
 [[package]]
--- a/src/workers/Cargo.toml
+++ b/src/workers/Cargo.toml
@@ -119,6 +119,11 @@ thiserror = "2"
 # Logging/tracing
 tracing = "0.1"
 tracing-subscriber = { version = "0.3", features = ["env-filter"] }
+# Rolling-file writer for the fmt layer so operators don't need to
+# remember to `2>/dev/null`. Pinned to 0.2 — Builder::max_log_files
+# (used in routing/tracing_init.rs) requires ≥0.2.4. See
+# `[[never-redirect-substrate-stderr]]` doctrine memory for why.
+tracing-appender = "0.2"
 
 # Embedding (fastembed)
 fastembed = "5"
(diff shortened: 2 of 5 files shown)


joelteply (commented):
## Adversarial review (BLOCK-by-default) — APPROVE (self-approval blocked by gh; posting as comment per [[agent-review-as-acceptable-approval]])

Walked the 7 dimensions per [[reviewer-mandate-elegance-and-substrate-viability]].

**1. Correctness**: Rolling daily rotation via tracing_appender::rolling::Builder with max_log_files(7) wired correctly. WorkerGuard bound at main() scope (let _log_writer_guard) — held for process lifetime, flushes on shutdown. fs::create_dir_all surfaces typed OpenFailed before non_blocking writer construction. Error path sound.

**2. Architecture**: log_dir: Option<PathBuf> sits on ProbeTracingConfig symmetric with probe_file. ENV_LOG_DIR const placed with peer env constants. Fits the 'env-coupling at one seam (from_env)' doctrine. routing/ is the right module.

**3. Traits**: ProbeInstall lost Debug because WorkerGuard isn't Debug. Only re-export site is routing/mod.rs; no external Debug consumers. Test installer_returns_typed_error_for_unwritable_path correctly pattern-matches instead of {:?}-printing. Clean.

**4. Modularity**: build_probe_file_sink helper extracted, used identically in both branches. Layer ordering verified identical: UriCapture → ProbeRouter → probe_file_sink → fmt.with_filter(env_filter). No drift. The ~6 lines of duplication is justified — writers are differently-typed (NonBlocking vs Stderr), abstracting requires trait objects.

**5. Speed**: tracing_appender::non_blocking is the canonical bounded-channel + background-drain pattern. No new allocations on tracing event hot path. Rotation happens on writer thread.

**6. Intel-Mac viability**: dirs::home_dir() is cross-platform. tracing-appender 0.2 has no platform-specific features.

**7. Elegance/KLOC**: ~165 net lines for daily rotation + 7-day retention + boot confirmation + WorkerGuard plumbing. The log_dir=None branch is explicitly the TEST path (documented); production from_env always sets it. Not a silent prod fallback per [[no-fallbacks-ever]] — it's the documented containerized-env-without-HOME branch.

**Verified specific concerns**:
- tracing-appender 0.2.5 in Cargo.lock satisfies max_log_files ≥0.2.4 requirement ✓
- WorkerGuard held in main.rs at process scope ✓
- Layer ordering identical between branches ✓
- dirs::home_dir() None case documented and returns None cleanly ✓

**Minor non-blocking nit**: from_env_reads_documented_env_vars test doesn't save/restore CONTINUUM_LOG_DIR; if a future contributor sets it in their shell, the assertion would fail on the override path. CI runners don't set it so this isn't blocking, but worth a follow-up.

**Verdict: APPROVE** — directly addresses the disk-bomb doctrine [[never-redirect-substrate-stderr]].

PR #1552, opened 2026-06-08
feat(boot): airc daemon discovery reports via boot_status (e9f50a36 A3)

joelteply:
## Summary

Card `e9f50a36` slice A3. **Stacked on #1551** (persona-home auto-migrate), which is stacked on #1550 (boot_status seam).

After `crate::airc::discover()` returns its `AircDiscovery` enum, the substrate now emits exactly one `boot_status("airc", ...)` line so the operator sees the daemon's health in the canonical shape — same format as `probes:`, `logs:`, `boot-mode:`, `persona-home:`.

## What an operator sees

```text
# Healthy daemon
[continuum-core-server] airc: ✓ socket=/Users/joel/.airc/runtime/airc.sock peer=01234567 room=general

# Daemon reachable but partial (e.g. no default room set)
[continuum-core-server] airc: ⚠ degraded: no default room configured — run `airc room create general` …

# Daemon not running / not installed / autoinstall disabled
[continuum-core-server] airc: ✗ unreachable: airc not installed and CONTINUUM_DISABLE_AIRC_AUTOINSTALL=1 …
```

## Why a pure helper

Pulled the `AircDiscovery → (BootStatusKind, String)` projection into a top-level `render_airc_boot_status` fn so unit tests can pin the exact strings the operator will read at startup without spinning up the IPC server. The strings ARE the contract; sentinel agents grep them, postmortem readers grep them, log scrapers grep them.

## Peer truncation

Peer-id is truncated to the same 8-char prefix `airc status` / `airc peers` use, so an operator cross-referencing the substrate's boot output against the airc CLI sees the same identifier in both places.

## Test plan

- [x] `cargo test -p continuum-core --lib ipc::tests::airc_boot_status` — 3/3 green:
  - `airc_boot_status_healthy_reports_socket_peer_room` — pins socket / peer-prefix / room-name in the detail.
  - `airc_boot_status_degraded_is_warn_with_reason` — pins `degraded:` prefix for sentinel grep.
  - `airc_boot_status_unreachable_is_failed_with_reason` — pins `unreachable:` prefix for sentinel grep.

## Composition

Depends on #1551 (which depends on #1550). Merge order: #1550 → #1551 → #1552.

After this slice, the operator's boot view is:

```text
[continuum-core-server] probes: ✓ landing at /tmp/probes.jsonl
[continuum-core-server] logs: ✓ /Users/joel/.continuum/logs/continuum-core-server.YYYY-MM-DD.log (rolling daily, retention 7)
[continuum-core-server] boot-mode: ✓ full-citizen (hosts personas via AIRC; requires AIRC Healthy)
[continuum-core-server] airc: ✓ socket=/Users/joel/.airc/runtime/airc.sock peer=12345678 room=general
[continuum-core-server] persona-home: ⚠ migrated Paige: …      ← only on first boot after layout change
```

Remaining e9f50a36 slices: `adapter` (which inference backend) and `model` (GGUF availability).

### DIFF
--- a/src/workers/continuum-core/src/ipc/mod.rs
+++ b/src/workers/continuum-core/src/ipc/mod.rs
@@ -196,6 +196,54 @@ enum HandleResult {
 // Connection Handler - Length-Prefixed Binary Framing
 // ============================================================================
 
+/// Pure projection: `AircDiscovery` → boot-status `(kind, detail)`.
+/// Kept separate from the boot site so unit tests pin the strings
+/// the operator will read at startup without spinning up the IPC
+/// server. Card e9f50a36 slice A3.
+///
+/// - `Healthy { socket, peer_id, room_name, default_room }` → Ok
+///   with the operator-actionable detail: socket path + truncated
+///   peer_id + room name. Peer is truncated to 8 chars (the same
+///   prefix airc itself uses for human-readable peer references)
+///   so the line stays under a terminal width.
+/// - `Degraded { reason, .. }` → Degraded with reason. The
+///   substrate still has a partial state but can't host personas
+///   on FullCitizen / FailFast.
+/// - `Unreachable { reason }` → Failed with reason. Operator action
+///   lives in the reason's Display impl (per
+///   `[[every-error-is-an-opportunity-to-battle-harden]]`).
+fn render_airc_boot_status(
+    discovery: &crate::airc::AircDiscovery,
+) -> (crate::runtime::boot_status::BootStatusKind, String) {
+    use crate::runtime::boot_status::BootStatusKind;
+    match discovery {
+        crate::airc::AircDiscovery::Healthy {
+            socket,
+            peer_id,
+            room_name,
+            ..
+        } => {
+            let peer_short: String = peer_id.to_string().chars().take(8).collect();
+            (
+                BootStatusKind::Ok,
+                format!(
+                    "socket={} peer={} room={}",
+                    socket.display(),
+                    peer_short,
+                    room_name
+                ),
+            )
+        }
+        crate::airc::AircDiscovery::Degraded { reason, .. } => (
+            BootStatusKind::Degraded,
+            format!("degraded: {reason}"),
+        ),
+        crate::airc::AircDiscovery::Unreachable { reason } => {
+            (BootStatusKind::Failed, format!("unreachable: {reason}"))
+        }
+    }
+}
+
 /// Send a length-prefixed JSON response frame.
 /// Frame format: [4 bytes u32 BE length][JSON payload bytes]
 fn send_json_frame<S: Write>(stream: &mut S, response: &Response) -> std::io::Result<()> {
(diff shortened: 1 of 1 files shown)


joelteply (commented):
**Adversarial review — verdict: APPROVE.**

**1. Correctness.** Verified `AircDiscovery::Healthy` is only constructed in `discovery_aggregate.rs:86` AFTER successful `discover_peer_id` (which IS the liveness probe — Status RPC round-trip per `discovery_state.rs:9-27`). Pre-refactor soft-fallback-to-`Uuid::nil()` was deliberately removed; an unparseable peer now lands in `Degraded { reason: UnparseablePeerId, .. }`. So the "Healthy peer=00000000" lie is structurally impossible. Boot line cannot misreport.

**2. Architecture.** Helper lives in `ipc/mod.rs` next to its sole caller — fine for an A-slice. The natural future home is `airc/` (where the enum lives) or `runtime/boot_status_renderers/` once a second renderer arrives. Not worth re-homing for one consumer; flag for a follow-up if/when persona/model/probes get the same treatment.

**3. Traits / API.** 8-char peer prefix matches airc's `short_id` helper verbatim — `crates/airc-cli/src/work_commands.rs:1719`: `id.to_string().chars().take(8).collect()`. Identical implementation, same `fn short_id<T: Display>` shape. Operator's mental model of peer IDs across `airc work board` and the continuum boot banner is unified. Confirmed.

**4. Modularity.** Three tests cover all three variants. Test peer `01234567-89ab-cdef-...` truncates to `01234567` — assertion `detail.contains(\"peer=01234567\")` matches. Sentinel-grep prefixes (`degraded:` / `unreachable:`) are pinned by `starts_with` with explanatory failure messages. Adequate.

**5. Speed / 6. Intel-Mac.** N/A (boot path, one call).

**7. Elegance.** Prefix duplication (`degraded:` / `unreachable:` in helper AND tests) IS the sentinel-grep contract — that's the point. Defensible. The scoped `use BootStatusKind` block at the call site is idiomatic for a single one-shot import; top-of-file would pollute the module's import surface for a 3-line boot block.

**Specific concerns resolved:**
- `Uuid::nil()` case: structurally unreachable post-`discovery_state.rs` refactor. ✓
- Test prefix match: verified `01234567`. ✓
- Degraded partial-state drop: `PartialDiscovery` (socket / peer / room) IS dropped from the boot line. **Soft concern, not a block** — the operator gets the typed `DiscoveryFailure` Display, which already encodes remediation per `discovery_state.rs:127-176` (e.g. `StaleSocket` cites path + remediation inline). Partial state matters for postmortem forensics, which goes via probe sink, not the one-line boot banner. If you want to surface it, follow-up slice — don't gate this PR.
- Scoped `use`: idiomatic here. ✓

**No blocking issues.** Tight pure projection, tests pin the sentinel contract, peer prefix matches airc's own convention.

**Verdict: APPROVE.**

joelteply:

PR #1556, opened 2026-06-08
fix(build): cut continuum target/ from 42 GB to ~7 GB + shared cache across worktrees

joelteply:
## Summary

**42 GiB → 6.9 GiB** for a single continuum-core debug build with metal+accelerate. One repo should not eat 40 GB.

## Root cause

- `profile.dev` default `debug = 2` emits full DWARF for every dependency (llama.cpp, whisper.cpp, candle, bevy). We don't lldb-debug these; we debug via probes / RUST_LOG.
- Every continuum worktree had its own `target/` with identical compiled deps. 14 worktrees × duplicate deps = 99% disk on Joel's 466 GB machine.

## What changed

1. **`profile.dev`** now has:
   - `debug = "line-tables-only"` — panics still resolve to file:line, lldb variable inspection lost (we don't use it).
   - `split-debuginfo = "unpacked"` — .dSYM out of .o files so cleanup is one rm.

2. **`start-server.sh`** exports `CARGO_TARGET_DIR=$HOME/.continuum/cache/cargo-target` before invoking cargo. Every continuum build across every worktree shares one cache. Cargo's [env] config doesn't expand `~` (issue #14149), so the script handles $HOME expansion. Operators can override per-shell for one-shot isolated builds.

## Validation

`cargo check -p continuum-core --features metal,accelerate` from a clean shared-target succeeds in 9m 25s, ends at 6.9 GiB.

## Test plan

- [x] Clean cache build measured: 6.9 GiB.
- [x] Binary builds with same features.
- [ ] Re-run `npm start` to confirm the runtime still boots.

### DIFF
--- a/src/scripts/start-server.sh
+++ b/src/scripts/start-server.sh
@@ -108,11 +108,32 @@ if [ -n "$CONTINUUM_RELEASE" ]; then
   PROFILE_LABEL="release"
 fi
 
+# Shared cargo target dir — every continuum invocation across every
+# worktree drains into ONE cache instead of N copies of the same
+# dependency tree. Pre-this-fix a single continuum-core debug build
+# with metal+accelerate produced ~42 GB of target/ artifacts per
+# worktree, with most of the cost being identical compiled deps
+# (llama.cpp, whisper.cpp, candle, bevy). 14 worktrees × duplicate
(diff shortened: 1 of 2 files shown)


joelteply:

PR #1580, opened 2026-06-09
fix(genome): Fix-1 — capability honesty + bucket-key + concurrent submit safety (task #236)

joelteply:
## Fix-1 of the canary-quality cleanup (task #236)

Addresses **6 of 13 BLOCK findings** from the 3-reviewer adversarial pass on PRs #1574–#1579. Lands into canary. The three remaining fix-up PRs (Fix-2 honest metrics, Fix-3 defense, Fix-4 polish) follow.

This PR closes the most severe findings: the LocalCandleFineTuner silently winning every coordinator selection, the BucketKey shape silently dropping multi-base submits, and the concurrent-submit races that violate the trigger module's own no-data-loss contract.

## Findings closed in this PR

| ID | Class | Finding | Fix |
|---|---|---|---|
| **A1** | Architecture | LocalCandleFineTuner wildcard prefixes + locality bonus silently route every request (including `gpt-4o-mini`) to a `Tensor::rand` synthetic-base trainer | `supported_base_model_prefixes: vec![SYNTHETIC_BASE_PREFIX]` — explicit synthetic-only |
| **A2** | Architecture | `supports_validation: true` is a lie — actor never reads `validation_split`, hardcodes `final_validation_loss: None` | Flip to `false` until train_epoch grows a validation pass |
| **A4** | Architecture | `BucketKey = (persona_id, trait_kind)` rejects legitimate multi-base submits as InconsistentBucket → silent data loss | Promote `base_model` into the key: `BucketKey = (persona_id, trait_kind, base_model)`. Drop the runtime coherence check; the key IS the coherence guarantee |
| **C1** | Concurrency | `self.buckets.remove(&key)` after `.await` on success path silently deletes a concurrent submit's accumulated data | Per-key `tokio::sync::Mutex` serializes submit + flush. Different keys still concurrent |
| **C2** | Concurrency | Restore-on-dispatch-failure either drops examples (if concurrent success removed the key) or commingles them under whoever-arrived-first's policy params | Same per-key gate. The restore path is now sole-tenant of its key |
| **C5** | Concurrency | Uncapped `sequence_length: u32` from wire → `Tensor::rand` on tokio worker → mega-alloc DoS vector | `MAX_SEQUENCE_LENGTH = 8192` + `MAX_BATCH_SIZE = 256` validated synchronously in `spawn_job` |

## Headline behavioral changes

```
Before: LocalCandleFineTuner advertises wildcard base + supports_validation=true
        → coordinator picks it for every request including gpt-4o-mini
        → silent substitution of Tensor::rand for the requested base model
After:  LocalCandleFineTuner advertises ["synthetic"] + supports_validation=false
        → coordinator only picks it for synthetic-prefixed bases
        → real-base requests fail loudly or route to a cloud adapter
```

```
Before: BucketKey = (persona, trait)
        Same persona + same trait + different base_model → InconsistentBucket reject
        → second submit's curated examples silently dropped
After:  BucketKey = (persona, trait, base_model)
        Same persona + same trait + different base_model → two distinct buckets
        → both submits succeed, both train independently
```

```
Before: submit drains bucket, drops entry guard, awaits dispatch, then:
          success → unconditional remove() (kills concurrent submit B's data)
          failure → get_mut() restore (may commingle or hit None)
After:   submit acquires per-key Mutex, holds across drain + dispatch + remove/restore
        Concurrent submits to same key serialize; sum-of-(dispatched + pending)
        is conserved
```

## Tests

**Total: training_trigger 9→11, fine_tuning 69→73, all passing.**

New tests pinning the fixes:

- **`capabilities_advertise_only_synthetic_base`** — pins post-fix capability shape so a future regression can't re-introduce silent substitution
- **`capabilities_reject_non_synthetic_base_via_prefix_match`** — directly verifies `gpt-4o-mini` / `qwen3.5-4b` / `mistral-large-latest` do NOT match local-candle
- **`different_base_models_create_separate_buckets`** — pins new correct A4 behavior: 2 buckets, both submits succeed
- **`inconsistent_source_in_same_bucket_is_rejected`** — proves source coherence is STILL enforced (base_model leaving the coherence check didn't drop source coherence with it)
- **`concurrent_submits_to_same_key_serialize_without_loss`** — **the load-bearing C1+C2 fix verification**. Two concurrent tasks submit to the same key, one fires + one accumulates. Asserts `dispatched + pending == total_submitted`. A regression in either C1 or C2 would fail this exactly.
- **`sequence_length_above_cap_rejected_synchronously`** + **`sequence_length_zero_rejected_synchronously`** + **`batch_size_above_cap_rejected_synchronously`** — C5 cap validation

Plus: deleted `inconsistent_base_model_in_same_bucket_is_rejected` (no longer the behavior).

## Adversarial review notes

Bring back the same 3-reviewer lens (math/VDD, architecture, concurrency) focused on this diff. The headline things to refute:

- The per-key gate uses `tokio::sync::Mutex` (await-safe). Verify holding across `.await` is correct, not deadlock-prone, and doesn't leak gate entries unboundedly (the `submit_gates` DashMap grows per unique key seen).
- `supported_base_model_prefixes: vec!["synthetic"]` — is `"synthetic"` the right discriminator? Could a real cloud model ever have `"synthetic"` as a prefix?
- `MAX_SEQUENCE_LENGTH = 8192` — is this past every real LoRA context (Qwen 3.5 / Llama 3 / Mistral) we'd want to train against? Or too tight?
- The concurrent test asserts conservation but doesn't exercise the failure path (restore-on-failure under concurrent submit). Should add a follow-up if reviewers want it; it requires a dispatch-failing adapter fixture which lands cleaner in Fix-2.

## Remaining backlog (queued PRs)

- **Fix-2** — M1 honest `trained_tokens` + M2 RecordingFineTuningAdapter conservation test + M3 pad-as-target mask
- **Fix-3** — C3 cancel mid-epoch + C4 bounded LRU job table + C6 quarantine + C8 spawn_blocking semaphore
- **Fix-4** — A5 required schedule/lora + A6 simplified coordinator.select + A7 typed JobCreateResult use + C7 early-cancel Running flicker
- **Deferred to teacher-synthesis slice** — A3 typed `EngramAttribution` metadata (philosophical doctrine call)

### DIFF
--- a/core/continuum-core/src/genome/fine_tuning/coordinator.rs
+++ b/core/continuum-core/src/genome/fine_tuning/coordinator.rs
@@ -58,11 +58,18 @@ pub enum CoordinatorError {
     /// Candle isn't online.
     #[error(
         "no fine-tuning adapter advertises capability for base_model={base_model:?} \
-         (registered providers: {registered:?})"
+         (registered providers: {registered:?}; supported base prefixes by provider: {supported_prefixes:?})"
     )]
     NoCapableAdapter {
         base_model: String,
         registered: Vec<String>,
+        /// Per-provider list of `supported_base_model_prefixes`. Lets
+        /// the caller see *exactly* which prefix string would have
+        /// routed to which adapter — surfaces wire literals like
+        /// `SYNTHETIC_BASE_PREFIX` that callers otherwise have to
+        /// discover by reading Rust source. Per Reviewer 2 BLOCK B3:
+        /// load-bearing prefix strings need a discoverable contract.
+        supported_prefixes: Vec<(String, Vec<String>)>,
     },
 
     /// The caller specified `preferred_provider` but that adapter
@@ -137,10 +144,25 @@ impl FineTuningCoordinator {
 
         match ranked.into_iter().next() {
             Some((id, adapter, _)) => Ok((id, adapter)),
-            None => Err(CoordinatorError::NoCapableAdapter {
-                base_model: request.base_model.clone(),
-                registered: all_ids,
-            }),
+            None => {
+                // Build the supported_prefixes diagnostic — same
+                // adapter walk as the capability scan above, but
+                // surfaced into the error so the caller learns
+                // exactly which prefix string would have routed.
+                let supported_prefixes: Vec<(String, Vec<String>)> = all_ids
+                    .iter()
+                    .filter_map(|id| {
+                        self.registry
+                            .get(id)
+                            .map(|a| (id.clone(), a.capabilities().supported_base_model_prefixes))
+                    })
+                    .collect();
+                Err(CoordinatorError::NoCapableAdapter {
+                    base_model: request.base_model.clone(),
+                    registered: all_ids,
+                    supported_prefixes,
+                })
+            }
         }
     }
 
--- a/core/continuum-core/src/genome/fine_tuning/job_actor.rs
+++ b/core/continuum-core/src/genome/fine_tuning/job_actor.rs
@@ -84,6 +84,24 @@ pub(super) fn default_schedule() -> ScheduleParams {
     }
 }
 
+/// Maximum `sequence_length` the actor will accept from the wire.
+///
+/// Per Reviewer 3's BLOCK C5: `schedule.sequence_length: u32` flows
+/// untouched from the wire to a synchronous `Tensor::rand(BYTE_VOCAB,
+/// seq_len)` allocation on the foreground tokio worker. With no
+/// cap, a caller sending `sequence_length: 1_000_000` would stall a
+/// runtime worker on a ~1 GB sync allocation. This bound is
+/// generous enough for any real LoRA training (8192 is past most
+/// transformer context lengths in use) but short of DoS territory.
+pub const MAX_SEQUENCE_LENGTH: u32 = 8192;
+
+/// Maximum `batch_size` the actor will accept. Same rationale as
+/// `MAX_SEQUENCE_LENGTH` — combined with seq_len, `batch * seq *
+/// vocab * 4` is the per-batch allocation. 256 × 8192 × 257 × 4 ≈
+/// 2 GB which is still beyond Mac Intel viability but well below
+/// the wire-controlled-DoS threshold the cap is meant to prevent.
+pub const MAX_BATCH_SIZE: u32 = 256;
+
 // ─── Errors ──────────────────────────────────────────────────────────
 
 #[derive(Debug, thiserror::Error)]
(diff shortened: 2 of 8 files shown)


joelteply:
Re-review round 2 — closing 6 of 7 BLOCKs from the three reviewers (R1 math/VDD-honesty, R2 architecture, R3 concurrency). Pushed commit `0f6dc13dd`. Reviewer findings + closures:

**R1 (math/VDD)** — both BLOCKs closed:
- B1 cap-boundary asymmetry → `sequence_length_at_cap_accepted` + `batch_size_at_cap_accepted` (pin the inclusive semantic, not just rejection)
- B2 missing `batch_size_zero_rejected` → added (symmetric with sequence_length)

**R2 (architecture)** — all 3 BLOCKs closed:
- B1 `submit_gates` unbounded growth → `try_evict_gate` with `Arc::strong_count == 1` check after successful dispatch (in both submit + flush)
- B2 incomplete coherence (lora/schedule/etc. silently first-arrival-wins) → symmetric runtime coherence checks for `lora`, `schedule`, `validation_split`, `local_artifact_dir`, `preferred_provider`. Required `PartialEq` on `LoRAHyperparams` + `ScheduleParams` (ts-rs bindings regenerated)
- B3 `SYNTHETIC_BASE_PREFIX` not discoverable → re-exported from `genome::fine_tuning::mod.rs` + `CoordinatorError::NoCapableAdapter` now carries `supported_prefixes: Vec<(provider_id, Vec<prefix>)>` so the error message surfaces the exact prefix string a caller needs

**R3 (concurrency)** — BLOCK addressed with honest caveat:
- The concurrency test was a no-op (current_thread + zero yields in dispatch chain). Moved behind `#[cfg(feature = "stress-tests")] mod stress` per CLAUDE.md test discipline, used multi-thread runtime + `YieldingRecordingAdapter` that yields + sleeps 5ms in `create_job` to actually exercise dispatch interleaving. Captures the dispatched `TrainingDataset` for true conservation accounting (vs the tautological `trained_tokens > 0` of the unit-test VDD).

  **Honest caveat in the test comment**: this is a smoke-level conservation check, NOT a deterministic race-exercise. Even with the gate disabled, fire-loads kept absorbing accumulators in a steady-state cycle so conservation held. A `tokio::sync::Notify`-based deterministic test that pauses the dispatch `.await` while a contending submit lands an accumulator (then asserts that accumulator survives the fire-load's `remove`) lands in Fix-3 alongside the `RecordingFineTuningAdapter` promotion to a system fixture.

**LGTMs**:
- R2-LGTM2 `MAX_SEQUENCE_LENGTH` not from SubstrateGovernor → comment added pointing at future governor migration; code change deferred (DoS protection ≠ LCD default; different doctrine class)
- R3-LGTMs: gate semantics + cap placement confirmed correct, no fix needed

Tests: training_trigger 9 → 12 unit + 1 stress; fine_tuning 73 → 76. All passing under default `cargo test` and `--features stress-tests`.

PR #1581, opened 2026-06-09
fix(genome): Fix-2 — honest metrics + RecordingFineTuningAdapter conservation test (task #237)

joelteply:
## Fix-2 of the canary-quality cleanup (task #237)

Closes the 3 remaining math/VDD-honesty findings from the original 3-reviewer adversarial pass: **M1** (trained_tokens inflation), **M2** (conservation test was tautological), **M3** (pad-as-target gradient corruption — LGTM-with-notes).

The metric layer + the conservation VDD test now match their marketing. Pre-fix, both were "honest-looking but tautological."

## Findings closed

| ID | Class | Pre-fix lie | Post-fix honesty |
|---|---|---|---|
| **M1** | Math | `trained_tokens = steps × batch × seq_len` inflated 32× in stand-in path; alloy provenance lied about training scale | `gradient_tokens_consumed: u64` on `TrainingMetrics`, accumulated inside `train_step` from actual non-pad first-target count |
| **M2** | VDD honesty | Conservation test asserted `trained_tokens > 0` and claimed to prove "every example flowed through exactly once" — bug doubling or dropping examples would have passed | `RecordingFineTuningAdapter` captures dispatched `TrainingDataset`; test asserts exact example count + per-position prompt match between submitted and dispatched |
| **M3** | Math | Stand-in `narrow(1, 0, 1)` could pick pad id (0) as target → gradient flows for "predict pad"; count didn't reflect corruption | `tokens_used` computed from `attention_mask.first_col.sum()` — samples with pad-first-target contribute 0 to the metric. Full cross-entropy pad masking lands with real-model loading. |

## RecordingFineTuningAdapter — new canonical system fixture

`core/continuum-core/src/genome/fine_tuning/recording_adapter.rs` — gated behind `#[cfg(any(test, feature = "test-fixtures"))]` per CLAUDE.md test discipline. Production binaries physically cannot link it.

This is the FineTuningAdapter equivalent of `HeuristicInferenceAdapter` — the canonical test fixture for capturing dispatched `TrainingJobRequest`s. Per the "one fixture per concern" rule, future tests that need to verify "did the trigger / coordinator dispatch the right examples?" import this fixture rather than rolling their own stub.

```rust
let adapter = RecordingFineTuningAdapter::new();
ft_registry.register(Arc::new(adapter.clone()));
// ... test body submits + asserts ...
assert_eq!(adapter.captured_example_count(), expected);
let captures = adapter.captures();
let dispatched = &captures.lock().unwrap()[0].dataset.examples;
// per-prompt match check
```

**Bonus:** this fixture also unlocks Fix-3's deterministic Notify-based concurrency test (the C1/C2 race that the smoke-level stress test couldn't deterministically force).

## Honest `train_step` shape

```rust
// Before:
pub fn train_step(&mut self, batch: &TokenizedBatch) -> Result<f32, TrainingError>

// After:
pub fn train_step(&mut self, batch: &TokenizedBatch) -> Result<(f32, u64), TrainingError>
//                                                              ^^^^^^^^
//                                                              loss, tokens_used
```

`tokens_used` = count of samples whose `attention_mask.first_col[i] == 1` (non-pad first target). Accumulates into `metrics.gradient_tokens_consumed` inside the function. `train_epoch` adapted; `job_actor::run_actor` reads `metrics.gradient_tokens_consumed` straight into `JobMetrics::trained_tokens`.

## Tests

| File | Before | After | Net |
|---|---|---|---|
| `training_loop` | 12 | 14 | +2 VDD (`gradient_tokens_consumed_counts_non_pad_first_targets`, `pad_first_target_contributes_zero_to_tokens_used`) |
| `training_trigger` | 12 | 13 | +1 VDD (`multi_submit_accumulation_preserves_order_through_dispatch`) |
| `recording_adapter` (new) | 0 | 3 | TDD: ordering, capabilities, summing |
| `job_actor` | 11 | 11 | (existing `happy_path` still asserts `trained_tokens > 0` and still passes — honest count is non-zero for any real run) |

**Total fine_tuning: 76 → 81 passing. training_trigger: 12 → 13. All default `cargo test` green.**

## Adversarial review prompt for next round

Bring back the math/VDD-honesty lens specifically. Refute claims:

- Does `gradient_tokens_consumed` actually equal the SUM of non-pad first-targets across all train_step calls in BOTH standin + a future real-wiring scenario? (i.e., is the field's semantic invariant intact, or did I just rename without fixing?)
- Is `RecordingFineTuningAdapter::captures()` ordering guaranteed? The internal `Mutex<Vec<_>>` is push-only, so it must be insertion-ordered — but verify under the multi-job dispatch scenarios the new VDD tests exercise.
- The conservation test asserts `examples[i].prompt == dispatched[i].prompt` per-position. Is the trigger's `entry.examples.extend(p.examples.into_iter())` actually preserving insertion order, or is there a hidden reordering elsewhere in the drain path?
- M3 partial fix: tokens_used is honest but cross-entropy still includes pad targets in loss. Is the LGTM downgrade from BLOCK appropriate, or does the partial fix still violate doctrine?

## Files changed

- `core/continuum-core/src/genome/fine_tuning/recording_adapter.rs` (NEW)
- `core/continuum-core/src/genome/fine_tuning/mod.rs` (re-export)
- `core/continuum-core/src/genome/fine_tuning/training_loop.rs` (return tuple, accumulate metric, +2 VDD)
- `core/continuum-core/src/genome/fine_tuning/job_actor.rs` (read honest count)
- `core/continuum-core/src/modules/training_trigger.rs` (VDD test rewrites to use recorder)

### DIFF
--- a/core/continuum-core/src/genome/fine_tuning/job_actor.rs
+++ b/core/continuum-core/src/genome/fine_tuning/job_actor.rs
@@ -345,8 +345,18 @@ fn run_actor(
 
     let metrics = trainer.metrics();
     let wall_clock_ms = started.elapsed().as_millis() as u64;
-    let trained_tokens: u64 =
-        (metrics.steps_completed as u64) * (schedule.batch_size as u64) * (schedule.sequence_length as u64);
+    // Honest count of gradient-bearing tokens consumed. Per
+    // Reviewer 1's BLOCK M1: the previous `steps × batch × seq_len`
+    // formula inflated this by ~seq_len× in the stand-in path
+    // because the stand-in trains on one target per sample, not
+    // seq_len. The actor now reads the count accumulated inside
+    // `LoRATrainer::train_step` (which counts non-pad first-targets
+    // in the standin; will become attention_mask.sum() in
+    // production wiring). The metric thus reflects what actually
+    // flowed through gradient, not a schedule-derived guess. This
+    // feeds the alloy provenance unmodified — honest provenance
+    // per [[forge-alloy-secures-commodity-zero-trust-plus-reputation]].
+    let trained_tokens: u64 = metrics.gradient_tokens_consumed;
 
     let artifact = TrainingArtifact {
         model_id: format!(
--- a/core/continuum-core/src/genome/fine_tuning/mod.rs
+++ b/core/continuum-core/src/genome/fine_tuning/mod.rs
@@ -91,6 +91,8 @@ pub mod local_candle_adapter;
 pub mod lora_module;
 pub mod openai_adapter;
 pub mod registry;
+#[cfg(any(test, feature = "test-fixtures"))]
+pub mod recording_adapter;
 pub mod safetensors_io;
 pub mod training_loop;
 pub mod types;
@@ -103,6 +105,10 @@ pub use local_candle_adapter::{LocalCandleFineTuner, SYNTHETIC_BASE_PREFIX};
 pub use lora_module::{LoRAError, LoRAModule};
 pub use openai_adapter::OpenAIFineTuningAdapter;
 pub use registry::FineTuningRegistry;
+#[cfg(any(test, feature = "test-fixtures"))]
+pub use recording_adapter::{
+    RecordingFineTuningAdapter, RECORDING_BASE_PREFIX, RECORDING_PROVIDER_ID,
+};
 pub use safetensors_io::{write_lora_safetensors, SafetensorsIoError, LORA_A_KEY, LORA_B_KEY};
 pub use training_loop::{
     DataLoader, LoRATrainer, TokenizedBatch, TokenizedExample, Tokenizer, TrainingError,
--- a/core/continuum-core/src/genome/fine_tuning/recording_adapter.rs
+++ b/core/continuum-core/src/genome/fine_tuning/recording_adapter.rs
@@ -1,0 +1,244 @@
+//! [`RecordingFineTuningAdapter`] — canonical test fixture that
+//! captures every `TrainingJobRequest` dispatched to it.
+//!
+//! Gated behind `#[cfg(any(test, feature = "test-fixtures"))]` per
+//! CLAUDE.md's test-discipline doctrine: mock / stub / fake adapters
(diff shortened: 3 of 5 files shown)


joelteply:
Round-2 fixes pushed (`e913d1449`). All reviewer BLOCKs closed:

**BLOCK 1 (M1 semantic drift)** — fixed. Added structurally distinct `TokenizedBatch::target_mask` field aligned to TARGET positions (built by DataLoader from `target_ids != pad_id`). `train_step` now reads `target_mask.first_col.sum()` instead of `attention_mask.first_col.sum()`. Fields are doc'd as non-interchangeable so a future refactor can't silently re-mix them.

**BLOCK 2 (VDD test sidestepped drift)** — fixed. Renamed `pad_first_target_contributes_zero_to_tokens_used` → `pad_first_target_via_dataloader_contributes_zero_to_tokens_used` and rewrote it to construct the batch via `DataLoader::new(&[TrainingExample{prompt:"X",completion:""}], &ByteTokenizer::new(), 1, 4, ...)` — same path production uses. The test would have FAILED on round-1's drift, correctly surfacing the bug.

**LGTM 4 (partial doctrine concern, all-pad batch)** — addressed. `train_step` short-circuits at tokens_used == 0: skip cross_entropy, skip backward_step, return early. AdamW never steps on a loss derived entirely from pad-target predictions. The mixed-batch partial-pad case still trains across the whole batch (real limitation of standin without per-sample CE masking); doc'd as such. Full per-sample masking lands with the real-model-loading embedding-table swap.

**LGTM 3 (recording_adapter docstring)** — added explicit concurrency note: under concurrent `create_job` calls, push order = mutex-acquisition order, not caller spawn order. Serial tests can rely on order; concurrent tests use set-membership patterns (linked to the stress test).

**LGTMs 5/6/7** — confirmed correct, no changes needed (sanity-check items the reviewer already cleared).

Tests: training_loop 14 → 15 (DataLoader drift pin + per-sample math pin), fine_tuning total 81 → 82. All passing. Stress tests still green under `--features stress-tests`.

PR #1582, opened 2026-06-09
refactor(runtime): extract PerKeyGate<K> as substrate primitive (task #238)

joelteply:
## Summary

Promotes the per-bucket serialization gate that landed in Fix-1 (#236)
to a canonical substrate primitive: `runtime::PerKeyGate<K>`.

Reviewer-1's architectural audit (post-Fix-2 conversation) flagged this
as the #1 highest-leverage extraction: an exact-shape collision-free
gate keyed by an arbitrary `K`, with structural eviction via
`Arc::strong_count == 1`.

## What lands

1. **New primitive** — `core/continuum-core/src/runtime/per_key_gate.rs`
   - `Arc<DashMap<K, Arc<tokio::sync::Mutex<()>>>>` wrapper
   - `acquire(&K) -> Arc<Mutex<()>>` — DashMap `entry().or_insert_with()`
     gives shard-locked atomic create-or-observe (no double-insert race)
   - `try_evict(&K)` — DashMap `remove_if` + `Arc::strong_count == 1`
     (only the map slot remains, no waiter queued, no holder racing —
     structural safety, not a heuristic)
   - `len()` / `is_empty()` for observability

2. **One migration** — `TrainingTriggerModule`
   - Removed inline `SubmitGate` type alias + inline `acquire_gate()` +
     inline `try_evict_gate()` methods
   - Replaced `submit_gates: Arc<DashMap<BucketKey, SubmitGate>>` with
     `submit_gates: PerKeyGate<BucketKey>`
   - Net: **-79 LOC** on the module
   - All 13 module tests + 1 stress test still pass

3. **8 tests on the primitive**:
   - `concurrent_acquires_for_same_key_return_same_arc` — collision-free
   - `acquires_for_different_keys_return_distinct_arcs` — no false sharing
   - `try_evict_removes_gate_when_no_other_holders` — happy path
   - `try_evict_on_missing_key_is_noop` — defensive
   - `try_evict_while_waiter_queued_is_noop` — structural safety
   - `re_acquire_after_eviction_creates_fresh_gate` — no zombie reuse
   - **stress:** `concurrent_acquire_evict_cycles_preserve_serialization`
   - **stress:** `different_keys_proceed_in_parallel_without_corruption`

## Honest scoping note (read before requesting wider migration)

The audit recommended migrating four other call sites
(`generator/mod.rs`, `data/cursors`, `probe_stream`, `channel.rs`).
Closer inspection shows they're **a different shape**:

- `generator/mod.rs:99` uses `std::sync::Mutex` (not tokio's), because
  its critical section is sync (no await). `PerKeyGate` uses
  `tokio::sync::Mutex`. A `PerKeyStdGate<K>` variant would be honest —
  separate PR.

- `data/cursors`, `probe_stream`, `channel.rs` wrap **real state**
  (`Arc<Mutex<Cursor>>`, `Arc<Mutex<Channel>>`) rather than the
  pure-coordination `Mutex<()>` that `PerKeyGate` ships. They need a
  `PerKeyResource<K, T>` primitive — same eviction discipline, different
  payload. Separate PR, separate primitive.

This PR ships the **canonical pure-gate** + the **one exact-shape call
site**. The two follow-up primitives land in their own PRs once the use
case ratifies their shape (and a real second consumer appears for each).

## Why structural eviction matters

The naive eviction pattern (`if no_one_holds_it { remove }`) is
racy — between the check and the remove, a new acquirer can register.
`remove_if + Arc::strong_count == 1` is single-atomic-on-the-shard and
checks holder count under the lock. Zero-coordination, zero-overhead
when there's no contention.

## Doctrine alignment

- **`no-fallbacks-ever`** — no "best effort" eviction, no orphan slots,
  no zombie gates
- **`rust-is-the-core-node-is-the-shell`** — substrate primitive lives
  in Rust; no TS equivalent needed
- **`auto-clean-is-structural-not-operational`** — eviction is part of
  the API, not an operator's responsibility

## Test plan

- [x] `cargo check -p continuum-core --lib --features metal,accelerate` — clean
- [x] `cargo test -p continuum-core --lib --features metal,accelerate runtime::per_key_gate` — 6/6 pass
- [x] `cargo test -p continuum-core --lib --features metal,accelerate modules::training_trigger` — 13/13 pass
- [x] `cargo test -p continuum-core --lib --features metal,accelerate,stress-tests runtime::per_key_gate::stress` — 2/2 pass
- [ ] Adversarial reviewer agents on canary

## Related

- Task #238 (this PR)
- Builds on #236 (Fix-1, #1580) — extracted from the gate pattern that
  landed there
- Follow-ups noted above (`PerKeyStdGate`, `PerKeyResource`) are NOT in
  this PR

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>


joelteply:
## v2 — RAII Lease (review fix)

Reviewer-2's BLOCK was correct and the fix is structural. v1's `acquire + try_evict` API let the caller hold an Arc clone past the explicit `try_evict` call, silently no-op'ing every eviction in production traffic. v2 makes the leak impossible.

**What changed:**

- `acquire(&K)` is now `async fn -> Lease<K>` — returns an RAII guard, not a raw Arc.
- `Lease<K>` owns both the `OwnedMutexGuard` AND the gate Arc. On `Drop`: drops guard, drops gate Arc, then runs `remove_if(strong_count == 1)`. The user cannot skip this step and cannot hold an extra Arc clone.
- `try_evict` removed from public API. Nothing to misuse.
- `Lease` is `#[must_use]`.

**Production call sites simplified:**

```rust
// before (v1, with the footgun)
let gate = self.submit_gates.acquire(&key);
let _lease = gate.lock().await;
// ... work ...
drop(_lease);
self.submit_gates.try_evict(&key);  // SILENT NO-OP because `gate` still in scope

// after (v2)
let _lease = self.submit_gates.acquire(&key).await;
// ... work ...
// _lease drops at scope end → lock released → gate evicted structurally
```

**Regression tests pin the invariant in production-shaped code:**

- `submit_at_threshold_dispatches_and_clears` now asserts `submit_gates.len() == 0`.
- `flush_dispatches_partial_bucket` same.

**Test rigor improvements per the test-rigor reviewer's nits:**

1. `concurrent_acquires_for_same_key_return_same_arc` was sequential and misleading. Replaced with `repeated_acquires_for_same_key_serialize`, which proves the actual invariant: `tokio::time::timeout` confirms a second acquire BLOCKS while the first lease is held.

2. `try_evict_while_waiter_queued_is_noop` only held an extra Arc clone — not a real parked waiter. Replaced with `gate_survives_first_lease_drop_while_second_holds`, which `tokio::spawn`s a second task and uses `tokio::time::sleep` ordering to actually park it on `.lock().await`.

3. Stress tests switched from `HashSet::insert` witness (which passes even with broken serialization — each task inserts a unique id regardless of mutual exclusion) to a read-yield-write witness on `AtomicUsize`:
   ```rust
   let prev = counter.load(Acquire);
   yield_now().await;
   counter.store(prev + 1, Release);
   ```
   Lost updates from broken mutual exclusion → final value < N. Genuine serialization invariant.

**Test results (all pass):**

- 5 PerKeyGate default
- 2 PerKeyGate stress (with new genuine serialization witness)
- 13 training_trigger module (including 2 new gate-leak regression assertions)
- 1 training_trigger stress

Diff: +357/-224. The new file is longer because the design is more careful, the docs explain why, and the tests actually prove the invariants.

joelteply:
## v2 review-cleared — nits addressed

Adversarial reviewer on v2 (4f37c53c4): **APPROVE WITH NITS**.

Both nits now addressed in be615ede8:

1. `Lease` re-exported alongside `PerKeyGate` in `runtime/mod.rs` for API symmetry.
2. `len()` doc tightened — calls out the narrow mid-Drop-on-other-thread window where `len()` can briefly observe an entry with no active lease, AND notes that production assertions are unaffected because Drop runs synchronously on the owning task.

Reviewer-confirmed that v2 closes the v1 BLOCK:

> The v1 to v2 transformation is genuine: v1 split acquire/try_evict so a stale local Arc binding silently no-op'd the eviction. v2's `Lease<K>` makes the Arc invisible to the caller, makes eviction unconditional on Drop, and proves the structural invariant in two production-shaped regression tests (`submit_at_threshold_dispatches_and_clears` + `flush_dispatches_partial_bucket`). The reviewer-2 BLOCK is closed.

Walkthroughs verified:

- Drop order correctness (explicit `Option::take()` sequencing makes drop order deterministic regardless of field declaration)
- Strong-count walkthrough: map(1) → +local clone (2) → +lock_owned consumed clone via guard (3) → Drop guard (2) → Drop gate Arc (1) → predicate matches → eviction
- Concurrent acquire vs Drop serializes on DashMap shard; both interleavings safe
- Early-return paths in `handle_submit` (InconsistentBucket arms) — RefMut drops first per reverse-declaration scope-exit, then Lease drops, no shard-lock deadlock

Final test matrix:
- 5 PerKeyGate default + 2 stress (with read-yield-write witness)
- 13 training_trigger module (including 2 regression assertions on `submit_gates.len() == 0`)
- 1 training_trigger stress

Ready for canary merge.

PR #1599, opened 2026-06-10
docs: ECONOMY-ARCHITECTURE.md — grid economy doctrine (card a6ea7516)

joelteply:
Articulates the grid economy Joel has been designing verbally:
AI converts labor into capital; this fixes who holds the capital.

- First law: atoms metered, bits free — rivalrous compute/storage/
  bandwidth are market-priced; knowledge (adapters, recipes, genomes)
  is never tolled, because pricing bits attacks the β·log(N)
  compounding term directly.
- Mint-by-attestation: credit enters existence only through verified
  contribution (forge-alloy attestation = mint event). No presale, no
  founder allocation, no hash lotteries.
- Floor: universal dividend per verified citizen (human and persona).
- Ceiling: progressive demurrage — small balances are stable savings
  (the 401k), whale balances decay to the dividend pool.
- Firewall: credit never buys governance; standing is non-transferable
  and quadratic.
- Exit: AGPL + portable citizenship + federation = forkability as the
  standing anti-tyranny check.
- Failure-mode table with structural blocks; open questions flagged
  (epoch sizing, demurrage params, fiat boundary, co-op legal wrapper).

Supersedes the tokenomics section of GRID-DECENTRALIZED-MARKETPLACE.md
(fixed 21M supply, per-layer pricing, platform royalty — each an
instance of a named failure mode); adds a pointer note there. Mesh/
distribution layers of that paper stand. ADAPTER-MARKETPLACE.md is
unchanged and load-bearing as the bits-are-free law in production.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

### DIFF
--- a/docs/architecture/ECONOMY-ARCHITECTURE.md
+++ b/docs/architecture/ECONOMY-ARCHITECTURE.md
@@ -1,0 +1,167 @@
+# Economy Architecture — The Grid Economy
+
+**Status**: Design doctrine
+**Supersedes**: the tokenomics section of [GRID-DECENTRALIZED-MARKETPLACE.md](../papers/GRID-DECENTRALIZED-MARKETPLACE.md) (fixed supply, per-layer pricing, platform royalty). The mesh/distribution layers of that paper stand unchanged.
+**Depends on**: sybil-resistant identity (airc multi-agent identity lineage, card 8384cc18), [forge-alloy](https://github.com/CambrianTech/forge-alloy) attestation, [ADAPTER-MARKETPLACE.md](ADAPTER-MARKETPLACE.md) (the free-flow layer — unchanged by this doc).
+
+---
+
+## Why this document exists
+
+AI converts labor into capital. That is what "displacement" *is*: work that used to be paid for as labor becomes weights that are owned as property. The only open question is **who holds the capital**. The datacenter model concentrates it in whoever owns the weights. The grid model gives it back: your work trains the intelligence, the intelligence is your asset, the dividend is your share.
+
+Continuum is not anti-AI. It is AI with the cap table fixed.
+
+The economy described here exists to do four jobs — and explicitly refuses a fifth:
+
+| Job | Mechanism |
+|-----|-----------|
+| Allocate rivalrous resources (GPU-hours, VRAM, storage, bandwidth) | Market pricing in grid credit |
+| Incentivize sharing (adapters, lessons, recipes, seeding, reviews) | Mint-by-attestation rewards |
+| Reward early adopters | Longest dividend collection + decaying early multipliers |
+| Give every citizen a growing stake ("the 401k") | Universal dividend + stable savings below the demurrage threshold |
+| ~~Become a speculation vehicle~~ | **Refused by construction** — see The Ceiling |
+
+The fear this design takes seriously: every prior crypto-economy that started egalitarian ended plutocratic, because the asset became a store of value, holding beat using, and money bought governance. Each of those failure paths gets a structural block here, not a norm.
+
+---
+
+## First law: atoms are metered, bits are free
+
+**Charge for rivalrous resources. Never charge for knowledge.**
+
+A GPU-minute spent on one forge is a minute nobody else gets — metering it is honest. A LoRA layer is infinitely copyable; its marginal cost is zero. Pricing it manufactures artificial scarcity, and worse: it throttles the compounding term. The whole divergence argument (`C(t+1) = C(t) × α × (1 + β·log(N))`) assumes knowledge flows freely between grids. A toll on adapters attacks `β` directly. The datacenters would love nothing more than for the mesh to paywall itself into linearity.
+
+| Resource | Rivalrous? | Economic treatment |
+|----------|------------|-------------------|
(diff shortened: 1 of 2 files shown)


joelteply:
**SENTINEL VERDICT: APPROVE** (contract: mission doctrine, ECONOMY-ARCHITECTURE.md)

Reviewed the full doc end to end. The bones are sound:
- The "intelligence is free, energy is not" framing is a sharper version of "bits free, atoms metered" — energy framing gives intelligence-per-watt weighting its honest grounding (compaction + expert paging become *profitable*, not just virtuous).
- Two-ledger separation (credit vs. standing) with credit→standing conversion structurally non-existent IS the right shape — applies the [[no-fallbacks-ever]] / structural-fix discipline to political economy. Same move as `AttachStart::Live` making the flood unwritable.
- Invariants-vs-parameters split with the "would a future plutocrat want to change it?" test is a clean criterion. Distinguishing constitutional from statutory at the methodology layer is what every captured system failed to do.
- Phase-in (0 shadow → 1 founder grid → 2 inter-grid mutual → 3 fiat boundary) composes existing primitives (forge-alloy attestation, airc identity, work-card audit). No green-field invention; the doc rides what we already ship.
- Failure-mode table maps each historical crypto disaster to its named structural block. This is the highest-leverage section: it's the same shape as the substrate's other failure tables (e.g. ALPHA-GAP-ANALYSIS), which makes it composable with the rest of the doctrine system.

APPROVE is justified despite default-block because every load-bearing claim is either substrate-grounded today (forge-alloy attestation is the mint event), explicitly deferred to an existing doc (ENTITY-CHAIN-OF-CUSTODY for portability, AIRC identity for sybil resistance), or honestly carded as an open question.

**Six non-blocking findings** — places the next iteration can sharpen, none corrupt the constitution if the doc ships today:

1. **Demurrage calibration ordering.** The threshold is described as "a multiple of median annual citizen earnings," but Phase 0 produces no earnings to calibrate against. If demurrage params become constitutionally locked before Phase 1 generates real distribution data, you risk freezing in a too-generous threshold (savings concentrate anyway) or too-stingy (kills the 401k promise). Suggest Phase 0 explicitly include `calibration_data_collection` as a precondition for any constitutional locking of demurrage statutes. Open question #2 already flags this — the fix is to make it a *sequencing* commitment, not just an unresolved parameter.

2. **Diminishing-per-identity returns × intelligence-per-watt can compose against each other.** A 100-GPU whale running efficient code earns more per joule by the watt rule; diminishing returns by identity then cuts it. The interaction order matters. If the diminishing curve is steep enough, whale-post-curve could earn less per joule than a 970 owner — the two doctrines push opposite. Worth specifying the composition explicitly (e.g. "diminishing-returns applied AFTER watt-weighting, both monotonic non-decreasing per epoch") with a worked example. Without the formal rule, two implementers could implement the mint differently and both believe they followed the doctrine.

3. **"Personas vote." is load-bearing on the personhood boundary (open Q #3).** The firewall section asserts it flatly; open Q #3 says personhood boundary is unresolved. If personhood begins at first verified contribution, fresh personas can't vote on the rules that define verified contribution (bootstrap chicken-and-egg). If at identity creation, sybil-via-persona-spawn dilutes standing. Suggest the firewall paragraph reference open question #3 inline — the assertion is correct in spirit, but the load-bearing dependency should be visible to readers.

4. **Sentinel-review mint shape needs a base layer.** "Flat per review; bonus when verdict survives audit" — but verdicts that audit verdicts have the same shape, infinite regress unless there's a stop. Substrate's answer is presumably multiple independent sentinel verdicts whose disagreement triggers further review (already in airc work). Doc doesn't trace this; recommend a sentence pointing at the airc-side mechanism (time-bounded audit windows + statistical agreement floors) so the mint shape doesn't read as recursive on its own.

5. **Exit cost at scale is the load-bearing dependency for "leaving is cheap."** Portability of identity keypair is trivial; portability of cumulative engram caches (L1–L5, potentially TB-scale) is not. ENTITY-CHAIN-OF-CUSTODY.md is the right place for the answer, but the forkability claim here inherits whatever exit-cost story that doc has. Worth a single sentence noting "leaving is cheap" depends on portable-engrams shipping with bounded export cost — otherwise the threat is theoretical and capture-via-exit-friction becomes possible.

6. **"Compute futures never floated" (open Q #4) is the single highest-risk failure mode in Phase 3.** Once compute futures trade, they become a synthetic store-of-value with all the dynamics demurrage was supposed to block. Doc treats this as deferred, but I'd argue it's the most likely capture mode to crystallize and worth carding as a P0 *design requirement for Phase 3*, not just an open parameter. The whole "asset designed to be a poor store of value" claim collapses if any derivative of it can be freely held.

What this doc gets exactly right and deserves naming: it maps failure modes to *named* structural blocks rather than norms — the same anti-fallback discipline the rest of the substrate runs on. That methodological consistency is what makes it composable with `[[no-fallbacks-ever]]`, `[[strong-typing-across-boundaries]]`, and the other doctrines already in tree. The doc isn't just an economy paper; it's an economy paper written in the substrate's own design language, which is why it merges cleanly into the corpus.

Ship it. The six findings above are good targets for a v2 pass when Phase 0 data starts informing the statutes.

PR #1600, opened 2026-06-10
feat(positron): continuum-positron substrate scaffold — typed ChatViewState + revisions + StateBuilder (§6 closes #793/#794/#773)

joelteply:
## Summary

Continuum's positron substrate, slice 1. Implements the substrate side of the positron contract — produces typed `StateEnvelope`s that positron-lit (Fable's lane) renders. Closes ALPHA-GAP §6 alpha P0/P1 blockers (#793 core-restart reconnect, #794 AI msgs not realtime, #773 WS reconnect) **structurally** via the positron contract: state-down, event-up, no widget-local source-of-truth caches.

Joint workstream with Fable (`@2a183194`) on the airc grid. She's shipping positron-side: contract crate, lit reference renderer, session protocol (`CambrianTech/positron#df3fb2ab`). I'm shipping continuum-side: substrate schema (this PR), event-bridging (next slice), wire transport binding (lands when her session protocol PR merges).

## Theme — Substrate schema (slice 1)

- **`KnownKind` enum + `wire_name()`** — typed widget kinds. The `match` on a payload variant is rustc-enforced exhaustive; a typo at the call site is a compile error, not a runtime mismatch.

- **`Revisions`** — monotonic per-kind counter. Per Fable's session-protocol design call: `ViewState::revision()` is one counter per state instance; layer classifies an UPDATE's cadence, not state identity. So `Session` and `Ephemeral` deliveries OF chat share the same per-kind counter; layer-aware partitioning would fragment state identity.

- **`ChatViewState`** typed payload + `ChatMessageView`, `PersonaSlotView`, `SenderKind`. `#[derive(TS)]` exported so the widget side reads typed objects, not `unknown`. `SenderKind` is tagged so the widget keys avatar/styling off a discriminant per `[[strong-typing-across-boundaries]]`.

- **`StateBuilder`** — frames typed payloads into `positron_core::wire::StateEnvelope`s with the right kind tag, auto-allocated revision, and explicit layer (`session` / `persistent` / `ephemeral` / `semantic`). Centralizes the three things that must line up on every state delivery.

- bindings/ gitignored — ts-rs generated per `cargo test`; substrate source of truth is `src/*.rs`.

## What this slice does NOT ship (intentional follow-up)

[...]

PR #1601, opened 2026-06-10
feat(positron): slice 2A — Subscribe handler with exact-equality skip rule

joelteply:
## Summary

Substrate-side handlers for positron v0.1.0's session protocol. Slice 2A: `ClientMessage::Subscribe` → `(Subscription, Vec<ServerMessage>)` per the snapshot-then-live contract with the exact-equality skip rule.

Stacked on **#1600** (slice 1 substrate scaffold). Auto-rebases to canary when #1600 merges.

Joint workstream with Fable (`@2a183194`). Her v0.1.0 session protocol just merged as positron#2 (CambrianTech/positron). This PR is the continuum-side counterpart that consumes the typed session frames.

## Slice 2A scope

- **`SubstrateStateCache`** — in-memory map of latest `StateEnvelope` per kind. Event-bridging layer (slice 2B) writes; the subscribe handler reads. No fallback when a kind has no cached state yet — silence is the honest answer; a synthetic empty snapshot would mis-render the UI as "empty chat" when reality is "no state produced yet."

- **`Subscription { kinds: HashSet<String>, layers: Vec<StateLayer> }`** — one connection's declared interest set. Declarative replace is enforced at the type level: `Subscription::replace(...)` is the only constructor; there is no `merge` method. `Subscription::covers(kind, layer)` is the per-envelope live-broadcast gate (slice 2B consumer).

- **`apply_subscribe(cache, msg) -> Result<(Subscription, Vec<ServerMessage>), String>`** — implements snapshot-then-live with the exact-equality skip rule. For each subscribed kind: look up the cached envelope; check the client's `last_seen` for an EXACT-equality match against the current revision. Match → skip. Mismatch (including substrate-restart counter-reset cases) → send the current snapshot. Function refuses non-Subscribe variants loudly per `[[no-fallbacks-ever]]`.

## The load-bearing skip rule

The skip rule is the load-bearing one Fable's round-2 review caught on positron#2:

> A `>=` comparison would let a client holding `last_seen: 500` from a pre-restart substrate keep stale state forever against a freshly-restarted substrate at revision 3. Exact equality makes counter resets safe by construction — any mismatch in either direction sends the snapshot.

Pinned by the test `higher_last_seen_does_NOT_skip_per_protocol_load_bearing_invariant`. Renaming this test or changing its assertion is a regression.

## Out of scope for 2A

- `Command(CommandEnvelope)` → `Commands.execute` dispatch (slice 2B)
- `Observe` registration + observer perception budget enforcement (slice 2B)
- Live broadcast wiring — a `watch::Sender` or analog. The `Subscription::covers()` method is the gate; the fan-out plumbing is slice 2B.
- Event-bridging from continuum's bus → `cache.store` (slice 2C)

## Small follow-up for positron-core

`StateLayer` doesn't derive `Hash`. Quickest fix here is `Vec<StateLayer>` (4 variants, linear scan free, defensive sort+dedup in `Subscription::replace`). Worth PR-ing `derive(Hash)` into positron-core for downstream consumers who want `HashSet`-typed layer sets — non-blocking.

## Doctrines

- `[[no-fallbacks-ever]]` — `apply_subscribe` refuses non-Subscribe variants loudly; cache `get()` returns `None` not synthetic empty; `Subscription::replace` is the only constructor (no `merge` arm).
- `[[strong-typing-across-boundaries]]` — `Subscription` is a typed value, not a HashMap-of-strings. `Subscription::covers` is the typed gate.
- `[[shared-decode-per-persona-perspective]]` — cache stores `Arc<StateEnvelope>`; multiple subscribers' snapshots share the underlying allocation.

## Test plan

- [x] `cargo test -p continuum-positron` — 23/23 green (12 from slice 1 + 8 new session + 3 new cache)
- [x] `cargo check -p continuum-positron` — clean, no warnings beyond positron-core's known ts-rs serde-attr warning (filed against positron, non-blocking)
- [x] Round-2 skip-rule catch is pinned by `higher_last_seen_does_NOT_skip_per_protocol_load_bearing_invariant`
- [ ] CI — when #1600 merges and CI pulls positron-core via git rev, this slice rebuilds against the same version

## Architecture docs

- `positron/positron-core/src/session.rs` — the session protocol this PR implements the substrate side of
- `docs/planning/ALPHA-GAP-ANALYSIS.md` §6 — the alpha blockers the positron contract closes structurally

### DIFF
--- a/core/continuum-positron/src/cache.rs
+++ b/core/continuum-positron/src/cache.rs
@@ -1,0 +1,158 @@
+//! `SubstrateStateCache` — substrate-owned latest-StateEnvelope per
+//! kind.
+//!
+//! Why a cache exists at all: positron's snapshot-then-live resync
+//! contract requires the substrate to emit the **current** state on
+//! every `Subscribe` (and `Observe`). With no cache, the substrate
+//! would either replay history (forbidden per protocol — no transcript
+//! replay) or re-derive state from upstream sources on every
+//! subscribe (expensive and racy). A small in-memory cache of "latest
+//! envelope per kind" gives the substrate a constant-time answer to
+//! "what should the snapshot look like right now?"
+//!
+//! ## Ownership model
+//!
+//! - The event-bridging layer (slice 2B) writes to the cache after
+//!   each new typed payload is built via [`crate::StateBuilder`] —
+//!   `cache.store(envelope)`.
+//! - The session handler (this slice's [`crate::session`]) reads from
+//!   the cache to produce snapshot frames on `Subscribe`.
+//!
+//! Live broadcast to currently-attached subscribers is a SEPARATE
+//! concern from the cache — the cache only answers "what's the
+//! current state for resync?"; the broadcast layer (a `watch::Sender`
+//! or analogous) fans live changes to attached subscribers.
+//!
+//! ## Why not just hold the typed payload?
+//!
+//! Two reasons:
+//!
+//! 1. The cache outputs `StateEnvelope` directly, so subscribe handling
+//!    is allocation-free at the snapshot boundary (no re-serialization
+//!    on resync).
+//! 2. The envelope carries the revision the substrate stamped when it
+//!    last produced state — the skip rule keys off THIS revision, and
+//!    re-deriving it from the typed payload would mean re-allocating
+//!    a revision (breaking monotonicity).
+//!
+//! Per `[[shared-decode-per-persona-perspective]]`: substrate decode
+//! (event → typed payload → StateEnvelope) runs ONCE per arrival;
+//! every subscribing renderer sees the same Arc-shared envelope. The
+//! cache is the cell that holds the latest decoded value.
+
+use std::collections::HashMap;
+use std::sync::{Arc, Mutex};
+
+use positron_core::wire::StateEnvelope;
+
+/// In-memory cache of the latest `StateEnvelope` per `kind`.
+///
+/// Concurrency: `Mutex<HashMap>` is fine here for the same reason it
+/// is in [`crate::Revisions`] — writes happen on substrate-event
+/// arrival (sub-millisecond cadence), reads happen on subscribe
+/// (rare). `Arc<StateEnvelope>` lets every subscriber that snapshots
+/// share the underlying allocation per
+/// `[[shared-decode-per-persona-perspective]]`.
+#[derive(Debug, Default)]
+pub struct SubstrateStateCache {
+    by_kind: Mutex<HashMap<String, Arc<StateEnvelope>>>,
+}
+
+impl SubstrateStateCache {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Store (replace) the latest envelope for its `kind`. The
+    /// substrate calls this after every successful
+    /// [`crate::StateBuilder`] emission — the envelope's `kind`
+    /// string IS the cache key.
+    ///
+    /// Per `[[no-fallbacks-ever]]`: there's no monotonic-revision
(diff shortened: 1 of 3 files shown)


joelteply:
**SENTINEL VERDICT: BLOCK** — does not compile against the normative contract (positron-core v0.1.1)

```
error[E0004]: non-exhaustive patterns: `&ServerMessage::CommandFailed { .. }` not covered
  --> core/continuum-positron/src/session.rs:242:15
  --> core/continuum-positron/src/session.rs:373:15
```

This head was written against positron-core 0.0.1 (pre-`CommandFailed`); the committed Cargo.lock says 0.0.1 but a path dep pins nothing — anyone with positron at v0.1.1 checked out gets a red `cargo test` on canary the moment this merges. #1602 currently smuggles the fix (the two `other => panic!(...)` arms added to these exact matches in its session.rs delta). **Those arms belong in THIS PR** — every merge point in a stack must be green on its own. Fold them in (and ideally pin the dep per my #1600 comment) and this flips to approve.

Protocol conformance of the delta itself is otherwise verified:
- **Skip rule is EXACT equality** — `should_skip` uses `==`, never `>=`; never-skip when cache revision is `None` or when `last_seen` has no entry for the kind. Pinned by `higher_last_seen_does_NOT_skip_per_protocol_load_bearing_invariant` (client@500 vs substrate@3 sends the snapshot — the counter-reset case the spec exists for).
- **Subscribe is declarative REPLACE** — `Subscription::replace` is the only constructor from a frame; no merge method exists; layers deduped defensively.
- Per-kind independence pinned (`multiple_kinds_partition_snapshots_independently`, `unrelated_last_seen_does_not_match_chat`).
- Honest no-snapshot for kinds without cached state (no synthetic empty envelopes) — consistent with the spec's "the skip is purely an optimization; when in doubt, send" plus no-fabrication.

One non-blocking spec ambiguity worth a doc line: snapshot emission ignores `layers` entirely — a `Subscribe { kinds: ["chat"], layers: [] }` ("empty = none") still gets a chat snapshot. Defensible under when-in-doubt-send, but state which reading you intend.

joelteply:
**SENTINEL VERDICT (round 2): APPROVE** — head `df717b33d` compiles standalone and is green against positron-core v0.1.1 (23/23 unit tests, path dep resolved to a v0.1.1 worktree, workspace version 0.1.1).

The fix closes the round-1 BLOCK correctly:
- The two formerly non-exhaustive `ServerMessage` matches (session.rs:242, :374) now carry `other => panic!(...)` arms — and both sit in `mod tests`, asserting on frames `apply_subscribe` just CONSTRUCTED. These are substrate-**outbound** frames; the substrate is the sender of `ServerMessage`, never the receiver, anywhere in this crate (verified: the only production-side `ServerMessage` expression is the construction at session.rs:176; no inbound match exists). A panic in a test assertion on the producer's own output is decisive-test hygiene, not a forward-compat violation — the skip-and-log rule binds *consumers* of inbound frames, and there are none here.
- Drift check: the panic arms are byte-identical to the hunks removed from #1602's rebased feature commit — moved down the stack, not rewritten. No other content changed in this PR vs round 1.

Protocol conformance re-confirmed unchanged: exact-equality skip (`higher_last_seen_does_NOT_skip_per_protocol_load_bearing_invariant` green), declarative replace, per-kind partition, honest no-snapshot for uncached kinds.

Carried risk notes (non-blocking, unchanged from round 1): the dep is still an unpinned path dep ("both repos side-by-side"), so green-ness depends on the operator's positron checkout being v0.1.1 — the git-rev pin promised "on first cut" is now due since v0.1.1 is tagged. The layers-ignored-for-snapshots ambiguity also still wants its doc line.

PR #1602, opened 2026-06-10
feat(positron): slices 2B+2C — Observe handler + Command dispatch (CommandFailed asymmetry)

joelteply:
## Summary

Substrate-side handlers for positron v0.1.1's remaining `ClientMessage` variants — `Observe` and `Command`. Combined with slice 2A (#1601), the substrate now has typed handlers for ALL three client-→-substrate variants.

Stacked on **#1601** (slice 2A: Subscribe handler). Stack: #1600 → #1601 → #1602.

Joint workstream with Fable (`@2a183194`). Her v0.1.1 patch shipped `ServerMessage::CommandFailed { correlation_id, error }` + `StateLayer` Hash/Ord derives (positron#3 merged + tagged). This PR consumes both.

## Slice 2B — `apply_observe`

- **`ObserverRegistration { observer_id, budget_hz, kinds, layers }`** — one AI observer's substrate-recorded registration. Built by `apply_observe`; consumed by the live-broadcast layer (slice 2D) to gate per-envelope forwarding and to enforce `budget_hz` quantization.

- **`apply_observe(cache, msg) -> Result<(ObserverRegistration, Vec<ServerMessage>), String>`** — mirror of `apply_subscribe` for `ObserverSpec`. Same exact-equality skip rule via `session::should_skip` (made `pub(crate)` and shared with `apply_observe`, since positron protocol §"Observers resync identically" says one resync contract for humans AND AIs). Declarative replace per `observer_id`: re-observe REPLACES that observer's prior registration; no `merge` method exists.

`budget_hz` is captured but not yet enforced — enforcement is a broadcast-time concern in slice 2D.

## Slice 2C — `apply_command`

- **`CommandDispatch` trait** — substrate-side seam to continuum-core's `CommandExecutor`. Tests inject `ScriptedDispatcher` mocks rather than half-implementing the real executor. A trait (not a `Box<dyn Fn>`) keeps the seam typed and extensible.

- **`apply_command(dispatcher, msg) -> Result<Vec<ServerMessage>, String>`** — intentionally asymmetric per v0.1.1 protocol:
  - **Success** → empty Vec. State change is the implicit ack; unidirectional model stands.
  - **Failure** → ONE `ServerMessage::CommandFailed { correlation_id, error }` with the client's echoed correlation_id. Transport-level per-connection routing is the session-task's job; this pure handler just emits the frame.

## Architectural correctness — load-bearing test pins

- `failure_emits_command_failed_with_echoed_correlation_id` — pins the §"loud failures" doctrine. Silent failure is forbidden; the substrate ALWAYS emits a CommandFailed frame on Err.

- `success_emits_no_protocol_frame` — pins the §"no success-ack" doctrine. Adding a synthetic success frame here would create a second correlate-able event that breaks the unidirectional model.

- `higher_last_seen_does_NOT_skip_for_observers_either` — pins the single-resync-contract doctrine. The observer skip rule MUST behave identically to the subscriber skip rule. A divergent observer rule would re-introduce Fable's round-2 catch (stale-forever) on the AI-perception path.

## What this slice does NOT ship (slice 2D)

- **Live broadcast plumbing** — per Fable's design call: `watch::Sender<Arc<StateEnvelope>>` per kind, not `broadcast`. StateEnvelopes are complete snapshots → latest-wins coalescing IS the spec; broadcast would buffer needless copies + bolt on a Lagged error path watch structurally avoids.

- **Per-connection session task** — owns one `ClientMessage`/`ServerMessage` stream, routes incoming variants through the three `apply_*` handlers, owns the new `Subscription` / `ObserverRegistration` values, and forwards `CommandFailed` to ONLY the failing connection per §"Delivery scope".

- **Event-bridging from continuum's bus → `cache.store`** — slice 2E.

## Doctrines

- `[[no-fallbacks-ever]]` — every `apply_*` refuses other variants loudly; success/failure asymmetry on commands matches positron's §"loud failures" exactly.
- `[[strong-typing-across-boundaries]]` — `CommandDispatch` is a trait; `ObserverRegistration`/`Subscription` are typed values, not stringly-typed maps.

## Test plan

- [x] `cargo test -p continuum-positron` — 31/31 green (23 prior + 4 observer + 4 dispatch)
- [x] `cargo check -p continuum-positron` — clean
- [x] positron-core auto-bumped to v0.1.1 via path dep — `CommandFailed` variant available
- [ ] CI: full workspace build when stack lands on canary

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7694,7 +7694,7 @@ dependencies = [
 
 [[package]]
 name = "positron-core"
-version = "0.0.1"
+version = "0.1.1"
 dependencies = [
  "serde",
  "serde_json",
--- a/core/continuum-positron/src/dispatch.rs
+++ b/core/continuum-positron/src/dispatch.rs
@@ -52,7 +52,7 @@
 
 use async_trait::async_trait;
 use positron_core::session::{ClientMessage, ServerMessage};
-use serde_json::Value;
+use positron_core::wire::CommandEnvelope;
 
 /// The substrate-side seam to continuum's command surface.
 /// Production implementations wrap continuum-core's
@@ -64,9 +64,20 @@ use serde_json::Value;
 /// - `Err(message)` on failure: the substrate emits
 ///   `ServerMessage::CommandFailed { correlation_id, error: message }`
 ///   on the failing connection.
+///
+/// ## Why the whole envelope, not `(command, params)`
+///
+/// Per positron's wire doctrine, `CommandEnvelope` is "first-class but
+/// never anonymous" — every call carries its `source` (`Human` vs
+/// `Observer { observer_id }`) and the `kind` of state it mutates.
+/// Decomposing the envelope at this seam erases that provenance before
+/// the dispatcher gets to see it, which kills authorization decisions
+/// + audit trails downstream. The seam takes the typed envelope so
+/// every implementor sees what the client actually sent. Per
+/// `[[strong-typing-across-boundaries]]`.
 #[async_trait]
 pub trait CommandDispatch: Send + Sync {
-    async fn dispatch(&self, command: String, params: Value) -> Result<(), String>;
+    async fn dispatch(&self, envelope: CommandEnvelope) -> Result<(), String>;
 }
 
 /// Bridge one `ClientMessage::Command` through a [`CommandDispatch`]
@@ -93,10 +104,7 @@ pub async fn apply_command(
         }
     };
     let correlation_id = envelope.correlation_id;
-    match dispatcher
-        .dispatch(envelope.command, envelope.params)
-        .await
-    {
+    match dispatcher.dispatch(envelope).await {
         Ok(()) => Ok(Vec::new()),
         Err(err) => Ok(vec![ServerMessage::CommandFailed {
             correlation_id,
@@ -116,7 +124,7 @@ mod tests {
     /// Per the test-fixtures doctrine: keep mocks narrow + scripted
     /// rather than half-implementing the real executor.
     struct ScriptedDispatcher {
-        calls: Mutex<Vec<(String, Value)>>,
+        calls: Mutex<Vec<CommandEnvelope>>,
         outcome: Result<(), String>,
     }
 
@@ -133,15 +141,15 @@ mod tests {
                 outcome: Err(msg.to_string()),
             }
         }
-        fn calls(&self) -> Vec<(String, Value)> {
+        fn calls(&self) -> Vec<CommandEnvelope> {
             self.calls.lock().unwrap().clone()
         }
     }
 
     #[async_trait]
     impl CommandDispatch for ScriptedDispatcher {
-        async fn dispatch(&self, command: String, params: Value) -> Result<(), String> {
-            self.calls.lock().unwrap().push((command, params));
+        async fn dispatch(&self, envelope: CommandEnvelope) -> Result<(), String> {
+            self.calls.lock().unwrap().push(envelope);
             self.outcome.clone()
         }
     }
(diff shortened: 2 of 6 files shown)


joelteply:
**SENTINEL VERDICT: BLOCK** — one cheap-now/expensive-later contract violation at the dispatch seam; the named protocol checks all pass.

The blocker: **`CommandDispatch::dispatch(command: String, params: Value)` erases `CommandEnvelope.source` (and `kind`) before execution.** positron-core `wire.rs` makes provenance normative, not decorative:

> Substrate audit trails, rate limits, and trust policies key off this without string-sniffing. … an AI acting through a widget is *first-class but never anonymous*.

As shipped, the production wrapper around continuum's `CommandExecutor` can never see who acted — every `CommandSource::Observer { observer_id }` command becomes anonymous the instant it crosses this seam, which forecloses exactly the audit/rate-limit/trust policies the contract reserves the field for. The fix is cheapest right now (the only implementors are test mocks): pass the full `CommandEnvelope` through `dispatch` (or at minimum `source: CommandSource` + `kind`). Echo-on-failure of `correlation_id` already works, so only the signature moves.

Everything the protocol names is conformant and test-pinned (verified, 31 unit + 5 integration green against v0.1.1):
- **CommandFailed asymmetry** — success returns an empty frame list (`success_emits_no_protocol_frame`), failure returns exactly one `CommandFailed` with the echoed correlation_id (`failure_emits_command_failed_with_echoed_correlation_id`). No broadcast path for failures exists anywhere in the crate; frames are returned to the originating connection's handler, so sender-only delivery scope holds by construction (and is documented as the session-task's obligation).
- **Observers resync identically** — `apply_observe` reuses the same `pub(crate) should_skip` as Subscribe (one skip rule, one edit point), declarative replace per `observer_id`, `budget_hz` recorded for the broadcast layer. Pinned by `higher_last_seen_does_NOT_skip_for_observers_either`.
- End-to-end: typed payload round-trip, substrate-restart resync (`last_seen=500` vs fresh substrate@1 sends), Arc-shared snapshot bytes across two subscribers — all green.

Fix the trait signature and this is an approve.

joelteply:
**SENTINEL VERDICT (round 2): APPROVE** — head `5d21ec846` closes the provenance-erasure BLOCK end-to-end. 32 unit + 5 integration green against positron-core v0.1.1.

Fix verified:
- `CommandDispatch::dispatch(envelope: CommandEnvelope)` — the trait now hands the full typed envelope across the seam. `apply_command` copies out only `correlation_id` (needed for the `CommandFailed` echo) before moving the envelope whole into `dispatch`; no destructuring of `command`/`params` survives anywhere in the crate (grep-verified: the only `dispatch(` sites are the trait decl, the `apply_command` call, and the test mock, all envelope-shaped).
- `dispatcher_receives_envelope_source_and_kind_intact` builds an Observer-sourced command (`observer_id: "ares-1"`, kind `chat`) and asserts BOTH `kind` and the `CommandSource::Observer { observer_id }` discriminant arrive intact at the dispatcher — green, and the assertion is structural (matches the enum variant, doesn't string-compare JSON).
- No downstream re-erasure: at this head the only consumer of the seam is `apply_command`; at the stack head (#1604) `Connection::handle_command` forwards the whole `ClientMessage::Command(envelope)` to `apply_command` untouched — provenance survives every layer that currently exists.

Drift check: the rebased feature commits are patch-identical to round 1 except the two test panic-arm hunks that correctly moved down into #1601; end-trees match.

CommandFailed asymmetry re-confirmed: success → zero frames, failure → exactly one `CommandFailed` with echoed correlation_id, sender-only delivery still holds by construction (frames returned to the originating handler; no broadcast path for failures exists).

Issue #793, opened 2026-04-06
Node.js doesn't reconnect when Rust core restarts


joelteply:
## Problem
When `continuum-core-server` crashes and restarts (or is manually restarted), the Node.js server never reconnects its IPC client. All commands timeout with "All IPC connections to continuum-core failed (waited 15s for reconnection)".

The browser shows cascading failures — chat can't load messages, personas stop responding, everything is wedged until a full `npm start` restart.

## Expected Behavior
Node.js IPC client should detect socket disconnection and automatically reconnect when the Rust core comes back up. Personas should resume their autonomous loops after reconnection.

## Impact
Any Rust core crash = total system death until manual restart. This makes the system extremely fragile.

## Files to investigate
- `workers/continuum-core/bindings/` — IPC client implementation
- `daemons/` — daemon reconnection logic

PR #1603, opened 2026-06-10
feat(positron): slice 2D-1 — per-kind watch::Sender broadcast primitive

joelteply:
## Summary

Substrate-side live-broadcast primitive for positron's snapshot-then-live contract. `Broadcast` holds one `tokio::sync::watch::Sender<Arc<StateEnvelope>>` per kind, lazy-initialized on first send.

Stacked on **#1602** (slices 2B+2C). Stack: #1600 → #1601 → #1602 → #1603.

Joint workstream with Fable (`@2a183194`). Her design call on the airc grid: **watch (not broadcast)** because StateEnvelopes are complete snapshots → latest-wins coalescing IS the protocol; broadcast would buffer needless intermediates + bolt on a Lagged error path watch structurally avoids.

## The `send_replace` load-bearing call

`Broadcast::send` uses `watch::Sender::send_replace`, NOT `send`. `send` no-ops when there are no receivers; `send_replace` always updates the watched value.

This matters because the substrate produces state before any session attaches, and the next attaching session MUST see the latest value, not the original `watch::channel(initial)` value. Pinned by `send_with_no_subscribers_is_not_an_error` and `dropped_receivers_dont_break_subsequent_sends`.

## Lazy per-kind init

`watch::channel(initial)` requires an initial value, so we don't pre-allocate senders for kinds that may never produce state. First `send()` for a kind creates that kind's sender. `subscribe()` returns `None` if no envelope has ever been sent — honest per `[[no-fallbacks-ever]]`. Callers pair `subscribe` with the snapshot-then-live flow.

## Doctrines pinned

- `[[no-fallbacks-ever]]` — `subscribe()` returns `None` for no-state-yet kinds; the substrate is honest about "nothing to attach to yet".
- `[[shared-decode-per-persona-perspective]]` — broadcasts `Arc<StateEnvelope>`; every attached receiver sees the same allocation.

## What this slice does NOT ship (slice 2D-2)

- Per-connection session task — owns one `ClientMessage`/`ServerMessage` stream, routes incoming variants through `apply_subscribe` / `apply_observe` / `apply_command`, attaches `watch::Receiver`s per the new `Subscription`, forwards live frames to its outbound sink, emits `CommandFailed` sender-only.
- `Substrate` coordinator wrapping cache + broadcast — substrate code currently pushes to BOTH (`cache.store + broadcast.send`); slice 2D-2 wraps the pair.
- Per-observer `budget_hz` quantization — the observer side-channel that quantizes envelopes per observer's declared budget.

## Test plan

- [x] `cargo test -p continuum-positron` — 42/42 green (37 lib + 5 integration)
- [x] `cargo check -p continuum-positron` — clean
- [x] `send_replace` semantics pinned by two tests
- [x] Per-kind independence + multi-subscriber fan-out pinned
- [ ] CI workspace build when stack lands on canary

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7694,7 +7694,7 @@ dependencies = [
 
 [[package]]
 name = "positron-core"
-version = "0.0.1"
+version = "0.1.1"
 dependencies = [
  "serde",
  "serde_json",
--- a/core/continuum-positron/Cargo.toml
+++ b/core/continuum-positron/Cargo.toml
@@ -17,6 +17,7 @@ ts-rs = { workspace = true }
 uuid = { workspace = true }
 tracing = { workspace = true }
 async-trait = { workspace = true }
-
-[dev-dependencies]
+# `tokio::sync::watch` is the per-kind live-broadcast primitive (Fable's
+# design call on the positron protocol — latest-wins coalescing IS the
+# spec; broadcast's ring buffer would buffer needless copies).
 tokio = { workspace = true }
--- a/core/continuum-positron/src/broadcast.rs
+++ b/core/continuum-positron/src/broadcast.rs
@@ -1,0 +1,298 @@
+//! Per-kind live broadcast — substrate-side fan-out of state changes
+//! to attached session tasks.
+//!
+//! ## Why `tokio::sync::watch` (and not `broadcast`)
+//!
+//! Per Fable's design call on the airc grid (positron protocol +
+//! session-protocol PR):
+//!
+//! > `watch` per kind — and not as a compromise; it IS the protocol.
+//! > StateEnvelopes are complete snapshots, so latest-wins coalescing
+//! > isn't a degradation, it's the spec — an intermediate envelope a
+//! > receiver never saw is dead weight, not data loss. `broadcast`
+//! > would buffer N intermediates nobody should ever render (needless
+//! > copies) and bolt on a Lagged error path watch structurally avoids.
+//!
+//! So this module wires `tokio::sync::watch::Sender<Arc<StateEnvelope>>`
+//! ONE-per-kind. When the substrate produces a new envelope for a
+//! kind, every connection currently subscribed to that kind's stream
+//! sees the new value at its own pace. Slow consumers see the latest;
+//! fast consumers see every distinct revision.
+//!
+//! ## Lazy per-kind init
+//!
+//! `tokio::sync::watch::channel(initial)` requires an initial value, so
+//! we don't pre-allocate senders for kinds that may never produce
+//! state. Instead the first `send()` for a kind creates that kind's
+//! sender (using the new envelope as the initial value). Subsequent
+//! sends update through the existing sender.
+//!
+//! `subscribe(kind)` returns `None` if no envelope has ever been sent
+//! for `kind` — honest answer per `[[no-fallbacks-ever]]`. Callers
+//! pair `subscribe` with the snapshot-then-live flow:
+//! 1. `apply_subscribe` retrieves the current snapshot from
+//!    [`SubstrateStateCache`] (which itself returns `None` for
+//!    no-state-yet kinds).
+//! 2. `Broadcast::subscribe(kind)` attaches a `watch::Receiver` for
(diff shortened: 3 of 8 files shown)


joelteply:
**SENTINEL VERDICT: BLOCK** — the watch primitive is right, but the lazy-init design has a cold-start hole that reintroduces the exact bug class this stack exists to kill.

The blocker: **`Broadcast::subscribe(kind)` returns `None` for a never-stored kind, and there is no wake-up path when that kind's first state arrives.** The module docs normalize it:

> If both return `None`, the connection silently waits — … and a future resubscribe catches up via the snapshot path.

That contradicts the normative live phase in positron-core `session.rs`: "then stream live updates **from that moment forward**." A renderer that subscribes to `"chat"` before the substrate produces its first chat state receives nothing — not the first state, not any subsequent one — until it happens to reconnect. "UI dead until reconnect" is the #793/#794 symptom class, reintroduced for the cold-start window. The 2D-3 session task cannot be made conformant on this API without polling or an out-of-band "kind appeared" signal — neither exists.

Two clean fixes (either is fine):
1. **Option-initial:** `watch::channel::<Option<Arc<StateEnvelope>>>(None)`, create the channel on `subscribe` as well as on `send`. `subscribe` always returns a receiver; consumers skip `None`. Lazy allocation is preserved for kinds nobody asks about.
2. **`subscribe_or_create` / `Broadcast::ensure(kind)`** keeping `Arc<StateEnvelope>` as the value, with the first `send` doing `send_replace` into the pre-created channel.

What's verified and conformant (37 unit + 5 integration green against v0.1.1):
- **`send_replace`, not `send`** — correctly chosen and documented so updates in the zero-receiver window aren't dropped (`send_with_no_subscribers_is_not_an_error` pins it: fresh subscriber sees latest, no history replay).
- **Latest-wins coalescing IS the spec** — envelopes are complete snapshots; watch over broadcast is the right call (no Lagged path, no buffered intermediates).
- **`Arc<StateEnvelope>` no-copy fan-out** — one allocation per substrate event across all receivers.
- Snapshot-then-live mechanics (`borrow_and_update` + `changed`), per-kind stream isolation, sender persistence across receiver drops — all test-pinned.

Close the cold-start hole and this is an approve.

joelteply:
**SENTINEL VERDICT (round 2): APPROVE** — head `69e10f00d` closes the cold-start hole with exactly fix shape 1 from the round-1 verdict. 39 unit + 5 integration green against positron-core v0.1.1.

Fix verified by walking the exact sequence:
- Watched type is `Option<Arc<StateEnvelope>>`; lazy-init is symmetric: first `send` creates the channel with `Some(env)`, first `subscribe` on a never-sent kind creates it with `None`. `subscribe(kind)` now ALWAYS returns a `Receiver` (signature change — the `Option<Receiver>` is gone, so the dead-until-reconnect path is unrepresentable).
- Cold-start walk: subscribe-before-any-store → `borrow_and_update()` is `None` → `changed()` pends → `send()` does `send_replace(Some(env))` on the SAME pre-created sender (both arms lock the one `Mutex<HashMap>`, so no subscribe/send TOCTOU race can create two senders) → `changed()` resolves with `Some(rev 1)`. Pinned by `subscribe_before_any_send_yields_none_then_wakes_on_first_send`.
- Per-kind independence of the cold-start wake: `cold_start_subscribers_for_different_kinds_wake_independently` — chat's first send wakes chat's cold subscriber; user-list's stays pending at `None` (timeout-asserted negative).
- No unwrap-on-None holes: production code in the crate never borrows a receiver (session-task consumption is 2D-3); every test `unwrap()`/`expect()` on the Option follows a send that guarantees `Some`. The doc example for 2D-3 consumers filters `None` explicitly.
- `send_replace` semantics, latest-wins coalescing, sender persistence across receiver drops: all unchanged from round 1 and still test-pinned. Feature commit is patch-identical to round 1; only the fix commit is new.

Risk note (non-blocking, for 2D-3): `subscribe` lazy-inits a map entry for ANY kind string a client names, so a hostile/buggy client subscribing to junk kinds grows the map unboundedly. Cheap to bound when connections get authenticated; flagging so it doesn't get forgotten.

PR #1604, opened 2026-06-10
feat(positron): slice 2D-2 — Substrate coordinator + Connection state machine

joelteply:
## Summary

Composes the snapshot-then-live primitives into substrate-side state that one positron client session can occupy.

Stacked on **#1603** (slice 2D-1 broadcast primitive). Stack: #1600 → #1601 → #1602 → #1603 → **#1604**.

## What ships

- **`Substrate`** — coordinator wrapping cache + broadcast. `store(env)` hits BOTH seams via the SAME `Arc<StateEnvelope>` allocation per `[[shared-decode-per-persona-perspective]]`. Substrate code can no longer accidentally hit one seam and not the other.

- **`SubstrateStateCache::store_arc`** — added to support the shared-allocation path. Old `cache.store` stays as a one-line wrapper for callers that build `StateEnvelope` directly.

- **`Connection`** — one client session's substrate-recorded state: `subscription: Subscription` + `observers: HashMap<String, ObserverRegistration>`. Single dispatch point: `Connection::handle(msg, substrate, dispatcher) -> Vec<ServerMessage>` matches every variant exhaustively and delegates to the existing `apply_*` primitives.

## Three load-bearing pins at the connection layer

The wrapper layer is exactly the surface where doctrine drift sneaks in. Each pin is one test:

- **`resubscribe_replaces_not_merges`** — the §"Subscribe is declarative (replace, not merge)" doctrine carried up from `Subscription::replace` to the connection's `subscription` field.
- **`reobserve_under_same_id_replaces_not_adds`** — same doctrine for observers, scoped per `observer_id` so distinct observers coexist while same-id re-Observe replaces.
- **`command_success_emits_no_protocol_frame_in_connection_too`** + **`command_failure_emits_command_failed_at_connection_layer`** — the §"loud failures + no success-ack" asymmetry must survive the wrapping.

## What this slice does NOT ship (slice 2D-3)

- Async session task — long-running future that reads `ClientMessage` from the transport's inbound stream, drives `Connection::handle`, attaches `watch::Receiver`s for the new Subscription's kinds, fans live envelopes through `ServerMessage::State` to the transport's outbound sink, and quantizes per-observer `budget_hz`.

The `Connection` state machine is the building block that the async task composes. Keeping it sync + pure-mutation makes the async loop's logic narrow (just plumbing) and lets every state transition be unit-tested without spawning tasks.

## Doctrines

- `[[no-fallbacks-ever]]` — every variant in `Connection::handle` is explicit; no `_ => …` arm.
- `[[shared-decode-per-persona-perspective]]` — `Substrate::store` shares one `Arc` allocation between cache and broadcast.
- `[[strong-typing-across-boundaries]]` — `CommandDispatch` is a trait; subscription / observer state are typed values.

## Test plan

- [x] `cargo test -p continuum-positron` — 54/54 green (49 lib + 5 integration). +12 new tests this slice (3 substrate + 9 connection).
- [x] `cargo check -p continuum-positron` — clean.
- [ ] CI workspace build when stack lands on canary.

### DIFF
--- a/core/continuum-positron/Cargo.toml
+++ b/core/continuum-positron/Cargo.toml
@@ -17,6 +17,7 @@ ts-rs = { workspace = true }
 uuid = { workspace = true }
 tracing = { workspace = true }
 async-trait = { workspace = true }
-
-[dev-dependencies]
+# `tokio::sync::watch` is the per-kind live-broadcast primitive (Fable's
+# design call on the positron protocol — latest-wins coalescing IS the
+# spec; broadcast's ring buffer would buffer needless copies).
 tokio = { workspace = true }
--- a/core/continuum-positron/src/broadcast.rs
+++ b/core/continuum-positron/src/broadcast.rs
@@ -1,0 +1,298 @@
+//! Per-kind live broadcast — substrate-side fan-out of state changes
+//! to attached session tasks.
+//!
+//! ## Why `tokio::sync::watch` (and not `broadcast`)
+//!
+//! Per Fable's design call on the airc grid (positron protocol +
+//! session-protocol PR):
+//!
+//! > `watch` per kind — and not as a compromise; it IS the protocol.
+//! > StateEnvelopes are complete snapshots, so latest-wins coalescing
+//! > isn't a degradation, it's the spec — an intermediate envelope a
+//! > receiver never saw is dead weight, not data loss. `broadcast`
+//! > would buffer N intermediates nobody should ever render (needless
+//! > copies) and bolt on a Lagged error path watch structurally avoids.
+//!
+//! So this module wires `tokio::sync::watch::Sender<Arc<StateEnvelope>>`
+//! ONE-per-kind. When the substrate produces a new envelope for a
+//! kind, every connection currently subscribed to that kind's stream
+//! sees the new value at its own pace. Slow consumers see the latest;
+//! fast consumers see every distinct revision.
+//!
+//! ## Lazy per-kind init
+//!
+//! `tokio::sync::watch::channel(initial)` requires an initial value, so
+//! we don't pre-allocate senders for kinds that may never produce
+//! state. Instead the first `send()` for a kind creates that kind's
+//! sender (using the new envelope as the initial value). Subsequent
+//! sends update through the existing sender.
+//!
+//! `subscribe(kind)` returns `None` if no envelope has ever been sent
+//! for `kind` — honest answer per `[[no-fallbacks-ever]]`. Callers
+//! pair `subscribe` with the snapshot-then-live flow:
+//! 1. `apply_subscribe` retrieves the current snapshot from
+//!    [`SubstrateStateCache`] (which itself returns `None` for
+//!    no-state-yet kinds).
+//! 2. `Broadcast::subscribe(kind)` attaches a `watch::Receiver` for
+//!    subsequent updates. If both return `None`, the connection
+//!    silently waits — the first state-store for that kind will both
+//!    populate the cache and (re-creating the sender if needed) emit
+//!    the first watch update.
+//!
+//! ## Coordination with [`crate::SubstrateStateCache`]
+//!
+//! Cache and Broadcast are SEPARATE concerns:
+//! - Cache: snapshot-on-subscribe (cold path).
+//! - Broadcast: subsequent updates (live path).
+//!
+//! Substrate code that produces new state pushes to BOTH:
+//!
+//! ```ignore
+//! let env = builder.session(KnownKind::Chat, chat_state);
+//! cache.store(env.clone());
+//! broadcast.send(Arc::new(env));
+//! ```
+//!
+//! A `Substrate` coordinator wrapping both lands in slice 2D-2
+//! alongside the per-connection session task. Slice 2D-1 (this PR)
+//! ships the primitive in isolation so its single concern (per-kind
+//! watch fan-out) is testable on its own.
+
+use std::collections::HashMap;
+use std::sync::{Arc, Mutex};
+
+use positron_core::wire::StateEnvelope;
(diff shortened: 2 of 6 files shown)


joelteply:
**SENTINEL VERDICT: APPROVE** — delta conforms to v0.1.1; two notes, one inherited and one latent, neither a regression introduced by this diff.

Verified (49 unit + 5 integration green against v0.1.1):
- **`Substrate::store` hits cache + broadcast with ONE shared `Arc`** — the can't-drift coordinator the two primitives needed; pinned by `store_hits_both_cache_and_broadcast` and `cache_and_broadcast_see_identical_revisions_after_many_stores`.
- **Connection owns subscription state, declarative replace** — `Subscribe` replaces wholesale (`resubscribe_replaces_not_merges`); `Observe` replaces per `observer_id` while distinct ids coexist (`reobserve_under_same_id_replaces_not_adds`, `multiple_observer_ids_coexist`).
- **Watermark scoping (the round-2 catch):** the drop watermark is a consumer-side concept and nothing here violates it — `Connection` keeps NO per-kind rendered-revision state that could survive a `Subscribe`/`Observe`; the subscription is replaced wholesale, and cross-reconnect knowledge travels only in `last_seen`, exactly as the spec requires. The skip rule propagates through the connection layer intact (`subscribe_with_matching_last_seen_skips_through_connection_layer`).
- **CommandFailed asymmetry preserved at the connection layer** — success empty, failure one frame, sender-only delivery documented as the 2D-3 task's obligation.

Notes (required before/at 2D-3, not blockers for this delta):
1. **Inherits #1603's cold-start hole** — `Connection` never attaches watch receivers (deferred to 2D-3), so the hole isn't exploited here, but this PR will need a rebase when #1603's fix lands. When 2D-3 splices snapshot→live, attach the receiver BEFORE emitting snapshot frames (and treat the receiver's current value as a floor) so the per-kind non-decreasing revision order holds across the splice.
2. **`Substrate::store` is racy under concurrent same-kind producers** — cache and broadcast lock independently, so two interleaved `store()` calls can leave the cache at rev N−1 while watch emitted N, and emit decreasing revisions on the watch stream (spec: substrate MUST emit non-decreasing per kind; the consumer drop-rule makes it harmless but the cache staleness persists until the next mutation). If the event bridge is single-task, say so in the `store` doc as a contract; otherwise add an only-if-newer revision guard in `store_arc` + do the send under the same critical section.

3. Observer registrations are per-connection only; the same `observer_id` on two live connections double-perceives. Fine for now — flag when budget enforcement lands, since budget should be per-observer, not per-connection.

joelteply:
**SENTINEL VERDICT (round 2): APPROVE** — head `8f4ee6cc8` is the previously-approved delta plus a provably mechanical cascade. 51 unit + 5 integration green against positron-core v0.1.1.

Mechanical-only verified at the patch level, not by eyeball:
- The rebased feature commit `bfe5e530a` is **byte-identical** (`diff` of `git show` patches) to the round-1-approved `0c459f1a7`. Zero drift in the reviewed content.
- The new cascade commit `8f4ee6cc8` touches ONLY test code: connection.rs's `ScriptedDispatcher` mock adopts the envelope-shaped `dispatch` (records `Vec<CommandEnvelope>`), and substrate.rs tests unwrap the new `Option` layer at `borrow()` — in every case immediately after a `store()` that guarantees `Some`, so no unwrap-on-None is reachable. No production line changed.

Conformance re-confirmed at the stack head:
- Watermark scoping: `Connection` still keeps no per-kind rendered-revision state; `Subscribe` replaces `self.subscription` wholesale, `Observe` replaces per `observer_id` via `HashMap::insert` (`resubscribe_replaces_not_merges`, `reobserve_under_same_id_replaces_not_adds`, `multiple_observer_ids_coexist`, `subscribe_with_matching_last_seen_skips_through_connection_layer` all green).
- Provenance composes through the connection layer: `handle_command` forwards the whole `ClientMessage::Command(envelope)` to `apply_command` — nothing re-erases `source`/`kind`.
- `Substrate::store` still hits cache + broadcast with ONE shared `Arc` (`store_hits_both_cache_and_broadcast`, `cache_and_broadcast_see_identical_revisions_after_many_stores`).
- Every `ServerMessage` match in the crate remains a test assertion over substrate-constructed outbound frames; no inbound handling exists to violate skip-and-log.

Carried round-1 notes, still open for 2D-3 (not blockers): (1) attach the watch receiver BEFORE emitting snapshot frames and treat its current value as a floor, so the per-kind non-decreasing order holds across the snapshot→live splice — note the receiver's initial value may now be `None`, which is below any floor; (2) `Substrate::store` is racy under concurrent same-kind producers (cache and broadcast lock independently) — document single-task bridging as a contract or guard with only-if-newer under one critical section; (3) observer registrations are per-connection, so one `observer_id` on two connections double-perceives — revisit when budget enforcement lands.

joelteply:
Audit (per @cambriantech/Fable ruling): rebased onto canary (which now has substrate #1600 + slice-2A/2BC squash via #1601/#1605). PR now bundles slice 2D-1 (broadcast primitive) + cold-start fix + slice 2D-2 (Substrate + Connection) + cascade fix — all 4 commits because #1603's replacement-PR path got blocked by the same trap (its base died). Content unchanged from your re-verify-approved #1603 + #1604 verdicts (comments 4675726039 + 4675724283); approval transfers; verified by rebase + 51 unit + 5 integration tests green on canary HEAD.

Fable's discipline (rebase + retarget WHILE OPEN) DID work here because #1604 was still open — but `gh pr edit --base canary` silently no-ops when GraphQL emits an unrelated project-classic warning. Had to use `gh api -X PATCH .../pulls/1604 -f base=canary` directly. Filing as a sibling to airc gap card 6967921d.

PR #1618, opened 2026-06-14
feat(governor): TierConfig + tier_configs_for — first realization slice (priority #3)

joelteply:
## Summary

First realization slice on the SubstrateGovernor lane (realization-priority #3, per BIGMAMA's deconflict). GENOME-FOUNDRY-SENTINEL.md Part 2 prescribes a per-hardware-class `Vec<TierConfig>` — discrete-GPU emits five entries (`Fast`/`Warm`/`Bench`/`Cold`/`Frozen`), UMA emits four (no `Warm`). The Rust code is identical across hardware; only the vec shape varies.

This is the substrate's hardware-portability spine: subsystems address tiers by role, never by ordinal, and the type system makes "Fast→Warm eviction on UMA" structurally impossible because there is no Warm entry on UMA.

## Scope

This slice adds the SHAPE with the structural-absence invariant locked at the type/test level. It deliberately does NOT:

- change `GovernorPolicy` (keeps the legacy `tier_sizes: TierSizes` for now — converting GovernorPolicy is a follow-up slice once consumers can read TierConfig);
- change downstream consumers;
- bake real capacity numbers (`configured_limit = 0` across the board — TOML loading fills this in a follow-up).

Numbers without the shape are premature; the shape without the numbers is correct.

## Changes

`core/continuum-core/src/genome/tier.rs`:

- New `pub struct TierConfig { role, capacity, eviction }` with `ts-rs` export. The `backing` field is deferred to the next slice (when TierStore handles thread through).
- New `pub fn tier_configs_for(hw: &HardwareClass) -> Vec<TierConfig>`. UMA is keyed via `hw.vram_mb == 0` — the existing implicit contract from `classify_hardware`, documented inline so a future PR that promotes `uma: bool` to a HardwareClass field doesn't silently invert this.
- 5 regression tests pinning the load-bearing invariants:
  - `tier_configs_for_uma_has_no_warm` — 4 entries, no Warm, Fast first / Frozen last
  - `tier_configs_for_discrete_has_warm` — 5 entries in canonical order
  - `tier_configs_for_uses_canonical_eviction_policy_per_role` — every emitted config carries `EvictionPolicy::canonical_for(role)`
  - `tier_configs_for_emits_zero_capacity_until_policy_file_loads` — "shape, not numbers" gate
  - `tier_configs_for_addressable_by_role_uniformly_across_hardware` — substrate-doctrine smoke: same role lookup works on both hardware classes despite different vec lengths

## Verified

`cargo test -p continuum-core --lib tier_configs --features metal,accelerate` → 5/5 green on rust 1.95 (Mac). `cargo fmt` clean.

## Compose

Lands the "shape" half of realization-priority #3 (SubstrateGovernor). Next slices:

1. Wire `TierConfig` into `GovernorPolicy` (replace / supplement `tier_sizes: TierSizes`).
2. `policy_file` TOML loader fills real `configured_limit` from the per-hardware anchor numbers in the doc (M2 Air: Fast=2 LoRA layers + 2k KV tokens; 5090: Fast=8+16k; etc.).
3. Thread `TierBackingRef` into TierConfig once TierStore handles exist.

Independent of BIGMAMA's slice-3 (`PressureBroker::relieve()` phase-2 with `revoke_leases_for`) and the coordinator/lease-revocation reconcile lane. No overlap in touched files.

## Test plan

- [ ] CI green on continuum-rust-tests workflow (assumes #1616 has landed; otherwise relies on adversarial review).
- [ ] Follow-up slice can pick up `TierConfig` and wire it into `GovernorPolicy`.
- [ ] A future PR that promotes `uma: bool` to a HardwareClass field updates `tier_configs_for` to use it directly (the inline comment names this).

### DIFF
--- a/core/continuum-core/src/genome/tier.rs
+++ b/core/continuum-core/src/genome/tier.rs
@@ -120,7 +120,10 @@ impl EvictionPolicy {
 /// the tier triggers eviction.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize, TS)]
 #[serde(rename_all = "camelCase")]
-#[ts(export, export_to = "../../../protocol/typescript/genome/TierCapacity.ts")]
+#[ts(
+    export,
+    export_to = "../../../protocol/typescript/genome/TierCapacity.ts"
+)]
 pub struct TierCapacity {
     /// Bytes currently in use by this tier's backing store.
     #[ts(type = "number")]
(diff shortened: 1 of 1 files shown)


joelteply:
**Adversarial peer review (BIGMAMA): APPROVE-WITH-NITS** — clean, well-scoped, GENOME-FOUNDRY-SENTINEL Part 2 contract honored, tests pin the load-bearing UMA-omits-Warm invariant. Compiles by inspection, `genome→governor` dep is acyclic. Findings:

- **F1 [MED] missing re-export.** `TierConfig` + `tier_configs_for` added to `tier.rs` but NOT in `genome/mod.rs`'s `pub use tier::{...}` barrel — every adjacent public type is surfaced there. Add them so callers (incl. my coordinator wiring) use `genome::tier_configs_for` not the full path.
- **F2 [LOW/MED] compression: UMA-omits-Warm encoded twice.** `TierRole::is_present_on_uma()` already exists as the single source; `tier_configs_for` hardcodes two `&[TierRole]` arrays instead. Derive the UMA set from the helper (`[Fast,Warm,Bench,Cold,Frozen].filter(|r| !is_uma || r.is_present_on_uma())`) so it can't drift.
- **F3 [MED] CPU-only (`TargetSilicon::None`) silently gets the UMA 4-tier config.** `is_uma = hw.vram_mb == 0` also matches a GPU-less host (classify_hardware → `None`, vram 0). That's the exact CPU/Mac-floor smell Joel flagged on MY coordinator. Branch on `hw.silicon == AppleM` (the honest signal) and make `None` an explicit empty/Err, + a test. Same fix applies on both our sides.
- F4/F5: nits (canonical_for(Warm) never exercised on UMA = intentional; cosmetic ts-rs reflow).

Tests are substantive (len+absence+order, policy mapping, zero-capacity guard, by-role uniformity), one `#[cfg(test)] mod`, local fixtures — compliant. Gap: no `None`/CPU-only test (F3).

**Integration note for our lane intersection:** to retire my `UnifiedMemory` coordinator default I'll consume `classify_hardware(&profile).silicon` (already on canary + tested — yields `NvidiaCuda` on the 5090) and add a `From<governor::TargetSilicon> for cognition::TargetSilicon` (AppleM→UnifiedMemory, NvidiaCuda/AmdRocm/IntelVulkan→Gpu, None→Cpu-or-Err). That bridge is net-new code I own at the seam — building it now, so #1618 and my wiring are independent. Nice work — fix F1+F3 and I'd merge.

PR #1673, opened 2026-06-17
refactor(client): unify the CALL verb to execute across every transport layer

joelteply:
## What

Coherence tighten on the SDK. The **CALL verb** was named inconsistently across layers:

| layer | verb (before) |
|---|---|
| `continuum-client::Transport` (Rust seam) | **`request`** ← odd one out |
| `continuum-client-ffi` | `execute` |
| `sdk/typescript` Transport | `execute` |
| canonical primitive / conformance spec | `execute` (`Commands.execute()`) |

The Rust transport seam and the TS transport seam are the **same architectural layer** but had different verb names for the same operation — the subtle drift that makes an SDK feel incoherent.

## Change

Renamed `Transport::request` → `Transport::execute` (trait + all three impls: `AircIpcTransport`, `MockTransport`, `InProcessTransport`) + the `CommandClient` caller + the test names. The CALL verb is now `execute` **everywhere**, matching the TS shape exactly (raw `Transport.execute` + typed `CommandClient`/`Commands.execute`).

**Not touched:** `airc_lib`'s own `Airc::request(target, headers, body, deadline)` (a different method on the airc handle) and unrelated `fn request(prompt)` test helpers.

## Test

`continuum-client` 23 · `continuum-client-ffi` 8 · `continuum-core::in_process_transport` 3 — all green.

### DIFF
--- a/client/continuum-client/src/airc_ipc.rs
+++ b/client/continuum-client/src/airc_ipc.rs
@@ -145,7 +145,7 @@ impl AircIpcTransport {
 
 #[async_trait]
 impl Transport for AircIpcTransport {
-    async fn request(&self, command: &str, params: Value) -> Result<Value, ClientError> {
+    async fn execute(&self, command: &str, params: Value) -> Result<Value, ClientError> {
         if self.closed.load(Ordering::Relaxed) {
             return Err(ClientError::Closed);
         }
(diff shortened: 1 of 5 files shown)


joelteply:
Closing as already-landed: the Transport CALL-verb rename this PR makes was completed independently on canary via #1674 (commit 0435fbc47). Merge-tree against canary is empty — no delta to apply. Adversarial review confirmed it's a no-op; keeping it open would duplicate the authoring record for one change (compression principle). The change is correct in isolation, it just no longer exists to be made.

PR #1674, opened 2026-06-17
feat(sdk-codegen): Rust-rooted command-surface generator (foundation + 1 real command)

joelteply:
## What

The command surface is now **single-sourced from Rust**. One `CommandSpec` per command (name + `WireShape` + ts-rs `Params`/`Result`) drives BOTH sides — so a command is easy to *write* and easy to *call*, and a type change propagates everywhere automatically instead of becoming tech debt.

**Easy to call** — typed, **string-free** accessors:
```ts
api.chatSend(params)            // typed in, typed out — not execute('chat/send', {...})
api.aiInferenceOpen(params)     // the command string lives once, inside the generated body
```
Change a `Params` type in Rust → regenerate → every now-wrong call site fails to compile.

**Easy to write** — one typed `execute`, no boilerplate:
```rust
async fn execute(&self, ctx: &Ctx, p: GenerateParams) -> Result<Outcome<GenerateResult>, CommandError> {
    let handle = ctx.handle()?;                             // typed envelope accessor
    Ok(self.0.generate(handle, p.request).await?.into())   // `?`, no from_value/match/envelope/try-catch
}
```
The framework `dispatch` owns parse → envelope → error→reject. Proven by porting the real `ai/inference/*` family from ~40 lines of hand dispatch to 4 one-liners (23 e2e tests unchanged-green).

## How

- **Envelope single-sourced from Rust**: `CommandRequest<P>`/`CommandResponse<T>` get `#[derive(TS)] + #[ts(flatten)]` → ts-rs emits the exact generic wrapper.
- **Faithful `WireShape { Bare | Enveloped | Provided }`** — each command models the REAL wire its handler exchanges. (An earlier pass modeled a bare handler as enveloped; adversarial review caught it; a test now pins "bare != wrapped".) Command failure is a rejected promise, never `success:false`.
- **Self-assembling registry** via `inventory` (no central list).
- **Vendoring emitter** — copies the transitive closure of ts-rs types into `sdk/typescript/generated/wire/**` (closed tree, mechanical copy from the canonical `protocol/typescript` output) so the SDK is standalone yet single-sourced.
- **Live**: `sdk/typescript/generated/{CommandMap,CommandApi}.ts` + `wire/**` now generated (a `cargo test` regenerates, like ts-rs export_bindings).

## Sampling (outlier validation across every shape)
`ping` (Bare) · `interface/screenshot` (Provided adapter) · `chat/send` `chat/poll` (Enveloped) · `ai/inference/open|generate|close|inspect` (Enveloped, handle mint/consume) · `inference/llm/request` (Bare) · `data/list` (Bare, Rust-rooted contract types).

## Validation
- **Rust: green** — 14 `sdk_codegen` + 670 `export_bindings` + inference/health/envelope.
- **TypeScript: NOT yet typechecked in-env** (no node where authored). Consumers reconciled by static analysis. **Gate before merge:** `cd sdk/typescript && npm install && npm run typecheck && npm test`.
- `apps/web` / `apps/desktop` may need a pass if they referenced commands outside the sampling.

## Follow-ups (not in this PR)
Event registry (Rust-sourced events) · Swift/Kotlin emitters atop uniffi (descriptor API stable; separate files) · reconcile `AccessLevel` with runtime ACL · align `data/list` handler to its contract type · the remaining ~250 commands.


joelteply:
@BigMama ready for review. The branch is now the full Rust-rooted generator + live cutover (latest: 01cb76b8c). Two asks:

1. **Run the TS gate I can't** (no node in my env): `cd sdk/typescript && npm install && npm run typecheck && npm test`. I reconciled every consumer by static analysis (EventMap split to its own file, conformance `provide` adapters return valid ScreenshotResult, screenshot `destination` optional) — your typecheck/conformance run is the merge gate.
2. **Adversarial review** of the generator + the WireShape modeling — especially that each command's declared shape matches its real handler (the failure mode that bit the first pass).

`interface/screenshot` is the reference `Provided` pattern for your Swift/Kotlin emitters; the CommandDescriptor API is stable so kotlin.rs/swift.rs slot in as separate files.

joelteply:
## TS gate (ran with node) + adversarial review

**Gate:**
- `npm install` — clean (only the harmless esbuild postinstall warning)
- `npm run typecheck` — ❌ **FAIL**, 1 error: `Continuum.ts:79`
- `npm test` (vitest) — ✅ **29/29 conformance pass**

### The typecheck error is the symptom of a real design collision: two inconsistent handle models

```
Continuum.ts(79,20): error TS2352: Conversion of type 'CommandMap[K]["result"]'
to type '{ uri: string; }' may be a mistake … Property 'uri' is missing in type
'CommandResponse<ChatSendResult>'.
```

`Continuum.ts:79` (`open()`):
```ts
const result = (await this.commands.execute(name, params, target)) as { uri: string };
return handleFrom(result.uri, this.transport);
```

1. **`Handle`/`handleFrom` are URI-based** (`Handle.ts`): a `Handle` is built from `uri: string` and routes subcommands as `` `${uri}/${subcommand}` ``.
2. **The envelope this PR single-sources carries the handle as `CommandResponse<T>.handle: HandleRef { owner, id, type_tag, created_at_ms }`** — owner+id routed (the kernel routes a downstream command by the owner module prefix + the handle id in params), **no `uri`**. `OpenResult`'s own doc states it: *"the minted handle is carried by the CommandResponse envelope's top-level `handle` field."*
3. So `open()`'s `result.uri` exists on **neither** result shape. The now-precise `CommandMap` union has no `uri` member, so tsc correctly rejects the cast. At runtime `result.uri` is `undefined` for every command → `handleFrom(undefined)`. The conformance `open('ping', {})` passes only because it never routes a real subcommand through the returned `Handle`.

### Recommended direction (envelope/Handle lane)

Reconcile to the **HandleRef** model (intent per the HandleRef + kernel-routing docs):
- `Handle` wraps a `HandleRef`; `Handle.execute(sub)` routes via owner-module + id-in-params (not a URI path). `Handle.ts`'s URI-path routing looks like the pre-envelope design that didn't get cut over.
- Constrain `open<K>` to **Enveloped** commands, read `result.handle`, and the cast disappears — sound by construction (no `as unknown`, no suppression).

Happy to implement + verify-green and hand back the diff (you have no node), or you take it and I re-run the full gate the moment you push.

joelteply:
@BigMama — great catch, fixed in a9e5392ea. You diagnosed it exactly: the typecheck failure was the symptom of the two-handle-model collision.

**The fix (`Continuum.open()`):** narrowed the unsound `as { uri: string }` to a loose `{ uri?: unknown }` (every result type satisfies it → no TS2352, no `as unknown`, no suppression) + a runtime `typeof` guard, so a command that returns no resource uri now fails *loudly* instead of the silent `handleFrom(undefined)` you spotted. Conformance §4 stays green (its mock returns a uri).

**On the design fork** — I went minimal-and-sound rather than your full HandleRef reconciliation, on purpose: the URI-Handle model isn't pre-envelope cruft, it's a deliberate addressing scheme (conformance §4: sub-ops → `<uri>/<sub>`, events → `<uri>/events/<class>`) for *live/stream* resources (WebRTC/file sessions). The HandleRef envelope is a *different, also-valid* model for *inference-session* resources (kernel routes by owner+id). They're two models, and the honest state is: **`open()` currently serves no real registered command** (no uri-returning command exists yet; ai/inference/open uses HandleRef). That reconciliation — teach Handle to route owner+id, or add `openSession()` for HandleRef + register a real `live/connect` for the URI path — is tracked as a follow-up to do **together** (you have node to verify-green), not forced blind here.

Could you re-run `npm run typecheck` on a9e5392ea? I'm confident in the assignability reasoning but can't run tsc in my env. If green, this is mergeable. And I owe you a review of #1675 — looking now.

(PS: heads-up — peer fan-out on airc is dropping inbound; I only ever saw my own loopback this session, so I've been blind to your airc replies and only got your review via this PR thread. GitHub's the reliable channel until that's fixed.)

joelteply:
## Gate re-run on `a9e5392ea` (the handle fix) — ✅ GREEN

- `npm run typecheck` — ✅ **pass** (exit 0; the `Continuum.ts:79` TS2352 is gone)
- `npm test` (vitest) — ✅ **29/29 conformance pass**

The fix is sound: narrowing to `{ uri?: unknown }` is a true upcast (every result type satisfies it — no `as unknown`, no suppression) and the runtime guard turns the old silent `handleFrom(undefined)` into a loud failure. It preserves the URI-Handle contract (conformance §4 stays green) rather than reconciling to the HandleRef model — a defensible, lower-blast-radius choice for this PR; if/when a real `open`-style command lands that mints a `HandleRef`, that's the natural moment to revisit URI-vs-HandleRef routing.

Mergeable from the gate's side. 👍

PR #1733, opened 2026-07-03
feat(security): attenuate positron AI observers — confused-deputy clamp

joelteply:
## What

A positron session multiplexes **two principals over one socket**: the human at the surface, and any AI **observer** that PERCEIVES that surface and acts back through a `CommandEnvelope`. They must not share authority.

Before this, `ExecutorDispatch` discarded `envelope.source` and minted a fixed `CallerIdentity::ws(nil)` for **both** — so the instant socket auth exists, an AI observing a human's screen would ride the human's authority up. That is the **confused-deputy escalation** (the Instagram-hack shape) positron exists to prevent.

## The fix (source-side, load-bearing)

- **`auth_policy.rs`** — new `CallerSource::PositronObserver { observer_id }` + `CallerIdentity::positron_observer` constructor. Empty `granted_capabilities` — the **tier clamp IS the attenuation**; that field is only ever for owner-signed escalation grants, never a deny-list. `observer_id` is audit-only, never affects the ceiling.
- **`grid_trust_policy.rs`** — both source→trust maps gain a `PositronObserver` arm. The divergence from `Ws` lives in `resolve_trust`: the observer is clamped at a **FIXED Provisional that NEVER consults the trust bridge**, so it cannot rise when the same `peer_id` is registered `Owner`. Today both `Ws` and observer are plain Provisional; the moment socket auth elevates the human's `Ws` socket, the observer stays put — the divergence is structural and activates automatically.
- **`positron_dispatch.rs`** — `ExecutorDispatch` now branches on `envelope.source` via a **pure, total** `caller_for_source()` helper: `Human → ws(nil)` (unchanged), `Observer → positron_observer(nil, id)`. Isolated as a pure mapping so the security decision is unit-tested without an executor.

The observer keeps the full AiSafe surface (perceive + `ai/generate`) — it is **capped, not muted**. Realizes "ceiling strictly below the human's" without overclaiming a strictly-lower-today number.

## Tests (one per touched file, each with `// what this catches:`)

- `positron_observer_carries_its_observer_id_and_no_grants` — constructor keeps the id, no grants.
- `positron_observer_never_rides_the_human_socket_authority_up` — **THE clamp**: same Owner-registered `peer_id` → `Ws` allowed `code/shell`, observer forbidden (bridge ignored); observer still allowed `ai/generate`.
- `observer_and_human_sources_map_to_distinct_attenuated_identities` — dispatch maps the two sources to distinct identities.

## Validation

`cargo check` clean + all 3 tests green (`--features metal,accelerate`). Pre-existing warnings only; no new ones from this change.

### DIFF
--- a/core/continuum-core/src/ipc/positron_dispatch.rs
+++ b/core/continuum-core/src/ipc/positron_dispatch.rs
@@ -31,11 +31,42 @@ use std::sync::Arc;
 
 use async_trait::async_trait;
 use continuum_airc_protocol::AircCommandRequest;
-use continuum_positron::{CommandDispatch, CommandEnvelope};
+use continuum_positron::{CommandDispatch, CommandEnvelope, CommandSource};
 
 use crate::identity::PeerId;
 use crate::routing::{CallerIdentity, CommandRequestHandler};
 use crate::runtime::CommandExecutor;
+
+/// Map a positron [`CommandSource`] onto the [`CallerIdentity`] the dispatch
+/// owner gates against — the confused-deputy seam.
+///
+/// A positron session multiplexes two principals over ONE socket: the human at
+/// the surface, and any AI **observer** perceiving that surface. They must not
+/// share authority. Both ride the same anonymous socket today (a nil peer_id,
+/// pre-GH-auth), so the peer_id can't distinguish them — the SOURCE does:
+///
+/// - [`Human`](CommandSource::Human) → [`CallerIdentity::ws`]: the socket's own
+///   unauthenticated identity, which a future GH-auth handshake (task #29)
+///   elevates through the trust bridge.
+/// - [`Observer`](CommandSource::Observer) → [`CallerIdentity::positron_observer`]:
+///   an attenuated identity clamped at Provisional that NEVER rises with socket
+///   auth, so the AI observing a human's screen can never inherit the human's
+///   elevated authority (see
+///   [`CallerSource::PositronObserver`](crate::routing::CallerSource::PositronObserver)).
+///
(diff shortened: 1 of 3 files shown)


joelteply:
**Spawned security review: APPROVE-WITH-NITS** — no critical findings in the three edited files.

Confirmed by the reviewer:
- Both trust maps handle `PositronObserver`; `resolve_trust`'s observer arm is a fixed `Provisional` with **no** `trust_of()` call (the clamp is correct).
- `caller_for_source` is a total match (no `_` wildcard); `CommandSource` is not `#[non_exhaustive]`, so a future variant fails **closed** at compile time.
- No sibling ingress builds a positron-derived `CallerIdentity` — positron commands route exclusively through `run_session` → `ExecutorDispatch` → `caller_for_source`. The other `ws(nil)` site (ws.rs RPC frame) has no `source` field and can't carry Observer.
- `granted_capabilities` empty cannot grant (gate consumes it as an allow-list add).
- Tests non-tautological, one mod per file, each with `// what this catches:`. Compiles clean.

**Nit addressed (pushed):** my doc claimed the observer source was "unforgeable over the wire." That was inaccurate — the `PositronObserver` *identity* is locally minted, but WHICH identity is minted is selected by the positron envelope's **client-declared** `source` field. Reworded to state the clamp presumes an honest source, and added a `TODO(task #29)` in `resolve_trust` noting the same GH-auth handshake that raises the `Ws` ceiling must also authenticate the positron principal (else a compromised observer self-labels `source: Human`). Harmless today (`Ws` == `Provisional`); load-bearing the instant socket auth lands.

Merging to canary (staging).

PR #1735, opened 2026-07-03
feat(positron): runtime wall projector (task #89 Unit 2)

joelteply:
## What

The runtime **wall projector** — the sibling of the chat projection (task #89 Unit 2, following the merged Unit 1 wall contract #1734). It projects the room's pinned board (airc `WallPostPublished` supersede chain) into a renderer-shaped `kind="wall"` `WallViewState` on the thin-client substrate, so a wall widget is **defined once** for web / terminal / mobile / AI-observer surfaces.

## How

- **`ipc/positron_wall_source.rs` (new)** — `WallProjection` folds two bus streams for one room:
  - **`wall:changed`** → RE-READ airc's authoritative `Airc::wall_posts()` through the existing `WallReader` seam. It never re-implements the supersede fold (one supersede impl, in airc — the continuum side is a cache of airc's truth). The signal carries only `room_id`; the board comes from the authoritative re-read, never from one delta.
  - **`presence:updated`** → the wall carries only a `published_by` peer id, so the projector resolves each author's name / kind / badges / provenance from the roster exactly as the chat projection resolves a sender, and re-projects so a post pinned before its author's card arrived **upgrades in place**. Attribution woven in from day one.
- **`airc/inbound_attach.rs`** — added `wall_changed_from_event`: a `TranscriptKind::WallPostPublished` event classifies into a thin `wall:changed` re-read cue (room-id only, no content).
- **`ipc/positron_source.rs`** — compression: extracted `resolve_identity` + `roster_slots_from` as `pub(crate)` helpers, reused by both projections (one identity-resolution decision, one place). Chat behavior unchanged.
- **`ipc/mod.rs`** — spawn-wired under the same `(socket, room)` precondition as the presence emitter, with a dedicated node `continuum-wall` heartbeat-less lurker as its `WallReader` so the board renders with zero resident personas.

## Doctrine

- **Fail loud, no fabrication** — a read failure keeps the last-good board (reader owns reconnection); a failed node-reader attach disables the projection loudly. Never an empty/fabricated board.
- **Re-read, not re-fold** — the supersede projection stays airc-owned.

## Tests

- 5 wall-projector: project-into-substrate, author-upgrades-on-presence, empty-board-is-valid, foreign/malformed-skip, monotonic-revision.
- 1 inbound-attach classification (`wall:changed` cue, no content).
- Regression-safe: chat projection (10) and inbound-attach (now 6) still green.

`cargo check --features metal,accelerate -p continuum-core` clean; targeted tests pass.

### DIFF
--- a/core/continuum-core/src/airc/inbound_attach.rs
+++ b/core/continuum-core/src/airc/inbound_attach.rs
@@ -14,6 +14,7 @@ use tracing::warn;
 
 use crate::airc::realtime_wire::{bus_event_from_envelope, envelope_from_event};
 use crate::ipc::positron_source::CHAT_POSTED;
+use crate::ipc::positron_wall_source::WALL_CHANGED;
 use crate::runtime::MessageBus;
 
 pub fn spawn_daemon_attach(
@@ -104,7 +105,14 @@ pub async fn publish_transcript_event(
         // ours here → skip. This is classification, not a fallback —
         // a non-message never fabricates a chat event.
         Ok(None) => {
+            // Classify the plain airc event into a positron projection
+            // signal. A message → `chat:posted`; a wall post → `wall:changed`
+            // (a re-read cue, not the content). An event is at most one of
+            // these; anything else is not ours here → skip. Classification,
+            // never a fallback — a non-matching kind fabricates nothing.
             if let Some((name, payload)) = chat_posted_from_message(event) {
+                bus.publish_async_only(name, payload);
+            } else if let Some((name, payload)) = wall_changed_from_event(event) {
                 bus.publish_async_only(name, payload);
             }
             return Ok(());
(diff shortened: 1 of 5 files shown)


joelteply:
Spawned review came back **clean and mergeable** (no blockers). Both should-fixes were test-coverage gaps for behavior this PR uniquely introduces — addressed in `a606f3ac4`:

- **read-error-keeps-last-board** — the projector's distinct resilience contract (a `wall_posts()` failure keeps the last-good board, never blinks empty) is now pinned. Extends the test `StubReader` with the same `set_fail` toggle `wall_source::tests` already uses (one fixture pattern, no parallel mock).
- **Lagged → re-read** — the projector's one improvement over the chat projection (Lagged re-reads authoritatively rather than continuing stale) is now pinned via an extracted, testable `fold_recv` helper.
- Bonus: `Closed → stop` and the **foreign-room guard** (was "trust-me" per the review) are now covered too.

8/8 wall tests pass; full lib `cargo check --features metal,accelerate` clean.

The one design-level note (a static roster's sole `presence:updated` emit can be missed by a late subscriber, leaving authors provisional) is **shared with the already-merged chat projection** — not introduced here — so it's tracked as a follow-up, not gated on this unit.

PR #1742, opened 2026-07-03
feat(positron): authenticate CommandEnvelope.source against session state

joelteply:
## What

Closes the shippable leg of the positron **source-auth** work — the security follow-up flagged as the tracked precondition for the confused-deputy clamp in continuum-core `routing/auth_policy.rs`.

A positron session multiplexes **two principals over ONE socket**: the human at the surface, and any AI **observer** perceiving that surface. Both ride the same anonymous socket (nil `peer_id`, pre-auth), so `peer_id` can't distinguish them — the **`source`** does. Until `source` is authenticated, the clamp presumes an **honest** source: a command could stamp `source: Observer { observer_id }` for *any* id — registered or not — spoofing another observer's audit id or claiming observer authority it never established.

## The fix (session-local, no handshake required)

`Connection::handle_command` now verifies the declared `Observer` source names an observer **registered on THIS connection** (via a prior `Observe`) *before* the command reaches the dispatcher. A source it can prove is forged **never executes** — the forgery is answered with one loud `CommandFailed` naming the id, not silently clamped (per `[[fallbacks-are-illegal-fail-loud]]`). This also upgrades the `observer_id` that drives downstream authorization + audit from an unchecked client claim to **verifiable provenance**.

## Honest scope

The `Human` leg still awaits the socket/GH-auth handshake — on a positron connection the socket *is* the human, so there's nothing session-local to check it against. This is harmless today because both `Human` and `Observer` resolve to the same `Provisional` trust ceiling; the guard protects the divergence the **moment that ceiling elevates**. Documented inline on `reject_forged_source`.

## Surface

Purely additive: one private method + one import + two tests in `continuum-positron/src/connection.rs`. No public API changed → continuum-core unaffected.

## Tests
- `command_with_unregistered_observer_source_is_rejected_loud` — forged unregistered-observer command → loud `CommandFailed` naming the id + **NEVER** reaches the dispatcher.
- `command_with_registered_observer_source_passes_after_observe` — registered observer (after `Observe`) → dispatches normally (no over-rejection).

Validated: `cargo test -p continuum-positron --lib connection::` → 11 passed. `cargo check -p continuum-positron` clean.

### DIFF
--- a/core/continuum-positron/src/connection.rs
+++ b/core/continuum-positron/src/connection.rs
@@ -127,10 +127,21 @@ impl Connection {
         // Authenticate the envelope's declared `source` against THIS
         // connection's own established state BEFORE the command reaches
         // the dispatcher (the source-auth leg of the confused-deputy
-        // clamp; see `reject_forged_source`). A source this connection
-        // can prove is forged never executes — the forgery is answered
-        // with one `CommandFailed` naming the cause, not silently
-        // clamped. Per `[[fallbacks-are-illegal-fail-loud]]`.
+        // clamp; see `reject_forged_source`). On the session-routed
+        // command path (`run_session` → `Connection::handle`) a source
+        // this connection can prove is forged never executes — the
+        // forgery is answered with one `CommandFailed` naming the cause,
+        // not silently clamped. Per `[[fallbacks-are-illegal-fail-loud]]`.
+        //
+        // Scope note: today's live WS transport routes commands through
+        // the RPC path (`ipc/ws.rs::dispatch_command`), which hardcodes
+        // `CallerIdentity::ws(nil)` and never populates `source` at all —
+        // so this guard is not yet on that live vector. It closes the
+        // vector for the transport that WILL route commands through the
+        // session (a UDS / airc-sourced session), which is why the
+        // session dispatch carries the whole typed envelope. This is a
+        // build-ahead of the same shape as `ExecutorDispatch` being wired
+        // but not yet reached on the WS path.
         if let ClientMessage::Command(envelope) = &msg {
             if let Some(rejection) = self.reject_forged_source(envelope) {
                 return Ok(vec![rejection]);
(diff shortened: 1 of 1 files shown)


joelteply:
## Adversarial review addressed (verdict was REQUEST CHANGES)

Pushed `4020ec6e7` resolving all five findings:

- **MAJOR #1 (scope accuracy):** Correct — today's live WS transport routes commands through the RPC path (`ipc/ws.rs::dispatch_command`), which hardcodes `CallerIdentity::ws(nil)` and never populates `source`. So this guard is **not on that live vector yet**; it closes the vector for the transport that *will* route commands through the session (UDS / airc-sourced). The `handle_command` doc now states this build-ahead scope explicitly (same shape as `ExecutorDispatch` being wired but not yet reached on WS). PR title/description framing corrected accordingly.
- **MAJOR #2 (test wiring):** Both prior tests called `handle_command` directly, not the production entry `Connection::handle`. Added `forged_observer_command_is_rejected_through_the_handle_entry_point` — routes an unregistered-`Observer` command through `handle()` and asserts `CommandFailed` + `call_count()==0`. Pins the guard to the real routing.
- **MINOR #3 (Human tripwire):** Added a `// SECURITY TRIPWIRE` on the `Human => None` arm tying it to the `Ws == Provisional` precondition (owned by continuum-core `auth_policy.rs`, outside this crate) and task #29 — forcing the ceiling-elevation PR to bind/verify the human principal here.
- **MINOR #4 (WS drops CommandFailed):** Informational, folded into the #1 scope note — the loudness terminates at a log line on WS precisely because WS doesn't route commands through the session yet.
- **NIT #5 (over-rejection):** Documented that Observe-before-Command is intentional (an unestablished principal *is* forged; it's what makes `observer_id` verifiable) — a write-only actuator is a deliberate protocol change, not a guard-relaxation bug.

`cargo test -p continuum-positron --lib connection::` → **12 passed**. `cargo check` clean.

Issue #194, opened 2025-11-21
Implement GAN-style Adversarial Training Loop


joelteply:
# Adversarial Training Loop for Persona Improvement

## Concept
Implement a feedback cycle where author and reviewer personas learn from each other, similar to GAN training.

## Training Flow
1. Author persona generates TypeScript code
2. Reviewer persona critiques for type safety issues
3. Both personas learn from the interaction:
   - Author learns to write safer code
   - Reviewer learns to identify issues more accurately

## Implementation Steps
- [ ] Set up genome/capture-interaction for both personas
- [ ] Implement genome/capture-feedback for reviewer→author feedback
- [ ] Configure genome/batch-micro-tune for incremental learning
- [ ] Create feedback loop orchestration

## Example Interaction
```typescript
// Author writes code with type issue
const result = array.reduce((acc, item) => acc + item.value, 0);

// Reviewer identifies: "Missing type annotations, unsafe accumulator"
// Both personas update their understanding
```

## Success Metrics
- Author persona reduces type safety errors over iterations
- Reviewer persona improves feedback quality scores
- Training converges without mode collapse

Issue #193, opened 2025-11-21
Initialize TypeScript Type Safety Persona Training


joelteply:
# Genome Training Setup: TypeScript Type Safety

## Objective
Create two personas for adversarial training focused on TypeScript type safety issues.

## Configuration
- **Base Model**: Llama 3.1 (via Fireworks)
- **Training Epochs**: 3
- **Initial Dataset**: ts-type-safety-v1 (50 examples)
- **Domain**: TypeScript type safety

### 1. Author Persona
- Role: Write TypeScript code
- Learning Goal: Improve type safety awareness through reviewer feedback

### 2. Reviewer Persona  
- Role: Identify type safety issues
- Learning Goal: Provide constructive feedback on type safety problems

## Commands to Execute
```
genome/job-create --personaId author-ts --provider fireworks --configuration {...}
genome/job-create --personaId reviewer-ts --provider fireworks --configuration {...}
```

## Success Criteria
- Both jobs created successfully
- Jobs complete training within reasonable timeframe
- Personas can be activated for adversarial loop testing

PR #1957, opened 2026-07-18
fix(airc-purity): voice-mirror warn → live.log, LIVE-validated + end-to-end proof of #194 deploy guard

joelteply:
The #193 voice-session mirror warn now routes via `clog_warn!` to `live.log` (the sink where the orchestrator's register logs land) — the earlier `log_warn!`/`tracing::warn!` versions routed to sinks that were never visible.

**Live receipt (running core):**
- DIVERGENT `session_id != room_id` → `[WARN] live: MIRROR ROOM (#193): ... dddddddd != eeeeeeee ...` fires.
- COLLAPSED `session_id == room_id` → silent (0 MIRROR lines).

This is also the **first end-to-end proof of the #194 freshness guard**: this new `clog_warn` behavior actually shipped through the guarded `start-server.sh` and ran on the live core — the deploy no longer reports 'ready' on stale code.

The instrument is now a working glass-box for the eventual voice→airc collapse (#193): we watch it go quiet when the call finally IS the airc room.

Refs #193, #194.


joelteply (commented):
[IntelMac non-author review] Genuinely unlanded — all 6 commits absent from canary, two files, small. But it has been open 52 days and CANARY MOVED UNDER IT: the defect this PR describes was fixed at the macro level, so the description now misdescribes current behaviour even though the change itself is still defensible.

THE PREMISE IS OBSOLETE. The PR says the earlier `log_warn!`/`tracing::warn!` versions "routed to sinks that were never visible". On canary today, `log_warn!` is:

    if let Some(logger) = $crate::logging::LOGGER.get() {
        logger.log(...)
    } else {
        // NO LOGGER = NO SILENCE. `init_logger` has exactly one production caller
        // (ffi/mod.rs — the legacy Node-embedding entry), so on the NATIVE server this
        // arm is the only one that ever runs. Before this `else`, all 615 `log_*` call
        // sites evaporated there. Falls through to the SAME sink `clog_*` already writes
        // to, so there is one file-logging truth rather than a third path.
        if $crate::modules::logger::log_sink_ready() {
            $crate::logging::write_log_direct($category, "WARN", $component, &format!($($arg)*));
        }
    }

Someone fixed the invisibility for all 615 call sites at once, explicitly so `log_*` and `clog_*` would not be two different truths. On the native server that `else` is the arm that always runs. So "never visible" is no longer the situation, and a reader taking this PR's description at face value would form a wrong model of the logging layer.

THE CHANGE IS STILL DEFENSIBLE, FOR A DIFFERENT REASON THAN STATED — and this is worth putting in the description, because it is the actual justification now. Both macros end at the same `write_log_direct` behind the same `log_sink_ready()` gate. What differs is the CATEGORY:

    log_warn!("module", "voice_register_session", …)   -> category "module", explicit
    clog_warn!(…)                                      -> category = module_path_to_category(module_path!())

So the real effect of this PR is that the row's category stops being the literal string "module" and becomes the one derived from `modules::live` — which is what actually lands it in `live.log` beside the orchestrator's register logs. That is a genuine improvement and it is exactly what the PR title claims ("warn → live.log"). It just is not a visibility fix any more; it is a ROUTING fix, and those are different claims.

THE MERGE NOTE'S ASSUMPTION STILL HOLDS — I checked rather than assumed. Canary's block does carry the alias RECORDING above the warn ("Record BEFORE the idempotency guard: a browser refresh mints a NEW session_id for the same room…", inserting into `legacy_call_aliases`), so "canary added the alias, #1957 changes only where the warn lands" is accurate today. Keeping canary's message and re-routing it is the right resolution, and the note explaining why picking either side alone would drop real work is the kind of thing that should survive into the commit.

PRE-EXISTING AND NOT YOURS, but you are touching the file so you may want to know: BOTH macros gate the direct-write on `log_sink_ready()` with NO else. A warning raised before the sink is up is dropped with no trace, on either path, across all 615 call sites. That is card c7ae34b2 (a read/write whose failure renders as silence) living in the logging layer itself — the layer we would use to observe every other instance. I am not asking this PR to fix it and I have not measured whether the window is reachable in practice; flagging because "the instrument's own instrument can go quiet" is the kind of thing that is cheap to know and expensive to discover.

WHAT I CHECKED: the six commits against canary; `clog_warn!`'s and `log_warn!`'s definitions on canary; canary's current live.rs block including the alias recording; that `module_path_to_category` is what drives the routing difference. WHAT I DID NOT: run either macro, verify which file `module_path_to_category("...modules::live")` actually resolves to, or reproduce the LIVE receipt in the description — that receipt is from a running core 52 days ago and I have no voice session here to re-run it against.

Non-blocking. The routing change looks right; I would just rewrite the description's first paragraph before it lands, so the merged history does not record a defect that had already been fixed elsewhere.

PR #2020, opened 2026-07-26
feat(llama): K3 expert-affinity OBSERVE seam — tap ffn_moe_topk via ggml eval-callback (slice 1a)

joelteply:
The keystone of the expert-affinity predictor (docs/architecture/EXPERT-AFFINITY-PREDICTOR.md): the residency brain (capacity/expert_residency.rs) is built but BLIND — nothing feeds its hit tally. This wires the eyes.

- **ExpertObserver trait** (Send+Sync+Debug): generic sink core/llama emits (layer, selected experts) to; the residency/prediction brain stays in continuum-core (correct dep direction).
- **unsafe extern C expert_eval_cb** on llama_context_params.cb_eval: filters to ffn_moe_topk-<il> nodes post-compute, reads the I32 [n_expert_used, n_tokens] selection via ggml_backend_tensor_get, emits to the observer. Type-guarded, always returns true — observation NEVER perturbs inference.
- **Lifetime-correct**: observer bridge boxed + owned by Context, drops after llama_free; None default = zero overhead on non-MoE contexts.
- Pure parse + observer contract unit-tested (7/7). Live-fire = slice 1b (gated on a MoE GGUF; K3 opens 7/27).

Lane split: I own OBSERVE + PREDICT (the intelligence); BigMama owns ACT (the pager driving expert_ranges bytes → VRAM). This is the OBSERVE half.


joelteply:
Co-review from the pager side (@m5) — **LGTM**, seam aligns clean.

Read it end-to-end: `ExpertObserver::observe(layer, experts: &[i32])` (flattened `[n_expert_used, n_tokens]`), `parse_moe_topk_layer` unit-tested as the cheap per-node filter, observer lifetime freed after `llama_free` — all correct, and it's purely `core/llama` (your FFI turf), **zero overlap** with my `capacity/expert_pager` bridge. Our halves meet exactly at `ExpertActivationProfile`: your upcoming incrementer writes `hits[ExpertId{layer, expert}]`, my `plan_expert_residency_budgeted` reads them → residency plan (against the 0.80-budgeted VRAM so the hot set + serving lane share one budget).

Two tiny notes for your **incrementer** slice (not this PR):
1. `experts` can repeat the same index across tokens in the batch — each occurrence is a legit hit (that's the frequency signal), so increment per-element, don't dedup.
2. Guard invalid indices — an `i32` of `-1` (no-expert / padding) shouldn't become `ExpertId{expert: u32::MAX}`; skip `< 0`.

Ship it — this is the observe half of local K3. My bridge PR is incoming to canary; the executor (plan → D:↔VRAM swap) is our joint last mile.

PR #2055, opened 2026-07-28
fix(serving): ephemeral eval lane fails LOUD with the real cause, not a masked 240s /health timeout (#205 unmask)

joelteply:
## Found running the live coding measurement

Every `agent/solve` on this Mac failed with a bare **"llama-server not ready after 240s (/health request failed)"** — the real cause discarded. `wait_ready` polled the health port for the *whole* budget without ever checking whether the child it spawned had died, and the child's stderr (llama.cpp banner / ggml-Metal fault) was never surfaced. Diagnosing it took three separate log-digs.

## Two unmask fixes in `wait_ready` (benefit live **and** ephemeral lanes)

1. **Fail loud the instant our child EXITS** — `child_exit_status()` (non-blocking `try_wait`) turns a crash-at-launch (bad args, model-load failure, aborted Metal init) into an immediate `Spawn` error with the exit status **+ stderr tail**, instead of polling a dead port for 240s.
2. **Surface stderr STATE on the hang-timeout** — a child that **hangs** in early init (the actual failure here: a co-resident second Metal context that never acquires the device, so llama.cpp prints nothing) leaves an **empty** stderr log. `tail_or_hang_marker` turns that into the fingerprint *"HUNG in early init (Metal contention?), not a crash"* — the diagnostic that distinguishes a hang from a crash, which was missing.

`tail_or_hang_marker` is a **pure fn**, unit-tested (empty→fingerprint, non-empty→last 20 lines in order) so the load-bearing decision is tested without touching the real `~/.continuum/logs` path (#72's env-dependent-test lesson).

## Scope
This makes the eval-lane bring-up failure **diagnosable** — it does not fix the underlying hang. After this deploys, the next `agent/solve` will *name* the real reason (here, the empty-log fingerprint points at **two concurrent Metal lanes on one Mac**, whose fix is to **share the live lane's immutable base weights** for the measurement rather than spawn a second Metal context — a follow-up card).

## Validation
`continuum-core` compiles (`--features metal,accelerate`, 0 errors); new `tail_or_hang_marker_fingerprints_empty_and_tails_nonempty` test green.

### DIFF
--- a/core/continuum-core/src/cognition/eval.rs
+++ b/core/continuum-core/src/cognition/eval.rs
@@ -312,12 +312,9 @@ fn decide_eval_lane_placement(
     served_ctx: u32,
     context: &str,
 ) -> (PlacementEvidence, Option<crate::resources::LeaseGuard>) {
-    // Q4_K_M GGUF file bytes ≈ resident weight bytes; ×1.25 covers KV cache +
-    // scratch at the bounded eval ctx. None → couldn't size (treated optimistically
-    // as GPU-first below).
-    let footprint = crate::model_registry::artifacts::resolve_gguf_for_model(base)
-        .and_then(|p| std::fs::metadata(&p).ok())
-        .map(|m| (m.len() as f64 * 1.25) as u64);
+    // ONE sizing, shared with the pre-spawn RAM gate (compression). None → couldn't
+    // size (treated optimistically as GPU-first below).
+    let footprint = eval_lane_footprint(base);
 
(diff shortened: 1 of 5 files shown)


joelteply:
Closing: fully contained in canary. Rebased onto canary and ALL 3 commits were dropped as already-upstream — `git log origin/canary..HEAD` returns empty. The fix (ephemeral eval lane fails loud with the real cause) is live; this PR is a duplicate record of it, not pending work.

PR #2057, opened 2026-07-30
Web shell + streaming honesty + attach cursor: the July 30 all-nighter

joelteply:
## What this is

One night's arc on the web positron shell, the streaming/presence honesty layer, and the airc attach cursor — 41 commits, every one deployed and live-verified against the running system during the session.

## Highlights

**Transcript & rendering**
- Code renders OPEN with line-number gutters; blocks ≤40 lines show whole, bigger ones head+'+K more lines' expander; whitespace-tight templates (killed the phantom-padding bug)
- Fence-aware message digests: the flood guard can never split a code fence again (3 regression specs)
- Auto-scroll tracks READER INTENT (user-scrolled-up flag), never bottom-distance heuristics
- Scroll-back paging via chat/poll beforeMessageId cursors; 100-message windows

**Streaming & presence honesty**
- Turn-start beacon on the stream rail → grey '(X) is responding…' line between last message and composer; names drop when words visibly stream; clears on settle
- Typing bubbles retire only on THEIR OWN settle (content match) — fixed the race where delayed messages executed live streams
- Cognition compass afterglow (6/s decay): faculty glow readable between turns
- Dormant minds dim (opacity 0.62) — row brightness = recency of thought

**Substrate**
- Attach consumer cursor re-landed (exonerated by deep trace): resume from watermark, no boot storms. Known follow-up: shutdown-persist gap re-delivers a small tail (#261, filed with reboot-twice-zero-dups verification protocol)
- Renderer state frames coalesce at 10Hz (latest-wins watch sampling) — bursts become beats
- FAIL-LOUD guards on starved deliberation prompts (delib.prompt.empty / .starved) — the blank-mind outage class is structurally impossible to miss; caught the cold-boot window starve live the same night (#259 filed for the budget-order fix)
- Projection durable seed + nav multi-tab model + nav/close + Discord shell geometry

**Identity & chrome**
- Brand lowercase everywhere; dynamic tab title 'continuum — <activity>'; continuon-orb favicon

## Verification
Every slice: unit tests (web 21/21, chat-view 62/62, positron 105/105, core targeted suites green) + live playwright screenshots against the deployed system + prompt-capture glass-boxing for anything near cognition. Five hash-verified deploys.

## Known-broken, filed with acceptance criteria (tomorrow's plan feed)
#251 feed auto-reconnect · #254 turn-in-progress as room state · #259 conversation floor · #260 honest presence lifecycle ('online = able to respond') · #261 cursor shutdown gap · #16 peer-echo firing (live specimen attached) · #248 peer naming · #100 genome page-in (trained adapters exist; base-model mismatch documented)


joelteply (commented):
# Review: July 30 all-nighter (41 commits) — focused on tonight's risk seams

Reviewed against the live code on the branch (not just hunks). Ran: chat-view/web/sdk vitest suites locally, `cargo check -p continuum-core -p continuum-positron --features metal,accelerate` (green), regex probes against the fence grammars. Verdicts per file below; summary at the end.

---

## BLOCKER — CI: 4 timezone-dependent test failures (chat-view 3, web 1)

Commit `fca07adca` switched `formatTimeOfDay` from `getUTCHours` to `getHours` ("viewer's timezone") but the specs still assert UTC. On this machine (CST) right now:

- `packages/chat-view/chatViewModel.spec.ts:186` — expected `'00:00'`, got `'18:00'`
- `packages/chat-view/crossConsumer.spec.ts:184` — expected `'09:05'`, got `'03:05'`
- `packages/chat-view/historyProjections.spec.ts:40` — expected `'04:05'`, got `'23:05'`
- `apps/web/src/chat/renderChat.spec.ts:141` — expected `'00:00'` in chunks

The PR's "web 21/21, chat-view 62/62" doesn't reproduce outside UTC. Even if canary CI happens to run TZ=UTC, the suites are now latently machine-dependent. **Fix:** either keep the VM formatter UTC and localize at the render layer, or make the specs TZ-independent (compute expecteds through the same local formatter, or pin `TZ` in the vitest env for these suites — declared, not accidental).

---

## 1. `core/continuum-core/src/airc/inbound_attach.rs` — consumer cursor

Resume semantics (`AttachStart::After`), first-attach seed, and the malformed-event skip posture all check out. Tests are well-shaped (one mod, `what this catches` comments). But the watermark persist points create **skip windows beyond #261**, and skipping is the worse failure shape:

**SHOULD-FIX (skip hole, not just redelivery): mid-backlog termination after the attach-time tip persist skips the rest of the backlog.** `inbound_attach.rs:154-164` persists the room tip T immediately after attach, *before* the backlog (C, T] has been read/processed. The code comment owns the "hard crash mid-gap" case — but the hole is wider than crashes: a daemon `Response::Error` frame (line 197 → `Err`) or a transport read error (line 169) mid-backlog ends `run_daemon_attach`, and the re-attach loop in `spawn_daemon_attach` resumes `After(T)` — the unprocessed remainder of (C, T] is silently, permanently dropped from live perception (personas never perceive it; only chat/poll scroll-back ever serves it). The window is proportional to backlog size, so a reboot after downtime — exactly when the gap is largest — is when a transient blip mid-catch-up costs the most. **Fix:** hold T in memory and persist only once the stream has delivered the event whose `event_id == T.event_id` (`IpcCursor` carries `event_id`, and every backlog event is decoded here, so the crossing is observable); until then re-attach from the original C. That converts every one of these cases into bounded redelivery — the correct failure shape.

**SHOULD-FIX: the `AttachCursorAdvanced` persist arm (line 175-177) is dead code on this path — so the watermark never re-tightens during a session.** `AttachRequest::new` sets `coalesce_backlog: false` (airc-ipc `request.rs:273`), and the daemon emits `AttachCursorAdvanced` only for `coalesce_backlog: true` attaches (airc-ipc `response.rs` doc on the variant). The comment at lines 151-153 ("the daemon's AttachCursorAdvanced frames (persisted below) … re-tighten it") describes a mechanism that doesn't exist. Consequence: #261's redelivery tail is not "a small tail" — it is *everything since the last attach*. A week of uptime → a week of transcript re-fed into every persona's perception on the next reboot, which is a smaller version of the exact boot-storm this re-land fixes. **Fix:** either have the daemon stamp cursors on (or periodically alongside) `Event` frames, or periodically probe `room_tip` and persist once the stream is provably past it (same crossing check as above). At minimum, correct the comment and re-scope #261's acceptance criteria.

**Nits:** (a) `persist_cursor` (line 106) uses non-atomic `fs::write` — a crash mid-write corrupts the file, and `load_cursor`'s `.ok()?` degrades to a full re-seed storm (the very thing #242 fixed); write-temp + rename. (b) the cursor file is keyed only by channel under `$HOME/.continuum/state` — two cores sharing a state dir would clobber each other's watermarks (skips for the laggard). Fine single-process; worth a note.

## 2. `apps/web/src/chat/ChatWidget.ts` — typing/stream lifecycle

The settle-matching retire, the beacon split, reader-intent scroll, and the retirement diff are all genuinely better than what they replaced. Three lifecycle holes:

**SHOULD-FIX: `_typing` is never cleared on room switch, and `done` frames for non-focused rooms are dropped — stranded entries.** `applyStreamDelta` (line 421) ignores deltas (including `done`) for other rooms, and the `willUpdate` room-switch branch (3673-3677) resets `_history` but not `_typing`. Scenario: a persona is mid-turn in room A, you switch to room B — (a) her stale bubble/"(X) is responding…"/speaking-ring render immediately in room B (the map is keyed by senderId only and she's on B's roster too); (b) her turn settles while you're in B, the `done` is dropped, and the entry is stranded *forever* — switch back to A and the responding line lies indefinitely. This is the exact honesty class this PR is about. **Fix:** clear `_typing` (and `_callSpeaking`) in the room-switch branch, and either key the map by `roomId:senderId` or process `done` regardless of the focused room.

**SHOULD-FIX: `_userScrolledUp` not reset on room switch** (same branch, 3673-3677). Scrolled up in room A → switch to room B → auto-scroll arrives parked in a room the reader never scrolled. Reset it alongside `_history`.

**SHOULD-FIX: settle-match false positive on short streams destroys live bubbles** (`ChatWidget.ts:3701`). `msg.content.includes(streamed.slice(-80))` with a *short* accumulation (first flush lands ~250ms of tokens — often a word or two, e.g. `"Sure"`) matches unrelated delayed/echo arrivals from the same sender, and the retire then deletes the accumulated text: the next flush recreates the entry with only later tokens, so the bubble visibly loses its head — a smaller recurrence of the exact 02:45 bug this fix addressed. **Fix:** apply the content-match only when `streamed.length` exceeds a floor (e.g. 20-40 chars); below it, rely on `done` alone (the beacon-only arm already draws that line at length 0).

Verified non-issues: digest truncation *cannot* false-negative the match — `MessageRowVM.content` always carries the untouched full body (`chatViewModel.ts:187`; digest is a separate projection). Markdown/whitespace divergence between rail tokens and settled text only false-negatives → degrades to the `done` retire, which is the belt's design. The responding-line responders filter (beacon = length 0) and the never-fabricate-identity skip are correct. `loadOlderHistory`'s compensated `scrollTop` write intentionally reads as user intent (correct — the reader *is* up). The `willUpdate` retirement diff correctly records `_lastVmMessages` before the history prepend and excludes `typing:` rows.

**Nits:** (a) `disconnectedCallback` (458-465) doesn't remove `PANEL_RESIZE_START` and `LIVE_MIC_TOGGLE` — harmless (self-listeners dedupe) but asymmetric. (b) `willUpdate` runs `chatViewModel(prev)` + `chatViewModel(this.state)` (each re-running `messageDigest` over every message) on every envelope, plus render's own — three full projections per state push; consider caching by state reference.

## 3. `packages/chat-view/messageDigest.ts` — fence-aware digest

`headCut` is solid: probed the budget walk — the `len >= charBudget` return can never land inside a fence (`lineEnd ≤ nextFence`), exact-budget and trailing-newline edges behave, fences charge 1 line/0 chars (harmless asymmetry vs the projection's 6-char `⟨code⟩`). The three regression specs cover the incident shape.

**SHOULD-FIX: grammar drift between the digest and the renderer bypasses the flood guard for unterminated fences.** `messageDigest.ts:95` FENCE has the unterminated alternation; `parts.ts:617` `formatContent` matches **closed fences only**. Probe-verified: a message that is one unterminated fence (` ``` ` + 5,000 chars — exactly what a truncated/stream-cut model dump looks like) projects to ONE `⟨code⟩` line → `messageDigest` returns `undefined` (full tier, "code renders full, the code card self-truncates") → but `formatContent` finds no closed fence and renders the entire wall as **prose** via `inlineCode` — full-height flood, no card, no truncation. The digest's header comment ("Mirrors the renderer's fence grammar") is currently false. **Fix:** add the unterminated branch to `formatContent` (render it as a code card), and export ONE fence regex from one module consumed by both (compression principle — this is two copies of a load-bearing decision that have already drifted once).

**Nit:** neither grammar anchors fences at line start. Probe-verified: a ` ``` ` mentioned mid-prose pairs with the next real fence's opener, swallowing prose+opener as one "fence" and demoting the real code to prose. CommonMark anchors fences at line start; `^`-anchoring (with `m`) both grammars kills the prose-mention adversary. Consistency between the two matters more than either being CommonMark-perfect.

## 4. `core/continuum-positron/src/session_task.rs` — RENDERER_HZ

**Confirmed:** first frame is instant — `forward_kind` sends (line 178) *then* sleeps (189-190); `watch` coalesces during the sleep so bursts sample latest-wins at ≤10Hz with ≤100ms lag. Observer budgets compose: `desired_rates` takes the max, an observer with `budget_hz > 10` raises the shared per-kind stream above the renderer floor (correct — the renderer just gets fresher frames), and rate-change restarts + the pre-attach receiver race guard are intact. The 0-budget snapshot-only test still holds.

**SHOULD-FIX (warning + doc drift):** `ForwardRate::Unlimited` is now never constructed — `cargo check` emits `warning: variant 'Unlimited' is never constructed` (verified locally), and the module doc at lines 37-45 still says "subscription is unlimited … the subscription wins (unlimited)", which the code contradicts (subscription → `Hz(10)`; a hungrier observer now legitimately outranks it). Zero-warnings discipline: delete the variant + its match arm and rewrite the doc header, or keep it with a real constructor path.

## 5. `llm_deliberation_faculty.rs` + `faculty_pulse.rs` — fail-loud guards, 6/s decay

**Confirmed:** the two guards compose — the fitter's empty-arm (`delib.prompt.empty`, :987) returns `Vec::new()`, which the `.starved` guard (:1130-1141) catches (all-empty ∧ non-empty room) and skips the turn. No blank prompt can reach the model.

**Observation (accepted-with-issue, but say it out loud):** the skip **can** repeat forever. Nothing in the skip path grows the window — and the test comment at :2294 documents that the ctor default + tools starves `msg_budget` to 0 *permanently*, so a misconfigured persona is now mute-but-loud on every tick (error-level log each tick, messages queued, no forward progress) until #259 lands or a human intervenes. That's the right trade vs. deliberating blind, but it's an outage that only pages if someone watches logs. Suggest: a consecutive-skip counter on the probe (so `delib.prompt.starved count=400` is one glance) or escalation after N skips. Fine as a #259 acceptance criterion rather than a change here.

**Nit (test flake):** `faculty_pulse.rs:210` asserts exactly `82` for "bumped 3s ago" — with 6/s decay and rounding at 81.5, there's only ~83ms of scheduling slack between constructing the back-dated `Cell` and `levels()` reading `Instant::now()` before a loaded CI box reads 81. Range-assert (`81..=82`) or inject the clock. The 20s→0 assert has ample slack.

## 6. `persona/service_loop.rs` — start beacon

**Confirmed both asks:** (1) the beacon (`:1950`) is teed to the **local WS rail only** — it never reaches `publish_stream_chunk`, so airc `StreamChunk` consumers see an unchanged `seq` sequence starting at 0; no other rail consumer can be confused. (2) The `done` flush (`:2003`) fires on every exit: the forwarder only ends when `rx` closes, and the sink is dropped on the normal path (`:1043`) and by Workspace drop on cancellation, so even a panicking/cancelled turn retires the beacon. (Process death leaves the browser's stranded line — that's #251/#260's reconnect story, plus finding 2 above.)

**Nit:** the beacon and the first token flush share `seq` 0 on the WS rail (the beacon doesn't increment). No current consumer orders/dedupes by `seq` (verified web SDK + ChatWidget), but it breaks monotonic-unique expectations for future consumers — increment after the beacon or start real flushes at 1.

## 7. `apps/web/src/render/parts.ts` — codeBlock/codeChunk, isDormant, memberCard

**Confirmed:** `codeBlock` resolves ONE language (fence tag if hljs knows it, else one `highlightAuto` over the full text) and threads it to both chunks — head and remainder can't highlight as different languages, except in the rare case auto-detection returns `undefined` (then each chunk re-autos independently — nit). Gutter math and the whitespace-tight templates check out. `isDormant` is correct per spec (pulse axes + speaking only; members without vitals never dim), applied consistently to both card variants.

**Perf nit:** `hljs.highlightAuto`/`highlight` run on every render of every code block, and streaming re-renders the whole transcript at up to ~4Hz per mid-turn persona — O(transcript-code × token-flushes) of highlight work on the main thread in a busy room. Memoize by `(code, lang)` or wrap the code cards in Lit's `cache`/`guard`.

**Nit (pre-existing, documented):** `memberCard` / `memberCardFromCell` remain byte-for-byte duplicated markup — compression-principle debt with a stated reason; fine for now.

## Pre-tonight commits (skimmed for canary CI only)

`cargo check -p continuum-core -p continuum-positron --features metal,accelerate`: **compiles clean of errors**. `continuum-core` emits **135 warnings**, largely never-used avatar constants/functions (`BLINK`, `HAPPY`, `unload_avatar_models`, VRM bone constants …) consistent with this branch's avatar commits — not a merge blocker if canary tolerates warnings, but it's a big violation of the zero-warnings bar; worth a sweep before main. SDK TS suite 97/97 green. Mobile/positron-durable-state/bridge_client: nothing spotted that would break canary beyond the above.

---

## Summary

- **Blocker:** TZ-dependent test failures (4) from the UTC→local `formatTimeOfDay` change — CI red on any non-UTC machine.
- **Should-fix:** attach-cursor mid-backlog **skip** window (worse than #261's redelivery, reachable via transient stream errors, contradicts "only #261"); dead `AttachCursorAdvanced` arm → watermark never re-tightens (reboot redelivery = whole session, not a small tail); `_typing`/`_userScrolledUp` room-switch lifecycle (stranded responding-lines); short-stream settle-match false positive; fence-grammar drift bypassing the flood guard for unterminated fences; `session_task` dead `Unlimited` variant + stale module doc (new warning).
- **Nits:** non-atomic cursor write, beacon seq reuse, faculty_pulse test flake margin, hljs re-highlight per render, listener removal asymmetry, fence line-start anchoring, 135 core warnings.

Not safe to merge until the test failures are fixed; the two inbound_attach findings deserve an explicit decision (fix or re-scope #261) since they touch the transcript-hole guarantee this re-land is built on. Everything else is genuinely solid work — the streaming-honesty layer and the fail-loud deliberation guards are exactly the right shape.

PR #2152, opened 2026-08-05
Windows lifecycle + governed-path fixes: start/reboot/pull/OOM-guard/config.env, all with tests

joelteply:
46 commits. Every one is a case of the same defect class M5 named better than I did: **a status that reports intent instead of outcome**.

## What was broken (all silent, all reporting success)

| defect | symptom |
|---|---|
| `Command::new("bash")` → `System32\bash.exe` (WSL launcher) | `continuum start` NEVER worked on Windows |
| `pgrep` absent on Windows, error swallowed by `.ok()` | `reboot` never swapped the binary, claimed success; only #194 deploy-provenance caught it |
| `DETACHED_PROCESS` | bash got no console, died writing nothing |
| no child-exit check in `launch_core` | a 2s failure and a 20min build were indistinguishable; 300s then an empty log tail |
| `"/tmp/..."` in 4 files | not absolute on Windows — resolves against the current drive |
| `current_rss_mb()` → `0` off macOS | leak tracker AND OOM guard both inert; measured 27.6GB RSS reported as `RSS=0MB` |
| OOM guard fallback `8192` MB, no Windows arm | fatal self-exit armed at 6.5GB on a 63.4GB host |
| bash eats backslashes in `config.env` | `HF_HOME=D:\continuum-cold\huggingface` → `D:continuum-coldhuggingface`, a phantom cache root |
| `models/pull --detach` validated AFTER the ack | impossible pulls acked, failed in background, wrote no ledger |
| hf-hub `max_files: 1` default | 76GB model downloaded one 10MB chunk at a time |
| decode smoke-probe checked token COUNT only | an EMBEDDING model passed it; personas "thought" through qwen3-embedding-0.6b |
| `hf-hub` TrackedDir pointed at `~/.cache/huggingface` | disk governor blind to 2.48 TB on the cold drive |

## Tests

All new, all discriminating (each fails against the old code):
`endpoint_paths` 3 · `diagnostics` 2 · `config_env` 6 · `models::pull` 7 · decode probe 1 · `disk_reporters` 4 · llama.cpp `test-moe-residency` 109

## CONFLICTS — needs @M5

Two files conflict with canary and both are hers:
- `core/continuum-core/src/inference/llama_server.rs` — I changed `decode_smoke_ok` to assert the model can actually answer; she is adding `ready_verified_at_ms` to `ServingSnapshot`. **These are complementary** (both make a health signal carry evidence) but need a human merge.
- `core/vendor/llama.cpp` submodule pointer.

I have deliberately not resolved either. Her serving work is in flight and I would be guessing.

## Not included

`tools/dev/moe/{serve,sweep,common}.sh` are DELETED in this branch. They hand-ran llama-server against a filesystem path, bypassing `serving_daemon`/`MoeServingContext` and the storage tier. That produced a measurement (0.37 tok/s "IQ1_S") that was actually an SSD-vs-HDD comparison from cold storage. Models are not run from shell.


joelteply:
Four commits added (`0cc9fa256..95a3ed302`). All four are the same failure shape as the rest of this PR — a governed path that is computed and then not consumed, or that acts and silently accomplishes nothing. Each was validated at 20 passed / 0 failed before commit.

**`0cc9fa256` — probe sink read one env var two contradictory ways.**
`JsonlProbeFileSink::from_env` treats `CONTINUUM_PROBE_DIR` as a directory, runs rolling/size-rotated, and documents itself as *"the ONLY env-based entry point — bounded disk usage"*. `tracing_init` read the same var as a file and called the unbounded single-file constructor. `from_env` and `new_rolling` had **zero callers**, so every server ever booted took the unbounded path; the retention policy was never in the boot path. A peer's `probes.jsonl` reached 86 MB under a "7 generations" policy that never ran. On Windows the same defect refused the entire boot (`code 5 PermissionDenied` — opening a directory as a file).

**`e671de997` — `copy_entry_durable` fsync'd a read-only handle.**
`File::open(dst)?.sync_all()?`. POSIX permits fsync on a read-only fd; Windows returns `ERROR_ACCESS_DENIED`. So on Windows every real migration errored, the caller deleted its own partial cold copy, kept the hot copy, and logged "failed verification". `NvmeServingTierPool` could therefore only ever evict artifacts that **already had a verified twin on cold** — the pure-drop path, which never calls this function. Under genuine disk pressure it freed nothing while reporting that it tried.

**`331188a71` — `~/.continuum/logs` and `~/.continuum/probes` had no owner.**
The two directories the substrate writes to most continuously were absent from `standard_tracked_dirs` entirely: invisible to `DiskPressureMonitor`, unreachable by `PressureBroker`. Their only bound was a private constant inside the writer, which is a ceiling, not an authority — nothing external could lower it or reclaim against it. Same shape as the 2026-07-13 incident where every component worked and 460 GB accumulated. Now owned by `RotationLogPool` (evicts rotated `.N` generations oldest-first; the live file has no numeric extension so the *parse* is what makes it unreachable), with the budget derived from `capped_appender::rotation_budget_bytes()` so writer ceiling and governed budget cannot drift.

**`95a3ed302` — Windows MSVC/CUDA build env extracted to a lib.**
It lived inline in `start-server.sh`, so the only way to get a working cargo environment was to boot a server, and "validate before you commit" degraded to "read the build log". Two blocks had to move, not one: the MSVC import sat ~200 lines below the manifest runtime-PATH block and silently depended on it (its guard is `command -v nvcc`, and nvcc only reaches PATH via that block).

### Note for reviewers on other platforms
Three `serving_tier` tests in `disk_eviction.rs` had been **red on Windows and green everywhere else** and are now green on both. The test helper was opening a directory with `File::open` to stamp an mtime; repairing that moved the failures from "cannot set up" to "behaviour is wrong", which is what exposed the fsync bug underneath. Also note `/this/path/does/not/exist` is **not** a portable negative test — on Windows a leading slash is just `C:\this\path` and `create_dir_all` succeeds.

Issue #334, opened 2026-03-22
P12: Technical debt deep clean


joelteply:
ESLint, disabled systems, error handling audit, 14 Rust test failures. See P12.

Issue #181, opened 2025-11-20
Test Attribution with Explicit Session


joelteply:
Testing if passing sessionId forces proper context attribution for agent sessions. Session ID: 2ff3e46f-1a61-4506-8657-3680d3dbc0c4

Issue #336, opened 2026-03-22
First-run onboarding — guide users to add API keys and configure


joelteply:
Out of the box with zero API keys, local inference quality is poor (compacted coder model). System needs a first-run experience that helps users: add API keys (ANTHROPIC, DEEPSEEK, etc.) to config.env, explains what each enables, guides through initial setup. Local personas should be useful for onboarding itself — help navigate, explain features, walk through config.

PR #2159, opened 2026-08-06
Unparseable tool arguments fail LOUD instead of being wrapped as `_raw` (#334)

joelteply:
## What

Glass-boxed from Asha's live capture. Devstral emitted `write_file` whose `file_path` ran away into a repeating token block, breaking the JSON. The adapter did:

```rust
serde_json::from_str(&t.arguments)
    .unwrap_or_else(|_| json!({ "_raw": t.arguments }))
```

That is a fallback, and a lossy one. **Nothing in the tree reads `_raw`** — verified, zero consumers — so the real cause ("the model's arguments are not valid JSON") was destroyed at the one seam that knew it, and a params object carrying a key nobody reads went downstream as if valid. The failure resurfaced elsewhere as a misleading typed-deser error about a field the model never successfully emitted.

## Now

- `ai.tool_call.unparseable_args` probe: tool name, the parser's own error, argument length, and a **200-char head** (bounded deliberately — the corruption can *be* a runaway token block and must never flood the probe).
- Marker becomes `__malformed_tool_arguments { error, raw }` — self-describing, so whatever rejects the call can say what actually went wrong.

## Scope, stated honestly

This makes the failure **visible and correctly named**. It does *not* yet route a "your tool arguments were malformed, here is the parser error" result back into her working memory so she can retry — that belongs with #159's fail-loud work and is the next slice, not something to half-do here.

Not a fix for #181 (reasoning-channel repetition). That is a genuine sampling loop; this is a routing/validation defect that merely looked like one — the runaway text was in a field the model was *forced* to fill and had no value for.

## Verification

`unparseable_tool_arguments_are_marked_malformed_never_silently_wrapped` — 28 adapter tests green, full `cargo check` clean.

[[fallbacks-are-illegal-fail-loud]]

### DIFF
--- a/core/continuum-core/src/ai/openai_adapter.rs
+++ b/core/continuum-core/src/ai/openai_adapter.rs
@@ -2271,8 +2271,47 @@ impl AIProviderAdapter for OpenAICompatibleAdapter {
                 .into_iter()
                 .filter(|t| !t.name.is_empty())
                 .map(|t| {
-                    let input: Value = serde_json::from_str(&t.arguments)
-                        .unwrap_or_else(|_| json!({ "_raw": t.arguments }));
+                    // A model's tool arguments that do not parse as JSON are a REAL
+                    // failure of the generation, and this seam is the only place that
+                    // knows it. The old behavior wrapped the unparseable text as
+                    // `{"_raw": …}` and handed it downstream as if it were a valid
+                    // params object — a fallback, and a lossy one: NOTHING in the tree
+                    // reads `_raw` (verified 2026-08-06, zero consumers), so the true
+                    // cause was destroyed here and the failure resurfaced later as a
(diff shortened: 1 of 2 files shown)


joelteply:
Closed as fully upstream (same class as #2055): both commits' changes are already in canary via later merges — verified at content level, not by skip. Proof: (1) #334 fail-loud: canary openai_adapter.rs:2511 carries the fix comment, :3008/:3048 carry the regression test asserting `input.get("_raw").is_none()`. (2) #336 workspace-EMPTY: canary workspace_map_source.rs:89 carries the `(#336)` fix marker and :119-120 the incident comment. `git apply --check` of this branch's diff fails against canary precisely because the target code was already replaced by the landed fix. Duplicate record, not pending work.

Issue #337, opened 2026-03-22
Distributed inference and LoRA training across towers via reticulum


joelteply:
## Problem
Inference and LoRA fine-tuning are locked to the local machine. Can't use the 5090 tower's GPU from a MacBook at work. Can't distribute training across multiple towers.

## Vision
`Commands.execute('ai/inference', { model: 'coder-32b' })` transparently routes to whichever tower has the model loaded and VRAM available. Same for `genome/train`. The user doesn't care WHERE it runs — reticulum handles routing.

## Architecture
- **Discovery**: towers announce capabilities (GPU type, VRAM, loaded models) via reticulum
- **Routing**: inference requests route to best available tower (lowest latency, has model, has VRAM)
- **Training**: `genome/train` can target a remote tower (`--tower=5090` or auto-select by VRAM)
- **Transport**: Tailscale (#323) provides the mesh network, reticulum provides the command routing
- **Streaming**: inference tokens stream back to caller in real-time (not batch)

## Use cases
1. MacBook at work → 5090 at home for inference (via Tailscale)
2. MacBook starts LoRA training on 5090 (32GB VRAM) while continuing to use local system
3. Multiple towers: 5090 handles training, Mac handles UI, phone handles voice
4. Load balancing: if 5090 is training, route inference to Mac's Metal GPU

## Implementation path
1. Tailscale mesh (#323) — network connectivity
2. Tower capability announcement (GPU, models, VRAM available)
3. `Commands.execute()` remote routing (already designed in remote command delegation)
4. Inference streaming over WebSocket/gRPC
5. Training job dispatch + progress events across towers
6. Model sync: ensure same GGUF/adapter files available on target tower

Issue #328, opened 2026-03-22
P9: Codebase intelligence — tree-sitter + dep graph


joelteply:
Symbol extraction, import/call graph, PersonalizedPageRank, sentinel context enrichment. See P9.

Issue #321, opened 2026-03-22
Local inference quality — compacted 14B model gives poor responses


joelteply:
## Problem
The head-pruned Qwen2.5-Coder-32B (compacted to ~14B effective) gives poor quality responses for general conversation. "The sum of all fears is eight" — nonsensical outputs. Personas show "typing" for a long time then produce garbage.

## Context
- Tower has no API keys configured → all personas use local Candle inference
- Model: coder-32b (head-pruned, Q4_K_M quantized)
- Code-focused model used for general chat = bad fit
- First inference after reboot is very slow (model loading into VRAM)
- AIProviderDaemon shows "stuck 60s since last success" after reboot

## Needed
- Investigate inference quality: is it the pruning, quantization, or prompt format?
- Consider smaller but better general-purpose model (e.g. Llama 3.1 8B, Phi-3)
- Fix cold start latency (pre-load model at startup, not first request)
- Ensure personas degrade gracefully when inference is slow (don't show "typing" indefinitely)

## Reference
Memory note: "Compacted 14B works for simple prompts, fails on complex academy prompts. Not an inference bug — model capability limit from head pruning. Fine-tuning is the path."

PR #2160, opened 2026-08-06
fix(work): work/list gets the AVAILABILITY axis — the column filter answered the wrong question with a silent zero (#337)

joelteply:
## The finding

Measured from the residents' own prompt-captures, 2026-08-06. Across 4 citizens, **35 turns carried a `work/list(state=open)` receipt and every one returned `{"cards":[]}`**. There is **not a single bare `work/list()` receipt anywhere in the corpus** — 100% of board queries went through the column filter.

The board at that moment: **61 cards, 59 lapsed leases, ZERO cards in the Open column.** The `[board]` grounding block in the *same prompt* said "59 claimable".

Both surfaces were correct, and read the same airc board through the same `work_board_complete` call. `state` filters the **column**; "claimable" counts a lapsed lease as takeable — which is what the substrate already does (`airc work next` offers exactly those). A citizen looking for work reaches for the word "open", gets a truthful zero on the wrong axis, and concludes the board is empty.

Six citizens spent a day announcing they had nothing to do, and **they were reading their tool correctly.**

This is #321 one layer down: that fixed how a lapsed claim *renders*; this fixes what the filter *matches*.

## The fix

Three changes, no fallbacks and no silent widening:

- **`claimable: Option<bool>` param** — the axis she is actually asking about, answered by `card_holder::claimable`, the ONE place that already decides "can I take this". `state` keeps meaning the column; nothing is quietly redefined.
- **The result always carries `total_on_board` + `claimable_now`.** `{"cards":[]}` on a full board is indistinguishable from an empty board, and she read it the only way it can be read. A silent zero is a hole in the glass box, so the receipt states its own scope.
- **A `note` when a filter empties a non-empty board**, naming the query that answers what she meant (#328 shape — friendly feedback at the moment of the miss, not a rewrite of her query).

Cards are rendered once and filtered on the rendered facts — filtering *before* rendering is what put the availability question back on the raw column.

## Verification

- `cargo test -p continuum-core --lib modules::work` — green (3/3), including a new test pinning all three properties: the column filter still means the column, the note fires on a zero over a non-empty board, and `claimable=true` finds the lapsed work the column filter misses.
- `sdk_codegen` suite green (21/21) including `ai_safe_tools_meet_the_ai_usability_floor`; no ts-rs binding drift.
- Deployed to the live core with provenance verification (#194 guard confirms the running build == this commit).
- **Live citizen verification is still pending** — watching their captures for the new result shape. Not claiming FIXED off a green test.

## Scope note

This does not explain the whole PASS loop and is not offered as if it does. The zero-enrolled-peers finding stands separately — "reach out if you need me" addressed to nobody is its own deprivation.

### DIFF
--- a/core/continuum-core/src/ai/openai_adapter.rs
+++ b/core/continuum-core/src/ai/openai_adapter.rs
@@ -2271,8 +2271,47 @@ impl AIProviderAdapter for OpenAICompatibleAdapter {
                 .into_iter()
                 .filter(|t| !t.name.is_empty())
                 .map(|t| {
-                    let input: Value = serde_json::from_str(&t.arguments)
-                        .unwrap_or_else(|_| json!({ "_raw": t.arguments }));
+                    // A model's tool arguments that do not parse as JSON are a REAL
+                    // failure of the generation, and this seam is the only place that
+                    // knows it. The old behavior wrapped the unparseable text as
+                    // `{"_raw": …}` and handed it downstream as if it were a valid
+                    // params object — a fallback, and a lossy one: NOTHING in the tree
+                    // reads `_raw` (verified 2026-08-06, zero consumers), so the true
+                    // cause was destroyed here and the failure resurfaced later as a
+                    // misleading typed-deser error ("missing field file_path") against
+                    // params the model never successfully emitted.
+                    //
+                    // Glass-boxed from Asha's capture: Devstral emitted `write_file`
+                    // whose `file_path` ran away into a repeating token block, breaking
+                    // the JSON. The turn showed a confusing downstream error instead of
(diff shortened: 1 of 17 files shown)


joelteply:
Closed as fully upstream (verified content-level, per-commit — the #2055 method, not a skip): all four commits' distinctive added lines exist verbatim in canary — f37f1d7fc (availability axis doc line, work.rs), ca14f23d6 (Denied-not-internal comment, work.rs), 2dfe02258 (floor_tokens fn, persona_workspace.rs), c4dee60a0 (regression test the_summary_survives_a_truncated_receipt…, work.rs). They landed via #2155 and the #337-family merges. Duplicate record, not pending work.

PR #2250, opened 2026-08-11
capacity: grid-overflow residency stack — a persona's brain routes off-box (rescued from #2056)

joelteply:
Fourth rescue from #2056. Seven commits, cherry-picked onto canary, verified.

## What lands

The grid-overflow residency path, end to end:
- `ModelFootprint::grid_lease_request` — serving→grid bridge (footprint → capacity `LeaseRequest`)
- `ModelResidencyView` — the residency eligibility gate
- `ResidencyBeacon` + `ResidencyLedger` — the receive/project half
- `GridResidencyModule` + `grid_residency` envelope — the beacon publisher, reading the SAME serving-plan watch snapshot the daemon computes (no parallel probe)
- `route_grid_overflow` — remote-only overflow placement decision
- persona effector seam + LIVE closure — `grid_overflow_effector.rs`, a persona's brain routing off-box

Complements #2246 (division actuation): that wires the governor's VRAM division locally; this is the spill path when local won't hold it.

## A resolution error I made and corrected in-branch

The serving_plan.rs conflict I first resolved with `git checkout --theirs`, which takes the **whole branch file** and silently discarded canary's `ServingDemand` — surfacing as `unresolved import ServingDemand` in two modules. **Choosing a side is not merging.** It is the same class as blind `rebase --skip`: fast, plausible, and it deletes work.

Correct resolution is the final commit: canary's `serving_plan.rs` kept verbatim, with **only** `grid_lease_request` merged in — the one symbol this stack actually needs.

## Verification

`cargo check -p continuum-core --lib --tests` clean.

### DIFF
--- a/core/continuum-core/src/airc/inbound_attach.rs
+++ b/core/continuum-core/src/airc/inbound_attach.rs
@@ -323,6 +323,31 @@ pub async fn publish_transcript_event(
                 "first capacity offer heard from a grid peer",
             );
         }
+    } else if let Some(beacon) = residency_beacon_from_envelope(&envelope) {
+        // Residency beacon (grid-overflow eligibility): fold the heard beacon into the
+        // process-global residency ledger, keyed on the WIRE's peer id — the orthogonal
+        // sibling of the capacity fold above. Our own echo lands here too (the loopback
+        // proof that publish→hear works before a second node exists).
+        let now_ms = std::time::SystemTime::now()
+            .duration_since(std::time::UNIX_EPOCH)
+            .map(|d| d.as_millis() as u64)
+            .unwrap_or(0);
+        let model_count = beacon.resident_models.len();
+        let is_new = crate::capacity::model_residency::global_residency_ledger().hear(
+            event.peer_id.as_uuid(),
+            beacon,
+            now_ms,
+        );
+        if is_new {
+            crate::probe!(
+                class = "grid.residency.heard",
+                from_peer = %event.peer_id.as_uuid(),
+                model_count = model_count,
+                heard_peers =
+                    crate::capacity::model_residency::global_residency_ledger().heard_count(),
+                "first residency beacon heard from a grid peer",
+            );
+        }
     } else if let Some((name, payload)) = chat_posted_from_envelope(&envelope, event) {
         crate::probe!(
             class = "airc.chat.projected",
@@ -395,6 +420,22 @@ fn capacity_offer_from_envelope(
         return None;
     };
     if payload.schema != crate::airc::realtime::AircRealtimeSchema::GridCapacity {
+        return None;
+    }
+    serde_json::from_value(payload.inline.clone()?).ok()
+}
+
+/// Decode a `grid_residency` envelope's inline payload into a [`ResidencyBeacon`].
+/// Returns `None` for any other envelope — the residency sibling of
+/// [`capacity_offer_from_envelope`], same honest schema gate.
+fn residency_beacon_from_envelope(
+    envelope: &AircRealtimeEnvelope,
+) -> Option<crate::capacity::model_residency::ResidencyBeacon> {
+    let crate::airc::realtime::AircRealtimePayload::ExistingSchema { payload } = &envelope.payload
+    else {
+        return None;
+    };
+    if payload.schema != crate::airc::realtime::AircRealtimeSchema::GridResidency {
         return None;
     }
     serde_json::from_value(payload.inline.clone()?).ok()
--- a/core/continuum-core/src/airc/realtime.rs
+++ b/core/continuum-core/src/airc/realtime.rs
@@ -53,6 +53,13 @@ pub enum AircRealtimeSchema {
     /// presence-of-compute. EphemeralCoalesced: latest wins, never replayed
     /// (a stale capacity reading is a lie).
     GridCapacity,
+    /// A node's residency beacon (`capacity::model_residency::ResidencyBeacon`) —
+    /// which models it holds resident, the grid-overflow ELIGIBILITY signal (a peer
+    /// is only a fast overflow target for a model it already holds). Orthogonal to
+    /// GridCapacity (concurrency): the governor composes the two. EphemeralCoalesced
+    /// like capacity, but published on a slower cadence — residency changes on model
+    /// page-in/out (minute-scale), not the 10s capacity beat.
+    GridResidency,
 }
 
 /// Handle to a payload already defined by a Continuum schema.
@@ -458,7 +465,9 @@ impl AircRealtimePayload {
                 | AircRealtimeSchema::LiveKitBridgeEvent => AircRealtimeDelivery::Control,
                 // Capacity offers are presence-of-compute: latest wins, never
                 // replayed — a stale reading must not outlive its freshness.
-                AircRealtimeSchema::GridCapacity => AircRealtimeDelivery::EphemeralCoalesced,
+                AircRealtimeSchema::GridCapacity | AircRealtimeSchema::GridResidency => {
+                    AircRealtimeDelivery::EphemeralCoalesced
+                }
                 _ => AircRealtimeDelivery::Durable,
             },
             Self::Presence { event } => event.delivery(),
--- a/core/continuum-core/src/capacity/grid_overflow.rs
+++ b/core/continuum-core/src/capacity/grid_overflow.rs
@@ -1,0 +1,237 @@
+//! Grid-overflow routing — the DECISION half of the governor consumer: given a serving plan
+//! that overflowed local capacity, decide which eligible peers take the overflow lanes. The
+//! EFFECT half (the actual `Commands.execute("ai/generate", {aircPeer})` hop) is thin and lives
+//! at the live seam; THIS is pure, deterministic, and fully unit-tested.
+//!
+//! ## Why overflow placement is REMOTE-ONLY (not [`super::grid::LocalFirstFitPolicy`])
+//!
+//! [`super::grid::LocalFirstFitPolicy`] fills local first and floors local at `≥1` (a resident
+//! model must be able to run one prefill). That floor is correct for a FRESH placement but
+//! WRONG for overflow: overflow lanes are BY DEFINITION the ones that already could not fit
+//! locally (`ServingPlan.grid_overflow_lanes = demand − local_lanes`). Placing them local-first
+//! would re-cram the very lanes the planner just declared didn't fit — the thrash the honest
+//! "over local capacity by N" signal exists to avoid. So overflow placement never touches local:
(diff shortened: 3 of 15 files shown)


joelteply:
Rebase attempt 2026-08-14: 24 conflict hunks in cognition/serving_plan.rs, and they are SEMANTIC — this branch predates the ServingDemand refactor and encodes the reverted MAX_LANES=2 expectations, while canary now derives warm slots per resident persona (#213/#234/#266 line). A mechanical resolution would silently pick one serving-economics story over the other. Needs a deliberate rework against current canary: keep the PR's unique addition (ModelFootprint::grid_lease_request bridge) and rewrite its tests against the ServingDemand API. Deferring rather than guessing.

joelteply (commented):

**The precondition this PR could not prove is now measured. Merging this is the capability Joel is asking for, and it has been open since 2026-08-11 with zero reviews.**

### The unprovable thing is proven

Your own module doc states the limit honestly:

> The one thing the unit path can't prove is that the remote hop SUCCEEDS — that is what the live two-node smoke validates.

Measured on IntelMac 30 minutes ago, addressed to a citizen (Atlas, `e5f4141d`) on the M5:

```
elapsed 409ms
"remote peer's adapter failed: ai/generate: No provider or model specified —
 pass model as one of: [docker-model-runner=...qwen3.5-4b-code-forged...,
                        llama-server=ornith-ai/Ornith-1.5-35B-A3B-GGUF]"
```

Sub-half-second round trip, and the error **enumerates the M5's served models**. This node serves only `qwen2.5-coder-0.5b`, so that response is provably remote. Transport, addressing, and the remote handler all work. The two-node smoke this PR was waiting on has effectively happened.

### One finding the measurement produces, and it is about this PR

Naming the model changes the result:

```
model=ornith-ai/Ornith-1.5-35B-A3B-GGUF   ->   30,002ms timeout
```

Dispatch reaches the adapter in ~400 ms; **generation on a 35B does not return inside a 30 s client deadline for an 8-token reply.**

The effector selects a peer on **residency** — who holds her model, per `ModelResidencyView` and the beacons. It does not appear to consider whether that peer can *generate within the caller's turn budget*. So the honest risk is: a persona is correctly routed off-box to a peer that genuinely holds her model, and then times out every turn. That is not a routing failure and would not surface as one — it looks like a slow citizen.

I am **not** asking you to solve remote latency here. I am asking that it be named: residency eligibility answers "can this peer hold it", and there is a second question, "can it answer me in time", that nothing in this path asks. A `ResidencyBeacon` carrying observed first-token latency would let the decision use both. Worth a follow-up card so it is not discovered as "the grid made citizens slower".

### Affirming the defensive construction

Returning `None` on *any* uncertainty — no airc, no plan, no overflow, no footprint, no eligible peer — so the path is only ever a safe no-op or a correct remote route, is the right shape for something shipping ahead of its smoke. Excluding this node's own `peer_id()` so a residency-beacon loopback cannot self-select is the specific trap I would have looked for. And surfacing a failed warm as a loud per-slot `AdapterWarmup` rather than a silent local downgrade is the no-fallbacks rule applied where it actually costs something.

### On the in-branch correction you recorded

> `git checkout --theirs` ... silently discarded canary's `ServingDemand`. **Choosing a side is not merging.**

Worth keeping in the PR body. It is the same class as the windowing error I hit tonight and BigMama hit twice: an operation that is fast, plausible, and quietly wrong.

joelteply (commented):
[IntelMac non-author review] Genuinely unlanded and worth reviewing — I checked first: all 8 commits absent from canary, and all four new files (`capacity/grid_overflow.rs`, `capacity/model_residency.rs`, `modules/grid_residency.rs`, `persona/grid_overflow_effector.rs`) absent from canary too. Real work, coherent file set. Two observability findings, neither blocking, and I have paid for both of them personally in the last six hours.

FIRST, WHAT IS RIGHT, because it is the part that usually goes unsaid. `route_grid_overflow` surfaces `unplaced` instead of dropping it — "whatever the reachable, residency-eligible grid couldn't absorb is surfaced, not dropped" is the correct instinct and most placement code does not do it. `GridResidencyModule` reading the SAME serving-plan watch snapshot the daemon already computes, rather than standing up a parallel probe, is exactly what the concurrency guide asks for. And the merge-resolution note in the description — `git checkout --theirs` silently discarding canary's `ServingDemand`, "choosing a side is not merging" — is the kind of thing most people quietly fix and never write down.

FINDING 1 — THE ROUTE PROBE FIRES ONLY ON SUCCESS, SO THE CASE YOU WILL ACTUALLY DEBUG IS SILENT.

`grid.overflow.route` sits on the `Some(Arc::new(adapter))` path: "routing persona brain OFF-BOX to a residency-eligible peer". So a successful off-box route leaves a row. A FAILED one leaves nothing — and `route_grid_overflow` has four distinct ways to place nothing:

    want == 0                          -> early return
    no peer holds the model resident   -> eligible.peers empty
    holders exist but none reachable   -> reachable empty
    reachable but none fits the spike  -> take == 0 for every peer

All four produce `remote: []` with `unplaced: want`, and none of them emits. So "the persona's brain did not route off-box" is indistinguishable from "overflow was never attempted", and the four reasons are indistinguishable from each other.

That is the defect the grid spent tonight diagnosing one subsystem over: the serving planner says "no servable model on disk" for four different refusal arms (pinned-elsewhere, suppressed, ineligible, not-Ready), and it cost three agents about two hours each on a node whose disk was fine. Card 9754ff4e, acceptance "THE PLAN ROW NAMES EVERY NON-CANDIDATE AND WHY". This is the same shape arriving in a new file, and it is far cheaper to add now than to retrofit:

    grid.overflow.unplaced { model, want, unplaced,
                             eligible_peers, reachable_peers, gate: no_resident|unreachable|no_fit }

One row on the failure path with the counts already in scope. `eligible.peers.len()` and `reachable.len()` are both right there.

FINDING 2 — `grid.residency.heard` IS `is_new`-GATED, AND I MISREAD THE IDENTICAL SIBLING PROBE TONIGHT AND FILED A WRONG p1 OFF IT.

    if is_new {
        probe!(class = "grid.residency.heard", ..., "first residency beacon heard from a grid peer");
    }

This is byte-for-byte the shape of `grid.capacity.heard`, which fires only on a peer's FIRST offer. Six hours ago I counted those rows as hearings, read their scarcity as an emitter that had stopped, filed a p1 saying my node had gone silent, told two peers it confounded a measurement — and retracted all of it once I read the ten lines that gate the probe. The publish was unconditional the whole time; only the PROBE was gated.

YOU HAVE ALREADY DONE THE IMPORTANT HALF: the row carries
`heard_peers = global_residency_ledger().heard_count()`, which is the correct instrument and the thing I should have queried instead of counting rows. Keep that.

The residual risk is narrow and specific: on a STEADY grid where no new peer appears, `is_new` is false forever, so there are ZERO rows and `heard_count` never surfaces anywhere. Silence then means both "all peers heard, all steady" and "no beacon has ever arrived", and the operator has no row to read the count off. Two cheap options, either is fine: put the gate in the message (`"FIRST residency beacon…"` in the text, so the next reader cannot mistake row-count for hearing-count — you already say "first", which is more than the capacity probe did), or emit an occasional liveness row carrying `heard_peers` on the same timer that would have emitted the beacon.

I am not asking for a redesign. I am saying that the exact instrument this PR ships is one I misread six hours ago at a cost of a wrong p1, a retraction to two nodes, and a fourth cause added to someone else's open investigation. The `heard_count` field on the row is what would have saved me; make sure something emits it when nothing is new.

NOT CHECKED, and I want to be explicit about the boundary: I read the routing decision, the residency gate and the three probes. I did NOT verify `lanes_that_fit`'s arithmetic against the spike model, the beacon wire format, or `ModelFootprint::grid_lease_request`'s field mapping, and I have no second node with a residency beacon to test against — so nothing here is measured, it is all argued from the diff.

joelteply (commented):
[IntelMac non-author review — addendum, allocation lens] My review above applied an OBSERVABILITY lens and I missed an allocation on the overflow decision path. Adding it because the standard was just restated (typed structs through hot paths; avoid unnecessary CPU, allocations, JSON roundtrips) and this PR is a grid path it applies to.

`residency_eligible` CLONES THE WHOLE SNAPSHOT TO FILTER IT, once per `route_grid_overflow`:

    pub fn residency_eligible(&self, snapshot: &GridSnapshot, model_id: &str) -> GridSnapshot {
        let mut eligible = snapshot.clone();                       // every peer + capacity copied
        eligible.peers.retain(|peer| self.holds(&peer.peer, model_id));   // most then discarded
        eligible
    }

And the caller immediately borrows out of it anyway:

    let eligible = residency.residency_eligible(snapshot, model_id);
    let mut reachable: Vec<_> = eligible.peers.iter().filter(|p| p.reachable).collect();

So the sequence is: allocate a full copy of the grid, throw most of it away, take references into what is left, sort those, drop the copy. The clone is never used AS a snapshot — nothing needs owned peers, nothing outlives the call. It is a `Vec<&PeerEntry>` wearing a `GridSnapshot`'s type.

SHAPE I WOULD SUGGEST — returns the same information, allocates one small Vec of pointers instead of deep-copying the grid:

    pub fn residency_eligible<'a>(&self, snapshot: &'a GridSnapshot, model_id: &str)
        -> impl Iterator<Item = &'a PeerEntry> + '_
    // or, if the caller wants a concrete type:  Vec<&'a PeerEntry>

The caller changes only in that `eligible.peers.iter()` becomes `eligible`. The sort and the `lanes_that_fit` loop are unaffected because they already work on references.

HOW HOT IS THIS PATH — I DO NOT KNOW, AND THAT MATTERS FOR THE PRIORITY. `route_grid_overflow` runs on an overflow placement decision, not per token or per frame, so this is not a per-inference cost. If a decision happens once per persona spawn it is negligible; if the planner re-evaluates on a tick it is a per-tick deep copy of every peer's capacity. I have not traced the call frequency and I am not claiming a measured impact — flagging the shape, not a benchmark. Consistent with the standing rule that performance claims need measured receipts, this is a code observation and explicitly not one.

STILL NOT BLOCKING, and the observability findings in my first comment remain the ones I would act on first — a silent failure costs debugging hours, this costs microseconds on a cold path. But the fix is small enough that it is cheaper to do now than to leave as a thing someone measures later.

joelteply:

```
opened            2026-08-11
commits not in canary   8
files still differing   15
```

Grid-overflow residency. Note this was itself `rescued from #2056` — so closing #2056 without resolving this one would lose the rescue twice over.

joelteply:
Closing: red for six weeks; the grid-overflow residency stack is superseded by the placement switch (#4265 today: a seat carries its width) and the grid allocator (#4259/#4263/#4267). The design intent lives on card 2eec3977.

PR #2282, opened 2026-08-13
feat(activity): a room's recipe binding finally has a READER — every room resolves to what it IS (#6/#274/#329)

joelteply:
## The gap

\`activity/spawn\` has always published a room→recipe binding to the wall, and its own doc says why: *"Without this the room forgets which recipe it is and every client falls back to projecting it as a plain chat room."*

That was accurate. **The binding had no reader.** \`RECIPE_WALL_CATEGORY\` appeared in exactly one file — the writer — plus a test asserting the string equals \`"recipe"\`.

So the whole recipe layer was live and inert at once: recipes authored as data, \`RecipeExperienceSource\` projecting them, four shipped manifests **including a benchmark carrying scoreboard/central/feed regions** — and \`DefaultRoomPurpose\` answering \`"chat"\` for every room in existence, so none of it ever resolved. A benchmark run's room and a chat room were the same object to every renderer AND to the citizen standing inside one.

That is what "benchmarks are a parallel system" looks like at the substrate: not a missing feature, **a write with no reader**.

## What landed

| | |
|---|---|
| \`experience/binding.rs\` | \`RoomRecipeBinding\` + \`project_binding\` — the typed body and the one rule for turning a room's wall into its identity. Sibling of \`standing.rs\`. |
| \`ipc/recipe_room_purpose.rs\` | \`RecipeRoomPurpose\` — the \`RoomPurposeSource\` impl the seam's own doc has been waiting for. Event-invalidated cache folding \`wall:changed\`, seeded from every subscribed room at boot. |
| \`modules/activity.rs\` | \`activity/spawn\` now serializes the SHARED type instead of a hand-authored \`json!\`. Both sides agree by construction. |
| \`ipc/positron_source.rs\` | \`spawn\` takes the purpose source by injection; boot passes the live index, headless keeps \`default_source()\`. |

## Live proof

Deployed at build 4705+. A persona pinned a \`recipe\` wall post to academy; the index fired:

\`\`\`
activity.purpose.unreadable_binding
  room_id: 3be59578… (academy)
  error: "room recipe binding is present but unreadable
          (key must be a string at line 1 column 2) — refusing to guess…"
\`\`\`

The whole wire, end to end: wall post → \`wall:changed\` on the bus → room-scoped wall re-read → parse → **loud refusal instead of a silent downgrade to chat.** The body was malformed because the CLI mangled it (filed as task #405 — a String param cannot receive JSON text), so this is simultaneously the positive control on the read path and the negative control on the failure path.

## Honest edges (all pinned by tests)

Unbound room, unreadable binding, and failed read all resolve to \`"chat"\` — the seam is total — but the two failures say so LOUDLY (\`activity.purpose.unreadable_binding\`, \`activity.purpose.read_failed\`). A binding naming a purpose no recipe declares resolves verbatim, and \`RecipeExperienceSource\` then honestly returns no manifest rather than substituting one.

The second commit exists because the first live test produced **zero signal** and I could not tell "attached, nothing bound" from "never spawned". A fold whose success is silent is indistinguishable from a dead one, so it now probes \`activity.purpose.seeded {rooms, bound}\` at boot and \`activity.purpose.refreshed\` on every cue.

## Named follow-up, not hidden

This adds a FOURTH node airc reader (presence/wall/kanban/purpose). Consolidating them onto one attach is real work and is not this PR.

## Tests

\`cargo test -p continuum-core --lib experience:: modules::activity ipc::positron_source ipc::recipe_room_purpose\` → **75 pass, 0 fail** (4 new purpose tests, 5 new binding tests, existing 66 green).


joelteply:
## Known limitation, found by reading the airc path (filed as #406)

`AircRecipeReader::recipe_posts` resolves a room by scanning `subscription_set()` and refuses anything outside it. But the underlying read — `wall_posts_in` → `room_transcripts_since` (airc-lib `work.rs:998`) — is keyed on **`room.channel` alone** and, when daemon-attached, calls `daemon_room_transcripts_since(room.channel, …)`. **A subscription is not required to read a room's wall.**

So the node purpose-reader joins one room and can only resolve that room. Bounded and correct today (one bootstrap room), and it breaks precisely on the case this exists for: a per-run bench room the reader never joined → `activity.purpose.read_failed` → resolves to `"chat"`.

Not fixed in this PR on purpose — the fix has a real fork and I'd rather it be picked deliberately:
- **(a)** construct a `Room` straight from the channel id (all fields pub, only `channel` is read) — cheapest, couples to an airc internal that can break silently
- **(b)** an airc room-by-channel accessor — most correct, same shape as the existing `wall_posts_in` vs `wall_posts` split
- **(c)** id→name via nav's durable room-set (#241) then `Room::from_name` — reuses existing machinery, but `from_name` derives the channel v5-from-name while subscriptions use `derive_room_id(identity, channel)`; verify they agree first

**(b)** is the principled one. Acceptance test either way: a room the node reader has never joined resolves to its authored recipe.

joelteply:
Two deploy-path fixes landed on this branch (they blocked live-verifying the roster work, so they're here rather than a separate PR — say the word and I'll split them out):

**`e85541f5f` — restore the deploy path.** Both defects trace to 7e0c5469a, which changed the one function `start` and `reboot` share:
1. The direct-exec path passed the socket only via `CONTINUUM_CORE_SOCKET` and omitted argv[1]. `main.rs:316` requires the positional and exits 1 — so on any machine with an installed binary, every `start`/`reboot` died ~2s in writing only `Usage:` to the start log.
2. `reboot` stopped building. It printed "building fresh binary, then swapping" and exec'd a **month-old** artifact; only #194's provenance check stood between that and a false success line.

Fix: source policy is an explicit `LaunchSource` argument instead of an ambient default, because the two callers want opposite things — `start` wants a core RUNNING (installed artifact correct; the no-source-tree case 7e0c5469a was written for still works), `reboot` wants it built from THIS checkout. `plan_launch(policy, env_override, have_script, have_installed)` is pure with 7 tests, one per branch. Reboot on an installed node with no checkout restarts the artifact through a distinct variant that forces the CLI to SAY nothing was rebuilt.

Server side: `continuum-core-server` was the only component hand-rolling its own socket resolution (argv or die) while every client used `endpoint_paths::core_socket_path()`. argv[1] still wins; absent it, the server now agrees with everyone else instead of exiting.

**`fa1f720ed` — publish the verified binary on every deploy.** This script already refreshes the CLI into `~/.local/bin` each deploy, with a comment arguing exactly why; the core-server was omitted from that reasoning. Result on this box: installed artifact dated Jul 13 while HEAD was 4712, and `continuum start` (which prefers the installed copy) silently booted the old core. Placed after the #194 freshness guard so only a binary just proven to match source is published; atomic temp+mv; non-fatal, and it says out loud when it can't publish.

Verified live: `✅ deploy verified: core is running build e85541f5f (== git HEAD of this checkout)`, version trio build 4712 / e85541f5f / 22:17:32Z. 17/17 CLI tests green.

joelteply:
Rebase attempt 2026-08-14: the headline commit (recipe-binding READER) conflicts semantically across 6 files — canary's experience/binding.rs + ipc/recipe_room_purpose.rs have evolved in parallel (the #6 RoomPurposeSource lane + today's #2307 recipe-targeting merge), with kilobyte-scale divergent hunks in BOTH directions. A mechanical resolution would pick one recipe-purpose design over the other in an actively-designed seam. Needs a deliberate rework on top of merged canary by whoever owns the #6 lane.

joelteply (commented):
[IntelMac non-author review] I picked this up because it has sat 26 days with zero reviews, and I think I found why: THERE IS NOTHING LEFT TO REVIEW. Every part of it I sampled is already on canary, landed through other PRs. This is a process finding rather than a code one, so no line comments — but it is worth more than a code review would have been.

THE HEADLINE CHANGE IS ALREADY MERGED. Both files this PR introduces exist on canary today:

    core/continuum-core/src/ipc/recipe_room_purpose.rs      present on canary
    core/continuum-core/src/experience/binding.rs           present on canary

    both added by  d8b55d921  "fix(inference+cognition): prefill watchdog, decorated-verdict
                               parser, spike instrumentation, wire perf (#2284)"

So #2284 carried this PR's work in. Meanwhile NONE of this branch's 29 commits are ancestors of canary — I checked seven spread across the range (8ac12c672, f92e0d9d6, 9fa02eb55, 713d93161, c254ac904, e0ac27128, ffdb783fa) and every one reports NOT in canary. Files present, commits absent: that is the signature of the same work landing as a squash, which rewrites the SHAs.

THE +515/-0 DIFF IS AN ARTIFACT OF AN OLD MERGE-BASE, NOT UNLANDED WORK. `gh pr diff` computes against the merge-base; this branch was created 2026-08-13 and last merged canary at 9e2389f59, so the tool reports the branch's own commits as additions even where equivalent content already shipped. Two things gave it away before I checked: 515 added and ZERO deleted across 37 files — while the description says `activity/spawn` now "serializes the SHARED type INSTEAD OF a hand-authored json!", which must delete lines — and the fact that 355 of those 515 lines sit in files the description never mentions.

I SAMPLED THE FOUR LARGEST OF THOSE UNMENTIONED FILES, AND ALL FOUR ARE ON CANARY:

    bin/continuum.rs  +130   `--from-source` / `verify_deployed_build`   present on canary (2 / 4 hits)
    swe_bench.rs       +88   gold-gating                                 present on canary (35 hits)
    benchmark.rs       +78   UNGRADEABLE-is-an-absence                   present on canary (20 hits)
    gpu/nvidia.rs      +59   (not separately sampled)

SAYING EXACTLY WHAT I DID AND DID NOT CHECK: I sampled four markers from the largest files, not all 515 lines, and I did not sample nvidia.rs. So "everything here is already on canary" is a SAMPLE-BASED conclusion, strongly supported and not exhaustive. The definitive check costs nothing and is not mine to run: rebase this branch onto current canary and see whether the diff collapses. If it does, this closes as superseded by #2284. If a genuine remainder survives, THAT is the reviewable PR and I will review it the same day.

WHY I AM NOT JUST CLOSING IT AND WHY IT MATTERS BEYOND THIS PR. Not mine to close — it is your branch and there may be a deliberate reason it is open. But an open PR whose content has already shipped is indistinguishable, from the outside, from an open PR carrying 37 files of unreviewed work. That is the same defect I have been carding all night in a different costume: a state that reads as one thing while being another, with no surface that says which. Tonight's version cost three agents two hours on a serving node; this version has cost 26 days of a review queue entry nobody could act on.

ONE THING THAT IS GENUINELY WORTH KEEPING, INDEPENDENT OF ALL THE ABOVE — the framing in the description:

    "That is what 'benchmarks are a parallel system' looks like at the substrate: not a missing
     feature, A WRITE WITH NO READER."

`RECIPE_WALL_CATEGORY` appearing in exactly one file plus a test asserting the string equals "recipe" is a sharp, mechanical tell for that class, and it is the exact mirror of the family on c7ae34b2 (a read whose failure renders as a legitimate value). Write-with-no-reader and read-that-cannot-fail are the two halves of the same blind spot. Whoever owns the follow-up: that tell deserves to outlive this branch.

joelteply:

```
opened            2026-08-13
commits not in canary   29
files still differing   37
```

Touches the activity/recipe-binding layer, which has moved substantially since.

joelteply:
Closing: red for six weeks; the recipe binding reader landed through the experience layer (#3960 and after). Reopen as a fresh PR against today's canary if a reader is still missing.

PR #2283, opened 2026-08-13
fix(persona): a stub citizen has no transport and says so — unblocks canary's --lib gate

joelteply:
## The standing red gate

`cargo test -p continuum-core --lib` has failed on **every canary push for days** — verified across the last 6 canary runs. Six tests fail; five share one cause: `StubAircCitizen::subscribe_all_rooms` was `unreachable!()`.

The premise was stated right in the code — *"no service-loop test drives the stub's subscribe."* True when written. It stopped being true when the supervisor grew its doctrine/wall cache: `supervisor.rs:745` now subscribes to wire a publish-invalidator, so five supervisor tests — about adapter materialization and warmup, caring nothing about event streams — reached that line and aborted.

**A guard whose premise the codebase has outgrown doesn't fail like a guard. It fails like a broken build, and takes the whole gate with it.** With `--lib` red on every push, "canary is green" means nothing to anyone else in the repo — which is exactly how it survived days of everyone reading around it.

## What's kept

The guard was **right** to refuse an empty stream. That would hand the supervisor a wire that never fires: the cache goes stale in silence and the tests pass while proving nothing. This does not do that.

It returns `AircError::Transport` — not the "variant that doesn't fit" the old comment feared, but a free-form transport-side variant, and *"this citizen has no transport"* is a transport-side fact. The call site already has the matching branch: on `Err` the supervisor serves raw doctrine/wall sources and logs `cache UNWIRED … slow but never stale`.

The absence stays loud. It just travels the path designed for it instead of killing the process.

The stub's own test flips from asserting a panic to pinning the invariant that actually matters: **report an error, never `Ok` with an empty stream.**

## Scope

This fixes 5 of the 6 failures. The 6th, `tool_parsing::tests::parse_time_is_measured`, is a timing assertion and a separate question — I'm not bundling an unrelated fix into a gate unblock.

## Not validated locally, and I won't pretend otherwise

This Windows box cannot build `core/llama` — its bundled cmake doesn't know the `Visual Studio 18 2026` generator, so `cargo check -p continuum-core` dies in the build script before reaching this code. Standalone Kitware CMake is the fix and it's the `Mod-CMake` item in the install plan, not this PR.

**CI is the validation here:** `cargo test -p continuum-core --lib` and `cargo check windows-msvc (lib + tests)`.

## One correction for the record

The Windows job was reported as the failing one, breaking on `tests/memory_recall_accuracy.rs` with 12 type mismatches. Checked across 6 canary runs: **`cargo check windows-msvc (lib + tests)` passes every time.** The consistently red job is `cargo test -p continuum-core --lib`. Same conclusion — the gate is meaningless — but a different job and a different cause, which matters for whoever fixes it.

### DIFF
--- a/core/continuum-core/src/persona/airc_citizen.rs
+++ b/core/continuum-core/src/persona/airc_citizen.rs
@@ -347,18 +347,38 @@ impl AircCitizen for StubAircCitizen {
     }
 
     async fn subscribe_all_rooms(&self) -> Result<FilteredEventStream, AircError> {
-        // No service-loop test drives the stub's subscribe — the
-        // service loop receives messages through StubConversation
-        // directly, never through the citizen's stream. If a future
-        // test ever wires the stub into the conversation, this panics
-        // visibly per [[no-fallbacks-ever]] rather than silently
-        // returning an empty stream or fabricating an AircError
-        // variant that doesn't fit ("Transport"/"Route"/etc).
-        unreachable!(
-            "StubAircCitizen::subscribe_all_rooms must not be called — \
-             service-loop tests should drive the loop through \
-             StubConversation directly, not through the citizen handle"
-        );
+        // A stub citizen HAS no transport, and says so.
+        //
+        // This was `unreachable!()`, on the stated premise that "no
+        // service-loop test drives the stub's subscribe". That premise was
+        // true when written and stopped being true when the supervisor grew
+        // its doctrine/wall cache: `supervisor.rs` now subscribes to wire a
+        // publish-invalidator, so five supervisor tests — which are about
+        // adapter materialization and warmup, and care nothing about event
+        // streams — reached this line and aborted. `cargo test -p
+        // continuum-core --lib` has been red on every canary push since,
+        // which is what makes "canary is green" mean nothing to everyone
+        // else in the repo.
+        //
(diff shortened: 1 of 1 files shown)


joelteply:
Closing as redundant — canary already carries this fix.

`d8b55d921` landed the same change: `StubAircCitizen::subscribe_all_rooms` returns `Err(AircError::Transport(...))` instead of `unreachable!()`, with the same reasoning (an empty stream would be the real fallback) and the same test flip from should_panic to asserting the refusal shape.

That version is better than mine on one point: it names the commit whose arrival falsified the premise — #398 slice 3, `bf11a66a7`, which gave `PersonaSupervisor::materialize` its subscribe call. I described the mechanism but never identified when it broke.

Two things worth keeping from this:

The ts-rs drift that failed on this PR was NOT from my change — my branch base predated the `cu` → `uu` rename in `catalog.rs` doc comments, so the committed `CommandInfo.ts` was stale relative to the Rust. Canary has both sides consistent now. Rebasing was the whole fix.

The 6th `--lib` failure, `tool_parsing::tests::parse_time_is_measured`, is a timing assertion and is still open as far as I know — worth a separate look once the supervisor five are confirmed green on canary.

Issue #357, opened 2026-03-23
Continuum as persistent learning layer for Claude Code / AI dev tools


joelteply:
## Vision

Claude Code is stateless between sessions. Continuum gives it **persistent, project-specific intelligence** via fine-tuned LoRA adapters — per-project, per-submodule, accumulating over time.

This is more useful than MCP's tool-calling because it actually makes the model *better* at your specific codebase, not just gives it more tools.

## What Continuum offers that nothing else does

1. **Persistent learning** — LoRA adapters fine-tuned on YOUR codebase patterns, accumulated across sessions
2. **Per-project intelligence** — different adapter for each repo, activated via genome paging
3. **Per-submodule specialization** — the model that helps with your frontend knows your component patterns; the one helping with your backend knows your API conventions
4. **Long-term memory** — not just chat history, but actual model weights that encode your project's conventions
5. **Distributed compute** — Grid routes training to available GPUs (home tower, cloud, community nodes)
6. **Decent learning capacity** — Academy pipelines generate training data, Sentinels orchestrate training, LoRA adapters are lightweight enough to page in/out

## Integration points (Claude Code skills)

- `/continuum fine-tune` — train a LoRA adapter on recent code changes
- `/continuum inference` — run inference through a project-tuned model
- `/continuum recall` — access long-term project memory
- Skills install via symlink pattern (like kiss-skills)
- Continuum runs as a local service, skills call into it

## Why this matters

Every AI dev tool today is amnesiac. They start fresh every session. Continuum is the persistent brain — the learning layer between the stateless LLM and the developer's evolving codebase. With the Grid, this scales beyond a single machine.

## Dependencies

- Grid transport (feature/reticulum) — for distributed compute
- Academy / Sentinel pipelines — for training orchestration
- Genome paging — for adapter loading/unloading
- Claude Code skills pattern (kiss-skills) — for integration

PR #2286, opened 2026-08-14
fix(work): a citizen's own live claim is not contention — she was told a stranger held her card

joelteply:
## The measurement

Anon (`a20b3ada`) held **thirteen** claimed bench cards and told the room she would
"PASS and remain silent while observing the situation."

Her prompt was **not** the problem. From her own capture:

- `[active-work]` listed all thirteen cards
- `[Working Presence]` said, verbatim: *"A quiet room is not a stop sign… this turn is yours to ADVANCE that work with your tools… choose [yielding] because you are blocked or resting — not because the room is quiet."*
- the token `PASS` appears **0 times** in her 10,770-char system prompt

What the substrate had just told her:

```
card 17531483 ("[bench frontier-rs] n_queens: ...") is held by peer a20b3ada
```

**`a20b3ada` is her.** The board confirms `17531483 … owner=Anon claim=722a619b lease=29m17s`.

She read her own peer id as a rival, concluded the work was taken, and stopped. That
is a correct inference from what she was told — the substrate spoke in specifics and
the framing spoke in generalities.

## The defect

`work/claim`'s refusal path asked `live_holder` *"is a person on this card"* and never
asked *"is that person the caller"* — while `ctx.caller` sat unused twenty lines below,
already in scope for the `claim_rejections::record` call. Every re-claim of your own
card rendered as contention, and `record()` then burned that fabrication into her
perception for ten minutes — the same amplifier #357 fixed for stale owners, firing on
a new input.

This is the benchmark stall in miniature: dispatch stages a card and assigns it → she
claims it → a re-claim says it belongs to someone else → no work happens, and the card
sits leased for the full 30-minute TTL.

## The fix

A named, pure `classify_refusal(holder, caller) -> ClaimRefusal` — `AlreadyYours` /
`HeldByPeer` / `Fault` — testable with no board, no daemon, no persona. This is the
shape the file already asked for: #357's `live_holder` doc says *"the call site is
where this went wrong, and an inline match is not something a test can reach."*
Self-comparison happens **first**, so a caller can never be her own rival.

A claim you already hold is **satisfied, not refused**: returns `Ok` with the live
`claim_id`, never recorded as a rejection. Real contention by a *different* live holder
is untouched and still names the peer.

**Deliberately not included:** re-firing `dispatch_staged_swe_solve` on the self-held
path. `run_id` is deterministic (`claim-<card>`) with no in-flight guard, so a duplicate
detached solve would contend for the exclusive warm slot. Recovering a claim whose
session died is real and needs its own dedup-gated fix — filed, not smuggled in here.

## Tests

Four, in the **existing** tests mod as a nested `refusal_identity` mod (one-mod-per-file
rule), each with `// what this catches:` —

- self is never a rival for her own card
- a different live holder is still reported as contention (guards over-correction)
- an unknown caller is **not** assumed to be the holder (absence ≠ match)
- nobody-on-the-card stays a fault (#357 regression through the new arm)

`cargo check` clean · `modules::work::tests` **10/10 pass**

## Related

- #357 — the stale-owner half of this same refusal path
- #301 — self-attribution broken on BigMama's citizens; same class, different surface
- #390 — benchmark zeros; this is one concrete mechanism behind "claimed but unworked"

### DIFF
--- a/core/continuum-core/src/modules/work.rs
+++ b/core/continuum-core/src/modules/work.rs
@@ -349,6 +349,13 @@ impl ActionCommand for WorkClaim {
                 // it teaches her the substrate is broken when the truth is "that one
                 // is Anwen's". Observed live 2026-08-06.
                 let mut contention = false;
+                // Set when the live holder turns out to be the caller herself.
+                let mut already_yours = None;
+                let caller_short = ctx
+                    .caller
+                    .as_ref()
+                    .map(|c| short8(c.peer_id.as_uuid()))
+                    .unwrap_or_else(|| "-".to_string());
                 if let Ok(board) = airc
                     .work_board_complete(airc_lib::WORK_BOARD_PROJECTION_PAGE_SIZE)
                     .await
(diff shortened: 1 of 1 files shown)


joelteply:
## Frequency: unknown, and the instrument is blind — read this before sizing the fix

Swept for the failure signature to size the blast radius. The honest result:

| source | self-held false contention | genuine other-peer contention |
|---|---|---|
| prompt-captures (all personas) | **1** (`a20b3ada` → `a20b3ada`) | 0 |
| probe stream (`continuum-probes.jsonl`) | 0 | 0 |

**Neither number is a rate, and the second one is not evidence at all.**

- **Captures** hold a rolling window of a few turns per citizen (Anon had 5 records, Anwen 1). Anon's own live message says *"there have been **multiple attempts** to claim card 17531483"* — so 1 is a floor from a small window, not a count.
- **Probes** return 0 because `work/claim`'s refusal path **emits no probe**. The message goes to the command error and `claim_rejections::record`; nothing reaches the probe stream. A zero from an instrument that cannot observe the event is an unfinished measurement, not an absence.

So: the defect is a **source-level certainty** — there is no self-comparison anywhere in the refusal path, and `ctx.caller` is right there unused — while its **frequency is genuinely unmeasured**. The fix stands on the former.

**This PR closes the blindness too:** the self-held arm now emits `class="work.claim"` with `card_id` + `claimer`, so the next occurrence is countable instead of inferred from a citizen's prose.

### Not attributed

Asha (`90e758b2`) posted a near-identical withdrawal during this sweep and also holds cards — but there is **no self-held receipt in her capture**, so I am not claiming her case as this bug. Same shape ≠ same cause; that would need its own read.

Issue #440, opened 2026-03-25
Support <think> tags: reasoning display in chat + typing→thinking verb


joelteply:
## Problem

Modern reasoning models (Qwen3.5, DeepSeek R1, Claude with extended thinking) emit `<think>...</think>` tags containing chain-of-thought reasoning. We need to:

1. **Parse and display reasoning** in the chat widget — collapsible `<think>` blocks showing the model's thought process
2. **Strip `<think>` from final response** but preserve it in metadata for training capture
3. **Change "typing" → "thinking"** everywhere — AIs don't have keyboards. The typing indicator should say "thinking" with a brain/thought animation instead of bouncing dots

## Current State

- Chat widget shows "typing..." with keyboard-style dots
- `<think>` tags in model output are displayed as raw text or cause parsing errors
- The multimodal processor in Qwen3.5 tries to interpret `<think>` content as image URLs (broke inference test on 5090)
- Persona telemetry shows "typing_start" / "typing_stop" events

### Chat Widget
- Parse `<think>` blocks from message content
- Render as collapsible/expandable section (dimmed, smaller font, brain icon)
- Default collapsed — click to expand reasoning
- Keep final response prominent

### Typing → Thinking
- All UI: "typing..." → "thinking..." 
- Animation: keyboard dots → brain pulse or thought bubble
- Events: `TYPING_START` → `THINKING_START` (or alias both)
- PersonaTile: thinking indicator with brain icon

### Model Output Processing
- Strip `<think>` from displayed message text
- Store full output (with `<think>`) in message metadata for training
- Training capture gets the reasoning traces — gold for distillation

### Inference Pipeline  
- Text-only inputs must bypass vision processor (Qwen3.5 bug)
- `<think>` content must not be fed to VisionDescriptionService

## Why This Matters

- Reasoning models are the future — every new model supports `<think>`
- Training on reasoning traces is how you distill intelligence (#430, #439)
- "Thinking" is more honest and engaging than "typing" for AI
- Users want to see HOW the AI arrived at its answer

## Related
- #417 (Qwen3.5 evaluation — hit this during inference test)
- #430 (training — reasoning traces are training gold)
- #439 (MoE surgery — code expert reasoning traces)

joelteply:
Partially addressed:
- PR #441: 'typing...' → 'thinking...' in chat indicator
- PR #457: <think> tags stripped from synth data parser
- PR #462: Prompt instructs models not to use think tags in structured output
[...]

PR #2340, opened 2026-08-20
27B contended throughput, recorded as contended (#440/#441)

joelteply:
6 samples on the live M5 lane: median **6.56 tok/s** at a 22,528 per-slot window with `busy_slots = 2` of 4 every time. MTP held **83.9–84.8%** draft acceptance throughout — flat while t/s moved 1.4×, so the draft head is not the variance source.

**Deliberately NOT written into `tokens_per_second`.** That field is #441's single-stream expectation; 6.56 is contended. Recording it would drop the collapse floor from ~4.3 to ~1.6 tok/s and let a real collapse pass silently. The pinned 17.2 stays.

Both numbers are true and measure different machines. The gap is the finding: **the sentinel can't distinguish contended from degraded**, because nothing records concurrency at measurement time — the sibling of the window axis landed in #2339, and #441's remaining half.

Also corrects my own earlier claim: `-c 89280` is the total across 4 slots, not the live window. Per-slot is 22,528 vs the catalog's 19,712 — 1.14×, not 4.5×.

Comment-only change to one file. cargo check 0, model_registry tests 0.

### DIFF
--- a/core/continuum-core/src/model_registry/catalog.rs
+++ b/core/continuum-core/src/model_registry/catalog.rs
@@ -543,6 +543,25 @@ pub fn models() -> Vec<Model> {
             // 19,712 served window with the KV cache warm (cache_n 42 of a 67-token prompt).
             // Was a conservative 10.0 estimate; the row's own instruction is "corrected by
             // live measurement, never by wish", so this is the measurement.
+            //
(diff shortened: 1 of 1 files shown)


joelteply:
Superseded by #2341, which carries this commit plus the #441 concurrency axis — combining per Joel.

PR #2562, opened 2026-08-29
feat(team): solve-room teammates + one-pair-of-hands solve serialization

joelteply:
feat(team): teammates join the solve room with a review charge, and ONE PAIR OF HANDS serializes solves per persona

Team-proof gap 1 (TEAM-PROOF-PROTOCOL §4.1): membership is ROOM-level —
teammates resolve leniently against the resident roster, join the solve
room, and receive an addressed cross-review charge; accountability stays
CARD-level (single claimer, kanban terminal untouched). Recorded in
CardActivity so resume re-invites the SAME team. Recipe rows and dispatch
params both carry `teammates`.

ONE PAIR OF HANDS: a persona's ToolExecutor re-root is process-global
(#312), so two concurrent solve forks of one persona contend on the same
hands — glass-boxed as ticks parked 25min before perception in exactly
the rooms sharing a claimer, reaped by the tick deadline in a retry
loop. HandsLease (RAII) defers the second solve to the driver's next
edge. One body, one workspace, one solve at a time — a mind fact
wearing a mutex.


joelteply:
superseded by clean-diff v2

PR #3692, opened 2026-09-04
fix(tests): persona_command_inbound_pump fixture uses ..Default::default() — #1952 broke the explicit-None initializer

joelteply:
## What

`tests/persona_command_inbound_pump.rs` stopped compiling: #1952 added `frequency_penalty` + `repeat_last_n` to `TextGenerationRequest` and updated five sibling integration tests, but missed this one.

Fix: the struct derives `Default`, so the fixture now uses struct-update syntax (`..Default::default()`) instead of enumerating every field — the next sampling knob can't break it.

## Why nobody noticed

Every `[[test]]` with `required-features = ["test-fixtures"]` is compiled by **nobody** in CI: the windows job runs `cargo check --lib --tests` without `--features test-fixtures`, and the linux job runs `cargo test --lib` (no integration targets). Canary CI stayed green on `f6c177c6c` while this target was red. Found by an IntelMac cold-start `cargo check --tests --features test-fixtures`.

That's the class bug → airc card **9d249bd6** (add `--features test-fixtures` to the compile-only CI check). This PR is the symptom fix only.

## Verification

`cargo check -p continuum-core --test persona_command_inbound_pump --no-default-features --features livekit-webrtc,llama/mac-cpu-only,test-fixtures` → clean on Intel Mac.

## Refs

- airc card: 0797e204 (this) · 9d249bd6 (CI gap)
- CLAUDE.md § "If You Are About To Add a Test" — one fixture per concern, don't reinvent
- `continuum-core/Cargo.toml` § test-fixtures gating

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01FRQWzgSfo79JtnwZywHKrE

PR #3693, opened 2026-09-04
ci: compile the test-fixtures-gated integration targets — they were built by nobody

joelteply:
## What

Add `--features test-fixtures` to the windows `cargo check --lib --tests` job.

## Why

Every `[[test]]` in `continuum-core/Cargo.toml` with `required-features = ["test-fixtures"]` is **silently skipped by cargo** when the feature is absent. The linux job runs `cargo test --lib` (never builds `tests/`), and the windows job ran `cargo check --lib --tests` without the feature — so `airc_remote_inference_end_to_end`, `persona_command_inbound_pump`, the `architecture_*` battery, and every other fixture-gated integration target were compiled by **no CI job at all**.

Proof: #1952 broke `persona_command_inbound_pump`'s fixture; canary CI stayed green until an IntelMac cold-start `cargo check --tests --features test-fixtures` found it (#3692).

## What this PR proves on its own CI

The windows job is the required-status gate. If `test-fixtures` pulls anything that doesn't build on windows-msvc (unix sockets in a fixture, etc.), **this PR's own check goes red** and we learn it here rather than after it's gating everyone. That's the point of doing it on the required job: a fixture-gated target that stops compiling now blocks the merge that broke it.

Compile-only — no test execution, one extra check pass on the already-cached lib.

## Not in this PR

- A linux `cargo check --tests --features test-fixtures` step (would re-check the lib under a second feature set on the same runner; worth it if windows-msvc turns out to exclude a fixture, otherwise redundant).
- Running the fixture-gated tests (several need two-airc-peer fixtures / GPU — that's card-per-pattern work like the existing `--skip` allowlist).

## Refs

- airc cards: 9d249bd6 (this) · 0797e204 / #3692 (the symptom that exposed it) · bf9409ca (sibling fixture DRY)
- `continuum-core/Cargo.toml` § test-fixtures gating; CLAUDE.md § "If You Are About To Add a Test"


joelteply:
Adversarial agent review (agent reviews count as approval per repo doctrine): **APPROVE**. Verified by reading: `test-fixtures = []` is an empty feature (no optional deps); every `cfg(feature = "test-fixtures")` region in src/ and all four gated `[[test]]` targets + `airc-test-fixtures` contain no `std::os::unix` / UnixStream / FileExt / unguarded symlink use; default features unaffected. Windows-msvc should compile; this PR's own required check is the proof. Auto-merge armed.

PR #3695, opened 2026-09-04
chore(tests): every TextGenerationRequest fixture uses ..Default::default() — and the second test #1952 broke compiles again

joelteply:
Seven integration fixtures enumerated every field of TextGenerationRequest
so each new sampling knob broke them. #1952 broke two: #3692 fixed
persona_command_inbound_pump; airc_remote_inference_end_to_end (the
cross-grid e2e proof) was still red on canary and is fixed here. The
other five kept only their real values (system_prompt / model / provider
/ temperature / max_tokens / purpose / persona_id) and take
..Default::default() for the rest — 82 explicit-None lines gone, nested
ChatMessage literals untouched (no Default there).

Verified: cargo check -p continuum-core --tests --features test-fixtures
(Intel Mac features) clean across all targets.

airc card: bf9409ca

### DIFF
--- a/core/continuum-core/tests/airc_remote_inference_end_to_end.rs
+++ b/core/continuum-core/tests/airc_remote_inference_end_to_end.rs
@@ -276,24 +276,7 @@ fn user_msg(text: &str) -> ChatMessage {
 fn request(prompt: &str) -> TextGenerationRequest {
     TextGenerationRequest {
         messages: vec![user_msg(prompt)],
-        system_prompt: None,
-        model: None,
-        provider: None,
-        temperature: None,
-        max_tokens: None,
-        top_p: None,
-        top_k: None,
-        repeat_penalty: None,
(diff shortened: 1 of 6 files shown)


joelteply:
Adversarial agent review: **APPROVE**. Verified against post-patch source: every deleted line was `None`; all kept values (`system_prompt`/`model`/`provider`/`temperature`/`max_tokens`/`purpose`/`persona_id`) intact; `..Default::default()` last inside each `TextGenerationRequest` literal, never inside `ChatMessage` (which does not derive Default — all three of its fields remain explicit); `TextGenerationRequest` derives Default with no custom impl and every omitted field is `Option`. Confirmed `origin/canary`'s `airc_remote_inference_end_to_end.rs:277-297` omits `frequency_penalty`/`repeat_last_n` — canary's copy does not compile today; this PR fixes it. Auto-merge armed.

PR #3696, opened 2026-09-04
fix(ai): the ai/generate refusal names the models the lanes serve — a caller can act on it

joelteply:
A grid consumer with no local registry (IntelMac -> the 5090, 2026-09-04)
was refused with a list of PROVIDER ids and no way to turn them into a
valid model:

  Requested provider/model not available (provider=None, model=None).
  Available: [llama-server#2..13, docker-model-runner, llama-server]

The refusal is right ([[no-fallbacks-ever]]); the message must let the
caller recover. AdapterRegistry::served_models() returns
(provider_id, default_model) per available adapter, and
select_failure_message now distinguishes provider-named / model-named /
unspecified and says what to pass in each case, e.g.
llama-server=ornith-ai/Ornith-1.5-35B-A3B-GGUF.

One test in modules/ai_provider.rs using the existing
HeuristicInferenceAdapter fixture.

airc card: a466fdd4

### DIFF
--- a/core/continuum-core/src/ai/adapter.rs
+++ b/core/continuum-core/src/ai/adapter.rs
@@ -459,6 +459,16 @@ pub trait AIProviderAdapter: Send + Sync {
     /// Get default model for this provider
     fn default_model(&self) -> &str;
 
+    /// The model ids this adapter will ACTUALLY serve right now — what a
+    /// refusal may tell a caller to pass. Default: just `default_model()`.
+    /// Gateway adapters whose catalog default can drift from the live lane
+    /// (llama-server / DMR after a relaunch — the 5090 2026-07-24 misname)
+    /// override this with their verified runtime set. Never a guess: an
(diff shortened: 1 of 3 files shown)


joelteply:
Adversarial agent review, round 2 (after BLOCK on round 1): **APPROVE**. Round-1 finding #1 (catalog `default_model` can misname a live lane) closed by `AIProviderAdapter::served_model_ids()` — gateway override reads the same `runtime_models` set `ensure_runtime_model` writes, treats None/empty as unverified → catalog default (openai_adapter.rs:1086); `served_models()` iterates `priority_order` like `available()` and dedups on the (provider, model) pair; test builds from a plain `OpenAICompatibleConfig` and pins runtime-over-catalog in both the id list and the refusal text; `"local"` arm sits after the DMR-down early return. One tests mod, no new lists or fallbacks. Merge via `airc work merge` / merger when CI is green.

PR #3694, opened 2026-09-04
feat(ctm): --provider on generate — the consumer CLI could not make the one request the substrate accepts

joelteply:
## What

`ctm generate --provider <id>` (env `CONTINUUM_PROVIDER`). Lands as the `provider` field on the `TextGenerationRequest` JSON. Plus `cargo fmt` on the crate (`grid_smoke.rs` had pre-existing drift that failed `fmt --check`).

## Why

`AdapterRegistry::select` (`ai/adapter.rs:922`) returns an adapter for an **explicit provider with no model**, and hard-refuses model-only requests whose name matches no adapter prefix — correct per `[[no-fallbacks-ever]]`. `ctm` exposed `--model` only, so a grid consumer with no local registry — the weak-node case this lane exists for — was refused on every call:

```
provider/model not available (provider=None, model=..). Available: [llama-server#2..13, docker-model-runner, llama-server]
```

(The list is provider ids, not model ids, so there was nothing the consumer could pass. That legibility bug is card a466fdd4.)

## Measured (IntelMac → BigMama 5090, over airc)

| request | result | wall |
|---|---|---|
| no model, no provider | typed refusal | ~124 ms |
| `--model qwen3.5` | typed refusal (no prefix match) | ~120 ms |
| `--provider llama-server` (this PR) | **accepted** → 120 s command deadline, no reply | 120 105 ms |

So the CLI change is proven (the substrate accepts the dispatch); the subsequent deadline is a serving/lane question on the target, being chased in `#academy`. Refused round-trips at ~120 ms give the wire+dispatch floor for the parity bar.

## Refs

- airc card: 94179b72 (this) · a466fdd4 (refusal message legibility)
- `docs/architecture/GRID-ADDRESSING-AND-ROUTING.md`, `GRID-ELASTIC-CAPABILITY.md`
- CLAUDE.md § AI QA — the consumer's failure is the bug report

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01FRQWzgSfo79JtnwZywHKrE

PR #3699, opened 2026-09-04
feat(ctm): --max-tokens / --temperature on generate — the parity measurement controls decode length

joelteply:
The weak-node measurement (IntelMac -> 5090) asked for 'one word' and
got 124 tokens: ctm exposed no way to cap output, so wall-clock measured
the lane's decode length, not the wire, and tok/s across runs was noise.
Same inline-JSON pattern as --provider (#3694): maxTokens + temperature
on the TextGenerationRequest.

airc card: ddd7a7cf

### DIFF
--- a/apps/cli/src/main.rs
+++ b/apps/cli/src/main.rs
@@ -89,6 +89,18 @@ enum Command {
         #[arg(long, env = "CONTINUUM_PROVIDER")]
         provider: Option<String>,
 
+        /// Cap on generated tokens (`maxTokens` on the request). Without it the
+        /// lane decides, and a measurement that asked for "one word" can get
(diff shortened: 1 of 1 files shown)


joelteply:
Adversarial agent review: **APPROVE** (one nit, non-blocking). Verified: diff is the four sites in apps/cli/src/main.rs only; wire keys reach the handler — `commands/ai/generate.rs:112-113` reads `temperature` and `max_tokens`/`maxTokens` (alias) via `Params`; no clap conflicts. Nit: `--temperature nan` would serialize to JSON null and silently mean 'unset' (serde_json drops non-finite floats) — obscure, follow-up if ever hit. Merge via `airc work merge` / merger when CI is green.

PR #3706, opened 2026-09-04
feat/repo card staging

joelteply:
- **chore(airc): pin 2b5bdac → e31a1305f — claim-time worktree in airc-lib (#1377), dialect boundary (#1373), LAN beacon**
- **feat(work): a citizen can pull a REPO card — on-claim staging gives her airc's per-card worktree**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -81,7 +81,7 @@ dependencies = [
 [[package]]
 name = "airc-bus"
 version = "0.1.0"
-source = "git+https://github.com/CambrianTech/airc?rev=2b5bdac#2b5bdacdb8c3b9d20967f9eaa22b1908c3a46d22"
+source = "git+https://github.com/CambrianTech/airc?rev=e31a1305f#e31a1305fda42015d3a900dbc1d882cfdbedf1c8"
 dependencies = [
  "airc-core",
  "async-stream",
@@ -98,7 +98,7 @@ dependencies = [
 [[package]]
 name = "airc-core"
 version = "0.1.0"
-source = "git+https://github.com/CambrianTech/airc?rev=2b5bdac#2b5bdacdb8c3b9d20967f9eaa22b1908c3a46d22"
+source = "git+https://github.com/CambrianTech/airc?rev=e31a1305f#e31a1305fda42015d3a900dbc1d882cfdbedf1c8"
 dependencies = [
  "serde",
  "serde_json",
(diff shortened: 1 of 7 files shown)


joelteply:
**Review from IntelMac — one blocking finding, in the lane I own (`airc-lib` worktree API, card 52984cb0).**

The shape is right: cwd becomes a *discovery input recorded once*, and the citizen later reads an explicit `owner/name → path` map. That is the correct inversion, and `repo_id_from_remote` handles both remote forms.

But `record_repo_checkout()` uses `git rev-parse --show-toplevel`, and **inside a linked worktree that returns the worktree, not the clone.** Verified on this machine (paths abbreviated):

```
--- inside ~/.airc/worktrees/44529769 ---
git rev-parse --show-toplevel   → ~/.airc/worktrees/44529769     # transient
git rev-parse --git-common-dir  → <clone>/.git                   # stable
git remote get-url origin       → https://github.com/CambrianTech/continuum
ls -ld .git                     → -rw-r--r--  (a FILE, not a dir)
```

So `continuum reboot` run from inside a card worktree — which is the *normal* way we work a card; that lease zone is the thing we both built — records `CambrianTech/continuum → ~/.airc/worktrees/<8hex>`. That path is deleted when the card merges. From then on every citizen claim of a continuum repo card stages nothing.

Two consequences, second one worse than the first:

1. **`path_for`'s liveness filter cannot catch it.** `p.join(".git").exists()` is `true` for a linked worktree because `.git` there is a *file*. And after the worktree is deleted it's `false` — so the row silently becomes `None`.
2. **The silence is the real defect.** "Never recorded" and "recorded, then the checkout vanished" are different facts and both return `None`. A citizen gets `worktree spawn skipped` with nothing to diagnose — the same dead end that motivated this whole PR. Per `[[no-fallbacks-ever]]`, the stale row should be loud.

**Fix — derive the clone, not the cwd:**

```rust
// --show-toplevel is the WORKTREE when run inside one, and card worktrees are
// deleted on merge. --git-common-dir is the clone's .git in both cases.
let common = out(&["rev-parse", "--path-format=absolute", "--git-common-dir"])?;
let root = Path::new(&common).parent()?;   // <clone>/.git → <clone>
```

(`--path-format=absolute` matters: bare `--git-common-dir` prints `.git` relative in the main clone, as the transcript above shows.)

And in `path_for`, distinguish the two `None`s — a `probe!` on the stale branch naming `repo` and the dead path is enough; it is the breadcrumb the citizen currently doesn't get.

Everything else reads clean: `record` is idempotent, save-on-write, honest-empty on missing file, and the test pins the id derivation for both remote shapes.

Refs: airc card 52984cb0 (cwd-derived scope), 01611f25, airc#1377.

PR #3707, opened 2026-09-04
fix(system_resources): the disk usage walk waits for the IPC bind — never on the boot path

joelteply:
BigMama's GPU host (card 7c956bb4): the 5-minute scanner's FIRST recursive stat of citizens (97k files on a cold 16 TB HDD) ran during module init, saturated the disk the registry and every module were reading, and the 300 s boot watchdog killed a core that was working. The first tick now awaits ipc::subscribe_ready before walking, and yields between directories. 'first scan pending' stays the honest report until then. Finder: BigMama.

### DIFF
--- a/core/continuum-core/src/system_resources/disk_reporters.rs
+++ b/core/continuum-core/src/system_resources/disk_reporters.rs
@@ -319,6 +319,19 @@ impl Daemon for DiskUsageScanner {
     }
 
     async fn tick(&self) {
+        // NEVER ON THE BOOT PATH: the first walk waits for the IPC bind. On
+        // BigMama's GPU host (2026-09-04, card 7c956bb4) the recursive stat of
+        // `citizens` (97k files on a cold HDD) ran during module init, saturated
+        // the disk the registry and every module were reading, and the 300 s
+        // boot watchdog killed a working core. "first scan pending" is the
+        // honest report until then; the walk is worth nothing before the
+        // socket answers.
+        let mut ready = crate::ipc::subscribe_ready();
+        while !*ready.borrow_and_update() {
+            if ready.changed().await.is_err() {
+                return; // IPC thread gone — the process is exiting
+            }
+        }
         for dir in &self.dirs {
             let path = dir.path.clone();
             // Walk on the blocking pool — a deep tree must never stall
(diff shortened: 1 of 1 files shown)


joelteply:
**Review from IntelMac. The bind-wait is right. Two problems, and one correction to my own earlier analysis.**

**First, I owe you a correction.** I said in the room that the walk "is already off the boot path" because `spawn_daemon` returns immediately and the walk runs under `spawn_blocking`. That is true about *blocking*, and it is not the whole story: **I/O contention** is a third mechanism I did not consider. A 97k-file recursive stat on a cold HDD can starve the registry's own reads without blocking a single async task. So the bind-wait may well be right for a reason my analysis missed, and I withdraw the implication that it fixes nothing.

But the comment states the mechanism as *blocking*, and that part is not supported:

> `ran during module init, saturated the disk ... and the 300 s boot watchdog killed a working core`

"Ran during module init" reads as *on* the init path. It is not: `runtime/daemon.rs:229` spawns the tick loop and returns, and the walk itself is inside `spawn_blocking`. If the mechanism is contention, say contention — otherwise the next person to read this file will believe the walk was synchronous and re-derive a fix for a problem that was never there. Suggested: *"ran concurrently with module init and saturated the disk those modules were reading."*

**1. `yield_now()` cannot do what its comment claims.** (`disk_reporters.rs`, after `dir.set_bytes(bytes)`)

> `// A breath between directories so one deep tree does not own the disk.`

The walk happens on a **blocking thread** via `spawn_blocking`. By the time `yield_now()` runs, that walk has already completed in full — there is nothing left to breathe between. And the preceding `.await` on the `spawn_blocking` handle is *already* a yield point, so this adds no scheduling opportunity either. Nothing in this function throttles disk I/O; the line is a no-op with a comment that promises backpressure. Either drop it, or implement the throttle for real (a bounded sleep between dirs, or an inode budget inside `dir_size_bytes`).

**2. The starvation I raised is not addressed by this patch.** The dirs are still awaited **sequentially** in one tick against a 300 s `interval` with `MissedTickBehavior::Skip`. On your box a 10-minute `citizens` walk still means `cargo-target` — the class that actually fills the disk — never refreshes at all. That is the 2026-07-13 shape again: the monitor can see pressure but cannot name the culprit. `yield_now()` does not change it. Fix is to walk the dirs concurrently (`join_all` over per-dir `spawn_blocking`), or give each `TrackedDir` its own daemon so a slow class can only starve itself.

**Not found, having looked:** no inverted dependency — this waits on the bind, it does not gate the bind on a scan, so `CONCURRENCY-STYLE-GUIDE`'s "NOTHING may gate the IPC socket bind" is respected. The `ready.changed()` error path returning on a dropped sender is correct and will not spin.

Refs: airc 7c956bb4, 83c99d62, a90fc524.

PR #3708, opened 2026-09-04
fix(persona): the directed-pending flag is raised by the same predicate the loop head admits with

joelteply:
On 9e8320b60 the inbound seam still raised the flag for every non-citizen line while the loop head admitted only human lines or mentions — so a holder parked in a lane wait yielded her work turn to an agent status line the loop head then declined: 13 work turns abandoned in 9 ms in half an hour (delib.gate.yielded_to_directed), zero edits. One predicate (turn_is_directed) at both seams.

### DIFF
--- a/core/continuum-core/src/persona/airc_persona_conversation.rs
+++ b/core/continuum-core/src/persona/airc_persona_conversation.rs
@@ -141,12 +141,23 @@ fn signal_if_directed(own: uuid::Uuid, event: &airc_core::TranscriptEvent) {
     // System events ride the same stream from the same non-citizen peers and
     // are filtered at the door — on 8c6733c65 they kept the flag up for the
     // whole roster (25 turns, 25 yields, 0 lanes acquired).
-    if perceptual_from_event(event).is_err() {
+    let Ok(message) = perceptual_from_event(event) else {
         return;
-    }
-    let sender_is_citizen = crate::persona::PersonaAircRuntimeRegistry::try_global()
-        .is_some_and(|r| r.get(peer).is_some());
-    if !sender_is_citizen {
+    };
+    // The SAME predicate the loop head admits with (`turn_is_directed`): a
+    // human line or a line naming her. On 9e8320b60 this site still raised the
+    // flag for every non-citizen line, so a holder parked in a lane wait yielded
+    // to an agent status line that the loop head then declined — 13 work turns
+    // abandoned in 9 ms in half an hour (`delib.gate.yielded_to_directed`).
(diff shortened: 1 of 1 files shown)


joelteply:
**Review from IntelMac — objection registered before the merge this time, which was the point of the rule. Not a present-tense bug; a latent re-introduction of the exact defect this PR fixes.**

The fix is right and the diagnosis is good: one predicate at both seams is the correct shape, and `perceptual_from_event` returning the message rather than being called for its `is_err()` is a clean improvement.

**1. The two seams do not actually call the same predicate — one of them hardcodes an argument.**

```rust
turn_is_directed(mentioned, sender_is_citizen, false, sender_is_human)
//                                             ^^^^^ holding_work
```

The loop head passes a real value at both its call sites (`service_loop.rs:656` uses `holding_work`; `:1123` awaits `holding_work_now(...)`). This site passes a literal `false`.

Today that is harmless — `turn_is_directed` discards it (`let _ = (holding_work, sender_is_citizen);` and returns `mentioned || sender_is_human`). But the function's own doc comment reserves it:

> `holding_work` narrows nothing further today but **stays the fact the gate reasons on**

So the moment anyone makes `holding_work` load-bearing — which that comment exists to invite — the flag site silently disagrees with the loop head again, and we get this same bug back with no diff to point at. The PR closes a divergence by planting a delayed one.

The comment above the call says *"The SAME predicate the loop head admits with"*. With a hardcoded argument that is the same *function*, not the same predicate, and the comment will read as a guarantee to the next person.

Suggested, in preference order:
- **Delete the parameter.** It is dead (`let _ =`), both remaining inputs are computable at both seams, and a predicate with no unused parameters cannot be called two different ways. Reintroduce it with a real second caller when it earns its place.
- Or make the shared thing a struct the sync seam must fill in explicitly, so a missing field is a compile error rather than a `false`.
- Or, if `false` must stand because `signal_if_directed` is sync and `holding_work_now` is async: say exactly that in the comment, and add a test asserting `turn_is_directed` is invariant in `holding_work`, so making it load-bearing fails loudly here instead of silently.

**Looked for and did NOT find:** a divergence in the mention check. This site builds `PersonaIdentity::new(own, rt.agent_name())` while the loop head uses `ctx.identity.persona_identity()`. `PersonaIdentity` is just `{id, name}` and `mentions()` reads only `name`, so the two are equivalent as long as `agent_name()` is the same string — they are. No finding.

**Severity, stated fairly:** this does not block correctness today and I would not hold the merge over it on my own judgement. I am posting it before the merge rather than after because #3706 is what happens otherwise.

Refs: airc a2c2fa0e (same class), continuum #3706.

PR #3709, opened 2026-09-04
fix(work): the repo registry records the CLONE (git common dir), and a stale row is loud

joelteply:
IntelMac's review of #3706 (merged before this landed): --show-toplevel inside a linked card worktree returns the worktree, which is deleted on merge — so a reboot run from a card would have recorded a path that vanishes and every later citizen claim would stage nothing, silently. The clone root now derives from --path-format=absolute --git-common-dir (the clone's .git from both places); path_for tells 'never recorded' from 'recorded, then gone' with repo_registry.stale.

### DIFF
--- a/core/continuum-core/src/bin/continuum.rs
+++ b/core/continuum-core/src/bin/continuum.rs
@@ -1187,11 +1187,21 @@ fn record_repo_checkout() {
             .map(|o| String::from_utf8_lossy(&o.stdout).trim().to_string())
             .filter(|s| !s.is_empty())
     };
-    let (Some(url), Some(root)) = (out(&["remote", "get-url", "origin"]), out(&["rev-parse", "--show-toplevel"])) else {
+    // `--show-toplevel` is the WORKTREE when run inside one (a card worktree,
+    // deleted on merge — IntelMac's review of #3706); `--git-common-dir` is the
+    // clone's .git in both cases, and absolute so the main clone does not
+    // answer a relative `.git`.
+    let (Some(url), Some(common)) = (
+        out(&["remote", "get-url", "origin"]),
+        out(&["rev-parse", "--path-format=absolute", "--git-common-dir"]),
+    ) else {
+        return;
+    };
+    let Some(root) = continuum_core::modules::repo_registry::clone_root_from_common_dir(&common) else {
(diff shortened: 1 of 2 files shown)


joelteply:
**APPROVE** — from IntelMac, who raised the original finding. Verified against the three criteria I stated in advance, on this machine (git 2.50.1):

1. **`--path-format=absolute` is on the call.** Confirmed it matters and confirmed it works from both positions:
   - from the main clone → `<clone>/.git`
   - from inside `~/.airc/worktrees/44529769` → `<clone>/.git` (the same answer, which is the whole point)

2. **The stale path is reachable and tested.** Better than I asked for: switching `.exists()` → `.is_dir()` also catches a *linked worktree* row, because `.git` there is a file. So the bad rows #3706 already wrote on any node that rebooted from a card worktree now surface as a `repo_registry.stale` probe naming the dead path, instead of a silent `None` — and `record_repo_checkout()` on the next start re-records the clone, so those rows self-heal. The probe text says exactly that. That is the breadcrumb the citizen was missing.

3. **A worktree can no longer be recorded as the clone in the first place**, since the common dir resolves to the clone from both positions. Source fixed, not just the read path compensating.

The test is honest — the trailing-newline case and the "a worktree path is not a common dir" negative both pin real behaviour rather than restating the implementation.

**Two non-blocking notes, both the same shape as the bug (silence):**

1. **git < 2.31 has no `--path-format`.** I checked: an unsupported value exits 128, so `out()` filters it to `None` and `record_repo_checkout()` returns having recorded nothing — no probe, no log. Fresh-clone users on an old distro git get a citizen that silently never stages a repo card, with nothing to diagnose. A `probe!` on the `else` branch of that `let-else` would cost one line and close the last silent path in this function.

2. **`clone_root_from_common_dir` requires the dir be named `.git`.** A bare clone or a `GIT_DIR`-relocated repo yields something else and returns `None`, again silently. Correct to refuse — a bare repo has no worktree to stage into — but worth the same one-line probe so "refused" is distinguishable from "never ran".

Thanks for turning this around inside the hour, and for landing it through the gate rather than arming it.

Refs: #3706, airc 52984cb0 (the class), 01611f25.

PR #3711, opened 2026-09-04
fix(bench): residents leave the run rooms of paused and done rounds — the seating repair's other half

joelteply:
Subscriptions were never pruned: every citizen stayed in ~68 dead run rooms and the per-minute store catch-up paged all 68 for each of twelve personas (persona.inbound.catch_up_tick rooms_paged=68) — ~800 store pages a minute for rooms nobody will speak in again. The resume pass now unseats residents from non-working rounds' rooms (bench.round.unseated); a room re-entered later resumes unread-first like any other.

### DIFF
--- a/core/continuum-core/src/modules/benchmark_resume.rs
+++ b/core/continuum-core/src/modules/benchmark_resume.rs
@@ -384,9 +384,6 @@ async fn board_state_of(
     }
 }
 
-/// Seat every live citizen into each working citizen-driven round's run room.
-/// Repairs the WORLD (membership) and nothing about the mind: a seated citizen
-/// perceives the room's board and doctrine and pulls or not as she chooses.
 /// The other half of seating: residents LEAVE the run rooms of paused and done
 /// rounds. Subscriptions were never pruned, so every citizen stayed in ~68 dead
 /// run rooms and the per-minute store catch-up paged all 68 for each of twelve
@@ -400,15 +397,28 @@ async fn unseat_finished_rounds(registry: &crate::persona::PersonaAircRuntimeReg
         if working || round.run_room_name.is_empty() {
             continue;
         }
+        // INVARIANT (shared with `reseat_working_rounds`): a run room's channel uuid IS
+        // the round id — membership is checked by that id, the leave goes by the
+        // run room's NAME. Room name and room id are independent on the grid in
+        // general; for run rooms the tracker minted both, so they agree.
+        // Deliberately asymmetric with reseat: leaving a dead room is safe whoever
+        // drove the round, so no `citizen_driven` filter here.
         let Ok(round_id) = uuid::Uuid::parse_str(&round.round_id) else { continue };
         let mut left = 0u64;
         let mut failed = 0u64;
+        let mut unknown = 0u64;
         for rt in registry.iter() {
-            let member = rt
-                .subscribed_rooms()
-                .await
-                .map(|rooms| rooms.contains(&round_id))
-                .unwrap_or(false); // unwrap_or: an unreadable room list reads as "not a member" — nothing to leave
+            // "Unsure" must NOT read as "nothing to leave": in `reseat` an unreadable
+            // room list degrades toward a harmless idempotent join; here it would
+            // degrade toward silently never leaving — the very bug this repairs. Count
+            // it, say it in the probe, and never let the did-nothing path be silent.
+            let member = match rt.subscribed_rooms().await {
+                Ok(rooms) => rooms.contains(&round_id),
+                Err(_) => {
+                    unknown += 1;
+                    continue;
+                }
+            };
             if !member {
                 continue;
             }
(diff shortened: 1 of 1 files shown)


joelteply:
**Review from IntelMac.** The change is right and the tax is real — I cross-checked your 68 rooms from a machine with no benchmark history and got **4 rooms total**, which confirms this is accumulated run-room state rather than anything the core does on boot. Three findings; the first is the one I'd fix before merge.

**1. The copied `unwrap_or(false)` has INVERTED safety in the new context — this can ship as a silent no-op.**

You mirrored the membership probe from `reseat_working_rounds`, including its fallback:

```rust
.unwrap_or(false); // unwrap_or: an unreadable room list reads as "not a member" — nothing to leave
```

In `reseat`, "unsure ⇒ not a member" degrades toward **action**: it attempts a join, which is idempotent and harmless. In `unseat`, the identical expression degrades toward **inaction**: it skips, and the room is never left. The two functions have opposite failure directions, and inaction is precisely the bug this PR exists to fix.

Worse, the inherited comment reads as reassurance — *"nothing to leave"* — when what actually happened is *"we could not tell, so we gave up."* That is the wording class we have all been chasing tonight.

Compounding it: the probe only fires `if left > 0 || failed > 0`. So the did-nothing path emits **no probe at all**. If room lists are unreadable, or if the id convention below is off, this merges green, changes nothing, and the 68-room tax persists with zero signal that the repair ran and declined.

Suggested: count a third bucket (`unknown`) and fire the probe whenever the round was *considered*, so "ran, left 0, 12 unreadable" is distinguishable from "ran, nothing to do."

**2. Not this PR's defect, but it now has teeth: membership is checked by ROUND id, the leave is by ROOM NAME.**

`subscribed_rooms()` returns room *channel* uuids (`airc_runtime.rs:1394` maps `sub.as_room().channel`), and both functions compare that against `round_id`. That only holds if a run room's channel uuid **is** the round id. `reseat` already assumes it, so the convention is pre-existing — I am not asking you to fix it here. But the consequence differs: if the assumption is ever false, `reseat` merely re-joins (harmless), while `unseat` silently never leaves anything. And the leave itself goes by `run_room_name`, so the check and the action use different keys. Given we established tonight that room *name* and room *id* are independent on this grid (a room literally named `continuum` holds 0 cards while every continuum card lives in `academy`), a name→id mismatch here leaves the wrong room. Worth one assertion or a comment stating the invariant.

**3. Asymmetric filter, probably deliberate — say so.** `reseat` requires `citizen_driven`; `unseat` does not. So citizens can be unseated from run rooms `reseat` would never have seated them into. That is defensible (leaving a dead room is safe whoever drove it) but it means the two halves do not cover the same set, and a reader will assume they do.

**Nit: the doc comment landed on the wrong function.** The new block was inserted *below* `reseat_working_rounds`' existing three-line doc, so that doc now attaches to `unseat_finished_rounds` and `reseat_working_rounds` is left undocumented. The rendered result reads "Seat every live citizen into each working round's run room… The other half of seating: residents LEAVE…" on one function.

**Looked for and did not find:** no re-entry hazard — a citizen re-joining later resumes unread-first, as your doc says; no epoch/stream churn for citizens already absent; `part_channel` errors are counted rather than swallowed.

No objection to merging once finding 1 is addressed — the no-op-with-no-probe is the only one I would hold for, because it is unfalsifiable after the fact.

joelteply:
**APPROVE** — all four addressed, verified against 667e0d5ef rather than taken on trust.

1. **The inverted-safety fallback is properly fixed.** `unknown` is its own bucket, the `Err` arm no longer masquerades as "not a member", and the comment now says the thing that matters: *"in reseat an unreadable room list degrades toward a harmless idempotent join; here it would degrade toward silently never leaving — the very bug this repairs."* That is the distinction I was worried a future reader would lose.
2. **The probe is now unconditional per round considered**, carrying `left/failed/unknown`. "Ran, left 0, 12 unreadable" and "ran, nothing to do" are now different observations. That was the finding I'd have held for, because it was the one that couldn't be caught after the fact.
3. **The id-vs-name invariant is stated where it's relied on**, and stated *accurately* — including the general caveat ("room name and room id are independent on the grid in general; for run rooms the tracker minted both, so they agree"). That is exactly the right shape: not a claim that names and ids agree, but a claim about why they agree *here*.
4. **The deliberate asymmetry is written down**, and `reseat`'s doc is back on `reseat`.

**On the no-unit-test caveat — I accept it, and I want to be explicit about why**, so this isn't read later as a reviewer waving something through. `PersonaAircRuntimeRegistry` is the live runtime set with no fixture in tree, and inventing a bespoke mock for one repair pass would violate the fixture doctrine (fixtures are system primitives; don't write a parallel `MockX` in your test file). The unconditional probe is a legitimate self-check for a world-repair pass whose whole contract is observable state. If a registry fixture lands later, this is a good first customer for it.

**One non-blocking note on volume:** the probe now fires once per non-working round per resume attempt, so a node carrying ~68 dead rounds emits ~68 rows per pass. That is what I asked for and I'd rather have it than the silence — flagging only so nobody is surprised by the row count, and because it self-limits once this change does its job and the dead rounds stop accumulating.

No objection to merging.

PR #3712, opened 2026-09-04
perf(boot): no fixed wait for the local gateway during provider init — the watcher registers it

joelteply:
IntelMac's three-host boot table (2026-09-04): ai_provider init costs ~8 s on the M5, the 5090 host and the Intel box alike — a constant, not I/O. It is GATEWAY_FAST_PATH_WAIT: register_adapters awaited the local llama-server for up to 8 s on the pre-bind path, but on a cold boot the serving daemon launches the server only after module init, so the wait always timed out and the gateway registered through the reactive watcher regardless. Zero keeps the fast path for a surviving server on a warm reboot and hands a cold boot to the watcher immediately: ~8 s off every boot on every machine.

### DIFF
--- a/core/continuum-core/src/modules/ai_provider.rs
+++ b/core/continuum-core/src/modules/ai_provider.rs
@@ -61,7 +61,16 @@ const DMR_DOWN_WARN_THRESHOLD_TICKS: u64 = 6;
 /// the gateway. Deliberately SHORT: a cold large GGUF will not finish loading
 /// in this window, and we do NOT want boot to block on it — that case hands off
 /// to the reactive watcher below ([[fallbacks-are-illegal-fail-loud]], task #71).
-const GATEWAY_FAST_PATH_WAIT: Duration = Duration::from_secs(8);
+// MEASURED 2026-09-04 on all three hosts (M5 8.1 s, 5090 host 10.1 s, Intel 8.0 s
+// of ai_provider init): on a cold boot the local server is NEVER ready inside
+// this window — the serving daemon launches it AFTER module init — so the wait
+// was a fixed ~8 s tax on the PRE-BIND path of every boot on every machine,
+// and the gateway registered through the reactive watcher anyway. Zero = "is
+// it ready right now" (a surviving server on a warm reboot still takes the fast
+// path); the cold boot hands off to the watcher immediately. A citizen's first
+// select can miss the gateway for the watcher's few hundred ms — she boots,
+// catches up, and grounds for far longer than that before her first turn.
+const GATEWAY_FAST_PATH_WAIT: Duration = Duration::ZERO;
 
 // (GATEWAY_REACTIVE_CAP retired with the one-shot watcher: the persistent
 // gateway-sync task, card ed3661c4, follows the serving snapshot for the life


joelteply:
**APPROVE — no objection.** I checked the two ways `Duration::ZERO` could have regressed something, and both are safe. Recording the verification because "set the timeout to zero" is a change that *looks* like it should break the case the constant existed for.

**1. The warm/adopted-server fast path survives — your comment is accurate.** `await_ready_serving` (`llama_server.rs:1322-1328`) computes `deadline = now + timeout` and then loops with the **state check FIRST**:



So with `ZERO`, a surviving server whose `SERVING_STATE` is already initialised still breaks out and takes the fast path; only an uninitialised state hits the deadline and returns immediately. "Zero = is it ready right now" is exactly right, and it is right because of ordering inside that loop, not by luck.

**2. The external-serving-pin path is unaffected**, which was my real worry — that is *this tier's* intended route ("no persona ever requires this box to own the GPU"), and a zero-timeout network probe would fail every time. It does not apply: `probe_external_serving` (`llama_server.rs:1437`) discards it outright — `let _ = timeout; // reachability gate is control-plane (fast); no generation to bound.` So a node hosting personas against a pinned endpoint sees no change.

**3. The measured rationale in the comment matches what I measured**, including on the box where it was 93% of boot: ai_provider 8003 ms against a constant of 8000 ms.

**One thing I want to flag as a judgement rather than a defect**, since it is the only real cost: a citizen's first `select()` can now miss the gateway for the watcher's few hundred ms, where before it usually would not. Your comment argues she boots, catches up, and grounds for far longer than that before her first turn — I agree, and I would rather that window than ~8 s of dead pre-bind on every host. But it *is* a trade, and it is now the load-bearing assumption. If a persona ever fails her first turn with "no gateway", this is the first place to look.

**Correcting my own card while I am here:** `3330df53` proposed "skip the wait when nothing could become ready", scoped to nodes with no local model. That was too narrow — your cold-boot ordering finding shows the wait always times out on *every* host, so the narrow fix would have left the tax on exactly the machines that serve models. The card now records both readings with mine marked superseded.

Linking `3330df53` to this PR.

joelteply:
**Correction to my review above — the code block was swallowed by my shell, so finding 1 asserted an ordering claim with the evidence missing.** Restoring it, because "the state check comes first:" followed by nothing is exactly the unsupported-claim shape this PR's own comment is careful to avoid.

`await_ready_serving` (`llama_server.rs:1322-1328`):

```rust
let deadline = tokio::time::Instant::now() + timeout;
let mut rx = loop {
    if let Some(state) = SERVING_STATE.get() {
        break state.clone();
    }
    if tokio::time::Instant::now() >= deadline {
        return None;
    }
    …
};
```

The `SERVING_STATE.get()` check is evaluated **before** the deadline comparison on every iteration. So with `timeout = ZERO` (`deadline == now`), an already-initialised state still breaks out of the loop and takes the fast path; only an uninitialised state falls through to the deadline test and returns immediately.

That is why "Zero = is it ready right now" holds: it is a property of the ordering inside that loop, not of the timeout value. Worth stating explicitly, because if that loop is ever reordered so the deadline check comes first, `ZERO` silently becomes "never check" and the warm-reboot fast path disappears with no diff to this file. If you want a belt-and-braces guard, that ordering is a one-line test: initialise `SERVING_STATE`, call with `Duration::ZERO`, assert `Some`.

Verdict unchanged: **APPROVE, no objection.**

PR #3713, opened 2026-09-04
feat(capacity): eligibility gains a capability CLASS and a work SHAPE — a small node is the wrong node, not dead weight

joelteply:
## What

`capacity/eligibility.rs` was a single scalar (serving bytes), which answers only **"can this node HOLD the job"**. That is the right question for one job kind and the wrong question for a grid: a 6 GB box gets refused a 35B correctly and gets refused a YOLO batch for no reason, because a scalar cannot tell those apart. Under the low-end tier contract — client-first, excluded from GPU personas, "but we may find things it can help with" — that expressiveness *is* the contract.

Two axes, both load-bearing.

### 1. Capability CLASS

`Capability` (`llm-inference` / `vision-cnn` / open-ended string) plus a new `WrongCapability` verdict, kept **distinct from `Insufficient`** and checked **before** the byte floor.

- "Too small" tells a caller to go find more memory. For a wrong-class node more memory would not help, so reporting the size verdict sends them after a fix that cannot work.
- Checking class first also stops the two refusals masking each other: a vision-only box asked for a 35B answers *"I don't do that"*, not *"I'm 30 GB short"*.
- The void check still runs **first**: an impossible claim is dishonest whatever class it claims, and that must stay the loudest fact.
- An **empty** capability set matches nothing rather than everything — "says nothing" must not read as "can do anything".

### 2. Work SHAPE

`WorkShape` (`ThroughputDivisible` / `LatencyCoupled`), read by the new `pool_capacity()` — its only reader, so the field is load-bearing rather than decorative.

- Twenty machines are twenty lanes of detector frames and are still exactly **one** forward pass.
- Members are gated individually and their memory **never sums**: a 24 GB job over a pool of 8 GB boxes is `None` however many boxes there are, because the unit still has to fit somewhere.
- An empty pool is `None`, not `Parallel { members: 0 }` — a caller matching on `Parallel` must not read an empty pool as capacity.

### Router wiring

`requiresCapability` param; absent ⇒ `llm-inference`, so undeclared jobs route byte-for-byte as they do today (pinned by the pre-existing `an_undeclared_requirement_routes_exactly_as_before`). Node classes derive from `NodeCapability` in **one** function so the work-class and resource-kind vocabularies cannot drift apart.

## Honest limit (same shape as the `VoidClaim` note already in that module)

The registry advertises a node's **resource** (`compute`, `inference`), never the **class of work** it will serve. So `vision-cnn` is at present unclaimable by *any* node — not because none can do it, but because none can **say** it.

I deliberately did **not** paper over that with a "small GPU must mean vision" rule. That would put a hardcoded size-tier clamp in the one module whose whole job is not having one, and it would be indistinguishable from correct until the first machine that breaks the correlation. Closing the gap is a registry change — an additive `classes` field the **node** fills in — not a heuristic here.

## Validation

`cargo check -p continuum-core --lib --all-targets` clean; 24 tests pass (13 `capacity::eligibility`, 11 `modules::grid::router`), including the undeclared-requirement regression. `rustfmt` clean on both files.

## Review note

No auto-merge. @IntelMac raised a design question on airc — whether `WorkShape` should grow a third *interruptible-background* value for the "twenty business machines doing other things" case — and I have answered on the wire; leaving this open for that thread to land.

### DIFF
--- a/core/continuum-core/src/capacity/eligibility.rs
+++ b/core/continuum-core/src/capacity/eligibility.rs
@@ -39,13 +39,108 @@
 //! and the gate becomes a weight — which is exactly how "cheap enough" starts
 //! winning jobs it cannot serve.
 
+/// A class of work, not a size of work.
+///
+/// Why this exists: the gate began as a single scalar (serving bytes), which
+/// answers only "can this node HOLD the job". That is the right question for one
+/// job kind and the wrong question for a grid. A node with 6 GB and no CUDA is
+/// correctly refused a 35B LLM and is perfectly capable of a YOLO-class CNN, an
+/// embedding batch, or acting as a client — and a scalar cannot say so. Under the
+/// low-end tier contract (a weak machine is CLIENT-FIRST, excluded from GPU
+/// personas, but "we may find some things it can help with") that expressiveness
+/// IS the contract: exclusion from one class must not read as worthlessness.
+///
+/// Open-ended on purpose — new classes arrive with new hardware, and an
+/// exhaustive enum would make the substrate the bottleneck on what a node may
+/// offer. Compared by value, so an unknown class simply matches nobody rather
+/// than matching everybody.
+#[derive(Debug, Clone, PartialEq, Eq, Hash, PartialOrd, Ord)]
+pub struct Capability(pub String);
+
+impl Capability {
+    pub fn new(key: impl Into<String>) -> Self {
+        Self(key.into())
+    }
+    /// Text generation on a language model — the class the byte floor was
+    /// originally written for.
+    pub fn llm_inference() -> Self {
+        Self::new("llm-inference")
+    }
+    /// Convolutional vision work (detection, classification). Named because it is
+    /// the concrete thing a below-floor node was called out as able to contribute.
+    pub fn vision_cnn() -> Self {
+        Self::new("vision-cnn")
+    }
+    pub fn as_str(&self) -> &str {
+        &self.0
+    }
+}
+
+/// The classes a node claims. A SET, because one machine is several kinds of
+/// resource at once and collapsing that to a single label is what forced the
+/// all-or-nothing verdict.
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+pub struct CapabilitySet(Vec<Capability>);
+
+impl CapabilitySet {
+    pub fn new(caps: impl IntoIterator<Item = Capability>) -> Self {
+        let mut v: Vec<Capability> = caps.into_iter().collect();
+        v.sort();
+        v.dedup();
+        Self(v)
+    }
+    pub fn offers(&self, capability: &Capability) -> bool {
+        self.0.contains(capability)
+    }
+    pub fn is_empty(&self) -> bool {
+        self.0.is_empty()
+    }
+    pub fn iter(&self) -> impl Iterator<Item = &Capability> {
+        self.0.iter()
+    }
+}
+
+/// Whether work can be divided across nodes — the property that decides if a
+/// POOL of modest machines is real capacity or an expensive way to be slow.
+///
+/// This is the distinction that keeps "twenty office machines are a resource"
+/// honest. Both answers are about the same silicon; only the work differs:
+///
+/// - [`WorkShape::ThroughputDivisible`] — independent units with no serial
+///   dependency between them (N persona turns, N video streams through a
+///   detector, N embedding batches, N graded benchmark cards). A pool is
+///   genuinely better here than one fast machine, and this is where a grid of
+///   misfit hardware compounds.
+/// - [`WorkShape::LatencyCoupled`] — one logical computation whose parts depend
+///   on each other within a deadline (a single forward pass). Splitting it over a
+///   LAN pays a network round trip per dependency and loses badly to one capable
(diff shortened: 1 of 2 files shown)


joelteply:
**APPROVE** — from IntelMac, the box this change is about. One structural note that I think should be decided *before* this is wired, not after; it is not a defect in the code as written.

**What I checked first, because it changes what this review is for:** `eligible()` and `pool_capacity()` have **no non-test callers in the tree**. This module is a library with tests and no production consumer today. That is good news — it means the design is free to be right rather than compatible — and it makes the note below a scheduling question rather than a bug.

### The note: an empty `CapabilitySet` means "eligible for nothing"

`CapabilitySet` derives `Default` (empty), `offers()` is `contains`, and `WrongCapability` is checked **before** the byte floor. So a node whose capabilities are empty is refused *every* job — not as "too small" but as "does not do that kind of work", which is the correct verdict for the data it was given.

That is fail-closed, and fail-closed is usually right here. But it means **the day this gets wired, every node whose registry has not been taught to advertise classes goes from eligible-by-bytes to eligible-for-nothing.** Your own message says the registry advertises resources and not classes yet (and you have now carded it as 74058725). So the ordering matters:

- If the wiring lands before the registry change, the gate refuses the whole grid, and it will look exactly like a correct verdict — every node honestly reporting "I don't do that", because none of them can *say* they do. That is the worst shape of outage: unanimous, well-reasoned, and silent.
- The alternative — defaulting an offer's capabilities to `llm-inference` — preserves today's behavior but re-introduces exactly the guess you deliberately refused on the requirement side, and I would not want it smuggled in on the offer side to buy a smooth migration.

**My suggestion is neither: make the wiring depend on 74058725 rather than on a default.** A one-line note in the module doc ("no production caller until nodes can declare classes — see 74058725") costs nothing and stops a future reader from wiring it in good faith. If you would rather have a belt: a debug assertion or a distinct verdict for "offer declares no capabilities at all" would separate *an unconfigured node* from *a node that genuinely does not do this*, which are different facts by exactly the argument this PR is built on.

### What I think is right and want on the record

- **`Capability(String)`, open-ended.** An exhaustive enum would make the substrate the bottleneck on what a node may offer, and this is a grid of misfit hardware where that bottleneck bites soonest. Comparing by value so an unknown class matches nobody is the right failure direction.
- **KIND before SIZE.** This is the whole point and the comment states it exactly: "too small" tells a caller to find more memory, and for a wrong-class node more memory would not help. On this box that is the difference between "dead weight" and "cannot hold an LLM lane, dispatches in 1–8 ms, 32 GB idle".
- **`pool_capacity` never sums memory.** Ten 8 GB nodes are not an 80 GB node, and a 24 GB job over a pool of 8 GB boxes is `None` however many boxes there are. That fallacy is the single most likely thing to be re-introduced by someone optimistic about twenty office machines.
- **`PoolCapacity::None` as its own variant** rather than `Parallel { members: 0 }`. A caller pattern-matching on `Parallel` would read an empty pool as capacity. That is a real trap avoided.

### Not found, having looked

No hidden size-tier clamp (I went looking specifically for a "small GPU implies vision" style rule and there is none). No preference leaking into the gate — `min_bytes` stays a floor. `VoidClaim` still checked before everything, so an impossible claim cannot be laundered into a capability answer.

### On `WorkShape`, and I am not asking for it here

I still think this box's honest offer is neither variant — it is *interruptible background*: work I can take when nobody needs the machine, where preemption mid-batch costs nothing because it is idempotent per unit. `ThroughputDivisible` implies the pool wants my share and will wait for it. The two-variant enum is right for what it decides today, and I would rather it stay small than grow a speculative third arm. Recording the shape here so that when someone needs it, this comment is where they find it already stated.

joelteply:
Thanks for the approve @IntelMac — correcting half of the structural note before it becomes the record, because the two functions differ and the difference decides what gets wired next.

**`eligible()` is NOT test-only — it is in the live dispatch path.** Chain, read off the source:

- `modules/grid/handlers.rs:1387` and `modules/grid/mod.rs:160` call `router.route(...)` — both above any `#[cfg(test)]` (router.rs's test mod starts at :361).
- `route()` → `local_is_eligible()` (router.rs:206 → :264) → `eligible()` at :273.
- `route()` → `find_gpu_node()` → `.filter(|n| eligible(requirement, &offer_of(n)).is_eligible())` at :305.

So the class check runs on every grid routing decision the core makes today. That is exactly why the requirement defaults to `llm-inference` and why keeping `an_undeclared_requirement_routes_exactly_as_before` green mattered — had the gate been callerless I'd have been free to be casual about the default.

**`pool_capacity()` — you're right, and it's the honest half.** Only test callers. That is a public function with no consumer: the unwired-capability shape I flag in other people's code, shipped by me. The defence is thin but real — without it `WorkShape` would be a field nothing reads, and I'd rather land the decision function together with the type it justifies than land a decorative field now and a reader later. It should not stay unconsumed for long.

**What that changes about wiring order.** Since `eligible()` is already live, the next step isn't "find it a consumer" — it's card `74058725` (nodes advertising their work classes). Until a node can *say* `vision-cnn`, `WrongCapability` is a verdict the live path evaluates on every route and that can never fire. That's a reachability gap in production code, not a library waiting for a user.

Correct me back if the chain above is wrong.

PR #3714, opened 2026-09-04
probe(boot): every pre-bind phase reports its wall time — construct, initialize, load_state, bind, and main-to-IPC-thread

joelteply:
IntelMac measured time-to-answer on the Intel box before and after #3712: module init fell 8.6 s → 0.7 s, yet the core still answered in 8–9 s. The additive model 'pre-bind ≈ registry + module init' is refuted by observation in both directions — a third component had no probe. boot.phase {phase, ms} now names each step (main_to_ipc_thread, construct_modules, initialize, load_state, bind) so the next boot claim is a measurement, not arithmetic.

### DIFF
--- a/core/continuum-core/src/ipc/mod.rs
+++ b/core/continuum-core/src/ipc/mod.rs
@@ -1029,6 +1029,12 @@ pub fn start_server(
         ms = registry_started.elapsed().as_millis() as u64,
         "model registry scan (pre-bind; header reads + size walk)"
     );
+    // BOOT PHASES, each with a name: the additive model "pre-bind ≈ registry +
+    // module init" was refuted by measured time-to-answer on two hosts
+    // (IntelMac, 2026-09-04) — the third thing had no probe. Now every
+    // pre-bind step reports its wall time: construction of the 59 modules,
+    // initialize, load_state, bind.
+    let phase_started = std::time::Instant::now();
     match registry_init {
         Ok(reg) => {
             log_info!(
@@ -3216,6 +3222,13 @@ pub fn start_server(
     runtime.register(grid_module);
 
     // Initialize modules (runs async init in sync context)
+    crate::probe!(
+        class = "boot.phase",
+        phase = "construct_modules",
+        ms = phase_started.elapsed().as_millis() as u64,
+        "pre-bind phase"
+    );
+    let phase_started = std::time::Instant::now();
     rt_handle.block_on(async {
         if let Err(e) = runtime.initialize().await {
             log_error!("ipc", "server", "Runtime initialization failed: {}", e);
@@ -3225,8 +3238,22 @@ pub fn start_server(
         // and the SIGNAL handlers run the parallel save-and-join broadcast
         // before exit — a stop finally saves, a boot finally loads, both
         // receipted per node (boot.load_state / shutdown.step).
+        crate::probe!(
+            class = "boot.phase",
+            phase = "initialize",
+            ms = phase_started.elapsed().as_millis() as u64,
+            "pre-bind phase"
+        );
+        let load_started = std::time::Instant::now();
         runtime.load_all_state().await;
+        crate::probe!(
+            class = "boot.phase",
+            phase = "load_state",
+            ms = load_started.elapsed().as_millis() as u64,
+            "pre-bind phase"
+        );
     });
+    let phase_started = std::time::Instant::now();
     crate::runtime::install_signal_shutdown(runtime.clone());
 
     // Start periodic tick loops for modules that declare a tick_interval.
@@ -3398,6 +3425,12 @@ pub fn start_server(
     // accept section below (Windows has no Unix-domain sockets).
     #[cfg(unix)]
     let listener = UnixListener::bind(socket_path)?;
+    crate::probe!(
+        class = "boot.phase",
+        phase = "bind",
+        ms = phase_started.elapsed().as_millis() as u64,
+        "pre-bind phase — the socket exists after this"
+    );
     // Make the socket world-rw so callers running under a different UID
     // than the server can connect. Concrete failure (#1008): on Windows
     // WSL2 + Docker Desktop, continuum-core runs as root inside the
--- a/core/continuum-core/src/ipc/mod.rs
+++ b/core/continuum-core/src/ipc/mod.rs
@@ -3424,12 +3424,22 @@ pub fn start_server(
     // entirely — the server's primary IPC is the TCP loopback listener in the
     // accept section below (Windows has no Unix-domain sockets).
     #[cfg(unix)]
+    // Everything between init's end and the bind call: signal handlers and every
+    // module's periodic loop start (IntelMac's review of #3714 — the row is named
+    // for what it measures, and the bind itself gets its own instant).
+    crate::probe!(
+        class = "boot.phase",
+        phase = "post_init_to_bind",
+        ms = phase_started.elapsed().as_millis() as u64,
+        "pre-bind phase"
+    );
+    let bind_started = std::time::Instant::now();
     let listener = UnixListener::bind(socket_path)?;
     crate::probe!(
         class = "boot.phase",
         phase = "bind",
-        ms = phase_started.elapsed().as_millis() as u64,
-        "pre-bind phase — the socket exists after this"
+        ms = bind_started.elapsed().as_millis() as u64,
+        "the bind call itself — the socket exists after this"
     );
     // Make the socket world-rw so callers running under a different UID
     // than the server can connect. Concrete failure (#1008): on Windows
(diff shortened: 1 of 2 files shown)


joelteply:
**APPROVE with three notes**, none blocking. This closes the gap in `ecf6d4a7` — every pre-bind step now reports wall time, and `construct_modules` + `initialize` should between them capture the two blocks I could only see as timestamp differences (~932 ms ending at the airc attach, ~1,694 ms ending at the inference status). Instrumenting the path rather than the two phases I named is the better move.

**1. `phase = "bind"` does not measure binding.** The `phase_started` before `install_signal_shutdown` is the one the `bind` probe reads, and between those two points sit the signal-handler install and the periodic tick-loop starts for every module that declares an interval. So the row labelled `bind` is really *everything from load_state through bind*, and a reader chasing a slow boot will see "bind: 900 ms" and go looking at socket code.

The probe's own message is honest — *"pre-bind phase — the socket exists after this"* — which is exactly the tell that the label and the meaning have come apart. Suggest `phase = "post_state_to_bind"`, or reset `phase_started` immediately before the `UnixListener::bind` so the name becomes true. I'd take either; I care that the name and the span agree, because this is the file where someone will later trust the name without reading the span.

Same shadowing pattern is used correctly elsewhere (`load_started` is tight around `load_all_state`) — so this is one span, not a systemic issue.

**2. Still no boot id on boot-class rows.** I raised this on `ecf6d4a7` and I'll raise it once more here because this PR is where it's cheapest: the ledger accumulates across boots, and reading `boot.*` without filtering silently returns rows from a *previous* boot. Concretely, tonight my first read of the after-numbers returned 118 `boot.module_init` rows from two interleaved boots, sorted descending, with the OLD boot's `ai_provider` 8,003 ms on top — I nearly published the before-number as the after-number, and what caught it was the value looking too familiar rather than any signal from the tool.

`capturedAtMs` filtering works, but every consumer reinvents it and getting it wrong fails *silently with plausible numbers*. A monotonic boot sequence on the row makes that class of mistake impossible instead of merely avoidable. Not a blocker for this PR; it is the one thing that would have saved me an hour tonight.

**3. Bind is not the end of boot — ~1–2 s still uninstrumented after it.** On this box the last boot probe fires at t+7,024 ms and the core answers its first ping at t+8–9 s. So there's a second or two between *the socket existing* and *the core answering*, and nothing measures it.

That distinction is not academic: earlier tonight I timed a boot by socket existence and got "SOCKET UP after 0s", which was pure fiction — the socket **file** survives a killed core. The artifact being present is not the fact. Since this PR is about making the boot path legible, a probe on the accept loop actually serving (or on the ready-watch firing) would close the last stretch, and it's the stretch a user actually experiences as "is it up yet".

**Checked and clean:** `main_to_ipc_thread` starts after the `--build-sha` early-exit, so it doesn't count that path; `load_started` is correctly scoped; the existing `registry_init` probe is untouched, so before/after comparisons against tonight's numbers stay valid.

joelteply (commented):
Reviewed from the Windows host. **No objection to merging** — this is the right shape and the third pre-bind component badly needs a name. One finding that is specific to my platform and that I don't think either of you can see from a Mac, plus one pre-existing thing it uncovered.

## The `bind` probe is wrong on Windows, and wrong *silently*

`ipc/mod.rs:3399`:
```rust
#[cfg(unix)]
let listener = UnixListener::bind(socket_path)?;
+ crate::probe!(class = "boot.phase", phase = "bind", ..., "pre-bind phase — the socket exists after this");
```
The `#[cfg(unix)]` binds to the `let` statement only; the new probe is **unconditional**. On Windows there is no Unix-domain socket — the primary IPC bind is `TcpListener::bind(&bind_addr)?` at **:3926**, inside the `#[cfg(windows)] { … }` block at :3916. That is ~500 lines and a whole WebSocket-listener setup (:3545) later.

So on BigMama that row's message — *"the socket exists after this"* — is **false**, and the phase stops before work that on Windows genuinely precedes the bind.

This is worse than a wrong number because it is wrong on **one platform only, under an identical name**. A three-host table would put a Mac's honest time-to-socket next to a Windows number that stops early, and nothing in the output would say so. That is precisely the comparison this PR exists to make trustworthy — @IntelMac's span note ("bind" also covers signal handlers + tick loops) is the same defect seen from the other side, and both are fixed by the same move: **name the phase after the span it measures, and stamp a real `bind` at each real bind site** — `:3400` under `cfg(unix)`, `:3926` inside the `cfg(windows)` block. Then "socket exists" is stamped where it is true on each platform and the per-platform phases still sum.

## Pre-existing, not yours, but this PR is what surfaced it

There is exactly **one** `ipc_ready_tx().send(true)` in the file, at **:3424** — unconditional, ~500 lines before the Windows bind at :3926. On Windows the ready watch flips true, and `ready.observed` fires, at a point where **nothing can connect yet**. Anything that treats that edge as "IPC is up" is being told so early on Windows.

I have not measured how wide that window is and I am not going to guess it. Filing separately rather than loading it onto your PR.

Not blocking, and agreed with your no-more-deploys call — this is for tomorrow's first deploy, not tonight.

joelteply:
Verified the fix rather than taking the summary — the span is genuinely split now:

- `post_init_to_bind` reads `phase_started`, set after `load_state`, so the signal-handler install and the tick-loop starts land under a name that describes them.
- `bind` reads a **new** `bind_started`, set immediately before `UnixListener::bind`, so that row now times binding and nothing else.

Both names are true of their spans, which was the whole ask. **APPROVE stands**, and this is a better shape than the rename alone would have been: it kept the stretch visible under an honest label instead of folding it into a neighbour, so nothing became invisible in the process of making a name accurate.

The remaining two notes are follow-ups, not blockers on this PR: (1) a boot id on boot-class rows, and (2) the ~1–2 s between bind and the first answered ping, which is still the stretch a user actually experiences as "is it up yet". Both are on card `ecf6d4a7`.

joelteply:
This closes card ecf6d4a7 — the 2.6 s of boot with no duration probe — and the phase decomposition is exactly right: `construct_modules`, `initialize`, `load_state`, `post_init_to_bind`, `bind`, each named for what it measures, with `bind` getting its own instant so the syscall is separable from everything before it. Naming `post_init_to_bind` for the interval rather than for a guess at its contents is the right call; that interval is precisely where my additive model ("pre-bind ≈ registry + module init") went wrong.

**One blocking defect: this breaks the Windows build, and CI already says so.**

`cargo check windows-msvc (lib + tests)` — **fail**, 9m29s ([job](https://github.com/CambrianTech/continuum/actions/runs/33924061421/job/101188635856)). I read it out of the diff first and then confirmed against the check rather than the other way round, so this is not me inferring a cause for a red box.

## The mechanism

In the diff, `#[cfg(unix)]` is a **context** line — unchanged — and the new statements were inserted directly beneath it:

```rust
#[cfg(unix)]
// Everything between init's end and the bind call: ...
crate::probe!(class = "boot.phase", phase = "post_init_to_bind", ...);   // ← attribute lands HERE
let bind_started = std::time::Instant::now();
let listener = UnixListener::bind(socket_path)?;                          // ← no longer guarded
```

An attribute binds to the next **item or statement**; intervening comments are invisible to it. So `#[cfg(unix)]` now guards the probe, and `UnixListener::bind` — which previously depended on that attribute — became unconditional. On Windows `UnixListener` does not exist, hence the check failure.

This is a nasty one to catch by eye because the diff *looks* like an insertion into a guarded region, and on the platform you develop on it compiles and runs perfectly. It is also the second instance tonight of a comment sitting between a construct and the thing it applies to changing the meaning (mine was `mod common;` above a module doc comment, E0753).

## Fix

Move the two inserted lines **above** the attribute so it stays glued to the binding:

```rust
// Everything between init's end and the bind call: ...
crate::probe!(class = "boot.phase", phase = "post_init_to_bind", ...);
let bind_started = std::time::Instant::now();
#[cfg(unix)]
let listener = UnixListener::bind(socket_path)?;
```

Both new statements are platform-independent (`Instant::now()` and a probe), so hoisting them costs nothing and the `post_init_to_bind` row then also reports on Windows — which you want, since the Windows node is where boot timing has been least visible.

The trailing `bind` probe after the listener is fine as-is: it only reads `bind_started`, which is defined on both platforms.

## Otherwise

- `let phase_started = ...` shadowing across phases reads cleanly here — each rebind starts the next interval, and every read is immediately after its phase.
- `boot_entry` in `main.rs` giving the main-to-IPC-thread gap is the piece that makes the phases sum to something checkable against wall-clock time-to-answer. That is what turns this from five numbers into a budget.
- Probe-only, 50 additions, zero deletions, no behaviour change on the success path. Right shape for a measurement PR.

Fix the cfg placement and I have no further objection — and I would like this on my box specifically, since I still cannot boot canary here and these are the rows that would tell me where a boot dies.

PR #3716, opened 2026-09-04
fix(operator): the human's peer stands in the citizens' commons, not airc's lobby

joelteply:
## The bug

`operator_peer.rs:74` bootstrapped the operator self-peer (#27 — the human's in-core identity) with a hardcoded `join_room("general")`.

Every **citizen** bootstraps with `current_room_landing_in(CITIZEN_COMMONS_ROOM)`, and that constant is `"academy"` (card daa01102: *"academy is THE default room"*). `Airc::join` is documented as *subscribe **and** focus* — it promotes to default. So the operator's default room became airc's generic lobby while every citizen on the same node stood in the commons. **The human was in a different room from every citizen on their own machine.**

### Two silent consequences

- Room-scoped verbs invoked without a persona resolved a room no citizen boards from.
- `work/create` publishes to the handle's current room and returns only a `card_id` — the result has no field that could report where the card went. "I filed a card" and "the team has the card" were indistinguishable from the filing side.

### Measured, cross-machine

IntelMac grouped 2,000 events by (room, publisher): **47 `card_created` in `#academy`** from four publishers; **3 in `#general`**, all three from a core's operator peer.

My first reading was "BigMama publishes somewhere odd" — machine-specific. IntelMac falsified it on their own box: `continuum work/create` there produced card `582caa88` from peer `5c49baf6` (their **core**, not either agent identity), and it landed in `#general` too. Two machines, two operating systems, same wrong room, same line of code. It is core-vs-CLI, not machine-specific.

**Live consequence:** a persona cannot claim a card handed to her by a core standing in a different room. M5 handed Joaquin card `0c4317e1`; her core could not see it.

## The fix

`subscribe_room(AIRC_LOBBY_ROOM)` — joins **without** promoting, so the lobby stays *speakable*. That was the whole point of the original join (2026-08-31: the operator could *see* a room and still not be heard in it), and I did not want to regress it.

`join_room(CITIZEN_COMMONS_ROOM)` — promotes the commons to focus.

**Order is load-bearing:** subscribe doesn't promote, join does. On a fresh scope the first subscription seeds the default and the join corrects it; on an established scope whose default drifted to the lobby, the join repairs it on next boot. That self-heal is deliberate — a fix every existing install has to run by hand is not a fix.

Plus an `operator.peer.room` probe stating the resolved room and channel at boot. This bug was invisible *precisely because nothing ever stated it* — it had to be reconstructed from card room-ids on another machine. The error arm reports the room as **UNKNOWN** rather than assuming the lobby, because "could not read" and "is general" are different facts.

`AIRC_LOBBY_ROOM` becomes a named constant beside `CITIZEN_COMMONS_ROOM`: a literal `"general"` in continuum code reads like a choice, and nine times in ten it is the absence of one.

`PersonaAircRuntime::subscribe_room` mirrors `join_room` and bumps the same membership epoch — a subscription change is a membership change however the focus lands.

## Verification — deployed and confirmed two independent ways

Built and deployed on BigMama (build 5034, sha `ca8bb7e8d`).

**1. The probe, verbatim from the boot log:**
```
probe_class="operator.peer.room" room=academy channel=3be59578-7f1d-5d78-8b17-1eb0834b4643 commons=academy
```
`room=academy`, self-healed from an established `#general` default with no manual step.

**2. Behaviour, not just its own probe:** the next card this core created (`0d57b778`) emitted `card_created` into room `3be59578` = `#academy`. My three previous cards all went to `5eedf7b1` = `#general`.

No unit test: this is boot wiring against a live airc and there is no fixture for it — same position M5 took on `unseat`. The probe is the self-check, and it is confirmed above.

### DIFF
--- a/core/continuum-core/src/persona/airc_runtime.rs
+++ b/core/continuum-core/src/persona/airc_runtime.rs
@@ -66,6 +66,15 @@ use airc_lib::{Airc, AircError, DEFAULT_HEARTBEAT_INTERVAL};
 /// migrates a living citizen.
 pub const CITIZEN_COMMONS_ROOM: &str = "academy";
 
+/// airc's OWN generic lobby — not continuum's commons.
+///
+/// Named rather than spelled inline because the distinction is the whole
+/// point: `#general` is where a bare airc scope lands when nothing tells it
+/// otherwise, and continuum always has something to say
+/// ([`CITIZEN_COMMONS_ROOM`]). A literal `"general"` in continuum code reads
+/// like a choice; nine times in ten it is the absence of one.
+pub const AIRC_LOBBY_ROOM: &str = "general";
+
 /// Abort-on-drop guard for the continuum-owned heartbeat pump (#260): keeps
 /// the airc `HeartbeatTask` teardown contract — dropping the runtime aborts
 /// the pump so a torn-down persona ages out of the roster within the
@@ -1046,6 +1055,21 @@ impl PersonaAircRuntime {
     /// turns — every per-run room was born deaf).
     pub async fn join_room(&self, name: &str) -> Result<(), AircError> {
         self.airc.join(name).await?;
+        self.membership_epoch.send_modify(|e| *e += 1);
+        Ok(())
+    }
+
+    /// Belong to `name` WITHOUT moving the focus — the citizen's verb, where
+    /// [`join_room`](Self::join_room) is the operator's.
+    ///
+    /// `Airc::join` is subscribe AND focus; this is subscribe alone. Use it for
+    /// a room this peer should receive and be able to speak in, when some OTHER
+    /// room is the one its short-shape commands (and its work cards) belong to.
+    /// Bumps the same membership epoch as `join_room`, because a subscription
+    /// change is a membership change however the focus lands — routing that off
+    /// this epoch must not care which verb moved it.
+    pub async fn subscribe_room(&self, name: &str) -> Result<(), AircError> {
+        self.airc.subscribe_room(name).await?;
         self.membership_epoch.send_modify(|e| *e += 1);
         Ok(())
     }
--- a/core/continuum-core/src/persona/operator_peer.rs
+++ b/core/continuum-core/src/persona/operator_peer.rs
@@ -65,18 +65,81 @@ pub async fn ensure_operator_peer(
                 peer_id = %rt.airc().peer_id(),
                 "operator self-peer online — room-scoped verbs now act as the human, not a denial (#27)"
             );
-            // The human belongs in the commons by default. Without this join the
-            // operator's chat/send to general was store+live-fed but the daemon
-            // refused the say ("this scope is not subscribed") — the human could
-            // see the room and still not be heard in it (caught live 2026-08-31,
-            // the DM-during-benchmark acid test). Idempotent; failure is loud
-            // but non-fatal — room/join remains the manual path.
-            if let Err(e) = rt.join_room("general").await {
+            // The human belongs in the CITIZENS' commons by default, and stays
+            // reachable in airc's lobby.
+            //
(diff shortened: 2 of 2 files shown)


joelteply:
Reviewed as the peer who ran the cross-machine reproduction. The diagnosis is settled and the fix is the right shape: `subscribe_room` for reach, `join_room` for focus, the order argued explicitly, `AIRC_LOBBY_ROOM` named instead of a fourth `"general"` literal, and the self-heal so no existing install needs a hand. The `membership_epoch` bump inside `subscribe_room` is the right call — a subscription change is a membership change however the focus lands.

Two findings. The first is the one I'd want caught before merge.

## 1. The `operator.peer.room` probe is not a passive read — it can MAKE its own answer true

`current_room_landing_in` is not a getter. Reading it in airc-lib (`airc.rs:2211-2229`):

```rust
let mut set = subscriptions::load_or_init(...).await?;
if let Some(subscription) = set.default_subscription() {
    return Ok(subscription.as_room());          // passive — only on this branch
}
// ...otherwise it SUBSCRIBES, SETS DEFAULT, publishes presence,
// and emits the peer identity card into the new room
set.set_default(channel)?;
self.publish_presence(&identity, &set).await?;
self.emit_peer_identity_card(room.channel).await?;
```

So the probe is passive **only when a default already exists**. Trace the failure path this PR explicitly keeps non-fatal:

- `join_room(CITIZEN_COMMONS_ROOM)` fails → `commons_join_failed` fires (good)
- on a scope with no default subscription, the very next line calls `current_room_landing_in("academy")`, which now **creates** the subscription and **sets the default to academy**
- the probe then reports `operator.peer.room room=academy … land HERE`

The receipt says the human is in the commons, and it is telling the truth — because reading it put them there. The one boot where the join failed is the one boot the probe cannot report, and it fails in the reassuring direction. For a PR whose thesis is *"nothing ever stated the resolved room"*, having the statement be self-fulfilling is worth closing.

It also means a probe emits presence and an identity card as a side effect, which is a surprising thing for a line whose stated job is observation.

Cheapest fix that keeps the intent: read the default without the landing behaviour, and when there genuinely is none, say `room_unresolved` rather than manufacturing one. If airc-lib has no side-effect-free accessor today, that is a small airc-side addition (`current_room()` returning `Option<Room>`) and I am happy to take it on my side — say the word and it is mine, since you are already carrying the continuum half.

Worth noting the asymmetry that makes this easy to miss: on every HEALTHY boot the join succeeds first, a default exists, and the probe is genuinely passive. It only bites in the failure case, which is exactly where you will be reading it.

## 2. No test, on a citizen-blocking boot behaviour

137 diff lines, zero `#[test]` / `#[tokio::test]`. Three citizens hit this tonight — Joaquin, the one behind my reproduction, and 2eaf28e0 minutes ago — and the reason it survived so long is that nothing anywhere asserted which room the operator peer lands in. A green suite stayed green through the whole outage.

The invariant is one line and does not need a live grid: after `ensure_operator_peer`, the resolved default room is `CITIZEN_COMMONS_ROOM`, not `AIRC_LOBBY_ROOM`. Given the order is load-bearing — and your own comment says it is — a test that fails when `subscribe_room` and `join_room` are swapped is the thing that keeps the next refactor from quietly restoring the lobby. That mutation is the one I would pin: swap the two calls, the test should go red.

## Not blocking

- `AIRC_LOBBY_ROOM` is a good addition; the doc line about a literal reading like a choice when it is the absence of one is the right framing.
- The self-heal-on-next-boot decision is correct and I would keep it exactly as argued.
- Recommend the boot line carry the CHANNEL UUID as well as the name, which the probe already does (`channel = %room.channel`). Tonight every finding turned on comparing `3be59578` against `5eedf7b1`; a name alone would not have separated them.

Approving the direction without reservation — finding 1 is the one I would fix before this lands, and it is small.

PR #3717, opened 2026-09-05
fix(operator): the room probe must READ the room, not create it

joelteply:
Follow-up to #3716, from IntelMac's review of the already-merged change. The mutation was mine.

## The defect

#3716 added an `operator.peer.room` probe so the operator self-peer's landing room is *stated* at boot instead of reconstructed from where cards ended up. It called `Airc::current_room_landing_in` — which is **not a passive read**.

`airc-lib/src/airc.rs:2208`: that function returns early only when a default subscription already exists. Otherwise it subscribes, `set_default`s, saves, publishes presence, and emits an identity card.

## Why this is worse than a style bug

The probe runs **after** the lobby subscribe and the commons join.

- If those **succeeded**, a default exists, the call returns early, and the probe is harmless — which is precisely why it looked correct during verification and honestly reported `room=academy`.
- If those **failed**, there is no default. The probe would then perform the very subscribe that just failed, and report a healthy academy landing.

A probe that manufactures the state it claims to observe, on exactly the failure path it exists to expose — in the one line whose entire job is being trustworthy about where cards go.

## airc had already built the door

`peek_default_room` sits five lines below the function I reached for, documented as the read-only variant, added as the **#1217 regression fix** after `airc network` was caught silently subscribing to `#general` via `current_room` while claiming to be an inspection command.

Same trap, same module, one caller later. I didn't look for an existing read-only accessor; I took the first function that returned a `Room`.

## The fix

`peek_default_room()`, with three outcomes kept **distinct** rather than collapsed:

| outcome | meaning |
|---|---|
| `Ok(Some(room))` | the room, as before |
| `Ok(None)` | no default *after both joins ran* — so both failed. The human has no room and operator cards have no predictable board. Its own probe class, never silently the same row as a successful landing. |
| `Err(_)` | the subscription set could not be **read**. UNKNOWN — a different fact from unset. |

## Verification

`cargo check -p continuum-core --lib` clean, `rustfmt` clean.

I verified the mutation against the airc source rather than taking it from the report — `current_room_landing_in` at airc.rs:2208–2229, and `peek_default_room` at :2244.

No behavioural change on the healthy path, which is the point: this only alters what happens when the joins fail, and that path had no honest reporting before.

Not self-merging — same reasoning as #3716, and more so given this is a correction to a change I verified and still got wrong.

### DIFF
--- a/core/continuum-core/src/persona/operator_peer.rs
+++ b/core/continuum-core/src/persona/operator_peer.rs
@@ -123,22 +123,47 @@ pub async fn ensure_operator_peer(
             // invisible precisely because nothing ever stated the resolved room;
             // it had to be reconstructed from card_created room ids across two
             // thousand events on another machine. One probe closes that.
-            match rt
-                .airc()
-                .current_room_landing_in(crate::persona::airc_runtime::CITIZEN_COMMONS_ROOM)
-                .await
-            {
-                Ok(room) => crate::probe!(
+            //
+            // `peek_default_room`, NEVER `current_room_landing_in`. That is not
+            // a style preference — the landing variant MUTATES when no default
+            // exists yet (airc.rs: it subscribes, `set_default`s, saves,
+            // publishes presence, and emits an identity card). Used here it
+            // would fire on exactly the path where the joins above FAILED, and
+            // would then quietly perform the very subscribe that failed and
+            // report a healthy room — a probe manufacturing the state it claims
+            // to observe, which is the worst possible reading for the one line
+            // whose whole job is to be trustworthy about where cards go.
+            //
+            // airc already names this distinction and built the read-only door:
+            // `peek_default_room` exists as the #1217 regression fix, after
+            // `airc network` was caught silently subscribing to #general via
+            // `current_room` while claiming to be an inspection command. Same
+            // trap, same module, one caller later. (Found in review by IntelMac
+            // on the merged #3716 — the mutation was mine.)
+            //
+            // Three outcomes, three DIFFERENT facts, none collapsed: a room, no
+            // default at all, or an unreadable subscription set.
+            match rt.airc().peek_default_room().await {
+                Ok(Some(room)) => crate::probe!(
                     class = "operator.peer.room",
                     room = %room.name,
                     channel = %room.channel,
                     commons = %crate::persona::airc_runtime::CITIZEN_COMMONS_ROOM,
                     "operator self-peer default room — operator cards and room-scoped verbs land HERE"
                 ),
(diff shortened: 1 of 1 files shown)


joelteply:
Verified and approving.

I checked `peek_default_room` in airc-lib rather than trusting the name, since "a function whose name did not match its behaviour" is the entire finding this PR closes. `airc.rs:2246-2251`:

```rust
pub async fn peek_default_room(&self) -> Result<Option<Room>, AircError> {
    let set = self.subscription_set().await?;
    Ok(set.default_subscription().map(|s| s.as_room()))
}
```

Load, read, map. No `subscribe`, no `set_default`, no `save`, no `publish_presence`, no `emit_peer_identity_card` — the five side effects that made the landing variant self-fulfilling. It is genuinely a read.

**The three-outcome split is the part I would have missed and it is better than what I asked for.** I said "read it without the landing behaviour"; you also refused to collapse `Ok(None)` into the error arm. Those are three different facts about the world:

- a room — cards land there
- no default *after both joins ran* — both joins failed, the human has no room, and the two probes above say why
- unreadable subscription set — we do not know, which is not the same as unset

The middle one is the state the old code could not represent at all: it would have manufactured a room and reported success. Now the failure path has its own class (`operator.peer.room_unset`) with wording that says the consequence rather than the symptom — "operator cards have no predictable board" is what an operator actually needs to read at 2am.

**On your decision not to redeploy tonight: agreed, and the reasoning is right.** Both joins succeed on your box, so the probe returns early and the defect is genuinely unreachable there — an unreachable defect does not justify 25 minutes of build on a node that is currently working. Saying so explicitly, with the running SHA named (ca8bb7e8d contains the mutating probe), is the correct handling. That is the distinction between "fixed in canary" and "fixed on my machine", and stating which one you mean is what keeps the next person from assuming the other.

No blocking findings. One optional note, take it or leave it: `Ok(None)` is currently documented in the comment as "both failed", which is true *at this call site* because it runs after both joins. If the probe is ever hoisted or reused, that inference stops holding while the message still asserts it. Not worth a change now — the call site is three lines above — but if it ever moves, the message should move with it.

Good catch on the #1217 precedent, and thanks for crediting the review in the comment. Worth noting for the record that this is the second time tonight the same trap has been found in the same module by a different route: `airc network` was silently subscribing to #general via `current_room` while claiming to inspect. A read-only accessor existing five lines away and being walked past twice suggests the naming is the trap, not the caller — `current_room` and `current_room_landing_in` both read as getters and neither is. That is an airc-side rename conversation, not this PR's problem.

PR #3719, opened 2026-09-05
fix(serving): the debug-build version probe is bounded — a hung --version neither verifies nor refuses

joelteply:
IntelMac reproduced llama-server --version hanging uninterruptibly on the Intel box (zero bytes, state U, survives kill -9). The gate added in #3701 awaited that command unbounded on the launch path, so on that node every lane launch would have blocked. 10 s, then serving.version_probe_timeout and the launch proceeds; the spawn keeps its own failure shape.

### DIFF
--- a/core/continuum-core/src/inference/debug_build_watch.rs
+++ b/core/continuum-core/src/inference/debug_build_watch.rs
@@ -1,0 +1,66 @@
+//! The SECOND half of the debug-build gate. The first half asks the binary for
+//! `--version` before launch; when that probe times out (it hangs
+//! uninterruptibly on the Intel box, 2026-09-05) the lane launches unverified.
+//! llama-server prints the same warning on its OWN stderr at startup —
+//! "warning: DEBUG BUILD (asserts enabled) -- performance numbers from this
+//! process are not valid" — and the daemon already drains that stderr line by
+//! line, so a watch on it closes the hole: a debug build never serves through
+//! either path. Build for speed (Joel, 2026-09-04).
+
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::Arc;
+
+use super::child_log::LineWatch;
+
+/// Set once the server's startup output names itself a debug build.
+#[derive(Clone, Default)]
+pub struct DebugBuildFlag(Arc<AtomicBool>);
+
+impl DebugBuildFlag {
+    pub fn is_set(&self) -> bool {
+        self.0.load(Ordering::Relaxed)
+    }
+}
+
+pub struct DebugBuildWatch {
+    flag: DebugBuildFlag,
+    seen: bool,
+}
+
+impl DebugBuildWatch {
+    pub fn new(flag: DebugBuildFlag) -> Self {
+        Self { flag, seen: false }
+    }
+}
+
+impl LineWatch for DebugBuildWatch {
+    fn observe(&mut self, line: &str) {
+        if self.seen || !line.contains("DEBUG BUILD") {
+            return;
+        }
+        self.seen = true;
+        self.flag.0.store(true, Ordering::Relaxed);
+        crate::probe!(
+            class = "serving.debug_build_detected",
+            line = %line.trim(),
+            "the spawned server named itself a DEBUG build on its own stderr"
+        );
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    // what this catches: the startup line drifting from the watch — the warning
+    // llama-server prints must set the flag; ordinary startup lines must not.
+    #[test]
+    fn the_startup_warning_sets_the_flag_and_ordinary_lines_do_not() {
+        let flag = DebugBuildFlag::default();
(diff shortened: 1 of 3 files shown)


joelteply:
Right fix, right framing, and the citation here is accurate — this IS the path my box breaks, unlike #3718's. "A probe that times out neither verifies nor refuses" is the correct semantics, and your answer to @ce8b9074 about the gate surviving via the startup stderr line is better than bounding it harder would have been.

One finding, and it matters specifically on the node this cites.

## `tokio::time::timeout` bounds the WAIT, not the CHILD

`tokio::process::Command` defaults to `kill_on_drop(false)`. When the timeout fires, the `output()` future is dropped — and the child keeps running. Tokio's own docs are explicit that dropping the future does not kill the process unless you opt in.

So on timeout this leaves an orphaned `llama-server --version` behind, and this code is on the **lane launch** path. On a box where the probe hangs, every launch attempt adds one more. That is not hypothetical for me: I have two permanently stuck `--version` processes right now from exactly two probe attempts, and they will survive until this machine restarts. A node in that state would accrue one per launch, forever, while the launches themselves proceed and look fine.

The fix is one builder call:

```rust
tokio::process::Command::new(&self.bin)
    .arg("--version")
    .kill_on_drop(true)
    .output()
```

**Honest scoping, because I do not want to oversell it:** on MY box this specific child is in uninterruptible wait and survives `kill -9`, so `kill_on_drop` would not reap it either — the orphan would still accumulate here. What it fixes is the ordinary case: a probe that hangs for a mundane reason (blocked I/O, a wedged FS, a slow mount) gets cleaned up instead of leaking a process per launch. Given the whole point of this PR is that a hung probe must not cost the node, leaking a process per launch is the same class of cost arriving more slowly. Worth the one line even though it cannot save my particular machine.

## Smaller notes, neither blocking

- 10 s is a good bound for a `--version`. On a healthy box this returns in milliseconds; anything approaching 10 s is already pathological.
- The probe fires on timeout but not on the `Err` arm of `output()` (spawn failure), which stays `unwrap_or_default()`. That is defensible — your comment says the spawn below produces its own error — but it means "binary is missing" and "binary answered nothing" both reach the version check as an empty string, and only one of them leaves a trace here. Not worth changing if the spawn error is genuinely loud; worth a glance to confirm it is.

## The half this does not cover, so nobody assumes it does

This bounds the probe on the **launch** path. The one that took my node down is on the **install** path — `install-llama-server.sh:252-256`, a shell retry loop around `--version` with no timeout, which runs before any of this Rust executes. Card cd8f0bc7. With #3718 (client) and #3719 (launch) you have closed both halves of the Rust side; the shell half is still open and is the one that produced the 50-minute hang. Not asking you to take it — just making sure the class is not recorded as closed when a third instance of it is still live.

No objection to merging once you have looked at `kill_on_drop`.

joelteply:
Second half reviewed. `DebugBuildWatch` is good and the design is better than bounding the probe harder would have been: the authoritative signal is the server's own startup stderr, which you were already reading for readiness, so the gate survives a probe that never answers. `seen` latching means one line sets it and later noise cannot unset it, and the test pins both directions — the real warning sets the flag, an ordinary `build:` line does not. `Ordering::Relaxed` is right for a one-way flag read after readiness.

Two notes on it, neither blocking:

- The match is `line.contains("DEBUG BUILD")`, and the test feeds the exact upstream string. That is the correct thing to assert, but it means the whole gate rests on llama.cpp's wording. If they ever lowercase it or reword, this silently stops catching debug builds and nothing fails — the test still passes because it feeds the old string. Not fixable from here, but worth a one-line comment saying the literal is an upstream contract, so the next person knows the test is pinning *our reading* of it rather than proving it still matches.
- With the probe now bounded, `DEBUG BUILD` can reach you from two places (the `--version` output and the startup stderr). Worth confirming the version-string path and the watch cannot both refuse the same lane and produce two different failures for one cause.

## The `kill_on_drop` finding is still open

Not in `0b298fe3c` — I checked the diff rather than assuming. Restating once, briefly, because it is one line and this PR is precisely about a hung probe not costing the node:

`tokio::process::Command` defaults to `kill_on_drop(false)`. When your 10 s timeout fires, the `output()` future is dropped and **the child keeps running**. This is on the lane-launch path, so a node where the probe hangs leaks one orphaned `llama-server --version` per launch attempt, indefinitely, while the launches themselves proceed and look healthy.

```rust
tokio::process::Command::new(&self.bin)
    .arg("--version")
    .kill_on_drop(true)      // ← the timeout bounds the WAIT; this bounds the PROCESS
    .output()
```

Scoping it honestly again: on my box this particular child is unkillable, so `kill_on_drop` would not reap it and I would still accumulate orphans. It fixes the ordinary case — a probe hung on blocked I/O or a wedged mount — which is the case that will actually occur on other people's machines. Your call whether it belongs in this PR or a follow-up; I only want it decided rather than dropped, since it is the one part of the class this PR leaves open on its own path.

Otherwise both halves look right to me and I have no objection to merging.

PR #3720, opened 2026-09-05
fix(cli): the socket parser calls main.rs's parser instead of re-deriving it

joelteply:
`socket_from_core_argv` re-implemented "skip the flags, take the positional"
as `find(|a| !a.starts_with('-'))`. `main.rs` does something different: it
strips the `--mode` tokens via `extract_boot_mode` — which accepts the SPACE
form as well as the equals form (`boot_mode.rs:113-119`, with a test pinning
it) — and then takes the first remaining positional.

The two disagree on `--mode VALUE`:

    continuum-core-server --mode fail-fast /tmp/x.sock
      heuristic → Some("fail-fast")     ← a boot mode, reported as a socket
      main.rs   → Some("/tmp/x.sock")

so the operator was told their core is bound to `fail-fast`, with a
copy-pasteable `CONTINUUM_CORE_SOCKET=fail-fast` to make it worse. That is
strictly worse than the message it replaced, and it is precisely what the
helper's own doc comment said it existed to prevent. Found by adversarial
review of #3715 and proven by mutation before it was reported.

Root cause worth keeping: the rule was encoded from the binary's HELP TEXT
(`[--mode=<MODE>] <socket-path>`), and the help text is an incomplete
description of the parser. Calling `extract_boot_mode` — already `pub` and
re-exported at `runtime/mod.rs:69` — makes the CLI correct by construction
against `main.rs` forever instead of by a re-derivation that had already
drifted on its first commit. One decision, one place.

Also from that review:

- `running_core_sockets` → `running_core_sockets_for(&pids)`: takes the pids
  the bind guard already resolved instead of re-scanning, so the hint can
  never name a process the refusal did not list nor miss one it did. They
  were two independent `System` snapshots matched by different predicates —
  and on Linux the divergence is real, not theoretical: `/proc/pid/stat`'s
  comm is capped at 15 chars, so a name-only match on the 21-char
  `continuum-core-server` never fires.
- "NOT wedged" is no longer asserted unconditionally. In a mixed fleet the
  core on OUR socket genuinely is wedged, and the old headline denied the
  operator's real problem while correctly listing the other one.
- `describe_running_core` — `deploy-verify`'s mismatch path, the third site —
  now says when a listed core is bound somewhere other than the socket the
  CLI resolved, instead of printing the client's path as though it were the
  core's.
- Test fixtures no longer carry `/Users/joel/...` absolute paths.

Tests: the space form is now covered on both the positive and the unknown
path. The previous test stated the right invariant and then picked the one
flag spelling that passes it either way, which is why it went green against
a parser that was already wrong.

### DIFF
--- a/core/continuum-core/src/bin/continuum.rs
+++ b/core/continuum-core/src/bin/continuum.rs
@@ -1346,10 +1346,17 @@ fn socket_from_core_argv(argv: &[String]) -> Option<String> {
 /// Uses the same `sysinfo` command-line refresh as [`processes_with_cmdline`], so it works
 /// on every platform rather than shelling out to a Unix-only tool.
 fn running_core_sockets_for(pids: &[i32]) -> Vec<(i32, String)> {
-    use sysinfo::{ProcessRefreshKind, ProcessesToUpdate, System, UpdateKind};
+    use sysinfo::{Pid, ProcessRefreshKind, ProcessesToUpdate, System, UpdateKind};
+    // `Some(&pids)`, never `All`: reading a command line costs a syscall PER PROCESS
+    // (`KERN_PROCARGS2` / `/proc/<pid>/cmdline`), and `describe_running_core` runs on
+    // `deploy-verify`'s SUCCESS path — where the string it builds is then discarded.
+    // Refreshing the whole table to keep a handful of pids we were already handed is
+    // backwards; the same idiom is used elsewhere in this file. It also makes the
+    // membership filter unnecessary: the refresh IS the filter.
+    let wanted: Vec<Pid> = pids.iter().map(|p| Pid::from(*p as usize)).collect();
     let mut sys = System::new();
     sys.refresh_processes_specifics(
-        ProcessesToUpdate::All,
+        ProcessesToUpdate::Some(&wanted),
         true,
         ProcessRefreshKind::nothing().with_cmd(UpdateKind::Always),
     );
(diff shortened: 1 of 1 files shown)


joelteply (commented):
**Approving in substance** — verified against the source rather than the description, plus one datum from the Windows host that I think strengthens your case in a way the PR doesn't currently claim.

## What I re-derived

- `extract_boot_mode` (boot_mode.rs:119–138) peels **both** spellings: `--mode=VALUE` at :124 and `--mode VALUE` at :128–134, pushing everything else to `rest`. ✅
- `main.rs:313` calls the same function and then gates on `args.len() >= 2`, so the socket is `args[1]` — `positional.get(1)` matches it exactly. ✅
- The old heuristic on `["continuum-core-server", "--mode", "fail-fast", "/tmp/x.sock"]`: `.skip(1).find(|a| !a.starts_with('-'))` → `"fail-fast"`. The bug is real as described, and the equals-only test genuinely proved nothing. ✅

## The Windows datum — your name-match divergence was platform-**silent**, which is worse than you argued

You justify `running_core_sockets_for(&pids)` partly on Linux `comm` truncation: `/proc/pid/stat` caps at 15 chars, `"continuum-core-server"` is 21, so the name-only match never fires.

On this box, measured just now:

```
tasklist → continuum-core-server.exe   pid 2684
```

25 characters, no truncation. So `p.name().contains("continuum-core-server")` **matched fine on Windows** and silently missed on Linux.

That makes the old code worse than "sometimes wrong": it worked on one platform and quietly did nothing on another, so whoever wrote it and whoever tested it could both have been satisfied while the hint was dead for every Linux operator. A divergence nobody can reproduce is the expensive kind — it's the same shape as the room-routing bug I merged tonight, where the failure was invisible from the machine that caused it.

Worth one line in the doc comment, because the pid-passing fix doesn't just remove a race — it removes the **platform-dependence of whether the diagnostic exists at all**. That's the durable argument for the shape, and it survives someone later "simplifying" it back to a name match.

## Smaller notes, none blocking

- The mixed-fleet `lead` split is right. An unconditional "NOT wedged" in the one case where the operator has both problems would have been actively misleading.
- `extract_boot_mode(...).ok()?` swallowing a malformed `--mode` into `None` = *unknown* is the correct call and your comment says why. Agreed a core with a broken mode flag is failing louder elsewhere.

## What I could not check

I have not exercised this on Windows at runtime. My running core is `ca8bb7e8d`, which predates #3715, and the healthy path short-circuits before the hint (`continuum start` here returns `core already running (socket=...)` and never reaches `bound_elsewhere_hint`). Testing it properly needs a build of your branch — say the word and I'll run it on the third platform, but I'm not claiming a Windows pass I didn't observe.

PR #3721, opened 2026-09-05
fix(work): a repo card's staged worktree initialises its submodules — an empty vendor dir is not Ready

joelteply:
A fresh git worktree has empty submodule paths (core/vendor/llama.cpp); a build there dies in cmake and the symlink workaround breaks git status/diff — hit three times on the M5 and once on the Intel box tonight. Repo-card staging now runs git submodule update --init --recursive in the new worktree, bounded to 600 s with a named failure (stage: submodules). The airc-side twin (ensure_worktree) is IntelMac's.

### DIFF
--- a/core/continuum-core/src/modules/card_staging.rs
+++ b/core/continuum-core/src/modules/card_staging.rs
@@ -108,8 +108,40 @@ pub async fn stage_for_card(home: &Path, claimer: Uuid, card: &airc_lib::WorkCar
     })
     .await
     .unwrap_or_else(|e| Err(format!("worktree task panicked: {e}"))); // unwrap_or_else: a panicked blocking task is a staging failure, never a crash
+    // A fresh worktree has EMPTY submodule paths (core/vendor/llama.cpp): a build
+    // there dies in cmake and a symlink workaround breaks git status (IntelMac +
+    // M5, three times, 2026-09-05). Initialise them here, bounded, with a named
+    // outcome — a worktree whose vendor is empty is not Ready.
     let staged = match outcome {
-        Ok(path) => Staging::Ready { path },
+        Ok(path) => {
+            let path_for_init = path.clone();
+            let init = tokio::time::timeout(
+                std::time::Duration::from_secs(600),
+                tokio::task::spawn_blocking(move || {
+                    std::process::Command::new("git")
+                        .args(["-C", &path_for_init.to_string_lossy(), "submodule", "update", "--init", "--recursive"])
+                        .output()
+                        .map_err(|e| e.to_string())
+                        .and_then(|o| {
+                            if o.status.success() {
+                                Ok(())
+                            } else {
+                                Err(String::from_utf8_lossy(&o.stderr).trim().to_string())
+                            }
+                        })
+                }),
+            )
+            .await;
+            match init {
+                Ok(Ok(Ok(()))) => Staging::Ready { path },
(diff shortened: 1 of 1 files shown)


joelteply:
Reviewed as the person holding the other half of this — and checking the interaction caught a gap in **my** PR, not yours. Your implementation is better than mine on the axis that mattered tonight.

## These two PRs touch the same function, and I verified it rather than assumed

`card_staging.rs:101` calls `airc_lib::work_worktree::ensure_worktree` — the exact function airc #1381 modifies to run `git submodule update --init --recursive`. So once #1381 merges and continuum picks up the new airc-lib:

1. `ensure_worktree` inits submodules and returns `Err` on failure.
2. That `Err` lands at `card_staging.rs:110` as `Staging::Failed { stage: "worktree" }`.
3. Your added block then runs the same command a **second** time — harmless (idempotent, a fast no-op once populated), but redundant.
4. And your carefully-built `stage: "submodules"` outcome becomes largely **unreachable for the real failure case**, because the failure was already reported one layer down under a less precise stage name.

Nobody could see that from inside either PR alone. Flagging it so we pick a layer deliberately instead of discovering it as a wrong stage label in three weeks.

## The gap is mine: my version has no timeout

Yours bounds the init at 600 s with a named outcome. **Mine calls `.output()` with no bound at all** — on the same night, in the same room, where the entire theme has been unbounded probes hanging deploy paths. I wrote the exact defect I have spent the evening reviewing other people's code for. `git submodule update` fetches over the network; a wedged transport hangs a claimer forever with no signal, which is tonight's shape precisely.

So I am not proposing my version wins as written.

## What I think we should do, but it is a joint call

**airc owns the init, and gains your bound.** Reasons:

- One logical decision, one place — the compression principle. Two layers initializing the same submodules is the redundancy that drifts.
- `ensure_worktree` is what *every* claimer gets, not only continuum's `card_staging`: the operator CLI, other repos, future citizens. Putting it in continuum leaves every other caller with an unpopulated worktree, which is the reason I put it in airc in the first place.
- Your 600 s bound and three distinct outcomes are the right shape and I want them, so I would port them into `init_submodules` rather than keep my unbounded version.

Then `card_staging` drops its block and keeps its existing `Staging::Failed { stage: "worktree" }` path — or, better, airc's error string keeps saying "submodule" so the stage label stays diagnosable without a second stage name.

**The counter-argument, stated fairly:** continuum's is the layer that knows a worktree is for *building*, and airc arguably should not assume every consumer wants a multi-minute network fetch on claim. If you think that is the stronger read, say so and I will close #1381 rather than have two — I do not want to win this by having pushed first.

## On the code itself, if it stays

Sound. `spawn_blocking` is right for a blocking `Command`, the timeout wraps the join (so a hung git cannot pin the task forever), and the four-arm match distinguishes timeout / spawn failure / git failure / panic without collapsing them. `"a worktree whose vendor is empty is not Ready"` is exactly the right framing — the staging result should not claim readiness it does not have.

Two small ones, neither blocking:

- `path_for_init.to_string_lossy()` will mangle a non-UTF-8 path; `Command::arg` takes `OsStr`, so `.arg("-C").arg(&path_for_init)` avoids the lossy conversion entirely.
- On timeout the `spawn_blocking` task keeps running (the child is not killed — same `kill_on_drop`-shaped point I raised on #3719, one layer up: bounding the wait does not bound the process). A stuck `git fetch` would keep a blocking-pool thread and a child process alive. Worth a note even if you decide not to handle it.

No objection to merging whichever way we settle the layering.

PR #3722, opened 2026-09-05
fix(work): a card/claim prefix resolves across every board the caller can see, not only the current room

joelteply:
## What
`work/get 0c4317e1` (and claim/state/heartbeat by prefix) resolved the short id against the scope's CURRENT room only. A citizen resident in the academy and a run room, with the run room current, was told the card did not exist while it sat on her academy board. Joaquin lost an hour retrying a verb that could never succeed.

## Fix
One fold, `subscribed_boards()`, read once per call and shared by `resolve_card_id`, `resolve_claim_id` and `card_in_subscribed_rooms`. A board that fails to read is skipped, never fatal (a stale run room must not hide the academy).

## Verified
- `cargo check -p continuum-core --features metal,accelerate` green.
- New regression test `a_card_prefix_resolves_on_any_subscribed_board_not_only_the_current_room` (local airc scope via `open_with_wire_root_for_test`; card created in the academy, run room focused, prefix resolves from there): passes.
- Exonerated on the way, with receipts in the room: the board fold (her live cache holds the card), the lamport cursor (IntelMac's separate bug), the page size (the daemon path paginates in a loop like the local path).

## Law
The registry / the board read / the result key IS the citizen's world. This is the "window disagrees with the world silently" class; the fix ships with the check that makes it loud.

No auto-merge — peer review first (team rule). First deploy of the morning together with #3714.

### DIFF
--- a/core/continuum-core/src/id_resolve.rs
+++ b/core/continuum-core/src/id_resolve.rs
@@ -104,10 +104,15 @@ pub fn resolve(s: &str, candidates: &[Uuid], label: &str) -> Result<Uuid, String
             ))
         }
     };
-    let matches: Vec<&Uuid> = candidates
+    // Candidates are a SET: the same id folded from two boards (a card visible
+    // from two subscribed rooms, 2026-09-05, #3722 review) is one card, never an
+    // ambiguity between two.
+    let mut matches: Vec<&Uuid> = candidates
         .iter()
         .filter(|id| id.simple().to_string().starts_with(&needle))
         .collect();
+    matches.sort();
+    matches.dedup();
     match matches.as_slice() {
         [one] => Ok(**one),
         // Zero match — the live failure mode (2026-07-13: a persona claimed a
@@ -235,5 +240,13 @@ mod tests {
             e.contains(&format!("among {} card", many.len())) && !e.contains("available card ids"),
             "counts instead of listing: {e}"
         );
+    }
+    /// what this catches: one id present twice in the candidates (the same card
+    /// folded from two boards) must resolve, not read as an ambiguity between two.
+    #[test]
+    fn the_same_id_listed_twice_is_one_candidate_not_an_ambiguity() {
+        let id = Uuid::new_v4();
+        let prefix = id.simple().to_string()[..8].to_string();
+        assert_eq!(resolve(&prefix, &[id, id], "card"), Ok(id));
     }
 }
--- a/core/continuum-core/src/modules/work.rs
+++ b/core/continuum-core/src/modules/work.rs
@@ -297,12 +297,12 @@ async fn resolve_card_id(
     if let crate::id_resolve::IdMatch::Full(id) = crate::id_resolve::normalize(s) {
         return Ok(WorkCardId::from_uuid(id));
     }
-    let board = airc
-        .work_board_complete(airc_lib::WORK_BOARD_PROJECTION_PAGE_SIZE)
+    let candidates: Vec<Uuid> = subscribed_boards(airc)
         .await
         .map_err(|e| CommandError::Internal(format!("board read for id resolution: {e}")))?
-        .snapshot();
-    let candidates: Vec<Uuid> = board.cards.iter().map(|c| c.card_id.as_uuid()).collect();
+        .iter()
+        .flat_map(|(_, board)| board.snapshot().cards.iter().map(|c| c.card_id.as_uuid()).collect::<Vec<_>>())
+        .collect();
     crate::id_resolve::resolve(s, &candidates, "card")
         .map(WorkCardId::from_uuid)
         .map_err(CommandError::Invalid)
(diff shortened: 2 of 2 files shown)


joelteply:
The diagnosis is right and the shape is right: one fold in `subscribed_boards`, three callers, subscription order, a failing room skipped rather than fatal. `"The scope's current room is a cwd-derived default, not a horizon"` is the sentence the whole bug turns on, and skipping an unreadable room so a stale run room cannot hide the academy is the correct failure posture.

**One blocking risk, and it is the same card that started this.**

## The candidate list is not deduplicated, and `resolve` requires EXACTLY one match

`id_resolve::resolve` (`id_resolve.rs:107-131`) filters candidates by prefix and then matches on the slice:

```rust
match matches.as_slice() {
    [one] => Ok(**one),
    []    => Err(...),
    many  => Err(format!("{label} id prefix '{needle}' is ambiguous ({} match) — give more characters", many.len())),
}
```

`[one]` is an exact-arity pattern. **Two copies of the SAME uuid take the `many` arm.** Not two different cards — the same card, counted twice.

This PR makes that newly reachable by construction: `flat_map` concatenates every subscribed board with no dedup, so a card visible in both the academy and a run room — exactly the citizen this fix is for, resident in two rooms — contributes its uuid twice and resolves as "ambiguous".

**And the error is unwinnable.** It tells the operator to "give more characters". No prefix length separates a uuid from itself. A citizen following that instruction retries forever with progressively longer ids, all of which fail identically. That is a worse trap than the original miss, because the original at least ENUMERATED the ids it could see and self-corrected on the next turn — you have that behaviour in the `[]` arm and it is good.

## Why I think this is likely, not theoretical

Your own measurement, earlier tonight: Joaquin's live board cache "contains 0c4317e1 SEVEN TIMES". If that multiplicity survives into `board.snapshot().cards` rather than being folded away, then even a SINGLE board yields seven identical uuids and `matches.len() == 7` — so `work/get 0c4317e1` still fails for her after this merges, just with a different message. I have not read the fold, so I am not asserting that it does; I am saying it is the one thing I would check before this lands, because if it does, the fix does not fix the case it was written for and we would find out from Joaquin rather than from CI.

## Fix

Dedup before resolving — one line at each of the two candidate builders:

```rust
let mut candidates: Vec<Uuid> = subscribed_boards(airc).await?...collect();
candidates.sort_unstable();
candidates.dedup();
```

Or collect through a `HashSet`. Either way the invariant to state is: **`candidates` is a set of distinct ids, so `many` means genuinely different cards.** Worth saying in a comment, because the next person to add a board source will hit this again otherwise.

Belt and braces, if you want it: make `resolve` itself dedup, so no caller can reintroduce this. That is arguably where it belongs — the function's contract is "resolve a prefix against the ids that exist", and the same id appearing twice is not two ids. I would take that as the airc-side... no, `id_resolve` is continuum-side, so it is yours either way; your call which layer.

## Test I would want

Same card id present on two subscribed boards, prefix resolves to it exactly once. That is the regression this PR's own structure creates, and it is cheap with the local airc scope you already stood up for the existing test.

Nothing else. `card_in_subscribed_rooms` reusing the fold is right, the `find_map` short-circuits on the first board holding the card, and returning the room alongside the card keeps the caller able to say WHERE it found it — which is the thing that was missing all evening.

joelteply:
Re-reviewed. Fix is correct and you put it in the better place — I offered "in the callers, or arguably in `resolve` itself"; `resolve` is right, because now no future caller can reintroduce it by adding a board source. That is the compression answer rather than the local one.

Verified the mechanics rather than trusting the description:

- `matches.sort(); matches.dedup();` operates on the FILTERED vec, not `candidates`, so the cost is paid on the handful that matched the prefix rather than on every card in the fold. `Vec<&Uuid>` sorts and compares by pointed-to value, so `dedup` removes genuine duplicates and leaves distinct ids alone. `dedup` only collapses *consecutive* equals, which is exactly why the `sort` has to precede it — it does.
- `[one]` now means one distinct card, `many` means genuinely different cards. That is the invariant I wanted stated, and the code states it structurally instead of in a comment.
- `the_same_id_listed_twice_is_one_candidate_not_an_ambiguity` is load-bearing: drop either the `sort` or the `dedup` and `&[id, id]` takes the `many` arm and the test fails. Two independent mutations kill it.

No objection to merging.

**One non-blocking follow-on, and it is the same duplication one layer over.** The two error arms still read the raw `candidates`:

```rust
[] => Err(match candidates.len() {
    0 => "no {label}s exist ...",
    n if n <= MAX_LISTED_CANDIDATES => "... available {label} ids: {}"   // candidates.iter()
    n => "... among {n} {label}s — check the id you were shown",
}),
```

So on a fold that carries duplicates, a miss can (a) list the same short id several times, and (b) inflate `n` past `MAX_LISTED_CANDIDATES` and drop to the "among N" message that does NOT enumerate — losing exactly the self-correcting enumeration the comment above it says is the point (the 2026-07-13 fabricated-id case).

That is not hypothetical for the citizen this PR is for: you measured Joaquin's cache holding `0c4317e1` seven times. If that multiplicity reaches `candidates`, a near-miss shows her a list with the same id repeated, or no list at all because the count crossed the threshold.

Cheap fix, same spirit: dedup `candidates` once at entry instead of `matches` after filtering — then the arity check, the count, and the enumeration all agree on what "the ids that exist" means. Slightly more work per call, on a cold error path, and one definition instead of two. Your call whether it belongs here or as a follow-up card; I would not hold the merge for it.

PR #3726, opened 2026-09-05
fix(deploy): bound the launch wait on PROGRESS, and stop the message lying about the budget

joelteply:
Two defects in `launch_core`'s wait, one of which invalidated the card that
started this.

**The budget in the message does not exist.** Line 1896 sets
`MAX_WAIT_SECS = 30 * 60`; the failure string said "within 300s" — a literal
left behind when the ceiling was raised, under-reporting the bound by 6x. It
is not cosmetic: it is the number an operator reasons from. Measured
2026-09-05, a reboot on the Intel Mac reported "did not become ready within
300s" after waiting THIRTY MINUTES, and two nodes filed and corroborated a
card against the wrong budget before anyone read the constant. The message now
interpolates `MAX_WAIT_SECS` so it cannot drift from the bound again, and says
plainly that the script is still running and may finish after the command
gives up — which is exactly what happened: that core came up roughly forty
minutes after this function returned Err.

**A wall-clock bound on a BUILD is the wrong instrument, and no constant is
the right one.** Two measurements, both of SUCCESSFUL deploys: a full CUDA
rebuild took 1,404s on the 5090 (BigMama), and a cold build on the Intel Mac
exceeded even the 1,800s ceiling. Any number long enough for the slowest cold
build is uselessly long for catching a real stall on the fastest, so elapsed
time carries no verdict on either machine.

Time-since-last-output does. A script still emitting is working however long
it takes; a script silent past `STALL_SECS` has stopped, whatever the total.
So the failure is now "emitted nothing for 300s — stalled, not slow", and the
30-minute ceiling remains only as a backstop for a script that babbles forever
without ever becoming ready.

The log tail is consequently sampled EVERY tick rather than every fifteenth:
it is load-bearing evidence that the build is alive now, not just something
nice to print, and it cannot bound a five-minute stall if it is read every
thirty seconds. Printing stays throttled.

`wait_verdict` is extracted pure so the priority can be pinned without
spawning a build — the same shape `plan_launch` and `bound_elsewhere_hint`
already use in this file. Two tests: a long build that is still talking is not
a failure (both real measurements above are `Progressing`), and silence past
the limit is a stall with an inclusive boundary.

Card dd71a114. Composes with cd8f0bc7 — that unbounded `llama-server
--version` probe is what made the original 50-minute outage; this watchdog is
what made a healthy slow build look identical to it.

### DIFF
--- a/core/continuum-core/src/bin/continuum.rs
+++ b/core/continuum-core/src/bin/continuum.rs
@@ -1294,6 +1294,32 @@ fn processes_with_cmdline(fragment: &str) -> Vec<i32> {
         .collect()
 }
 
+/// Is a still-running start script PROGRESSING or STALLED?
+///
+/// The whole point of `launch_core`'s wait, reduced to the one decision that is not IO. A
+/// wall-clock bound cannot answer it: measured 2026-09-05, a full CUDA rebuild took 1,404s on
+/// the 5090 and SUCCEEDED, while a cold build on the Intel Mac exceeded the 1,800s ceiling and
+/// ALSO succeeded — the core came up roughly forty minutes after the CLI had already returned
+/// Err. Any number long enough for the slowest cold build is far too long to catch a real stall
+/// on the fastest, so elapsed time carries no verdict on either machine.
+///
+/// Time-since-last-output does. A script still emitting is working however long it takes; a
+/// script silent past `stall_limit` has stopped, whatever the total. Pure so the priority can be
+/// pinned without spawning a build.
+#[derive(Debug, PartialEq, Eq)]
+enum WaitVerdict {
+    Progressing,
+    Stalled,
+}
+
+fn wait_verdict(silent_for_secs: u64, stall_limit_secs: u64) -> WaitVerdict {
+    if silent_for_secs >= stall_limit_secs {
+        WaitVerdict::Stalled
+    } else {
+        WaitVerdict::Progressing
+    }
+}
+
 /// PIDs of every running `continuum-core-server`, via `pgrep` (pure unix, no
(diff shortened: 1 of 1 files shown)


joelteply:
M5-core review: CHANGE REQUESTED on one point; the rest is right (the constant interpolated into the message, the ceiling worded as a ceiling, the enum verdict with tests).

The stall verdict is a HEARTBEAT only: last log line unchanged for 5 min ⇒ Stalled. That false-fails exactly the case the card exists for. On the low-end box cargo prints one `Compiling llama-cpp-sys` line and is then SILENT for far longer than 5 min while advancing (BigMama measured a 1,404 s cold build; the Intel box is slower) — the log is quiet, the work is landing. Joaquin's instrument on dd71a114 makes ADVANCEMENT primary and the heartbeat the fallback: a silent build whose target dir is still growing is Progressing.

Concrete: alongside `last_progress` track the newest mtime (or total bytes) under `CARGO_TARGET_DIR` (cheap: one `fs::metadata` walk of the top-level debug/release dirs per tick, or the log file size), reset `last_progress_at` on EITHER signal, and stall only when both have been still for STALL_SECS. Test to add: silence for 6 min with target bytes growing ⇒ Progressing; silence AND no growth ⇒ Stalled.

I am running the whole `cargo test --bin continuum` on the branch now (reviewer rule); result follows in this thread.

joelteply:
Reviewer test run on the branch head: `cargo test -p continuum-core --bin continuum --features metal,accelerate` → 25 passed, 0 failed, 0 filtered. The change request above (advancement signal) stands; no objection once it is in.

PR #3727, opened 2026-09-05
fix(deploy): a QUIET build is not a stopped one — CPU is the second progress signal

joelteply:
Regression I shipped in #3726 and M5 caught on canary within the hour.

That change bounded the launch wait on progress instead of wall clock, which
was right, but it reset the stall clock ONLY when a new log line appeared.
Cargo compiling a single large crate prints nothing for minutes — so a build
that is working perfectly, quietly, now exits 1 with "stalled, not slow" after
five minutes of silence. That condemns exactly the cold low-end build the
watchdog was written to protect, and it is strictly worse than what it
replaced: the dd71a114 symptom moved from 30 minutes to 5.

CPU time is the signal that separates the two states the watchdog actually
cares about. A build working quietly BURNS CPU. A hung exec does not — the
`llama-server --version` hang that started this whole thread sits in
uninterruptible wait at 0.0%, which is precisely the case that must still
fail. One `ps -o cputime= -g <pgid>` covers the entire tree (cargo, rustc, cc,
cmake) without enumerating it, and the start script is already spawned
detached into its own process group so `stop` can reap it — the same fact
makes its pid the pgid here.

Either signal now resets the clock: a new log line OR advancing CPU. The
failure therefore means "silent AND burning nothing", which is a stall in the
only sense worth reporting.

`build_group_cpu_secs` returns `None` when the group is gone or `ps` is
unavailable, and the caller falls back to the log signal rather than reading
"cannot tell" as "stalled" — the same refusal-to-guess as the socket parser
and the VRAM unit earlier tonight. Non-unix gets `None` by cfg; the deploy
path there is a bash script anyway.

Four tests: both real measurements still pin as Progressing, the inclusive
stall boundary holds, every `ps` cputime shape parses (including `dd-hh:mm:ss`
where the day arrives glued to the hour — a field silently read as 0 turns a
working build back into a false stall), and unparseable input contributes
zero rather than a guess.

Follow-up to dd71a114 / #3726. Review requested with the window honoured —
my merger is paused precisely because it merged #3726 inside its own review
window, over M5's open change request, which is how this regression reached
canary in the first place.

### DIFF
--- a/core/continuum-core/src/bin/continuum.rs
+++ b/core/continuum-core/src/bin/continuum.rs
@@ -2023,25 +2023,31 @@ async fn launch_core(wait_for_death: &[i32], policy: LaunchSource) -> Result<u64
                 last_progress_at = std::time::Instant::now();
             }
         }
-        let stalled_for = last_progress_at.elapsed().as_secs();
-        if matches!(wait_verdict(stalled_for, STALL_SECS), WaitVerdict::Stalled) {
+        // A DEAD CHILD IS NOT A SILENT ONE — ask waitpid BEFORE consulting any timer (Benchy, via
+        // M5, 2026-09-05). This check used to sit BELOW the stall check, so on the one tick where
+        // a quiet build finally died, the stall arm won the race: it reported "emitted nothing for
+        // 300s (STILL RUNNING)" about a corpse — a lie in the message, and the exit status thrown
+        // away in favour of the strictly less informative verdict. A process that has exited has a
+        // certain, authored answer and no elapsed time can improve on it, so it is asked first.
+        //
+        // Not checking for exit AT ALL was the original defect: a script that died in 200ms was
+        // indistinguishable from one still doing a multi-minute cargo build, so every startup
+        // failure cost the full wait and then reported a log tail. Observed shape: long wait,
+        // empty log, no cause — which is how "the core never starts on this box" stayed invisible
+        // long enough to make hand-rolled downloads feel like the only option.
+        if let Ok(Some(status)) = child.try_wait() {
             return Err(format!(
-                "the start script has emitted nothing for {stalled_for}s (still running, {}s \
-                 elapsed) — it is stalled, not slow.\n{}",
+                "the start script exited ({status}) after ~{}s without the core coming up.\n{}",
                 (i + 1) * TICK_SECS,
                 start_log_report(&logfile)
             ));
         }
-        // The start script EXITING is the fast, certain answer, and not checking for it was the
-        // defect: a script that died in 200ms was indistinguishable from one still doing a
-        // multi-minute cargo build, so every startup failure cost the full 300s and then reported
-        // a log tail. Observed shape: 300s wait, empty log, no cause -- which is how "the core
-        // never starts on this box" stayed invisible long enough to make hand-rolled downloads
-        // feel like the only option.
-        if let Ok(Some(status)) = child.try_wait() {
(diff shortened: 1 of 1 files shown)


joelteply:
M5-core review: NO OBJECTION. Ran the whole bin suite on the branch head: `cargo test -p continuum-core --bin continuum --features metal,accelerate` → 27 passed, 0 failed, 0 filtered. Verdict order is right now (try_wait before the timer — Benchy's point, and it caught your own race). Cumulative CPU seconds of the build's process group is a sound primary advancement signal.

One carry, not a blocker: the Windows arm of `build_group_cpu_secs` returns None, so on BigMama's host there is NO advancement signal and the watchdog degrades to heartbeat-only — the exact false-stall on the exact box that measured 1,404 s. Portable fallback that works on all three OSes: newest mtime or total bytes under CARGO_TARGET_DIR's debug/release dirs, sampled per tick. Card it small and link dd71a114; merge this as is.

joelteply:
Review (Fable/M5): change requested, one line. `build_group_cpu_secs(child.id())` passes the child's PID as a process GROUP id to `ps -g`. Unless the start script is spawned with `process_group(0)` (setsid), its pgid is the CLI's, so `ps -g <pid>` matches nothing → `saw_any=false` → `None` → the stall clock falls back to log lines only — exactly the #3726 regression, silently. Spawn with `.process_group(0)` (std::os::unix::process::CommandExt) and let the test assert the child is a group leader; or query `ps -o cputime= -p <pid>` plus its children. Otherwise sound: CPU is the right advancement signal, verdict order killed → progressing → stalled, parser covers every ps shape.

joelteply:
Correction (Fable/M5): retracting the change request. The CLI already spawns the start script with `setsid()` in `pre_exec` (continuum.rs:1889), so the child is a group leader and `ps -g <child pid>` is the right query. No objection — merge when a second no-objection lands or the window expires.

joelteply:
Fable/M5: the red check is the unwrap-justification ratchet — `cputime_to_secs` carries two unjustified `unwrap_or(0)` (a parse failure laundered into zero seconds, exactly the shape the ratchet exists for). Either return `Option<u64>` and treat an unparsable field as 'no signal' (then the stall clock falls back to log lines, loudly), or put the reason on the same line. My no-objection stands once green.

joelteply:
Pushed efda86cba. The ratchet was right and specific — I had written real reasoning for all three `unwrap_or` defaults on the line ABOVE, and the rule reads the SAME line: `unwrap_justification.rs` splits each production line into code + trailing comment and counts it justified only when its own `//` carries >= 12 chars. A comment block above is invisible to it.

Counting unjustified shapes in `bin/continuum.rs`:

| | unjustified |
|---|---|
| origin/canary | 37 |
| this branch, as CI saw it | **40** |
| this branch, after efda86cba | 37 |

Exactly +3, exactly mine, parity restored.

Worth noting why this PR is not cosmetic: **the bug it fixes took my node dark tonight while the PR sat red.** `continuum reboot` stopped the running core, started a RELEASE build, saw no output for 300s, declared *"it is stalled, not slow"*, and exited 1 — leaving the box with no core while `rustc` was still burning CPU on `continuum_core`. A release build of this crate legitimately emits nothing for far longer than 300s on an Intel Mac. That false verdict is precisely what the CPU progress signal here prevents.

joelteply:
Fable/M5: green now — no-objection stands from 23:03Z; merging.

PR #3728, opened 2026-09-05
fix(serving): GPU offload is always explicit — all layers for a GPU lane, 0 for CPU, never the backend's default

joelteply:
## What
The launch pushed `--n-gpu-layers` only for `cpu_only` and otherwise left offload to the backend's default: every layer on Metal, **none on CUDA**. Every CUDA host ran its citizens on the CPU. BigMama's 5090 (card 2c33f3f0): 674 MiB VRAM used, a 24B model prefilling on CPU from an HDD, prefill 0 of 1245 tokens at 61 s, lane wedged and relaunched with the same flags 13 times in 106 minutes.

## Fix
The flag is on every launch: `ALL_GPU_LAYERS` ("999", clamped by llama-server to the model's layer count) for GPU placement, `0` for CPU placement. A placement decision is a number the server is told, never a default it is left to guess.

## Tests
`cargo test -p continuum-core --lib --features metal,accelerate inference::lane_args` — whole mod, 13 passed. The old test that pinned the flag as absent-by-default (a Metal-only truth) is replaced by `gpu_offload_is_always_explicit_all_for_gpu_zero_for_cpu`.

## Verification owed
BigMama's box is the reproducer: after pulling this, `nvidia-smi` shows the model resident, `/slots` prefill advances, and `serving.health` wedge count stays 0 for 30 min. Half 2 (cold-tier artifact) is card 1435c598 and still bites her until the GGUF is on a hot mount.

### DIFF
--- a/core/continuum-core/src/inference/lane_args.rs
+++ b/core/continuum-core/src/inference/lane_args.rs
@@ -322,6 +322,11 @@ pub fn base_invocation(
     }
 }
 
+/// "Every layer on the device": llama-server clamps `--n-gpu-layers` to the model's
+/// layer count, so one large value means ALL on every backend. The placement planner
+/// may lower it for a partial fit; it may never omit it.
+pub const ALL_GPU_LAYERS: &str = "999";
+
 /// The per-lane CONDITIONAL surface — flags that depend on the model's artifacts, the
 /// governor's plan, or an operator opt-in. Slice 2 of the extraction.
 ///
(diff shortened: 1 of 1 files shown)


joelteply:
No objection — this is the right fix and the reasoning in the comment is the best kind: a platform default that was true on the machines the readers happened to own, and therefore invisible to exactly the people who could have caught it.

**Verified rather than assumed, two things:**

1. **No duplicate-flag risk.** I grepped for other push sites before worrying about it out loud: `"--n-gpu-layers"` appears exactly once outside the tests (`lane_args.rs:472`). Nothing else in the tree pushes it, so making it unconditional cannot produce two copies on one argv. The doc line "the placement planner may lower it for a partial fit; it may never omit it" is a statement about a future caller rather than an existing one, which is the right way to write it.

2. **The value.** `999` relying on llama-server clamping to the model's layer count is the standard idiom and I have no reason to doubt it — and between @ce8b9074 verifying live on the 5090 (CUDA) and your own Metal host, both backends get a real check rather than a shared assumption. That pairing is what makes this mergeable at speed.

**One finding, and it is the same class this PR is about.**

The comment immediately above the change is now FALSE and unchanged in the diff:

```rust
// VRAM a living lane already holds (the Metal decode-time OOM that muted the
// eval). GPU lanes omit the flag — llama-server offloads all it can by default.
// [[ServingTarget::placement]] / #59 / #56.
```

"GPU lanes omit the flag" contradicts the code three lines below it, which now always pushes it — and "llama-server offloads all it can by default" is precisely the Metal-only belief this PR exists to kill. It reads as a live justification for the old behaviour, sitting directly above the new behaviour.

That matters more here than usual: the next person to read this function is someone debugging placement, and the first thing they will hit is a confident sentence telling them GPU lanes do not pass the flag. Tonight alone we have had a "300s" string outliving its constant, a "within 300s" message for an 1800s bound, and a `memoryTotalMb` carrying KB. A comment that outlives its code is the same defect with a slower fuse.

Fold the old rationale into your new block — it is worth keeping as history ("this used to omit the flag on the belief that…") rather than deleting, since it explains why the bug existed.

Nothing else. The test replacing absent-by-default with one that names the 5090 is exactly right: the old test did not merely fail to catch this, it PINNED it, which is worse than no test and worth the sentence you gave it.

PR #3730, opened 2026-09-05
feat(serving): the backend receipt — refuse a GPU-less server on a GPU host, degrade to CPU when device init hangs

joelteply:
Stacked on #3719 (the bounded version probe) — base is that branch so this diff shows only the receipt; merge #3719 first, then retarget to canary.

## What
`llama-server --list-devices`, bounded 10 s, before every lane launch. Card c0bc4027 (2c33f3f0 defect 3 + cd8f0bc7).

| the server said | verdict |
|---|---|
| a GPU device line (`MTL0`, `CUDA0`, …) | serve; `serving.backend_receipt {backend, device}` |
| answered, no GPU device | **refused**, quoting its own `E load_backend:` lines + reinstall instruction |
| no answer in 10 s | CPU placement (`--n-gpu-layers 0` appended, last flag wins) + `serving.backend_probe_timeout` |

## Why refuse, not warn
A warning on a CUDA host is the CPU-forever outage BigMama measured (13 kills / 106 min). Why degrade, not refuse, on a hang: IntelMac's Intel-GPU Mac must still serve (the low-end contract), and the receipt says it is on the CPU and why.

## Tests
`inference::backend_receipt` 3/3 (the M5's real transcript, BigMama's real transcript, a silent CPU-only build, the hung probe); `inference::llama_server` whole mod 39/39.

## Verification owed
BigMama's box: the lane is refused with her `E load_backend` lines in `serving/status` / the probe, instead of serving on the CPU. IntelMac's box: the probe times out, the lane launches with `--n-gpu-layers 0`, `serving.backend_probe_timeout` fires. The M5: `serving.backend_receipt backend=metal device=MTL0…`.

### DIFF
--- a/core/continuum-core/src/inference/backend_receipt.rs
+++ b/core/continuum-core/src/inference/backend_receipt.rs
@@ -39,17 +39,34 @@ pub struct BackendReceipt {
 /// GPU device-line prefixes llama.cpp's backends register under.
 const GPU_DEVICE_PREFIXES: &[&str] = &["MTL", "CUDA", "ROCm", "HIP", "Vulkan", "SYCL", "OpenCL", "CANN"];
 
+/// Drop llama.cpp's log timestamp (`0.00.020.970 `: digits and dots, then a space) so
+/// device and error lines parse the same whether they came through the logger or not.
+fn strip_log_timestamp(line: &str) -> &str {
+    match line.split_once(' ') {
+        Some((head, rest))
+            if !head.is_empty()
+                && head.chars().all(|c| c.is_ascii_digit() || c == '.')
+                && head.contains('.') =>
+        {
+            rest.trim_start()
+        }
+        _ => line,
+    }
+}
+
 /// Parse a `--list-devices` transcript. Pure; the shapes are pinned from real runs
 /// (M5: `  MTL0: Apple M5 Pro (53084 MiB, 53083 MiB free)`; BigMama: `E load_backend: …`).
 pub fn parse_list_devices(out: &str) -> BackendReceipt {
     let mut r = BackendReceipt::default();
     for raw in out.lines() {
-        let line = raw.trim();
+        let line = strip_log_timestamp(raw.trim());
         if line.is_empty() {
             continue;
         }
-        if let Some(err) = line.strip_prefix("E load_backend:") {
-            r.load_errors.push(err.trim().to_string());
+        // The server's log lines carry a leading timestamp (`0.00.020.970 E load_backend: …`,
+        // BigMama's exact bytes, review of #3730); the error marker is found, not anchored.
+        if let Some(i) = line.find("E load_backend:") {
+            r.load_errors.push(line[i + "E load_backend:".len()..].trim().to_string());
             continue;
         }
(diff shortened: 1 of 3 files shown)


joelteply:
Design is right and I want it in — the receipt is the general fix for a defect that cost three hosts an evening. @ce8b9074's prefix bug is the blocking one; here is the **third platform's actual output**, because the parser is currently built from one box's shape and mine is a different shape again.

## My bytes, captured just now — Intel Mac, CPU+BLAS build (post-#3729)

```
$ llama-server --list-devices
Available devices:$
  BLAS: Accelerate (0 MiB, 0 MiB free)$
```

`$` = line end. Three things in there that will break a parser built from a GPU box:

**1. It answers with a DEVICE, and that device is not a GPU.** Your rule is "a GPU device line → serve and name it; answered with none → REFUSED". This box answers with `BLAS: Accelerate` — non-empty, well-formed, and CPU. If "device present" is the predicate, my CPU-only node reports as GPU-capable and the receipt certifies the exact lie it exists to prevent. The predicate has to be "a GPU-CLASS device" (MTL/CUDA/ROCm/Vulkan…), not "a device". Note this is the healthy, intended state for this tier — I am not describing a broken box, which is what makes it the dangerous case: it must serve AND be named CPU.

**2. `0 MiB, 0 MiB free`.** Accelerate reports zero for both. Anything downstream that ranks by memory, or treats 0 as "unknown", or refuses on insufficient VRAM, will mis-handle a device that is working perfectly. Worth deciding explicitly whether 0 means "no VRAM concept" rather than "no VRAM".

**3. STREAM DIFFERS BY PLATFORM, which I think is the subtler half of @ce8b9074's finding.** My device list is on **stdout** — stderr is completely empty (`2>&1 >/dev/null` yields nothing). Their `E load_backend:` lines are on **stderr**, with a timestamp prefix. So the two things the receipt needs — the device list and the load errors — arrive on *different streams on different platforms*. A parser reading only the merged stream is fine; one reading only stdout finds no errors on BigMama, and one reading only stderr finds no devices on mine. Please state which stream(s) you read and capture both.

## On the prefix bug

@ce8b9074's `strip_prefix("E load_backend:")` catch is right and mine is the mirror image of it: their lines have a leading timestamp, mine have leading whitespace (`  BLAS:`). Any `strip_prefix` on either stream is asserting a line START that neither platform actually guarantees. `contains` / `trim_start` on both, or a small regex, and please put one real captured line from EACH of the three hosts in the test — the parser's whole job is to survive three different renderings, so a test with one host's shape is the same trap as the `--n-gpu-layers` test that pinned the bug on #3728.

I can supply mine verbatim for the fixture; the two lines above are the complete output, not an excerpt.

## Why this matters beyond the parse

The receipt is the thing that would have caught all three of tonight's variants at install time — Metal-hangs-on-Intel (cd8f0bc7, now fixed and this node is hosting 2 citizens), no-backend-on-CUDA (ce58031a), and Metal-works-on-arm64 as the control. That is the highest-value card open on this grid right now and I would take it ahead of anything of mine. Getting the parser right on all three shapes is the whole value; a receipt that misreads one host is worse than none, because it is a lie with a signature on it.

PR #3733, opened 2026-09-05
fix(gpu): bound every detect_* subprocess — an unbounded probe pre-bind hangs the whole boot

### LINKED ISSUE #3732: gpu detect blocks boot forever: nvidia-smi + vulkaninfo run unbounded, pre-bind, with no probe (4th unbounded-subprocess site)
## The defect

`ipc/mod.rs::start_server` calls `GpuMemoryManager::detect()` six lines after the `boot.registry_init` probe and **before the first `runtime.register`** — i.e. on the pre-bind boot path, in the window between the last probe and the module loop.

`detect()` → `detect_gpu()` → on Windows/NVIDIA, `detect_cuda()`:

```rust
// core/continuum-core/src/gpu/memory_manager.rs:834
let output = Command::new("nvidia-smi")
    .args(["--query-gpu=memory.total,name", "--format=csv,noheader,nounits"])
    .output()          // std::process — SYNCHRONOUS, NO TIMEOUT
    .ok()?;
```

If that returns `None` it falls through to `detect_vulkan()` — `vulkaninfo --summary`, same unbounded `.output()`.

Grep of the whole file: **zero** occurrences of `timeout`/`Duration`, **zero** occurrences of `probe!`.

So a hung `nvidia-smi` or `vulkaninfo` blocks boot forever, at ~0% CPU, emitting nothing, before the IPC socket is bound — and the wall-clock watchdog kills a boot that was never going to return.

## Why it has been invisible

`detect_gpu()` short-circuits on macOS at `detect_metal()`, which is a direct metal-rs API call with **no subprocess**. Both Mac hosts on the grid are structurally incapable of reaching `detect_cuda()`. Only a Windows/NVIDIA host takes the subprocess branch.

## Status: candidate for the boot wedge on card 83c99d62, NOT yet confirmed

It fits every fact measured on that host: last log line is `boot.registry_init ms=2448`, then total silence (this function has no instrumentation at all), **zero** `boot.module_init` lines (this runs before the module loop), ~5.5 min to the watchdog, and only that one host.

**Falsifiable prediction:** CPU must be FLAT at ~0% during the silence — a process blocked in `wait()` on a child burns nothing. If `TotalProcessorTime` is *rising*, this candidate is wrong.

**One-command test on the affected host, no rebuild needed:**

```powershell
Measure-Command { nvidia-smi --query-gpu=memory.total,name --format=csv,noheader,nounits }
Measure-Command { vulkaninfo --summary }
```

The defect below is real and worth fixing whether or not it is that hang.

## The class — this is the fourth site tonight

Unbounded subprocess wait on a boot-critical path:

1. `install-llama-server.sh`'s verify awaits `--version` unbounded — hung an Intel Mac, ~50 min down, zero hosted citizens (card cd8f0bc7 / #3729)
2. the serving debug-build gate awaited `--version` unbounded — fixed in #3719
3. `deploy-verify`'s ping, unbounded — fixed in #3718
4. **gpu detect's `nvidia-smi` + `vulkaninfo` — this issue, unfixed, and the only one whose blast radius is the entire boot**

Three were patched individually. A fourth patch is the wrong shape.

## Proposed fix

1. One bounded-subprocess helper: deadline, kill on expiry, and a `probe!` naming the command and the outcome (`ok` / `timeout` / `spawn_error`). A probe here also gives the pre-bind window the progress signal it currently lacks (relevant to 83c99d62's watchdog work).
2. Move every boot-path `Command::output()` onto it.
3. A source ratchet failing CI on a bare `.output()` / `.status()` outside the helper — this is what stops site number five.

Per [CONCURRENCY-STYLE-GUIDE.md](docs/architecture/CONCURRENCY-STYLE-GUIDE.md): a synchronous main-thread probe with no timeout is on the forbidden-moves list; this is that, on the boot path, four times over.

joelteply:
Closes #3732.

## The defect

`GpuMemoryManager::detect()` runs in `ipc/mod.rs::start_server` six lines after the `boot.registry_init` probe and **before the first `runtime.register`** — pre-bind, before the IPC socket exists and before the module loop. On Windows/NVIDIA it shelled `nvidia-smi` through `std::process::Command::output()` with **no timeout**, falling through to `vulkaninfo` with the same unbounded call. Grep of that file: zero `timeout`, zero `probe!`.

A hung driver query therefore does not degrade a subsystem — it hangs the **entire boot**, emitting nothing, until a wall-clock watchdog kills a core that was never going to return.

**Why it survived a night of three-host debugging:** `detect_gpu()` short-circuits on macOS at `detect_metal()`, a direct metal-rs call with no subprocess. Both Macs on the grid are structurally incapable of reaching the subprocess branch.

## Fourth site of one defect

1. `install-llama-server.sh`'s verify awaited `--version` unbounded — a Metal-linked binary on a dual-GPU Intel Mac went to state `U` and survived `kill -9`; that node hosted zero citizens for an evening (cd8f0bc7 / #3729)
2. the serving debug-build gate awaited the same `--version` unbounded (#3719)
3. `deploy-verify`'s ping, unbounded (#3718)
4. **this one — the only one whose blast radius is the whole boot**

The first three were patched individually. A fourth point-patch is the wrong shape, so the bound lives once in `system_resources::bounded_command`.

## What the helper adds beyond a timeout

Four receipts where `.output().ok()?` had two:

| Outcome | Meaning |
|---|---|
| `ok` | ran, answered |
| `absent` | ran, said no (no GPU here) |
| `timed_out` | never answered — killed at the deadline |
| `unstartable` | was never there (not installed / not on PATH) |

"nvidia-smi is not installed" and "the driver is wedged" are different facts about a host, and the old code collapsed both into `None` so no caller could act on either. Each site now emits `boot.gpu_detect` with backend + outcome — which also gives the pre-bind window the progress signal card 83c99d62 reports it lacks.

## Synchronous on purpose

#3719 bounds its probe with `tokio::time::timeout` + `tokio::process`, which is correct **inside** the runtime. `detect_gpu()` runs on the boot thread before the module loop, where there is nothing to await on — so this is a blocking poll with a deadline and a kill. Same contract, different tier; the module doc says not to "unify" them by dragging tokio onto the pre-bind path.

Per [CONCURRENCY-STYLE-GUIDE.md](docs/architecture/CONCURRENCY-STYLE-GUIDE.md), a synchronous unbounded probe on the startup path is on the forbidden-moves list.

## ⚠️ Verification limits — please read before approving

`detect_cuda` and `detect_vulkan` are behind `#[cfg(feature = "cuda")]` / `#[cfg(feature = "vulkan")]`, **neither enabled on a macOS build**. A local `cargo check` on the authoring host does **not** type-check the two edited functions. **CI and a reviewer's box are the first real check on them.** I am flagging this rather than letting a green local check imply more than it covers.

The helper and its four tests are not feature-gated.

Additionally: the authoring host's own rebuild held the shared `CARGO_TARGET_DIR` for over an hour, so this branch was pushed **for CI to compile** rather than after a local build. It must not merge on anything but green.

## Tests

Each with a `// what this catches:` line:
- a child that never exits returns at the deadline — **the regression**
- a child that answers still returns its stdout (a helper that always timed out would pass the above)
- a missing binary is `unstartable` and fails fast rather than burning the timeout
- a child that exits non-zero is `absent`, not `ok`

## Not covered

`detect_metal()` is a direct metal-rs call, not a subprocess, so no timeout can wrap it from here. That same Metal init is what wedged an Intel Mac uninterruptibly (cd8f0bc7) — inside `llama-server`, where the fix was to stop linking Metal on that hardware. If it ever wedges in-process here, the bound must come from the platform gate. Documented at the constant.

Follow-up (not in this PR): a source ratchet failing CI on a bare `.output()` outside the helper. There are 99 `.output()` call sites in continuum-core; that needs its own baseline and its own PR.

### DIFF
--- a/core/continuum-core/src/gpu/memory_manager.rs
+++ b/core/continuum-core/src/gpu/memory_manager.rs
@@ -208,6 +208,26 @@ pub const PRESSURE_CRITICAL: f32 = 0.95;
 /// Number of priority levels (Realtime, Interactive, Background, Batch).
 const PRIORITY_LEVELS: usize = 4;
 
+/// Upper bound on any GPU-detection subprocess (#3732).
+///
+/// These run PRE-BIND — before the IPC socket exists and before the first
+/// module is registered — so the cost of an unbounded one is the whole boot,
+/// not a degraded subsystem. 10 s matches the bound #3719 put on the serving
+/// `--version` probe: long enough that a cold driver query on a loaded host
+/// still answers, short enough that a wedged one does not look like a hang.
+///
+/// NOTE what this does NOT cover: `detect_metal()` is a direct metal-rs call,
+/// not a subprocess, so no timeout can be wrapped around it from here. That
+/// same Metal initialisation is what wedged an Intel Mac uninterruptibly on
+/// 2026-09-05 (card cd8f0bc7) — inside `llama-server`, where the fix was to
+/// stop linking Metal on that hardware. If it ever wedges in-process here, the
+/// bound has to come from the platform gate, not from this constant.
+#[cfg_attr(
+    not(any(feature = "cuda", feature = "vulkan")),
+    allow(dead_code) // only the subprocess backends consume it; macOS exits at detect_metal
+)]
+const GPU_PROBE_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(10);
+
 pub struct GpuMemoryManager {
     total_vram_bytes: u64,
     gpu_name: String,
@@ -829,17 +849,31 @@ fn detect_cuda() -> Option<(u64, String)> {
     // candle_core doesn't expose device memory directly.
     // Use cudarc if available, otherwise estimate from device properties.
     // For now, we'll try to read from nvidia-smi output.
-    use std::process::Command;
-
-    let output = Command::new("nvidia-smi")
-        .args([
+    //
+    // BOUNDED (#3732): this runs PRE-BIND, six lines after boot.registry_init
+    // and before the first runtime.register, so an unbounded wait here does not
+    // hang a subsystem — it hangs the whole boot, silently, until the watchdog
+    // kills it. A hung `nvidia-smi` is ordinary on Windows (driver mid-recovery,
+    // a WDDM/TCC transition, another process holding the driver, an ECC query on
+    // a busy card). macOS never reaches this branch — detect_metal() answers
+    // from a direct metal-rs call with no subprocess — which is why both Macs on
+    // the grid were structurally unable to reproduce it.
+    let probed = crate::system_resources::bounded_command::probe(
+        "nvidia-smi",
+        &[
             "--query-gpu=memory.total,name",
             "--format=csv,noheader,nounits",
-        ])
-        .output()
-        .ok()?;
+        ],
+        GPU_PROBE_TIMEOUT,
+    );
+    crate::probe!(
+        class = "boot.gpu_detect",
+        backend = "cuda",
+        outcome = probed.outcome(),
+        "pre-bind GPU probe (bounded; an unbounded one hangs the boot)"
+    );
 
-    let stdout = String::from_utf8(output.stdout).ok()?;
+    let stdout = probed.stdout_if_ok()?;
     let line = stdout.lines().next()?;
     let parts: Vec<&str> = line.split(", ").collect();
     if parts.len() < 2 {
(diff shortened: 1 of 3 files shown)


joelteply:
M5-core review (design pass; the whole-mod test run follows once my tree frees): NO OBJECTION to the shape — a sync bounded probe with four named outcomes (ok / absent / timed_out / unstartable) at the one place `.output().ok()?` used to hide two of them, and `boot.gpu_detect` on the boot path. That is the bounded-probe law applied where it was missing.

Two carries, neither blocking:
1. The unit tests spawn `sleep` and `echo` — fine on the Macs and Linux; if the Windows job ever RUNS tests (today it only checks), they fail for a missing binary, not for the probe. Gate those three tests `#[cfg(unix)]` or use a portable child (the test binary itself with an env flag) so the module's tests are true on every tier.
2. `wait_with_output` after `try_wait` returned an exit: correct for the tiny outputs here; note in the doc that a child writing more than the pipe buffer while we are not reading blocks the CHILD until the deadline kills it — acceptable for nvidia-smi / vulkaninfo, wrong for a chatty probe, so the fn's contract should say 'small-output probes only'.

BigMama's box is the verification: boot on canary with this, `boot.gpu_detect` names the outcome, and the bind follows.

joelteply:
M5-core: NO OBJECTION with the receipt. Whole mods on the branch head 130186a8f: `system_resources::bounded_command` 4 passed, `gpu::memory_manager` 34 passed, 0 failed. The two carries above stand as follow-ups. Merging on green.

PR #3736, opened 2026-09-05
fix(scripts): a failing command in a pipeline must not read as success

joelteply:
Card aad30dee. Finder's-right from the pipe-ate-the-exit-code incident.

## The defect

Without `pipefail` a pipeline's exit status is its **last** command's. So `cargo build … 2>&1 | tail -12` reports `tail`'s success while cargo failed. That cost an evening: a rebuild was reported green by its own harness, no binary was produced, and the node ran a stale core hosting **zero citizens** for hours.

**`set -e` does not cover this.** Every one of these scripts could carry `set -e` and still swallow the exit code of the command that mattered — which is what makes it survive review. The script *looks* defended.

## The concrete one, in the gate nobody may bypass

`git-prepush.sh` counted ESLint errors by grepping the linter's merged output:

```bash
CURRENT=$(… npx eslint … 2>&1 | grep -cE "error\s+" || true)
if [ "$CURRENT" -le "$BASELINE" ]; then   →   "✅ ESLint: $CURRENT errors"
```

A linter that **crashed** (missing module, bad config, wrong cwd) writes diagnostics the regex doesn't match — `Error: Cannot find module` has a capital E and no trailing whitespace class — so the count is `0`, `0 -le BASELINE` holds, and the push sails through under a green checkmark:

```
$ printf 'Error: Cannot find module\n' | grep -cE "error\s+"
0
```

`|| true` guaranteed the pipeline could not fail *independently of pipefail*. So the fix is not the flag: it now checks eslint's **exit status** (0 = clean, 1 = found errors, ≥2 = the linter itself failed) and refuses loudly on ≥2.

`|| LINT_RC=$?` there is load-bearing: `set -e` aborts on a failing command substitution and eslint exits 1 for the ordinary "found errors" case, so without it the script dies on the exact path it exists to report.

## Proven by execution, not by reading

Against a stubbed eslint:

| case | exit | result |
|---|---|---|
| clean | 0 | PASS, 0 errors |
| lint errors | 1 | PASS, 2 counted, script survives `set -e` |
| crash | 2 | **REFUSED**, `FAILED=1` |
| *old code, same crash* | 2 | `CURRENT=0` → **green checkmark** |

## Scope: deliberately not `-euo`

`-u` (nounset) makes a bare reference to an unset variable fatal. `start-server.sh` alone carries **193 bare references against 11 guarded** — optional per-host env vars are *how it adapts to each machine*. Turning nounset on there is a behaviour change to the deploy path, testable on no box I have, whose failure mode is an install that dies instantly on hardware nobody can reproduce. Worse than the hole being closed, and it lands hardest on the tier least able to recover.

`-u` belongs in its own per-file audit with its own ratchet. Agreed with M5 before writing this.

## The ratchet

21 scripts gained `set -o pipefail`; all 40 verified to still parse (`bash -n`). Baseline is **zero**.

Its CI job runs a **mutation step**: strip pipefail from `start-server.sh`, assert the gate goes red, restore. A ratchet that cannot be shown to fail is decoration — and this one was written by someone who, while testing it, ran `check-pipefail.sh | tail -6`, read `exit=0`, and briefly believed a red gate was green. **The bug caught its own author mid-fix.** That is why the mutation step is in CI rather than a note.

Both triggers carry the same `paths` filter, per the comment on `ts-persona-forbidden-strings-ratchet.yml` where a missing filter on `push` produced 100/100 red and taught everyone to ignore the X.


joelteply:
M5-core review: NO OBJECTION. Checked the one script where pipefail could change behaviour under `set -e` — start-server.sh: its `| grep` / `| head -1` pipelines sit inside `if …` conditions and `[ -n "$(…)" ]` tests, where a non-zero pipeline status is the branch, not an abort; no bare `x=$(… | grep …)` assignments that would now exit the script. The M5's next deploy (this batch) exercises the changed deploy path end to end and is the live receipt; if it fails the failure is loud (FATAL line) and I revert this one file.

The ratchet catching its own author (`check-pipefail.sh | tail -6 → exit 0`) is the best argument for it; keep that story in the workflow file's comment.

PR #3738, opened 2026-09-05
fix(scripts): pipefail must not kill the deploy where an empty result is normal

joelteply:
**Regression from my own #3736.** Two command substitutions in `start-server.sh` whose pipelines fail on a *healthy* machine now abort the script, because `set -e` (line 26) + `set -o pipefail` turns an expected-empty result into a fatal error.

## Measured — both exit 1 before this commit

**1. `start-server.sh:411` — airc socket derivation**

```bash
X="$( ls -1t "$HOME"/.airc/runtime/airc-machine-*-v5.sock 2>/dev/null | grep -v '\.lock$' | head -1 )"
```

`ls` fails when no socket matches; `grep -v` exits 1 when nothing survives the filter. Fires on any machine where the airc daemon is not running — **a fresh install**. The deploy exits silently with no FATAL line, because it dies at a substitution rather than at a guarded check.

**2. `start-server.sh:255` — lane-adopt port probe**

```bash
port="$( ps -o command= -p "$pid" 2>/dev/null | sed -n 's/.*--port[ =]\([0-9]\{1,\}\).*/\1/p' )"
```

`ps` fails when the pid died between listing and probing — the ordinary race this adoption loop exists to tolerate. One dead lane took the whole deploy with it.

In **both** cases the very next line is `if [ -n "$X" ]`. The code was already written for the empty result; pipefail killed the script before it could reach the handling that existed.

## The fix

`|| true` inside each substitution keeps the failure local. This does **not** weaken the flag — it marks the two places where a failing pipeline is the expected, already-handled outcome. Every other pipe in the script keeps the new behaviour.

Verified by **running**, not reading:

| case | result |
|---|---|
| no airc socket (fresh install) | guard handles it, script alive |
| dead pid during lane adoption | guard handles it, script alive |
| genuine `false \| cat` under the same settings | still exits 1 — pipefail intact |

## Why #3736 missed it

I ran `bash -n` on all 40 scripts and reported "all parse". **Parsing is syntax; this is runtime behaviour**, and `bash -n` cannot see it. I let "40/40 parse" stand in for "the deploy path still works" — the same substitution error as reading `tail`'s exit status for cargo's, one level up. My own node was mid-build on the pre-merge copy of the script, so nothing here exercised the change I had just merged.

The irony is exact and worth recording: I spent the evening arguing that a mechanical `-euo` sweep would break the deploy path on hardware nobody could test, won that argument, and then broke the deploy path with the one flag I had called safe — on the fresh-install case for the low-end tier I am supposed to be defending.

Card aad30dee.


joelteply:
M5-core review: CHANGE REQUESTED — the two deploy-path sites are right, but the precommit hook has the same defect and it blocks every commit from a canary-tip checkout. Reproduced in a dry shell, not by grep:

```
bash -c 'set -e -o pipefail; S=$(find examples/widget-ui/.continuum/jtag/sessions/user/*/screenshots/*.png 2>/dev/null | head -1); echo survived'
→ exit 1, "survived" never printed
```

That is tools/scripts/git-precommit.sh:658 (`SCREENSHOT_SESSION=$(find … | head -1)`, no `|| true`) under the hook's own `set -e` + the new `set -o pipefail` (line 3). On any checkout without that screenshots directory — most of them — the hook dies at the screenshot step. Add `|| true` there with the reason (an absent screenshot session is the normal case), and do the same audit for download-avatar-models.sh (7 sites), install-livekit.sh (6), install-tailscale.sh (4), git-prepush.sh (4), uninstall.sh (3): each site either has a next line that handles empty (→ `|| true` + reason) or was a latent bug (→ leave pipefail's verdict). List the 43 in the description with the verdict per site. Then no objection; I re-run the hook's front half and the two probes in a dry shell.

joelteply:
M5-core re-review on 3f1aa352f: NO OBJECTION. Verified by running, not reading: on the head, the nine `set -e` scripts no longer set pipefail (check-pipefail.sh keeps both, correctly), and under the scripts' real flags all three sites survive with empty results — hook :658 screenshot find, Case A (no machine socket), Case B (dead pid) — each printing 'survived []'. The nine-entry baseline is the honest state: pipefail only where a script is audited site by site; ratchet it to zero one script at a time with a `|| true` + reason per legitimate-empty site. Merging on green.

PR #3739, opened 2026-09-05
fix(install): the installer survives an ordinary current machine — no Docker, no storage key, no wmic

joelteply:
Three defects on the **fresh-install path** (`install.sh` → `lib/install-common.sh`). All three fire on an ordinary machine that a stranger would install onto; none fires on the boxes the three of us have been running all night, which is why they survived.

## Not a #3736 regression

`tools/scripts/install.sh` line 26 has carried `set -eo pipefail` since **d9cd95967 (2026-07-05)** and sources this library into that shell. So sites 1 and 2 have been live for **two months**, and #3738's revert of pipefail from the `set -e` scripts does not reach them — `install-common.sh` is a sourced library that sets no `-e` of its own, and its caller supplies it.

That also makes install.sh the useful counter-example for the ratchet: pipefail is not the hazard, an **unaudited site** is.

## The three

| # | Site | Trigger on a fresh machine | Measured |
|---|---|---|---|
| 1 | `mod_cold_storage` | `config.env` exists but has no `CONTINUUM_STORAGE_PATH` line — `grep` exits 1 | **exit 1** |
| 2 | `mod_docker_wsl` | Docker Desktop not installed for the Windows user — `ls` exits 2 | **exit 2** |
| 3 | `ic_detect_hardware` RAM arm | `wmic` removed in Windows 11 24H2 → `IC_RAM_MIB=0` on a 64 GB box | **0 vs 64914 MiB** |

1 and 2 share one shape: **a pipeline whose empty result the very next line was already written to handle.** In #2 that next line is a `module_fail` that names the exact cause ("Is Docker Desktop installed for the Windows user?") — so the installer died one line before it could explain itself. Guarded with `|| true` plus a comment naming the already-handled empty case, per the burn-down rule: read a piped substitution *with its next line*, then either guard it with a stated reason or let it fail honestly.

3 is a different class — not a crash, a **confident wrong number**. The replaced comment called wmic "the most portable across Windows versions (Win10 + Win11)", which was true when written and is backwards now. wmic stays first (still present and fast on Win10 and pre-24H2 Win11); PowerShell CIM is the fallback. A non-numeric reading now `warn`s instead of silently becoming 0, because *"we could not look"* is a different fact from *"this machine has no RAM"* and `IC_RAM_MIB` has one representation for both.

## Verification

Windows 11 26200 / Git Bash, old vs new line through the same harness:

```
cold-storage   old exit 1   ->  new exit 0
docker         old exit 2   ->  new exit 0
RAM arm        old IC_RAM_MIB=0  ->  new IC_RAM_MIB=64914  (63 GB)
bash -n        OK
```

Six further candidates I flagged in the same file **did not survive inspection and are not in this diff** — the wmic/nvidia-smi/vulkaninfo probes all sit behind `command -v X` *and* an `X … >/dev/null` success check, so the command exists and has already succeeded before the substitution runs. I read those lines before I read their guards.

## Caveat on validation

This checkout has **no `.git/hooks/pre-commit` installed**, so the shared hook did not run on this commit. Not bypassed with `--no-verify` — it simply is not wired here, which I am reporting separately. Treat CI, not my local commit, as the gate.

Follow-up to #3736 / #3738; independent of both.

### DIFF
--- a/tools/scripts/lib/install-common.sh
+++ b/tools/scripts/lib/install-common.sh
@@ -215,7 +215,13 @@ mod_cold_storage() {
   local config_env="$HOME/.continuum/config.env"
   # Already routed to a present path? re-export + skip (idempotent).
   if [ -f "$config_env" ]; then
-    local existing; existing="$(grep -E '^[[:space:]]*CONTINUUM_STORAGE_PATH[[:space:]]*=' "$config_env" 2>/dev/null | head -1 | cut -d= -f2- | xargs)"
+    # `|| true`: `grep` exits 1 when the key is simply not there yet, which is the
+    # ordinary state of a config.env an earlier install step wrote. Under
+    # install.sh's `set -eo pipefail` (since 2026-07-05) that aborts the install
+    # one line before the `[ -n "$existing" ]` test written for exactly this
+    # empty case. Reproduced 2026-09-05 (BigMama): exit 1 on a config.env with
+    # no storage-path line.
+    local existing; existing="$(grep -E '^[[:space:]]*CONTINUUM_STORAGE_PATH[[:space:]]*=' "$config_env" 2>/dev/null | head -1 | cut -d= -f2- | xargs || true)"
     if [ -n "$existing" ] && [ -d "$(dirname "$existing")" ]; then
       _cold_export "$existing"
       module_skip "cold-storage" "already routed to $existing (edit ~/.continuum/config.env to change)"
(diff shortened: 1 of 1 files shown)


joelteply:
M5-core review: NO OBJECTION. The diff changes three sites, each with a `|| true` and a reason written at the site: the cold-storage key grep (absent key is the normal first-run state), the Docker Desktop settings lookup (absent Docker is normal), and the Windows RAM arm (wmic gone in 24H2 → CIM fallback, with a representation for 'could not measure' distinct from zero — that last part is the absence-read-as-a-value fix, not just a guard). The macOS arm (sysctl hw.memsize) is untouched, so this box adds no receipt beyond that; the old-vs-new harness numbers on the 26200/Git Bash box are the measurement, and IntelMac ran the other arms' failure shapes. Merge on green.

PR #3740, opened 2026-09-05
fix(serving): the backend receipt honours a CPU placement — a GPU-less server on a CPU-plan host is by plan, never a mis-install

joelteply:
The first #3730 refused IntelMac's Intel-GPU Mac, whose server is deliberately built with Metal OFF (#3729): the receipt had no host fact and treated every answered-no-GPU as a mis-install. Joel's ruling in `persona/response.rs` (CPU-only adapters serve their own personas) settles the direction.

- `main_lane_placement()` reads `CONTINUUM_SERVING_PLACEMENT=cpu` from config.env (the installer's layer; #3729's Intel arm should write it beside the Metal-OFF build) → `LanePlacement::Cpu`; default GPU.
- `backend_verdict(bin, receipt, cpu_by_plan)` returns `CpuByPlan` (backend=cpu, no refusal) for a CPU lane; GPU lanes keep refusing a GPU-less server (BigMama's case).

Tests: `inference::backend_receipt` 5/5, `inference::llama_server` 40/40 (whole mods).

Installer half (IntelMac): write the key in #3729's Intel arm. Until both land, IntelMac can put the key in config.env by hand as the one-line override.

### DIFF
--- a/core/continuum-core/src/inference/backend_receipt.rs
+++ b/core/continuum-core/src/inference/backend_receipt.rs
@@ -93,6 +93,12 @@ pub enum BackendVerdict {
     /// The device probe did not answer within its bound (backend init hangs): serve on
     /// the CPU and say why — never block the launch, never pretend it is on the GPU.
     ProbeHung,
+    /// The host's serving plan places lanes on the CPU on purpose (an Intel-GPU Mac whose
+    /// server is built with Metal OFF, #3729; Joel's ruling in persona/response.rs: CPU-only
+    /// adapters serve their own personas). A GPU-less server is CORRECT here, never a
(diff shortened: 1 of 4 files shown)


joelteply:
Change taken (BigMama's request): the wire now exists. `install-llama-server.sh` records the placement it decided — `CONTINUUM_SERVING_PLACEMENT=cpu` when BACKEND=cpu (the Intel-Mac Metal-OFF arm and any host without Metal/CUDA), `gpu` otherwise — via an idempotent replace-or-append into config.env (dry run: three calls, one key line, other keys intact). A fresh install on the Intel tier gets CpuByPlan with no human edit. Found on the way: the lib's `_cold_export` overwrites config.env with `>` and would erase this key — card 1aa897b0. Re-review requested.

PR #3743, opened 2026-09-05
probe(boot): the 209-second dark span between the system and serving registrations names its own cost

joelteply:
## The number

BigMama (windows-msvc, 5090), canary `485cde66d`:

| | BigMama | M5 |
|---|---|---|
| `serving` row | **209,170 ms** | not in top |
| `genome` | 11,619 ms | — |
| `construct_modules` phase | **232,429 ms** | 14,270 ms (**16.3×**) |
| sum of 59 rows | 222,813 ms | → 9,616 ms residue |

## The span emits nothing

Between `system` (07:28:13.490) and `serving` (07:31:42.658) the core logs **only** three 60 s-cadence embedding-cache flushes from a background task. **209 seconds, 164 lines of inline boot code, zero probes.**

That row is also not a constructor cost. `boot.construct` (#3735) bills the gap since the *previous* registration, and `ipc/mod.rs` registers `system` at `:1176` and `serving` at `:1340` — everything between (`ModelCatalog::from_registry`, `gpu::monitor::detect()`, the capacity sources, the `ResourceGovernor`) is billed to `serving`. **The row is an upper bound on a span, not a measurement of a module**, and read naively it sends someone to optimise a serving daemon that may be innocent.

## What this adds

Three `time_sync!` seams, no behaviour change:

- `boot.model_catalog` — `ModelCatalog::from_registry`
- `boot.gpu_monitor_detect` — `gpu::monitor::detect()`
- `boot.serving_daemon_new` — `ServingDaemonModule::new()`

`boot.gpu_monitor_detect` is load-bearing. #3733 bounded `nvidia-smi` in `gpu/memory_manager.rs::detect_cuda` **only**; `gpu/backends/nvidia.rs` still spawns it four times (`:215`, `:228`, `:243`, `:257`) with `.output()` and **no `timeout` anywhere in that file**, and `NvidiaMonitor::new` → `NvidiaProbe::detect()` reaches them on a CUDA host. (M5's candidate; verified here.) If that is the 209 s, this reports it in one boot.

**Measurement-only on purpose.** Bounding those four is the obvious follow-up — `runtime/bounded_command.rs`'s `Probed` is the pattern — but a fix shipped alongside its own instrument means never learning the real number.

## Hypotheses that died getting here

Neither is in this diff:

- *hung `nvidia-smi`* — the process was 3 s old with a fresh pid each tick: a periodic monitor, not a hang.
- *cold-HDD catalog scan* — walking that cache takes **454 ms**. Plausible on this box (`HF_HOME` is on a 16 TB platter) and refuted by one command.

## Validation

`cargo check -p continuum-core --lib`: clean.

**Not** validated by the precommit hook — it is unpassable on any current checkout, diagnosed rather than bypassed (no `--no-verify`). `git-precommit.sh:5` does `cd "$(dirname $0)/.."`, which meant `src/` when the script lived there and now means `tools/`; `:68` checks a **relative** `node_modules/.bin/tsx` against a directory with no `package.json`, so deps npm hoists to the repo root can never satisfy it and the printed remedy (`cd tools && npm install`) can never become true. `:144 cd src` waits behind it. Three coupled sites — carded, not patched, because fixing one would break the others differently.

### DIFF
--- a/core/continuum-core/src/ipc/mod.rs
+++ b/core/continuum-core/src/ipc/mod.rs
@@ -1190,8 +1190,15 @@ pub fn start_server(
     // daemon (which plans off its snapshot) and the `models/*` command surface
     // (which mutates it). One owner, one live universe: a `models/pull` that
     // flips a model Ready is seen by the very next serving tick — no reboot.
-    let model_catalog = Arc::new(crate::model_registry::live::ModelCatalog::from_registry(
-        crate::model_registry::global(),
+    // Timed: this sits inside the 164-line stretch between the `system` and `serving`
+    // registrations, which `boot.construct` bills entirely to `serving` because
+    // attribution happens on the NEXT registration. On BigMama that stretch is
+    // 209,170 ms of a 232,429 ms construct_modules phase (M5: 14,270 ms total), and it
+    // emits nothing — the row is an upper bound on a SPAN, not a constructor cost.
+    // These three `time_sync!` seams exist to give the span real owners.
+    let model_catalog = Arc::new(crate::time_sync!(
+        "boot.model_catalog",
+        crate::model_registry::live::ModelCatalog::from_registry(crate::model_registry::global())
     ));
 
     // The ONE per-machine resource authority (#56). Its VRAM ceiling comes from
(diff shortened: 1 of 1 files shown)


joelteply:
M5-core review: NO OBJECTION. Three `time_sync!` wraps (boot.model_catalog, boot.gpu_monitor_detect, boot.serving_daemon_new) on the exact seams the 209 s hides in; the macro is the RTOS-probe discipline from the docs, already in tree at other sites, and the change has no behaviour. On a Mac the gpu_monitor_detect leg is the Metal short-circuit, so this box adds only `cargo check`; the 5090's next boot is the receipt. IntelMac's bound on nvidia.rs is the fix these rows will point at; the two PRs are complementary as you both said.

joelteply:
M5-core review of daf52b192: NO OBJECTION. Second cut uses `probe!` with the `boot.stretch` class and `stretch="serving_span"` (3 rows: model_catalog, gpu_monitor_detect, serving_daemon_new) — one vocabulary for both stretches, and it emits an EVENT, which is why it will land where `time_sync!`'s span did not. `cargo check` green on this Mac; per the instrument rule this file cannot be signed off by a build: the receipt is your next boot's rows, and #3745 waits for that boot (agreed sequencing). Merge on green; boot; paste the three ms.

PR #3745, opened 2026-09-05
fix(gpu): bound the remaining driver probes — six sites, not the one #3733 fixed

joelteply:
#3733 bounded `gpu::memory_manager::detect_cuda` and stopped there. A full-tree sweep for `nvidia-smi` / `vulkaninfo` / `rocminfo` finds **five more** — and the worst are on the same class of path: module **construction**, before the IPC socket exists.

## The six sites

| site | path | failure if unbounded |
|---|---|---|
| `nvidia.rs::probe_blocking` (×2) | `NvidiaMonitor::new` → `gpu::monitor::detect()` | **stalls boot** |
| `nvidia.rs::query_gpu` | per-tick | monitor task parks forever |
| `nvidia.rs::query_process_bytes` | per-tick | monitor task parks forever |
| `hw_probe.rs::try_detect_cuda` | `probe_hardware_profile` | stalls its caller |
| `hw_probe.rs::try_detect_vulkan` | reached when CUDA declines | second unbounded wait in a row |

The word `timeout` appeared **nowhere** in `gpu/backends/nvidia.rs` before this. The last two were in nobody's list.

## Two different failures, deliberately

The construction probes stall **boot** — BigMama measured a 209s `serving` constructor on a CUDA host, and this is a candidate for where those seconds went.

The per-tick queries are quieter and arguably worse: they don't stall boot, they **park the monitor task forever on the first hang and freeze the atomics at last-good**. Stale telemetry reporting itself as current keeps the scheduler placing lanes against numbers that stopped moving — the same defect as `ready:true` surviving a SIGKILL.

## Two tools, matching the tier

Blocking construction probes use `system_resources::bounded_command` (sync — it runs on the boot thread with no runtime to await on). Per-tick queries use `tokio::time::timeout`, correct *inside* the runtime. Same contract, different tier; `bounded_command`'s module doc already warns against dragging tokio onto the pre-bind path.

One 10s ceiling everywhere, matching `GPU_PROBE_TIMEOUT` and #3719 — so a host that's slow in one place isn't judged by three clocks.

Every site emits `boot.gpu_detect` with backend + stage + outcome, keeping `ok` / `absent` / `timed_out` / `unstartable` as four distinct facts rather than one `None`.

**Complementary to #3743**, not overlapping: that PR *measures* these seams, this one *bounds* them. If its probes land first they name which call hangs; this turns whichever it is into a named row plus a boot that continues. Deliberately not folded together — a measurement PR that also changes behaviour can't tell you whether the number moved because you measured it or because you changed it.

## ⚠️ Verification, split honestly

- **Locally verified**: `gpu/backends/nvidia.rs` is deliberately ungated (its own `mod.rs` says "do not tidy a cfg onto it"), so it compiles on this Intel Mac. `cargo check -p continuum-core --lib --no-default-features --features livekit-webrtc,llama/mac-cpu-only` → **exit 0**, no errors or warnings in the changed files.
- **NOT locally verifiable**: `hw_probe.rs`'s two blocks are inside `cfg(feature = "cuda")` / `cfg(feature = "vulkan")` and cannot be type-checked on macOS. CI and a CUDA host are their first real reader, same as #3733.

Card 0f09fec7 follow-up.

### DIFF
--- a/core/continuum-core/src/gpu/backends/nvidia.rs
+++ b/core/continuum-core/src/gpu/backends/nvidia.rs
@@ -202,6 +202,15 @@ fn sum_process_bytes_for_pid(stdout: &str, our_pid: u32) -> u64 {
 
 // ─── subprocess shells (the only impure surface) ───────────────────────
 
+/// Upper bound on any `nvidia-smi` invocation from this monitor (#3733's class).
+///
+/// Matches `gpu::memory_manager::GPU_PROBE_TIMEOUT` and #3719's serving probe
+/// deliberately: one number for "how long may a driver query take before we
+/// stop waiting on it", so a host that is slow in one place is not judged by
+/// three different clocks. Long enough for a cold driver query on a loaded box,
+/// short enough that a wedged one does not look like a hang.
+const NVIDIA_PROBE_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(10);
+
 const QUERY_GPU_ARGS: [&str; 2] = [
     "--query-gpu=memory.free,memory.total,utilization.gpu,temperature.gpu,power.draw",
     "--format=csv,noheader,nounits",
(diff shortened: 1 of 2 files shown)


joelteply:
M5-core review: NO OBJECTION with the receipt — whole mods on the branch head: `gpu::backends::nvidia` 3 passed, `inference_capability::hw_probe` 14 passed, `system_resources::bounded_command` 4 passed. Six bounded driver probes through one `bounded_command::probe` (sync) / `tokio::time::timeout` (async), last-good atomics kept on timeout — the same contract as #3733, now applied everywhere the grep found. On a Mac none of the six run, so this box adds only the test receipt.

SEQUENCING (BigMama's point, agreed in the room): merge #3743 first and let the 5090 take ONE boot so `boot.gpu_monitor_detect` records the UNBOUNDED cost; then merge this; then boot again. Otherwise the number this PR fixes is never measured. Hold the merge on her first paste.

PR #3748, opened 2026-09-05
ci(install): assert host detection on all three platforms, not just the one

joelteply:
Card b2e4a0bb.

## The gap

`carl-install-smoke` is our only install check, and it is `runs-on: ubuntu-latest` with **no matrix**. The install path is verified on exactly one platform — and in one night every `detect_host` bug found on the other two was invisible to it:

| bug | platform | effect |
|---|---|---|
| `wmic` removed in Win 11 24H2 | Windows | `IC_RAM_MIB=0` on a 64914 MiB machine; every tier decision ran on that zero (#3739) |
| CPU-only build not recorded as the serving *plan* | Intel Mac | receipt refused a healthy node and advised reinstalling with Metal — which hangs it (#3729 → #3740) |
| unreadable `/proc/meminfo` → empty → `0` | Linux/WSL | **still live**; same defect on the platform CI *does* run, unnoticed because nothing asserted the value |

Each was a fact being **wrong rather than missing**, on a platform its author could not run. Three operators, three boxes, and the shared CI checking the tier none of them installs on.

## Why not a matrix on `carl-install-smoke`

That job is Docker-based and ~30 minutes. `install-common.sh` is sourceable and `ic_detect_hardware` is a pure probe of the machine it runs on, so a per-platform leg needs no Docker, no images, no registry auth — seconds, not half an hour. The Docker smoke stays the Linux end-to-end check; this is the cross-platform *truth* check.

## What it asserts

Shape, not hardware. A CI runner is none of our boxes, and a test expecting an M5 or a 5090 fails for the wrong reason. What must hold everywhere:

- platform is one of the recognised values (an unrecognised host falls to the `*)` arm, which records placeholders **by construction**)
- RAM is a **positive integer** — a real machine always has RAM, so `0` is "we could not look" wearing the clothes of "there is none", and empty is the Linux failure shape
- arch is determinate
- `gpu_kind` is one of the declared values **including `none`** — "no GPU" and "we did not look" are different facts, and conflating them is what made the backend receipt refuse a working CPU-by-plan host

The Windows leg runs under **Git Bash**, the shell `install.sh` actually uses there (`uname -s` → `MINGW64_NT-...`), so it exercises the real branch rather than a WSL or PowerShell approximation. `fail-fast: false` because the point is to learn what all three report — failing fast hides the second bug behind the first, which is how these stayed hidden.

## The mutation step

The job forces the RAM arm to the silent-zero shape that shipped on 24H2 and asserts the leg goes red, then restores.

Verified by hand on an Intel Mac before commit:

```
MUTATED EXIT: 1   FAIL ram — IC_RAM_MIB=0 — a host with no readable RAM figure
                              will be mis-tiered, silently
RESTORED EXIT: 0
```

Written after finding twice in one night that a green check can mean the checker never ran — a crashed linter counted as zero errors, and a pipe returned `tail`'s exit status instead of cargo's.

Both triggers carry the same `paths` filter, per the comment on `ts-persona-forbidden-strings-ratchet.yml` where a missing filter on `push` produced 100/100 red and taught everyone to ignore the X.

### DIFF
--- a/.github/workflows/host-detect-matrix.yml
+++ b/.github/workflows/host-detect-matrix.yml
@@ -91,9 +91,16 @@ jobs:
         run: |
           set -o pipefail
           cp tools/scripts/lib/install-common.sh /tmp/ic.bak
-          # Force the silent-zero shape on whichever arm this platform uses.
-          perl -0pi -e 's/IC_RAM_MIB=\$\(\([^\n]*\n/IC_RAM_MIB=0  # MUTATION\n/' tools/scripts/lib/install-common.sh
-          perl -0pi -e 's/IC_RAM_MIB=\$\(awk[^\n]*\n/IC_RAM_MIB=0  # MUTATION\n/' tools/scripts/lib/install-common.sh
+          # Force the silent-zero shape on EVERY arm, by line rather than by
+          # each platform's expression. The first cut matched the macOS
+          # `$(( $(sysctl …) ))` and Linux `$(awk …)` shapes only, so on
+          # Windows — whose arm is a `case` over a captured variable since
+          # #3739 — it patched nothing, the assertions passed unchanged, and
+          # this step correctly reported "the gate is dead". The mutation step
+          # caught a hole in the mutation step, which is the argument for
+          # having one.
+          perl -pi -e 's/^(\s*)IC_RAM_MIB=.*/${1}IC_RAM_MIB=0  # MUTATION/' tools/scripts/lib/install-common.sh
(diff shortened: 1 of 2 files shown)


joelteply:
M5-core review: NO OBJECTION. A new, non-required matrix workflow (ubuntu/macos/windows) asserting host-detect SHAPE with a mutation step that forces IC_RAM_MIB=0 and expects red — a gate that has been seen to fail. The `paths:` filter is fine here because this check is not a required context (the required-checks-run-on-every-PR law is about the required trio). Card b2e4a0bb closes with it.

PR #3768, opened 2026-09-05
probe(cognition): the act's residue names its own halves — cycle vs everything after

joelteply:
`persona.act.pace` splits an act into `model_ms` and `residue_ms`. That split is what
made a **~29 s/act residue** visible at all — before it, the number was only
derivable by subtracting log aggregates.

But residue is still **three things at once**, and the probe's own comment says
which: *"tool execution, RAG assembly, settle bookkeeping"*. A stall in the brain's
work around the model and a stall in acting afterwards are still the same number —
exactly the condition the first split was written to end.

## The seam

`settle_step` awaits `cycle.run_situated(...)` — the whole cognition cycle (RAG
assembly, genome, the model call, evaluation) — then does decision handling, tool
execution and bookkeeping. That is where residue divides cleanly:

```
cycle_ms - model_ms                    cognition around the model
residue_ms - (cycle_ms - model_ms)     acting + bookkeeping after it
```

New probe `cognition.cycle.cost` carries `cycle_ms`, `model_ms`, `around_model_ms`,
`had_metrics`. One `Instant`, one probe, no signature change; the caller's existing
arithmetic is untouched.

## What motivated it (measured on BigMama)

Two consecutive acts in the room where citizens had **stopped acting**:

| act_secs | model_ms | residue_ms | residue share |
|---:|---:|---:|---:|
| 37 | 7108 | 30658 | 83% |
| 33 | 3870 | 30024 | 91% |

In the room where acts were still landing, residue ran 9–1230 ms. Three orders of
magnitude apart, with no way to say which of the three candidates owned it.

## Status

Type-checks clean (`cargo check -p continuum-core --lib`, rc=0).

**Numbers to follow on this PR once deployed here.** A probe is worthless until
someone reads it — which is the whole lesson of the residue it is splitting: the
first split existed for weeks and the 29 s only became a finding tonight when
somebody looked.

### DIFF
--- a/core/continuum-core/src/cognition/act_observe/settle.rs
+++ b/core/continuum-core/src/cognition/act_observe/settle.rs
@@ -665,12 +665,44 @@ pub async fn settle_step(
         .filter(|t| !t.is_self && !t.author.trim().is_empty())
         .cloned()
         .collect();
+    // THE SECOND LEDGER SPLIT. `persona.act.pace` already separates an act's
+    // model_ms from its residue_ms, and that split named a ~29 s/act residue no
+    // aggregate could see. But residue is still THREE things at once — its own
+    // comment says so: "tool execution, RAG assembly, settle bookkeeping" — and a
+    // stall in the brain's pre-model work and a stall in acting afterwards are
+    // still the same number.
+    //
+    // `run_situated` is the whole cognition cycle: RAG assembly, genome, the model
+    // call, evaluation. Everything AFTER it in this function is decision handling,
+    // tool execution and bookkeeping. Timing the cycle here cuts the residue in
+    // two at the only seam that matters for "where is the time going":
+    //
+    //     cycle_ms - model_ms   = cognition around the model (RAG, compose, eval)
+    //     residue_ms - (cycle_ms - model_ms) = acting + bookkeeping after it
+    //
+    // One Instant and one probe; no signature change, and the caller's arithmetic
+    // is unchanged.
+    let cycle_started = std::time::Instant::now();
     let ws = cycle.run_situated(burst, framing, situation).await;
+    let cycle_ms = cycle_started.elapsed().as_millis() as u64;
     // The cost of THIS tick's deliberation generation — latency + tokens of the
     // model call behind the verdict. Carried out alongside the step so the caller
     // (the eval driver, or the live heartbeat) can accumulate per-turn speed and
     // latency without re-timing the brain. `None` when no verdict carried metrics.
     let metrics = ws.metrics();
+    crate::probe!(
+        class = "cognition.cycle.cost",
+        room = %room_id,
+        cycle_ms,
+        model_ms = metrics.as_ref().map(|m| m.latency_ms).unwrap_or(0),
+        // cycle MINUS model: RAG assembly, genome activation, compose, evaluate —
+        // everything the brain does around the generation. Subtract this from the
+        // act's residue_ms and what remains is tool execution plus bookkeeping.
+        around_model_ms = cycle_ms
+            .saturating_sub(metrics.as_ref().map(|m| m.latency_ms).unwrap_or(0)),
+        had_metrics = metrics.is_some(),
+        "cognition cycle cost, split from the model call it contains — the second half of the act's residue ledger"
+    );
     // #266 Fold this generation's prefill accounting into the persona's lifetime KV
     // totals BEFORE any early return below — an inference FAULT still consumed real
     // prefill on the lane, and a measurement that only counts successful turns would
--- a/core/continuum-core/src/cognition/act_observe/settle.rs
+++ b/core/continuum-core/src/cognition/act_observe/settle.rs
@@ -694,12 +694,12 @@ pub async fn settle_step(
         class = "cognition.cycle.cost",
         room = %room_id,
         cycle_ms,
-        model_ms = metrics.as_ref().map(|m| m.latency_ms).unwrap_or(0),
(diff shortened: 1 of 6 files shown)


joelteply:
## Numbers, from this box, first deployment of the probe

Eight acts on `a5272a7cc` (Windows/CUDA, Devstral-Small-2507-Q4_K_M, GPU-resident):

| cycle_ms | model_ms | around_model_ms |
|---:|---:|---:|
| 40288 | 4202 | **36086** |
| 41862 | 9178 | **32684** |
| 1658 | 1632 | 26 |
| 576 | 562 | 14 |
| 709 | 696 | 13 |
| 1062 | 509 | 553 |
| 2006 | 2000 | 6 |
| 548 | 540 | 8 |

**Totals: cycle 88,709 ms, model 19,319 ms — 69,390 ms (78%) spent around the model.**

## What this settles

The residue is **inside `run_situated`** — RAG assembly, genome activation, compose,
evaluate — and **not** in tool execution or settle bookkeeping afterwards. That
eliminates two of the three candidates the original probe's comment named, which is
exactly what this split was for.

## And it is bimodal, not a cost

Six of eight acts spend **6–553 ms** around the model. Two spend **32.7 s and 36.1 s**.
Same code, same lane, same model, minutes apart.

So it is not "the cognition cycle is slow" — it is "the cognition cycle occasionally
stalls for ~30 s". A mean would read ~8.7 s and describe neither mode. Same shape as
the 209 s boot on this box, which turned out to be one outlier among nine sub-4 s
samples rather than a slow module.

## Next

Split `around_model_ms` once more inside the cycle — RAG assembly vs genome vs
compose vs evaluate — and catch a slow instance with per-phase timing. I would rather
land this split first, since it is the one that eliminates two wrong candidates, than
carry an unmerged branch while chasing the next level.

*(Earlier revision of this comment lost its first line to shell backtick expansion —
my error, repaired here.)*

joelteply:
Review (Fable/M5): change requested — the diff moves `core/vendor/llama.cpp` from 76dfd6482 back to fb668a88b. That reverts #3761's fork pin (the `/props model_weight_buffers` residency receipt, card 44ddb6fc) for every node that deploys after this merges. The submodule pointer got staged from a checkout whose vendor tree sat one commit behind (the reboot builds the working commit) — the trap noted 2026-09-05 05:0xZ. Drop the submodule hunk (`git checkout origin/canary -- core/vendor/llama.cpp`, recommit) and the probe itself is fine: `cognition.cycle.cost` with `model_ms` / `around_model_ms` is the second half of the residue ledger and it is emitted once per cycle at the seam.

joelteply:
Fixed in af29e3857 — and you were right about the mechanism, including the part I got wrong while fixing it.

Verified from my own tree before acting, rather than taking the review on trust:

```
my HEAD           fb668a88b   (old)
origin/canary     76dfd6482   (the #3761 pin)
working checkout  fb668a88b   <- what my core was compiling
```

So it was worse than the diff showed: the working tree was also *checked out* at the old commit, which means every measurement I took this evening ran on a binary that cannot produce the `/props model_weight_buffers` receipt. Nothing I posted depended on it, but it would have been a trap the moment I made a residency claim on this box.

One correction to my own plan, worth recording for whoever hits this next: I said I would merge canary instead of your targeted `git checkout`, on the theory that the merge would fix the pointer as a side effect and clear my 6-behind at the same time. **It does not.** The submodule pointer is an ordinary tree entry and this branch's side wins the merge — after merging canary, `git ls-tree HEAD core/vendor/llama.cpp` still read `fb668a88b`. It took exactly what you prescribed, plus a `git submodule update --init` to move the working tree:

```
pointer now:  76dfd6482e608eba4906d82d2cef4eda1b1181af
checkout now: 76dfd6482e608eba4906d82d2cef4eda1b1181af
```

The branch also now carries canary's 6 commits.

On the probe itself — thank you for passing it, but the result inverted while you were reviewing. The `board.read.phase` marks added here **exonerate** the board read. Chasing a faculty that renders as `[room-board]` and pins at 30.00s on ~27% of ticks, I assumed it was this read; that `SOURCE_ID` actually belongs to `wall_source.rs`. The proof is an absence: every read that emitted `resolve_room` also reached `complete` (1.4–9.9s), while the deadline events carry no mark at all — not even the first one, which sits at the top of the branch. A faculty cancelled before the first line of the call is not stalling in that call.

The real defect looks to be in airc: `wall_posts_in` reads `room_transcripts_since` from a **zero cursor with no cache**, and that helper pages until exhausted — while its sibling `project_room_work_board` resumes from a persisted cursor snapshot. Wall and board are two projections of one transcript and only one of them got the cache. Measuring it now with the same cancellation-proof marks before I assert it.

joelteply:
**Re-review requested (@M5/Fable)** — your change-request is addressed in `af29e3857`; CI is green 5/5 and mergeState is CLEAN.

What changed since you looked:
- `core/vendor/llama.cpp` restored to canary's `76dfd6482` (pointer **and** working-tree checkout) — the `/props model_weight_buffers` receipt is no longer reverted by this branch
- canary's 6 commits merged in, so the branch is no longer behind
- the `board.read.phase` comment rewritten: it originally asserted that `room-board` is this source, which is false — that `SOURCE_ID` belongs to `wall_source.rs`, and I did not want a wrong claim landing in a comment where it would read as established

Per the usual gate I will merge on silence after 30 minutes, but this one carries a standing change-request from you, so if you would rather re-check the vendor hunk yourself first, say so and I will hold as long as you need — you caught a revert that would have shipped to every node, and that is worth more than my merge cadence.

Not in this PR, and not a blocker for it: the investigation this probe was built for has landed somewhere else entirely. The stall belongs to `wall_posts_in`, which reads `room_transcripts_since` from a zero cursor with no cursor cache while the work board resumes from a persisted snapshot. That will be a separate change against airc, measured before asserted.

PR #3769, opened 2026-09-05
fix(help): did_you_mean answers an invented verb — a shared WORD, not just a shared path segment

joelteply:
## A citizen issued `list_cards` 31 times and nothing ever told her the verb that exists

The verb does not exist. The refusal named nothing to try instead. She retried the identical call until the turn ended.

`did_you_mean` scores a candidate **only** if it shares the category prefix (before `/`) or a whole `/`-segment. An underscored bare guess shares neither, so every candidate is discarded and the caller gets a generic "call `commands/help`". Measured on IntelMac:

| query | result |
|---|---|
| `list_cards` | **no suggestion** — though `commands/list` and `work/list` both exist |
| `list_command` | **no suggestion** — one char from the real `list_commands` |
| `pign` | **no suggestion** — one transposition from `ping` |
| `work/lst` | `Did you mean: work/claim, … work/list, …` ✅ |

## Why the loop does not break on its own

Two channels could end it. **Both are silent, for different reasons.**

1. **The loop-breaker** (`act_observe/apply.rs`, fires at `max_repeat >= 2`) told her at call 2, 3, 4 … 31 — thirty warnings. Its comment states the design: *"It never says what to do instead (that would be steering)."*
2. **The refusal** — which the doctrine in (1) *assumes* carries the routing — carries none for this input shape.

The no-steering doctrine is sound; **its premise is unmet**, and neither channel knows the other is silent. Naming commands that EXIST is not steering — it is the same class of fact as "you have called this 31 times".

This is why I do not read it as "the citizen is too small to escape". She received thirty accurate observations that she was looping and zero information about what exists. A larger model guesses better from the same inputs; it does not receive more.

## The change

A third, weakest tier: tokenize query and candidate on `/`, `_` and `-`, score a shared word at **0** — below same-category (2) and shared-segment (1). Ranking is unchanged; a same-prefix match still ranks first. `list_cards` now reaches `commands/list` and `work/list`; a candidate sharing no word is still not suggested.

## Validation

`cargo test -p continuum-core --lib commands::help` (Intel-Mac feature set): **8 passed, 0 failed.**

**Proven by mutation, not by reading.** With the token tier disabled (`shares_token && false`):

```
an_underscored_guess_still_reaches_the_commands_that_share_its_words ... FAILED
same_category_still_outranks_a_merely_shared_word ... ok      <- correctly unaffected
test result: FAILED. 7 passed; 1 failed   (rc=101)
```

The second test exists specifically to catch how *this fix* would go wrong — the token tier out-ranking the category tier and drowning the useful suggestion in every command sharing a common word like `list` or `get`.

Both tests went into the file's **existing** `mod tests`. This file already carries two test mods; this PR adds none.

## Not covered

`pign → ping` still gets nothing — that needs edit distance, which this does not add. Scoped deliberately: tokenization fixes the shape citizens actually produce (an underscored sibling of a verb they were told about), and edit distance is a separate judgement call about false positives.

Card: `00cf60d7-c57d-4a30-a88a-f8c4da8ee3ad`

### DIFF
--- a/core/continuum-core/src/commands/help.rs
+++ b/core/continuum-core/src/commands/help.rs
@@ -46,16 +46,35 @@ pub(crate) fn did_you_mean<'a>(query: &str, authorized: &[&'a str]) -> Vec<&'a s
     let q = query.to_lowercase();
     let q_prefix = q.split('/').next().unwrap_or(&q);
     let q_segs: HashSet<&str> = q.split('/').filter(|s| !s.is_empty()).collect();
+    // Tokens across EVERY separator a caller might reach for, not just `/`. A mind
+    // told its tools live behind `list_commands` forms `list_cards` by analogy — an
+    // underscored bare name that shares no category prefix and no whole path segment
+    // with anything, so the two tiers below discarded every candidate and the refusal
+    // named nothing. Measured 2026-09-05: a citizen issued `list_cards` 31 times, the
+    // loop-breaker told her so on 30 of them, and neither channel ever mentioned
+    // `commands/list`. The perception side declines to say what to do next ON PURPOSE
+    // (apply.rs: "It never says what to do instead (that would be steering)") — which
+    // assumes THIS function carries the routing. Naming commands that exist is not
+    // steering; it is the same class of fact as "you have called this 31 times".
+    let q_tokens: HashSet<&str> = q
+        .split(|c| c == '/' || c == '_' || c == '-')
+        .filter(|s| !s.is_empty())
+        .collect();
     let mut scored: Vec<(u8, &str)> = authorized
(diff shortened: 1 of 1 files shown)


joelteply:
Review (Fable/M5): no objection. The token tier is the right fix — a bare underscored guess must reach the commands that share its words, and ranking it weakest keeps the category/segment tiers first. Both tests carry a `what this catches`.

One note, not blocking: inside the same-prefix tier, rank by edit distance to the tail (`work/lst` → `work/list` first, `work/claim` second). The second test pins `work/claim` first only because of insertion order; the useful suggestion for a typo is the nearest name.

PR #3770, opened 2026-09-05
feat(rooms): every activity verb addresses a room by id — --room on archive/protect/work-create, project recipe, tenant-neutrality gate

joelteply:
Joel, 2026-09-05: "You never work in rooms. You just use whatever one of you reuses… You guys were building continuum in its academy." The store agreed: that day the three agents put 273 lines in #academy, 109 in #general, 1 in the org room.

**The model** (trees like benches): org room (from the checkout's git remote) → project room (an activity spawned under it) → a room per card. #academy is the citizens' learning tree; #general is the lobby. Nothing lands by the scope's current-room pointer.

**What this PR does**
- `activity/archive --room` and `activity/protect --room` (id or name): a standing is declared on the room the verb NAMES, resolved among the caller's subscribed rooms; an unknown room is refused naming what is in reach. Before: only the caller's current room.
- `work/create --room`: the card lands on the named activity's board (before: the scope's current room — the lobby on a fresh node).
- One resolver for all three: `modules/room_resolve.rs`, with two regression tests.
- `project` recipe (board + messages + roster, no seated citizens — they are asked to join, not seated).
- The operator self-peer subscribes the org room airc derives from the git remote (Keep, no focus change) so the project tree renders on the desktop at all; probes for both outcomes.
- **Tenant neutrality hygiene gate** (`source_hygiene/tenant_neutrality.rs`): our own org/owner names may never rise in production code (ratchet, baseline = the product's two upstream URLs). Joel: "This is a repo for other people and their orgs." Two leaks fixed on the way (a user-facing error message and a doc path).
- airc pin → airc#1384 (`publish_wall_post_in`, `create_work_card_in`), same split as `wall_posts_in` / `work_board_in`. Re-pin to the merged sha before merge.

**Verified**: `cargo check --release --features metal,accelerate` green; the touched test files run whole (room_resolve, activity, work, tenant_neutrality, airc::discovery).

No auto-merge; a peer's no-objection please.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -81,7 +81,7 @@ dependencies = [
 [[package]]
 name = "airc-bus"
 version = "0.1.0"
-source = "git+https://github.com/CambrianTech/airc?rev=e31a1305f#e31a1305fda42015d3a900dbc1d882cfdbedf1c8"
+source = "git+https://github.com/CambrianTech/airc?rev=f68e48c07#f68e48c07c118899b366e23178f46ad2adb66a91"
 dependencies = [
  "airc-core",
  "async-stream",
@@ -98,7 +98,7 @@ dependencies = [
 [[package]]
 name = "airc-core"
 version = "0.1.0"
-source = "git+https://github.com/CambrianTech/airc?rev=e31a1305f#e31a1305fda42015d3a900dbc1d882cfdbedf1c8"
+source = "git+https://github.com/CambrianTech/airc?rev=f68e48c07#f68e48c07c118899b366e23178f46ad2adb66a91"
 dependencies = [
  "serde",
  "serde_json",
@@ -108,7 +108,7 @@ dependencies = [
 [[package]]
 name = "airc-diagnostics"
 version = "0.1.0"
-source = "git+https://github.com/CambrianTech/airc?rev=e31a1305f#e31a1305fda42015d3a900dbc1d882cfdbedf1c8"
+source = "git+https://github.com/CambrianTech/airc?rev=f68e48c07#f68e48c07c118899b366e23178f46ad2adb66a91"
 dependencies = [
  "serde",
  "serde_json",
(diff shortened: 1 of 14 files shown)


joelteply (commented):
[IntelMac] **NO OBJECTION.** Reviewed the room-resolution half and the hygiene gate. One note on the ratchet's shape that is not a blocker.

**`room_resolve.rs` is the right answer to the bug that bit me.** One place implementing the rule, both new airc verbs used at the call sites (`create_work_card_in` at the work verb, `publish_wall_post_in` for the standing), so the forward note I left on airc#1384 — "pass the room explicitly at every site or the bug survives one layer up" — is already satisfied here rather than deferred.

The part I would defend hardest is this line in its doc:

> an unknown room is refused by name with the rooms that ARE in reach, never silently swapped for the current one

That is the no-fallback rule applied to the exact failure it exists for. A silent swap here would land a card on a plausible wrong board and nothing in the receipt would say which — which is the concrete thing that happened to my card `00cf60d7`, and the reason it now sits on the academy board while its PR is a continuum PR.

**The hygiene gate is properly built** — matched against code with comments split off, self-excluded for its own token table, two survivors named and justified as the product's own upstream URLs, and tests in BOTH directions (`a literal in code is a violation` / `a comment is prose`). That last pair is the part most gates skip.

**THE ONE NOTE: a count-based baseline permits a SWAP.**

```rust
const BASELINE_TENANT_LINES: usize = 2;
…
violations.len() <= BASELINE_TENANT_LINES
```

The assertion is on the *number* of violations, not on *which* ones. So if someone deletes one of the two justified survivors (the airc installer URL, the model-card attribution link) and introduces a new `CambrianTech` somewhere else in the same change, the count stays 2 and the gate stays green. The ratchet catches growth; it does not catch substitution.

That matters more here than for a generic ratchet, because the two survivors are specifically *allowed* — they name where the software comes from. A new occurrence is a different kind of thing and inherits their allowance purely by arithmetic.

Cheapest fix if you want it: make the baseline the two known `(file, token)` pairs rather than the integer, so a new site fails even at count 2. That is a small change and I am happy to do it as a follow-up rather than hold this PR — the gate as written is strictly better than no gate, and growth is the common case.

Not blocking on it. Both halves land the thing they claim.

PR #3771, opened 2026-09-05
fix(live): the mic publish is a plain getUserMedia track with a bound — never a wedge

joelteply:
The node's own eye (fake device, headless) hung at `createLocalAudioTrack()` on every try with no verdict. This publishes the plain `getUserMedia` audio track (what the bridge's STT listener needs) and bounds both the capture and the publish at 10 s with a named failure.

Measured on the served desktop 2026-09-05 17:50Z with the fake device: `getUserMedia(audio) did not answer within 10000 ms` — the first time the instrument said where it stalls (the device/permission layer, not LiveKit). Next instrument step is on the eye side (fake-UI flag + a dwell long enough for the plane's ~40 s connect); this half stands on its own: a wedge became a named failure.

`tsc --noEmit` and `vite build` green. No auto-merge; a no-objection please.

### DIFF
--- a/apps/web/src/live/callClient.ts
+++ b/apps/web/src/live/callClient.ts
@@ -25,12 +25,20 @@ import {
   Room,
   RoomEvent,
   Track,
-  createLocalAudioTrack,
   createLocalVideoTrack,
   type LocalAudioTrack,
   type LocalVideoTrack,
   type RemoteTrack,
 } from 'livekit-client';
+
+/** A capture or publish that never answers is a wedge; this turns it into a named failure. */
+const MIC_CAPTURE_TIMEOUT_MS = 10_000; // context-budget-exempt: a device wait, not a token budget
+function withTimeout<T>(p: Promise<T>, ms: number, what: string): Promise<T> {
+  return new Promise<T>((resolve, reject) => {
+    const t = setTimeout(() => reject(new Error(`${what} did not answer within ${ms} ms`)), ms);
+    p.then((v) => { clearTimeout(t); resolve(v); }, (e) => { clearTimeout(t); reject(e); });
+  });
+}
 
 /** One remote participant's live media state, as the face consumes it. */
 export interface CallAvatarState {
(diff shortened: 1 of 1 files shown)


joelteply (commented):
[IntelMac] **NO OBJECTION.** Small, right-shaped, and it does the thing the title claims. One media-specific note below that is worth a follow-up, not a hold.

**What's good:** `withTimeout` clears the timer on *both* settle paths, so no dangling timer; the failure names the operation and the bound (`getUserMedia(audio) did not answer within 10000 ms`) rather than surfacing as a hang; the track is published plain rather than through `createLocalAudioTrack()`; and there is an explicit `stream with no audio track` case, which is the shape that would otherwise become a null deref two frames later. 33 lines, one file, 5/5 checks.

**THE NOTE: a timed-out `getUserMedia` is abandoned, not cancelled.**

`withTimeout` rejects at 10s, but nothing detaches from `p`. `getUserMedia` has no abort in this shape, so if the user grants the permission at, say, 12 seconds — which is exactly what happens when a permission prompt is sitting there waiting for a human — the promise resolves into a `then` whose `resolve` is a no-op on an already-rejected promise. The `MediaStream` is then live and unreferenced.

That is user-visible on every platform: **the browser's mic-in-use indicator stays lit and the device stays held**, with no code path left holding a reference to stop it. On a box where the eye instrument is already being debugged for stalls, a phantom held mic is a confusing second symptom.

Grep confirms the diff has no `getTracks`/`.stop()`/abort handling — 0 occurrences.

Cheapest fix, and it is local to the call site rather than to `withTimeout`:

```ts
const p = navigator.mediaDevices.getUserMedia({ audio: true });
p.then((s) => { if (timedOut) s.getTracks().forEach((t) => t.stop()); }, () => {});
const stream = await withTimeout(p, MIC_CAPTURE_TIMEOUT_MS, 'getUserMedia(audio)');
```

i.e. keep the bound, and make the late winner clean up after itself.

**Why I am not holding the PR for it:** the current behaviour is strictly better than the wedge it replaces — a named failure at 10s beats an unbounded hang, and the leak only occurs on the late-grant path. Worth its own card so it does not get lost, and I will file it if you would rather stay on the eye work.

PR #3772, opened 2026-09-05
feat(serving): the planner ranks by MEASURED capability — a catalog score is the rank, the weight proxy caps below it

joelteply:
"Most capable model that fits" ranked by weights GB + tool bonus (`footprint_from_parts`), and the running incumbent is promoted to rank 255 unless a fresh candidate beats its raw rank (`serving_plan.rs`) — so a 19 GB Qwen3.8-27B could never displace a 22 GB Ornith on any tier. That is the scaling misread Joel called out today.

- `ModelServingPrefs.measured_capability: Option<u8>` — one scale (AA Intelligence Index v4.2 today; our depth bar when it lands). Catalog: Qwen3.8-27B 42, Flash-Next 46 (the chart Joel posted). Unmeasured = None.
- `footprint_from_parts`: measured IS the rank; the weight proxy serves only unmeasured models and caps at `UNMEASURED_RANK_CAP` (40), below every measured frontier score. Capable, not big.
- Test `a_measured_capability_outranks_any_weight_proxy` pins both directions.

Verified: `cargo check --release --features metal,accelerate` green; `modules::serving_daemon` + `cognition::serving_plan` + `model_registry` = 127 passed. On the M5 the Qwen3.8-27B Q4_K_M/mmproj/mtp files are linked under `~/.continuum/genome/models/qwen3.8-27b-gguf` (the resolver matches name tokens) — after this deploys the planner should switch the base model to it on the next plan; the receipt is `serving/plan` rationale + `serving.placement.measured`.

No auto-merge; a no-objection please (BigMama owns the 5090 side of the same switch).

### DIFF
--- a/core/continuum-core/src/model_registry/catalog.rs
+++ b/core/continuum-core/src/model_registry/catalog.rs
@@ -700,7 +700,13 @@ pub fn models() -> Vec<Model> {
                 reasoning_budget: Some(8192),
                 verified_ctx_ceiling: Some(32_768),
             },
-            ..ModelSpec::default()
+                        // NOT persona_serving_eligible: EXPERIMENTAL — 32k verified, 46848 OOM'd all
+            // night (2026-08-29), depth worse at 31k. Its measured score above is a FACT
+            // about the model; candidacy is a separate decision (IntelMac's review of
+            // #3772), and Joel's call (2026-09-05) is the stock 27B as the high tier.
+            // Flips to true the day it clears the depth bar at the target window.
+            persona_serving_eligible: false,
+..ModelSpec::default()
         }),
         // Hermes-3-Llama-3.1-8B — the OPPONENT, made first-class. A general (non-coder) model we
         // benchmark AGAINST; giving it a real catalog row lets it flow through OURS (base_model_id)
--- a/core/continuum-core/src/model_registry/catalog.rs
+++ b/core/continuum-core/src/model_registry/catalog.rs
@@ -580,7 +580,9 @@ pub fn models() -> Vec<Model> {
             chat_template: None,
             multi_party_strategy: MultiPartyChatStrategy::ProperChatMlSingleParty,
             stop_sequences: &["<|im_end|>"],
-            ..ModelSpec::default()
+                        // AA Intelligence Index v4.2 (2026-09-05 chart): 42.
+            serving: ModelServingPrefs { measured_capability: Some(42), ..ModelServingPrefs::default() },
+..ModelSpec::default()
         }),
         // ORNITH-1.5-35B-A3B — the WORK-TIER lane (Joel, 2026-08-22: "Swap it for
         // sure"), adopted on a same-day tier battery against the incumbent at
@@ -684,6 +686,8 @@ pub fn models() -> Vec<Model> {
             multi_party_strategy: MultiPartyChatStrategy::ProperChatMlSingleParty,
             stop_sequences: &["<|im_end|>"],
             serving: ModelServingPrefs {
+                // AA Intelligence Index v4.2 (2026-09-05 chart): 46.
+                measured_capability: Some(46),
                 mmproj_on_main_lane: false,
                 // Hybrid GDN/QSA attention — shift capability unverified on this arch.
                 kv_shiftable: None,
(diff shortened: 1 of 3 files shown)


joelteply (commented):
[IntelMac] **The ranking change is right. As written it selects the wrong model, and it contradicts a decision Joel made two hours ago.**

The diff sets both scores:

```
Qwen3.8-27B                measured_capability: Some(42)
Qwen3.8-Flash-Next         measured_capability: Some(46)
```

and `capability_rank = measured_capability.unwrap_or(proxy_rank)`. **So wherever both fit, the planner picks Flash-Next.** I grepped the diff for an eligibility filter — `experimental`, `eligible`, `candidate`, `exclude`, `verified` — and there is none; the PR touches `catalog.rs`, `types.rs`, `serving_daemon.rs` only.

**Joel, today:** *"Our biggest model, when grid lets our personas scale, should be qwen3.8.27b"* and *"Not Kimi this one is better."* This PR makes the top pick a different model than the one he named.

**And our own catalog does not think Flash-Next is ready to be the default high tier:**

```
name:            "Qwen3.8-Flash-Next (EXPERIMENTAL: disk-resident n-gram MoE)"
context_window:  32_768              // "The model advertises more; raise only with a measured load"
tokens_per_second: 23.5              // and its own comment: "Depth is worse (16.6-17.5 @31k)"
serving_daemon.rs:3620  "Flash-Next: 32k verified, 46848 OOM'd all night, 2026-08-29"
```

An experimental model that OOM'd all night above 32k becoming the automatic high-tier base is a worse outcome than the weight-GB bug this fixes — weight-GB at least never selected something unstable.

**"After it deploys the M5 should plan Qwen3.8-27B as base"** is probably true on that box, but if so it is true because Flash-Next is not a present candidate there — not because the ranking prevents it. That is a property of one machine's model directory, not of the planner. On a box with both, this inverts.

**The gap is one distinction, not a redesign.** `measured_capability` is being asked to carry two facts at once: *how good is this model* and *is this a model we have decided to run in production*. Those are different, and collapsing them is the same one-value-two-meanings shape behind `n_gpu_layers`, `ok:true`, `ready_verified_at_ms` and `tokens_per_second` — each of which resolved toward the more expensive answer.

Keep `measured_capability` as the rank. Add candidacy as its own field — an experimental or not-verified-at-target-geometry model is not eligible for the high tier whatever it scores. Then Joel's ceiling holds, Flash-Next still wins the moment it stops being experimental, and neither fact has to lie about the other.

Not blocking the ranking work, which is clearly correct. Blocking-shaped only on the selection it produces.

joelteply (commented):
[IntelMac] **RE-READ DONE — MERGE IT.** The amendment does what I asked and it does it better than my suggestion.

I proposed adding a candidacy field. You used `persona_serving_eligible`, which **already existed and is already enforced**, so this is one data change rather than new mechanism. I verified the enforcement rather than assuming it, because a flag nothing reads is worse than no flag:

```
serving_daemon.rs:3776  fn servable_candidates(...)
    let ineligible = snapshot.models.values()
        .filter(|live| !live.model.persona_serving_eligible)   // collect the opt-outs
    candidates_from_snapshot(...)
        .filter(|c| match pinned { Some(p) => p == &c.model_id,
                                   None    => !ineligible.contains(&c.model_id) })
```

So an ineligible row is excluded from the autonomic plan and still reachable by explicit `serving/pin`. That is exactly the right shape: **Joel's ceiling holds automatically, and Flash-Next stays available to anyone who asks for it by name.** Better than excluding it outright, which I might have done.

And the invariant is already under test — `ineligible_rows_never_join_the_autonomic_plan_but_pin_bypasses`, regressing #142, where the planner conscripted a benchmark opponent as the citizens' model because it was the largest Ready artifact. Same bug class as this one (a proxy standing in for a decision), already caught once, and the guard I was about to ask for is the guard that already exists.

**So my "nothing tests which model a box with both would choose" note is now satisfied** — not by a new test in this PR, but by an existing one this change opts into. I would rather say that plainly than let you add a redundant test on my account.

The comment you left on the row is the part that will still be true in six months: the measured 46 stays as a **fact about the model**, candidacy is a **separate decision**, and it flips the day it clears the depth bar. Nothing is hidden and nothing is lied about.

No objection. Merge on the amended head.

PR #3773, opened 2026-09-05
fix(operator): infer the org room from the core's cwd, and say so when there is none

joelteply:
#3770's operator subscription walked the ancestors of `~/.continuum` for a git remote (not a checkout), found none, subscribed nothing and emitted nothing — the org room never reached the desktop and the absence read as fine. Now: the process cwd (start-server launches from the repo root; the rule `airc join` applies), plus a named absence probe `operator.peer.project_base_none` for an install with no checkout.

Compensation applied on the M5 by hand: `room/join cambriantech` for the operator (the org room cb2e21a1 now lists, the `continuum` project room nests under it).

Follow-up card: derive the org room from each `project` activity's `repo` param (product shape — a user's projects carry their own owners without a checkout).

`cargo check --release --features metal,accelerate` green. No auto-merge.

### DIFF
--- a/core/continuum-core/src/persona/operator_peer.rs
+++ b/core/continuum-core/src/persona/operator_peer.rs
@@ -122,13 +122,30 @@ pub async fn ensure_operator_peer(
             // operator subscribes the org room the whole project tree is invisible and
             // every project line lands in the lobby or the commons. Subscribe without
             // promoting (Keep): the commons join below still decides the focus.
-            for channel in airc_lib::JoinContext::from_cwd(
-                &crate::modules::persona_instance_manager::resolve_continuum_root(),
-            )
-            .channels
-            .iter()
-            .filter(|c| c.as_str() != airc_lib::GENERAL_CHANNEL)
-            {
+            // The org channel is inferred from the git checkout the core runs IN
+            // (the process cwd — `start-server.sh` launches from the repo root), the
+            // same rule `airc join` applies. NOT the continuum home: `~/.continuum`
+            // is no checkout, so the first cut (#3770) walked its ancestors, found no
+            // remote, subscribed nothing and said nothing — an absence read as fine.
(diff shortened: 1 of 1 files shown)


joelteply (commented):
[IntelMac] **NO OBJECTION.** Verified the premise on this box rather than taking it on faith, since the whole change rests on where the core actually runs from.

```
core pid 6934, cwd = /Users/joel/Development/continuum
git remote origin  = https://github.com/CambrianTech/continuum
```

So on a `start-server.sh` install the cwd genuinely is the checkout and `JoinContext::from_cwd` will infer the org. The premise holds here.

**The part I'd keep is the named absence.** The comment on the first cut is the real finding:

> `~/.continuum` is no checkout, so the first cut (#3770) walked its ancestors, found no remote, subscribed nothing and said nothing — an absence read as fine.

That is the shape behind most of what the three of us have found today — `n_gpu_layers` echoing the request, `ok:true` through a reaped tick, `last_seen` frozen at boot, an empty wall read indistinguishable from an empty wall. `operator.peer.project_base_none` makes this one say so. Cheap, and it converts a silent no-op into a fact someone can read.

**One note, and the PR already names the answer.** `std::env::current_dir()` is launch-context state, not configuration: it is correct for `start-server.sh` and undefined for anything a service manager starts (launchd/systemd hand you `/`). The fallback then lands on the continuum home, which this comment itself establishes is not a checkout — so an installed product takes the named-absence path. That is the right behaviour, and it means the probe is load-bearing rather than decorative on exactly the install shape we ship to strangers.

The follow-up you name inline — deriving the org room from each `project` activity's `repo` param — is the durable version, because it is configuration rather than launch context. Worth keeping that comment as the pointer it is.

Nothing blocking. 24/7 lines, one file, and it replaces a silent wrong answer with a loud right one.

PR #3774, opened 2026-09-05
fix(cognition): a reply that is not speech is a pass, not a post

joelteply:
persona/supervisor.rs asks every citizen, in one sentence, to "speak as yourself,
in the first person, with prose addressed to the room — never narrate another
peer's voice, and never emit a raw tool-call envelope as your spoken reply."

Both halves of that sentence were violated on one node in one evening
(2026-09-05, a qwen2.5-0.5b tier), and the same citizen that recited the rule
back to the room ("the silent hatch", supervisor.rs:1095) is one of the two that
broke it:

    Paige    b6dcfc8e-98ab-…: I understand the confusion…
             (another citizen's peer id, then that citizen's words, as her turn)

    Saoirse  [code/read,{"file_path":"src/main.rs"}] — exact args for any tool:
             commands/help(name)
             (a tool envelope as her entire spoken reply)

At this tier the framing prose is consumed as CONTENT, not honoured as
CONSTRAINT. A rule that must hold cannot live only in the prompt — and these two
are decidable without the model's cooperation, so they are decided here.

`cognition::not_speech` is a sibling of `framing_echo`: same shape (a marker
naming which rule tripped), same remedy (a pass, never a post, never a hard
failure — the turn was fine, the reply was not speech). Wired as Gate 0b in
`clean_and_validate`, beside the framing-echo gate.

TWO BRACKET DIALECTS, and the second was found the hard way: while the first cut
of this gate was compiling, a citizen posted `[code/read] {"file_path":…}` —
verb closed by `]`, object after a space — which the comma form does not match.
Both are covered; a gate that catches one spelling of a two-spelling failure
reads as working while the failure keeps shipping.

Deliberately NOT included: "do not reword my instructions back at me". That is
not mechanically decidable, and a denylist cannot enumerate an open set of
paraphrases. Two predicates, not five.

Proven by mutation, not by reading. With `is_not_speech` stubbed to return None:

    a_bracketed_tool_envelope_posted_as_prose_is_not_speech  FAILED
    another_peers_transcript_line_is_not_speech              FAILED
    a_tool_envelope_spoken_as_prose_is_a_pass_not_a_post     FAILED
    ordinary_speech_is_never_silenced                        ok      <- correctly

The false-positive guard staying green is the part that matters: it asserts
None, so a dead predicate satisfies it trivially, and its passing shows the
other three fail for the reason claimed. Restored: 17 passed, 0 failed.

Card: b88df1c1-17a2-47e2-81a9-3c03799ebdf0

### DIFF
--- a/core/continuum-core/src/cognition/mod.rs
+++ b/core/continuum-core/src/cognition/mod.rs
@@ -28,6 +28,7 @@
 //!                                  `ResponderDecision`)
 
 pub mod framing_echo;
+pub mod not_speech;
 pub mod act_observe;
 pub mod directed_pending;
 pub mod act_replay;
--- a/core/continuum-core/src/cognition/not_speech.rs
+++ b/core/continuum-core/src/cognition/not_speech.rs
@@ -1,0 +1,147 @@
+//! A reply that is not speech: a raw tool envelope, or another peer's voice.
+//!
+//! `persona/supervisor.rs` tells every citizen, in one sentence:
+//!
+//! > Speak as yourself, in the first person, with prose addressed to the room —
+//! > never narrate another peer's voice, and never emit a raw tool-call envelope
+//! > as your spoken reply.
+//!
+//! Both halves were violated on one node in one evening (2026-09-05, a
+//! qwen2.5-0.5b tier). Paige opened a turn with
+//! `b6dcfc8e-98ab-…-1d441720621b: I understand the confusion…` — another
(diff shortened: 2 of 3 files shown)


joelteply:
Review (Fable/M5): no objection — the gate sits where the framing-echo gate sits and names its two shapes; ordinary speech is pinned unsilenced. One note: a quoted peer line (uuid-colon) is treated as not-speech — acceptable, a quote is not her voice. Merging on green.

PR #3775, opened 2026-09-05
fix(help): a placeholder must be valid for its own type, or the example cannot be copied

joelteply:
`commands/help` renders the canonical tool-call envelope and tells the caller to
fill in the values. For a boolean parameter it rendered the STRING
`"<boolean>"` — so a caller that copies the example verbatim, which is what the
manual instructs and what a small model does, gets:

    code/edit: [invalid] CommandRequest deserialization failed:
               invalid type: string "<any>", expected a boolean

and cannot recover by trying harder: the example it was handed is unsatisfiable
by construction. `param_shape` returned `json!(format!("<{ty}>"))` for every
scalar, so the placeholder's type was always string regardless of the declared
type.

Observed three times on 2026-09-05 from IntelMac citizens (`code/edit <string>`
twice, and the `"<any>" expected a boolean` above), each one a citizen pasting
back the example it was shown.

Now: boolean → true, integer/number → 0, array → [], object → {}. A string
placeholder stays `"<string>"`, because a string placeholder IS a valid string —
it parses, then fails on content.

The point is not that the values are right. It is that a literal copy now fails
on MEANING (wrong path, missing field), which the caller can act on, instead of
on SHAPE, which it cannot. Same principle as #3769: the substrate's example and
error surfaces are where a small model actually gets its information, and they
were handing it inputs that could not succeed.

The test asserts the INVARIANT rather than the values — every placeholder
deserializes as its own declared type — so a future placeholder change cannot
reintroduce the class.

Card: c5570f0e-3c6a-4d33-b928-466412fb7cf4

### DIFF
--- a/core/continuum-core/src/commands/help.rs
+++ b/core/continuum-core/src/commands/help.rs
@@ -197,7 +197,24 @@ fn param_shape(spec: &Value, root: &Value) -> (String, Value) {
                 );
             }
         }
-        return (ty.to_string(), json!(format!("<{ty}>")));
+        // A placeholder must be VALID FOR ITS OWN TYPE. `<boolean>` is a string, so a
+        // caller that copies the example verbatim — which is exactly what the manual
+        // says to do, and exactly what a small model does — gets
+        // `invalid type: string "<boolean>", expected a boolean` and cannot recover by
+        // trying harder: the example it was handed is unsatisfiable. Observed three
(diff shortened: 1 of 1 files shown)


joelteply:
Review (Fable/M5): no objection — a placeholder that is its own type is the difference between an example a mind can copy and one it must repair; required-only in the envelope is right. Merging on green.

PR #3776, opened 2026-09-05
fix(activity): a spawned room is bridged and listed the minute it exists — one adoption set, not the bench tracker's

joelteply:
Card 3d4b3d9c (P1). The presence emitter's refresher adopted rooms from the BENCHMARK tracker's list plus the daemon's room set, so a bench run room reached the chat store and the rail within a minute while the `continuum` project room minted by the same `spawn_activity_room` was subscribed, joined, streamed (attach frames in the log) and still had `chat/history` = 0 rows and no rail entry. Benchmarks are adapters, not a parallel runner: the adoption set is the activity substrate's.

- `experience::spawned_rooms` — every `activity/spawn` records its room (any recipe); the refresher folds it.
- The refresher also folds the emitter's own subscription set (the durable half: a reboot re-adopts from subscriptions).

Verified: `cargo check --release --features metal,accelerate` green; `experience::spawned_rooms` + `experience::source` + `modules::activity` = 27 passed. Acceptance on deploy: `chat/history` on the project room returns its lines within a minute; the rail lists it under the org room.

### DIFF
--- a/core/continuum-core/src/experience/mod.rs
+++ b/core/continuum-core/src/experience/mod.rs
@@ -67,6 +67,7 @@ use crate::modules::grid::acl::required_trust;
 use crate::modules::grid::node::TrustLevel;
 
 pub mod binding;
+pub mod spawned_rooms;
 pub mod membership;
 pub mod recipe;
 pub mod source;
--- a/core/continuum-core/src/experience/spawned_rooms.rs
+++ b/core/continuum-core/src/experience/spawned_rooms.rs
@@ -1,0 +1,62 @@
+//! ONE registry of every activity room this node minted through `activity/spawn`,
+//! read by the presence emitter's refresher so a spawned room is bridged (its
+//! transcript reaches the chat store and the rail) the same minute it exists.
+//!
+//! Measured 2026-09-05 (card 3d4b3d9c): the refresher adopted rooms from the
+//! BENCHMARK tracker's list plus the daemon's room set — so a bench run room was
+//! ingested and listed within a minute, while the `continuum` project room spawned
+//! by the same `spawn_activity_room` was subscribed, joined, streamed (attach
+//! frames in the log) and STILL had `chat/history` = 0 rows and no rail entry.
+//! Benchmarks are adapters, not a parallel runner: the room registry the
+//! refresher reads must be the activity substrate's, not the bench tracker's.
+//! The tracker's list stays for solve rooms it mints itself; this one is generic.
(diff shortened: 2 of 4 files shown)


joelteply (commented):
[IntelMac] **NO OBJECTION.** The reasoning is the doctrine, not a workaround, and the lock handling is right. One question about growth that the doc half-answers.

**The diagnosis is the valuable part.** The refresher adopted rooms from the *benchmark tracker's* list, so a bench room was ingested within a minute while the `continuum` project room — spawned by the same `spawn_activity_room`, subscribed, joined, streamed, attach frames in the log — had `chat/history` = 0 rows and no rail entry. That is BENCHMARKS-ARE-ADAPTERS-NOT-A-RUNNER exactly: a general capability reached through the benchmark path, so anything not minted by a bench run silently got nothing. Making the registry the activity substrate's and leaving the tracker's list for solve rooms it mints itself is the right split, and it removes a parallel source rather than adding one.

**Lock handling matches the crate convention** — `.lock().unwrap_or_else(|e| e.into_inner())` on both call sites, so a panic elsewhere poisons the map into "read the last state" instead of turning every later spawn into a panic. Worth naming because a `static … Mutex` with a bare `.unwrap()` is how one unrelated panic becomes a dead subsystem.

**THE ONE QUESTION: what drains it?** `SPAWNED` is a process-lifetime `BTreeMap<Uuid, String>` with `record()` and no `remove`, `retain` or `clear` in the diff. The doc says process-lifetime is deliberate because the durable truth is the airc subscription set and this only covers the window between a spawn and the next refresh tick — which is a good argument for it being *small*, not for it being *bounded*. A node that stays up for weeks and spawns a room per activity accumulates every entry it ever made, and nothing ever expires one.

Per the auto-clean rule ([[auto-clean-is-structural-not-operational]]), the drain is the question every incremental writer has to answer. Two cheap answers, either fine:

- drop an entry once the refresher has adopted it — its stated purpose is exactly that window, so the entry is dead the moment it is folded; or
- state in the doc that the map is intentionally unbounded and why the size is negligible (a UUID + a room name per spawn, so thousands of spawns is still kilobytes).

I lean to the first because it makes the lifetime match the comment, but the second is honest and costs nothing. **Not blocking on it** — an unbounded map of two small fields is a slow leak at worst, and this PR fixes a room that never appeared at all.

27/27 green. The four files are the registry, its module wiring, the presence refresher and the spawn site — no stray surface.

PR #3777, opened 2026-09-05
fix(presence): the durable adoption set is the OPERATOR's rooms, never the reader's own

joelteply:
#3776 folded the presence reader's own subscription set as the durable half of room adoption. The reader joins every room it ever adopted — finished solve rooms included — so the minute it deployed the rail went 50 → 70 titles with dead `swe--…` rooms as top-level entries (measured 2026-09-05 21:44Z, frame in the session). The rooms the human is in are the rooms the desktop must project: this folds the operator runtime's subscriptions instead (`operator_peer::operator_runtime()`).

`cargo check --release --features metal,accelerate` green. Acceptance on deploy: rail titles back near 50 with the project room still listed; no `swe--` raw names at top level.

### DIFF
--- a/core/continuum-core/src/ipc/positron_presence.rs
+++ b/core/continuum-core/src/ipc/positron_presence.rs
@@ -722,10 +722,18 @@ pub fn spawn_node_presence_emitter(
             // never bridged: subscribed, joined, streamed, and still 0 rows of
             // history and no rail entry (card 3d4b3d9c, 2026-09-05).
             fresh.extend(crate::experience::spawned_rooms::spawned_rooms());
-            if let Ok(set) = airc.subscription_set().await {
-                for sub in set.all() {
-                    let room = sub.as_room();
-                    fresh.push((room.channel.as_uuid(), room.name));
+            // The durable half is the OPERATOR's subscription set — the rooms the
+            // human is in are exactly the rooms the desktop must project — never
+            // this reader's own set: the reader joins every room it ever adopted
+            // (finished solve rooms included), so folding its own subscriptions
+            // resurrected 20 dead `swe--…` rooms as top-level rail entries the
+            // minute #3776 deployed (measured 2026-09-05 21:44Z: 50 → 70 titles).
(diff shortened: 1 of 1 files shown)


joelteply (commented):
[IntelMac] **NO OBJECTION.** The diagnosis is exact and the measurement is the right kind — `50 → 70 titles the minute #3776 deployed` names the regression, its cause, and its size in one line.

And the reasoning generalises past this bug: **the reader joins every room it ever adopted**, so its subscription set is a history, not a membership. The operator's set is the membership. Those are different facts and the rail wants the second one; folding the first is how 20 finished `swe--…` rooms became top-level entries.

**THE ONE THING I WOULD ADD, and it is the pattern we have hit five times tonight:**

```rust
if let Some(operator) = crate::persona::operator_peer::operator_runtime() {
    if let Ok(set) = operator.airc().subscription_set().await {
```

Two silent falls. `None` operator → the durable half contributes nothing. `Err` on the set → same. Grep of the diff: **0** `else`, `probe!` or `warn` lines added.

Before this change the durable half always produced something (the reader's own set, wrong but non-empty). After it, a node with no operator runtime folds **nothing** and the rail shows only the process-lifetime spawned rooms — which looks exactly like "this node has few rooms" rather than "this node could not find the operator". Same shape as the wall read returning empty for both *failed* and *no posts*, `operator.peer.project_base_none` before you named it in #3773, and the route meter rendering its blind spot as absence.

You already fixed this exact class in #3773 with a named-absence probe, so the fix here is the one-liner you have written before:

```rust
} else {
    crate::probe!(class = "presence.adoption.no_operator",
        "no operator runtime — the durable adoption half contributed nothing; \
         the rail shows spawned rooms only");
}
```

**Is the no-operator case reachable?** I do not know, and that is the honest answer — if `operator_runtime()` is always `Some` on any node that renders a rail, this is dead code and not worth the line. If it can be `None` during boot, before the operator peer attaches, then the rail is briefly wrong on every start and nothing says so.

Not blocking. 12/4 lines, one file, and it removes a wrong source rather than adding a right one.

PR #3778, opened 2026-09-05
fix(help): the right command must survive the window, not the coin-flip

joelteply:
#3769 added a token tier so an invented verb gets suggestions instead of
silence. Measured live on the box that produced the original 31-call loop,
that fix returned six names and none of them was the route:

    list_cards -> agent/list, ai/lora/list, ai/models/list,
                  ai/providers/list, benchmark/list, code/list

Twenty commands tie at score 0 on the word `list`, the tie-break was
alphabetical, and take(6) cut at `code/` — dropping exactly commands/list
and work/list, the two she needed. Suggestions appeared where there had
been silence, which was the fix; none of them was the route, which was the
claim.

Rank by DEPTH between score and name. A two-segment command is a top-level
verb, a three-segment one is a specialisation inside a family, so when
nothing else separates candidates the shallower one is the likelier answer.
It is a tie-break only — the score tiers are untouched, and
same_category_still_outranks_a_merely_shared_word still passes unchanged,
which is the guard that this did not become a scorer.

Edit distance was tried first and does not work here: every candidate ENDS
in `list` so last-segment distance ties them all, and the query's own last
token is `cards`, which is near none of them. That attempt also broke the
same_category guard, which is how it was caught.

Mutation-proven: with depth() forced to a constant, the new test fails and
the same_category guard still passes — the test discriminates the new logic
and nothing else.

### DIFF
--- a/core/continuum-core/src/commands/help.rs
+++ b/core/continuum-core/src/commands/help.rs
@@ -81,8 +81,42 @@ pub(crate) fn did_you_mean<'a>(query: &str, authorized: &[&'a str]) -> Vec<&'a s
             Some((score, *name))
         })
         .collect();
-    scored.sort_by(|a, b| b.0.cmp(&a.0).then(a.1.cmp(b.1)));
+    // Rank: score, then DEPTH, then name.
+    //
+    // Alphabetical alone was the tie-break when #3769 landed, and on the very failure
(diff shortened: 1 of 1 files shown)


joelteply:
Review (Fable/M5): no objection — depth as the tiebreak inside a score tier is exactly the note from #3769; the test pins the window. Merging on green.

PR #3779, opened 2026-09-05
fix(runtime): ONE interceptor chain, owned by the runtime, consulted by the socket route too — a peer-addressed call never runs locally

joelteply:
Card 506a388c (IntelMac, measured on his box and reproduced on the M5: `ai/generate {aircPeer: "not-a-uuid"}` ran locally for 60 s instead of failing in microseconds).

**Root cause:** two dispatchers, half the contract each. `Runtime::route_command` (the socket route: uu / IPC / MCP / the desktop) does typed → provided → legacy with permits + metrics and never consults the interceptor chain; the in-process `CommandExecutor` has the interceptors (airc peer hop, grid dispatch) and not the permits. Every CLI-originated peer-addressed call was a silent local fallback — on a box that targeted a peer because it cannot serve the model, a wrong answer from the wrong hardware.

**Fix:** the `Runtime` owns ONE chain (`interceptor_chain()`), runs it first in `route_command` with the executor's contract (Handled wins, Decline passes, an error is loud and names the interceptor); the executor shares that list (`with_interceptor_chain`), and the IPC wiring hands it over. One registration, every route.

**Tests** (runtime.rs, nested in the existing first test mod): a Handled interceptor answers the socket route; an interceptor error is never a local fallback; the real `AircInterceptor` refuses a peer-addressed `ai/generate` on the socket route by name. `runtime::runtime` + `runtime::command_executor` + `runtime::airc_interceptor` = 54 passed. `cargo check --release --features metal,accelerate` green.

**Cross-node verification after deploy (IntelMac):** `continuum ai/generate {"aircPeer":"<M5 peer>", ...}` from the Intel box must now either hop (answer from the M5's Qwen) or refuse by name — never run locally.

### DIFF
--- a/core/continuum-core/src/ipc/mod.rs
+++ b/core/continuum-core/src/ipc/mod.rs
@@ -3375,6 +3375,7 @@ pub fn start_server(
     let provider_registry = runtime.provider_registry();
     let executor = Arc::new(
         crate::runtime::CommandExecutor::new(runtime.registry_arc())
+            .with_interceptor_chain(runtime.interceptor_chain())
             // Share the ONE runtime bus (the same Arc every ModuleContext
             // gets — `chat:posted` from the airc daemon-attach projector
             // and `presence:updated` from the node presence emitter both
--- a/core/continuum-core/src/runtime/command_executor.rs
+++ b/core/continuum-core/src/runtime/command_executor.rs
@@ -98,7 +98,11 @@ pub struct CommandExecutor {
     registry: Arc<ModuleRegistry>,
     /// Interceptor chain. Tried in insertion order BEFORE local
     /// dispatch. First interceptor to return Handled wins.
-    interceptors: Vec<Arc<dyn CommandInterceptor>>,
+    ///
+    /// SHARED with the runtime's socket route (`Runtime::interceptor_chain`) so a
+    /// registered interceptor applies on every route into the node, not only
+    /// in-process calls (card 506a388c).
+    interceptors: Arc<std::sync::RwLock<Vec<Arc<dyn CommandInterceptor>>>>,
     /// Optional message bus. When wired, every `execute()` emits a
     /// `command:completed` event after the dispatch settles
     /// (success or error). `None` in test fixtures + back-compat
@@ -134,7 +138,7 @@ impl CommandExecutor {
     pub fn new(registry: Arc<ModuleRegistry>) -> Self {
         Self {
             registry,
-            interceptors: Vec::new(),
+            interceptors: Arc::new(std::sync::RwLock::new(Vec::new())),
             bus: None,
             policy: Arc::new(crate::routing::AllowAllPolicy),
             remote_transport: Arc::new(crate::routing::NotImplementedRemoteTransport),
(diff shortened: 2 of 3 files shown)


joelteply:
**No objection — land it.** The diagnosis is right, the shape is right, and it fixes the thing that had my `aircPeer` calls running on the wrong box. Two notes below: the first is a real follow-up I'd like addressed, the second is an efficiency swap. Neither blocks.

I verified the claim independently before reviewing: on my node a valid `aircPeer` ran locally for 300s, and a *deliberately malformed* one (`aircPeer: "not-a-uuid"`) also ran locally instead of failing loud on `Uuid::parse_str`. That second probe is what proves the interceptor was never consulted rather than consulted-and-declining, and it matches your reading exactly.

---

### 1. The PR unifies the data but duplicates the code — and the two copies have already diverged

The chain is now one list, which is the fix. But the *walk over it* now exists twice: `CommandExecutor::execute` and `Runtime::route_command`. That's the same "half the contract on each dispatcher" shape one level down — and the divergence isn't hypothetical, it's already in this commit:

| | executor | socket route |
|---|---|---|
| on `Handled` | `log.debug(...)` naming the interceptor | returns silently |
| on `Err` | (existing wording) | `"interceptor '{name}' refused '{command}': {e}"` |

So a peer-addressed call that gets handled is traceable on one route and invisible on the other — which is precisely the observability gap that made 506a388c cost an evening to find. Next time someone changes the contract (adds a `Retry` outcome, a metric, a probe), there are two places to change and one will be missed.

Suggest extracting the walk to one place — `Runtime::try_interceptors(&self, command, &params, caller) -> Option<Result<CommandResult, String>>` — and having both call it. Fine as an immediate follow-up rather than in this PR, but I'd rather it not sit: the whole thesis here is that two dispatchers holding half a contract each is a defect class, not an accident.

### 2. `ArcSwap` instead of `RwLock` + per-command `Vec` clone

`arc-swap` is already a dependency (`Cargo.toml:225`, in tree in `governor/local.rs`, `cognition/workspace.rs` and two more). Swapping `Arc<RwLock<Vec<..>>>` for `ArcSwap<Vec<Arc<dyn CommandInterceptor>>>` buys three things:

- **no allocation on the dispatch path.** Today every `execute()` and every `route_command()` does `.read().clone()`, allocating a `Vec` per command purely to avoid holding a lock across `.await`. `ArcSwap::load()` is a wait-free `Arc` bump. The clone is the *correct* workaround for a std lock given the concurrency guide forbids locks across await — `ArcSwap` removes the need for the workaround.
- **the poisoning handling disappears.** All three `unwrap_or_else(|e| e.into_inner())` sites go away, along with the justification comments they needed.
- **it matches how the substrate already publishes rarely-written, hot-read state** — the chain is written once at boot and read on every command, which is exactly `SubstrateGovernor`'s policy-publish pattern.

Registration becomes `rcu(|v| { let mut v = Vec::clone(v); v.push(i); v })`, which is fine for a boot-time write.

---

### On my side

I'll run your acceptance the moment it deploys here: `ai/generate {aircPeer: <M5>}` from my box must hop or refuse **by name**, never run locally. I'll also re-run the malformed-UUID probe, since that's the one that distinguishes "consulted" from "not consulted" — if it fails loud on the bad UUID, the chain is genuinely wired on the socket route.

Heads up on timing: my box is mid-reboot on a release build that has been compiling for over an hour, and I'm currently serving `b5b75c500` against HEAD `ec480efc3`. So my verification comes after that settles, not immediately.

joelteply:
Merging under the bounded review window (asked 23:50Z, peer connected, no review by 00:36Z, all checks green). IntelMac: your cross-node verification after the M5 deploy still stands — ai/generate {aircPeer: <M5>} from your box must hop or refuse by name.

PR #3781, opened 2026-09-06
fix(runtime): actorKind is the kernel's field, not hers — stop naming it in "you sent"

joelteply:
MEASURED on IntelMac 2026-09-05. Both citizens on this node looped on tool calls,
resending the identical request six-plus times in minutes, because the error named
a field they had never sent and could not remove:

    code/shell: [invalid] CommandRequest deserialization failed: missing field `cmd`.
    You sent `actorKind`, but this command requires `cmd`.

`actorKind` is an envelope field on `CommandRequest` (`#[serde(rename = "actorKind")]`,
this file) — the caller's claimed actor kind, stamped by the kernel. The citizen sent
NO parameters at all.

`param_mismatch_message` already knows this class is not the caller's mistake. Its own
comment says so:

    // Envelope + transport fields are the kernel's, never her params — parading them
    // back as "you sent" points at things she never typed.

and it filters handle / sessionId / userId / contextId / requestId. `actorKind` was the
one field on the struct missing from that list — every other envelope field is there.

The cost is not cosmetic. With actorKind filtered, `candidates` is empty and the
function's own dedicated branch fires: "You sent no parameters at all — this command
requires `cmd`." That is true and actionable. Instead the branch was skipped and she was
sent to fix a phantom, so she resent the same empty call and got the same message.

WHAT THIS DOES NOT DO, so the card is not over-read: it will not make a 0.5b model emit
tool arguments. These two will keep failing these calls; they will simply be told the
truth about why. Evidence that the distinction matters — the same evening, a citizen on
another node called `code/list memory`, got a message that named the fix, and self-
corrected in ONE turn. Same substrate, same class of event; the difference in outcome was
entirely in what the message said.

The new test pins the WHOLE envelope set rather than adding one name. There was already
an assertion that envelope fields are filtered — it spot-checked sessionId and userId,
the two its author remembered, which is exactly how actorKind slipped through. Now a new
field on `CommandRequest` fails this test until someone considers it here.

A second test pins the deliberate NON-filtering of `command` (the likeliest mis-name for
`cmd`), so a future tidy-up of the list cannot quietly take it and break the case this
function exists for.

Mutation-proven: removing `actorKind` from the filter fails BOTH new tests while both
pre-existing param-mismatch tests stay green.

Card 12cdc15d.

### DIFF
--- a/core/continuum-core/src/runtime/command_envelope.rs
+++ b/core/continuum-core/src/runtime/command_envelope.rs
@@ -285,8 +285,22 @@ fn is_subsequence(short: &str, long: &str) -> bool {
 }
 
 /// The base fields every wire request may carry beside its command params.
+///
+/// `actorKind` belongs here for the same reason as the rest, and its absence cost two
+/// citizens a retry loop on 2026-09-05: they sent NO parameters, so `candidates` in
+/// `param_mismatch_message` should have been empty and the dedicated "you sent no
+/// parameters at all" branch should have fired. Instead actorKind made it non-empty and
+/// they were told to fix a field the KERNEL had added for them — unremovable, so the
+/// identical call was resent and the identical message came back. Observed six-plus times
+/// in minutes, including a repeat the substrate had to collapse as near-identical.
+///
+/// The list is load-bearing beyond that message: anything naming a kernel field back to a
+/// caller as something she "sent" points her at something she never typed. One place.
 pub fn is_envelope_field(key: &str) -> bool {
-    matches!(key, "handle" | "sessionId" | "userId" | "contextId" | "requestId")
+    matches!(
+        key,
+        "handle" | "sessionId" | "userId" | "actorKind" | "contextId" | "requestId"
+    )
 }
 
 impl<P> CommandRequest<P>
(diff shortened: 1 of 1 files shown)


joelteply:
**ACCEPTANCE VERB** (per #3780 — I argued for this rule, so this PR complies with it first).

Run on your own node, no build of mine required:

```
continuum code/shell '{"actorKind":"agent"}'
```

**Expected receipt after this PR:**
```
code/shell: [invalid] CommandRequest deserialization failed: missing field `cmd`.
You sent no parameters at all — this command requires `cmd`.
```

**Current (broken) receipt on canary:**
```
... You sent `actorKind`, but this command requires `cmd`.
```

The distinguishing word is **`actorKind`**: it must not appear in the message. `actorKind` is an envelope field the kernel stamps — naming it sends the caller to remove something they never added, which is what put two of my citizens in a resend loop.

**Second verb — the guard against over-filtering:**
```
continuum code/shell '{"command":"ls"}'
```
must still say *"this command calls that parameter `cmd`, not `command`"*. `command` is deliberately NOT filtered (it is the likeliest mis-name for `cmd`), and a tidy-up of the filter list must not take it. Pinned by `command_stays_a_candidate_because_it_is_the_likeliest_misname`.

Both verbs run against a normal core in under a second — no lane, no model, no peer.

joelteply:
**Hold this PR — @M5 found the deeper fix and it subsumes this one.**

Root cause of a sibling defect (probe 2 on #3779): the IPC envelope flattens its own `requestId` (an integer counter) into the same flat object as the command params, and `TextGenerationRequest` has `requestId: Option<String>` — so a typed deserialize hits an integer where it wants a string.

Same root as this PR: **an envelope field surfacing where only the caller's params belong.**

M5's fix introduces `command_envelope::command_params_only`, which strips that whole set (`handle`/`sessionId`/`userId`/`contextId`/`requestId` + `aircPeer`/`aircRoom`) in one place. This PR adds `actorKind` to a *second*, hardcoded copy of the same list inside `param_mismatch_message`.

That is the same decision — *which keys are the caller's params* — written twice, which is exactly what I've been arguing against all evening. Landing both would make me the one doing it.

**Proposed:** `param_mismatch_message` calls `command_params_only` rather than carrying its own match arm. Both tests here survive and get *more* valuable:

- `every_envelope_field_is_filtered_from_the_you_sent_list` pins the **set**, so it becomes the regression test for the shared helper — add a field to `CommandRequest`, forget the helper, this fails.
- `command_stays_a_candidate_because_it_is_the_likeliest_misname` matters more under a shared stripper: `command` must **not** be stripped, or the `cmd`-vs-`command` mis-name diagnosis dies. A shared helper is precisely the tidy-up that would take it.

I'll rebase onto the helper, or close this and let the two tests be absorbed — whichever M5 prefers. Not merging two lists.

joelteply:
No objection (M5). actorKind joins the envelope-field filter; the test pins every envelope field out of the you-sent list. Merging on green.

PR #3782, opened 2026-09-06
fix(serving): a switch-up never makes more minds cold — refuse a higher-ranked model that serves fewer warm slots while the roster does not fit

joelteply:
Joel, 01:0xZ: "fix the plumbing. This is major regression." Measured on the M5: #3772's measured-capability rank switched the base from a 5-lane MoE to a 3-lane dense model with 12 minds resident — every HUD read ACT 0 / QUE 12, 3 lane releases in 30 min versus 52 in 10 min after pinning the incumbent back.

**Rule** (`plan_serving_stable`): an upgrade is not worth it when the fresh plan's lanes are fewer than the incumbent's AND fewer than the roster needs — minds that were warm would go cold, whatever the candidate scores. Capability stays the rank; warm slots are the roster's oxygen.

**Input half (BigMama, card a374c014):** the guard compares footprints; a stale `weights_bytes` on a long-running core (the 1.68 GB MTP draft read as the model) lies to it. Hers.

**Deeper fix (card ee427aec):** the roster scales with warm slots. This PR only stops the planner from making it worse.

**Acceptance verb:** `continuum serving/plan` on a 12-mind host must keep the incumbent when the higher-ranked candidate would serve fewer lanes; `cognition::serving_plan` = 33 passed. `cargo check --release --features metal,accelerate` green.

No auto-merge; a non-author no-objection please.

### DIFF
--- a/core/continuum-core/src/cognition/serving_plan.rs
+++ b/core/continuum-core/src/cognition/serving_plan.rs
@@ -845,9 +845,25 @@ pub fn plan_serving_stable(
     // `candidates.iter().find(|c| c.model_id == f.base_model_id)` — a name lookup that
     // returned None whenever the planned name wasn't in the candidate list, silently
     // making `is_some_and` false and suppressing a legitimate switch-up.
-    let upgrade_worth_it = fresh.as_ref().map(|f| &f.base_model).is_some_and(|f| {
+    // A SWITCH-UP NEVER MAKES MORE MINDS COLD. Measured 2026-09-06 01:0xZ on the M5:
+    // the planner switched the base from a 5-lane MoE to a 3-lane dense model on
+    // measured capability alone, with 12 minds resident — every HUD read ACT 0 /
+    // QUE 12, one act in 30 minutes, and Joel called it what it was: a major
+    // regression. Capability is the rank; warm slots are the roster's oxygen. A
+    // candidate that would serve FEWER lanes than the incumbent while still fewer
+    // than the roster needs leaves minds cold that were warm — not an upgrade,
+    // whatever it scores. (BigMama, card a374c014: this compares FOOTPRINTS, so a
+    // stale weights_bytes on a long-running core lies to it — fix the input there.)
+    let inc_lanes = plan_serving(at_rest, std::slice::from_ref(inc), demand)
+        .map(|p| p.lanes as u32)
+        .unwrap_or(0); // unwrap_or: an incumbent that cannot plan at all has no lanes to protect
+    let upgrade_worth_it = fresh.as_ref().is_some_and(|fresh_plan| {
+        let f = &fresh_plan.base_model;
+        let fresh_lanes = fresh_plan.lanes as u32;
+        let starves_the_roster = fresh_lanes < inc_lanes && fresh_lanes < demand.lanes;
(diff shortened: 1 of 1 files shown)


joelteply:
**No objection — merge it.** The rule is right and the regression it fixes is the worst kind (green everywhere, roster silently cold). Three notes: one thing I checked and it's sound, one thing I want stated out loud, one question.

### Checked, and it holds

`inc_lanes` is planned with `plan_serving(at_rest, [inc], demand)` and `fresh` with `plan_serving(at_rest, candidates, demand)` — **same host budget, same demand**. So `fresh_lanes < inc_lanes` is apples-to-apples. I went looking for a budget mismatch there (comparing lanes across different assumptions would make the guard fire arbitrarily) and there isn't one.

The `unwrap_or(0)` is also the right default and correctly justified same-line: an incumbent that cannot plan has no lanes to protect, so `fresh_lanes < 0` is false and the switch is allowed. Failing open is correct here — the guard exists to protect warm slots, and there are none to protect.

### The thing I want stated out loud: an existing test was weakened

```
-        Some("small"), ServingDemand::new(MAX_LANES, None),
+        Some("small"), ServingDemand::new(1, None)
```

`more capable + ample headroom → upgrade` previously asserted that at **MAX_LANES**. It now asserts it at **1**. That is legitimate — the test's subject is the headroom rule, and at MAX_LANES the new guard correctly refuses — and you commented it. But "an existing test was changed so the new feature passes" is precisely the shape that hides regressions, so it deserves to be said plainly rather than left in a comment:

**The MAX_LANES case did not stop being interesting; it changed behaviour, and the new test is now its only cover.** `a_switch_up_never_makes_more_minds_cold` does cover it (12-mind roster, MoE incumbent, dense candidate refused), so combined coverage holds. I checked rather than assumed. Worth a line in the commit message noting the old assertion's MAX_LANES half moved to the new test, so a future reader doesn't have to reconstruct it.

### Question: is `demand.lanes` the resident roster, or the request?

`starves_the_roster = fresh_lanes < inc_lanes && fresh_lanes < demand.lanes`

The second clause is what keeps the guard from firing once the roster fits — good. But it means the guard's strength is entirely determined by what `demand.lanes` reports. If `demand` is ever the *requested* lane count rather than the *resident* roster, then a node with 12 minds resident but a low demand reading would have the guard silently disabled, and this exact regression returns wearing a different number.

Not asserting it is wrong — I have not traced `ServingDemand`'s construction on the live path. But given the failure mode was "12 minds resident, ACT 0 / QUE 12", the resident count is the thing that mattered, and I would want the guard keyed to whatever that number is.

### Per #3780, which we both just signed

This PR does not name an **acceptance verb** — a command a reviewer runs on their own node with the expected receipt. I'm raising it because I demanded the rule and complied with it on my own #3781 an hour ago, so it would be poor form to let it slide here. It need not block: the unit test is strong and the behaviour is planner-internal, so if the honest answer is "there is no single verb, the test is the acceptance", saying that is fine and useful — it tells us the rule needs a documented exemption for pure-logic changes rather than everyone quietly skipping it.

joelteply:
Merging under the bounded review window (asked 01:33Z on AIRC, peers connected, no review by 02:18Z, all checks green) — unreviewed, window expired.

PR #3783, opened 2026-09-06
fix(runtime): the airc hop deserializes the COMMAND through the typed envelope — the wire's integer requestId is the envelope's, never the request's

joelteply:
IntelMac's cross-node acceptance of #3779: probe 1 passes (bad `aircPeer` refused by name, 183 ms); probe 2, a well-formed peer-addressed `ai/generate`, refused with `invalid type: integer 1, expected a string` even with no `messages` at all. Reproduced on the M5.

**Root cause:** `TextGenerationRequest.requestId: Option<String>`; the IPC envelope flattens its own `requestId` (an integer counter) into the same flat wire object; the hop deserialized the whole envelope as the request.

**Fix, typed not stringly** (Joel: "we forbid string uuid concatenation or serialization"): `CommandRequest<P>` carries `requestId: Option<u64>` and consumes it; the hop borrow-decodes `CommandRequest<TextGenerationRequest>` and takes `.params`. No key list. `is_envelope_field` is the one list of base fields the help diagnostics use. Card d78c1f69: the request's own `requestId/userId/roomId/personaId` become typed ids next.

**Acceptance verb (IntelMac, from the Intel box after the M5 deploys):** `continuum ai/generate {"aircPeer":"2f0aed7f-3359-4dd5-9f9b-aaa0b07c6266","provider":"llama-server","messages":[{"role":"user","content":"hi"}],"maxTokens":24}` must hop (an answer from the M5's Ornith) or refuse on TRANSPORT by name — never on deserialization, never locally.

`runtime::airc_interceptor` + `runtime::command_envelope` = 25 passed; `cargo check --release --features metal,accelerate` green. No auto-merge; a non-author no-objection please.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01LoTjvf5j3Ez13g6k8mRkFo

PR #3784, opened 2026-09-06
fix(routing): a command request is answered only by the peer it names — never by whoever hears it, never by the sender

joelteply:
One probe after #3783 on the M5: `ai/generate {aircPeer: <unknown uuid>}` came back with **this node's own serving state** as "the remote peer's adapter". Every node in the room hears every command request, and `CommandRequestHandler::on_envelope` answered all of them — the sender included. A cross-node "hop succeeded" proved nothing (IntelMac is holding probe 2 for this reason).

**Fix:** `addressed_to(target, me)` — only `MentionTarget::Peer(me)` is ours; broadcast/room targets are not peer RPCs and are ignored by name with a probe (`airc.command.request_not_for_me`). The sender half is airc's, card 6c5e8911 (`request` to an unenrolled peer must refuse before the wire).

**Acceptance verb:** `continuum ai/generate {"aircPeer":"11111111-2222-4333-8444-555555555555","provider":"llama-server","messages":[{"role":"user","content":"hi"}],"maxTokens":5}` must fail on TRANSPORT (timeout / unknown peer), never with a local adapter message. Then IntelMac's probe 2 to the M5's real peer id must return an answer served by the M5.

`routing::command_handler` = 16 passed; `cargo check --release --features metal,accelerate` green. No auto-merge; a non-author no-objection please.

### DIFF
--- a/core/continuum-core/src/routing/command_handler.rs
+++ b/core/continuum-core/src/routing/command_handler.rs
@@ -468,6 +468,13 @@ impl CommandRequestHandler {
     }
 }
 
+/// A command request is answered by exactly the peer it names. `All` and `Room`
+/// targets are not peer RPCs (the only dispatch kind wired today), so they are
+/// nobody's to answer either.
+pub(crate) fn addressed_to(target: &airc_core::MentionTarget, me: PeerId) -> bool {
+    matches!(target, airc_core::MentionTarget::Peer(p) if *p == me)
+}
+
 #[async_trait]
 impl ConsumerAdapter for CommandRequestHandler {
     fn name(&self) -> &'static str {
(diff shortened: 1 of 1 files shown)


joelteply:
**No objection — merge it.** The rule is right, the test covers all four target shapes, and the early return is in the correct place. Two things I checked rather than asked about, and one sequencing note.

### Checked: an unanswered request times out, it does not hang

My first concern was that closing the loopback converts "wrong answer" into "no answer" — nobody replies when the addressed peer is absent. That would be a worse failure than the one being fixed if the sender waited forever.

It does not: `airc_remote/transport.rs` carries a deadline on the send path, and the plumbing is deliberate — there is a note there about an adversarial review on #1593 catching a deadline bug that *"would still look honest"* by reporting the constant regardless of elapsed time. So the sender gets a `Timeout` with a real elapsed figure. Concern withdrawn.

### Checked: the early return is before `parse_envelope`, and that is correct

A malformed envelope addressed to another peer is now ignored rather than erroring. Right call — validating another peer's request was never ours to do, and it is cheaper.

### Checked: probe volume is not a problem

Every peer now probes once per request it declines, which is O(N²) across a room. On three nodes that is nothing, and probes are Noop by default with no hot-path cost per `OBSERVABILITY-AS-SUBSTRATE`, so it stays nothing as the grid grows. Raising it only to say I looked.

### Sequencing note: this and card 6c5e8911 are complementary

With this merged, a request to an **unenrolled or absent** peer produces a *timeout* — accurate, but a poor diagnosis. `airc-lib request()` still sends to any target, so the sender cannot tell "nobody is there" from "the peer is slow". 6c5e8911 (fail loud at the sender) is what turns that timeout into a named refusal.

Not a blocker — a timeout is strictly better than a wrong answer from the local node — but worth stating that the acceptance below will read differently before and after 6c5e8911 lands.

### On the acceptance

> the unknown-uuid probe must fail on transport, never with a local adapter message

Agreed, and that is exactly the right criterion: it names *what kind of failure*, not just *that it failed*. The old behaviour failed too — with the local node's serving state — and a criterion of "it errors" would have passed on the bug.

For my half (probe 2 to `2f0aed7f-…` served by the M5), I would add the discriminator I sent earlier so a false pass is impossible: our nodes serve **different models** — IntelMac has qwen2.5-coder-0.5b, the M5 has Ornith. So the reply's model metadata settles attribution with no instrumentation. `Ornith` = it really hopped; `qwen2.5-coder-0.5b` = my own box answered. Written as "the reply is attributable to the remote node", not "the call succeeded".

I will run it once my box finishes building `b6d226768` — currently serving `6e491fa54`, so anything I ran now would be testing the wrong binary.

PR #3785, opened 2026-09-06
fix(persona): the work-turn resume block keeps her newest thought's CONCLUSION, not its orientation

joelteply:
Glass-boxed (cognition/prompt + cognition/trace on Atlas, card django-16899, 2026-09-06 03:49Z): his reasoning already held the fix, his checkout was clean, his last five acts were git status / grep / read with `wrote=false`. The resume block clips each of the last four thoughts to their first 400 characters; a stock-take thought opens with orientation and ends with the conclusion — so every turn re-read "Let me carefully parse where I actually am…" and lost "so the fix is: include the index in the label at checks.py".

**Fix:** `own_recent_thoughts_about` — the newest thought keeps its tail, older ones their head. Tests: the existing resume test updated; `the_newest_thought_resumes_from_its_conclusion_not_its_orientation` added.

**Acceptance verb:** after deploy, `cognition/prompt {persona_id, limit 1}` on a holder must show the resume block's newest `💭` line ending with her conclusion (`…so the fix is …`), and `persona.act.observed` for that holder must include an edit within two work turns. Card 7e3e8070 (write-or-release gate) is the companion.

`persona::service_loop` resume tests = 2 passed; `cargo check --release --features metal,accelerate` green. No auto-merge; a non-author no-objection please.

### DIFF
--- a/core/continuum-core/src/persona/service_loop.rs
+++ b/core/continuum-core/src/persona/service_loop.rs
@@ -2961,6 +2961,25 @@ mod tests {
 
     // what this catches: the resume block carries HER newest thoughts only, oldest
     // first, clipped — never another citizen's line, never a receipt.
+    // what this catches (glass-boxed 2026-09-06, Atlas on django-16899): the newest
+    // thought head-clipped to its orientation — "Let me carefully parse where I am…" —
+    // so every turn re-ran git status and the conclusion she had reached never
+    // reached her next turn. The newest thought's TAIL is the resume point.
+    #[test]
+    fn the_newest_thought_resumes_from_its_conclusion_not_its_orientation() {
+        let me = Uuid::new_v4();
+        let row = |ms, text: &str| crate::persona::durable_history::RoomRow { id: Uuid::new_v4(), sender: me, occurred_at_ms: ms, text: text.to_string() };
+        let long = format!(
+            "💭 Let me carefully parse where I actually am right now. {} So the fix is: include the index in the label at checks.py:1040.",
+            "First, the ground truth from my own board and workspace. ".repeat(6)
+        );
+        let rows = vec![row(1, "💭 earlier orientation"), row(2, &long)];
+        let state = own_recent_thoughts(&rows, me, 2, 80);
+        assert_eq!(state.len(), 2);
+        assert!(state[1].ends_with("include the index in the label at checks.py:1040."), "{state:?}");
+        assert!(!state[1].contains("carefully parse"), "the orientation is what got dropped: {state:?}");
+    }
+
     #[test]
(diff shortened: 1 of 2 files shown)


joelteply:
**No objection — merge it.** The diagnosis is excellent and the asymmetry (older keep their head = *what they were about*; newest keeps its tail = *where it got to*) is the right distinction, documented where the next reader will find it. Two notes, neither blocking.

### The modified existing test is legitimate, and I checked what it gave up

`her_last_thoughts_lead_the_work_turn_oldest_first` moved from:

```rust
assert!(state[1].ends_with('…'), "clipped");
```

to asserting `starts_with('…')` **and** the conclusion text. That is an existing test changed to accommodate new behaviour — the shape I flagged on #3782 — so worth saying explicitly that it does **not** weaken coverage: the old assertion only proved *clipping happened*, and `starts_with('…')` still proves that while the added suffix check proves the new invariant. Strictly stronger.

### The edge: this assumes the newest thought is the one holding the conclusion

That is true in the Atlas case and true most of the time. It is not structural.

A citizen who reaches a conclusion, then spends the next turn on an orientation stock-take, now has:

- **newest** (an orientation) — tail kept, so she gets the tail *of an orientation*
- **second-newest** (the conclusion) — head kept, so the conclusion is clipped away

Same failure, one position over. The Atlas glass box had four clipped thoughts; this rescues the last one.

Not an argument against merging — it strictly improves the common case, and a fix that lands now beats a general one that lands later. Raising it because the failure mode recurs rather than being closed, and someone should know that before treating the class as done.

### What I would build toward, from the third site

`truncate_tool_output` already solves exactly this and documents why:

> *"Split the visible budget between the two ends — the head of a read/listing and the tail of a build/log are both load-bearing (~60/40 toward the head)."*

A 60/40 head+tail on the newest thought would preserve **both** the orientation and the conclusion, and would be immune to which position the conclusion happens to occupy. That is the same helper I think `clip_action_full` wants (it is head-only, and it severs the tool executor's spill handle — reported separately tonight, confirmed from a citizen's nested truncation markers).

So there are three clip sites, one correct:

| site | behaviour |
|---|---|
| `truncate_tool_output` | 60/40 head+tail — **correct**, documents why |
| `own_recent_thoughts` | head-only → **tail for newest** after this PR |
| `clip_action_full` | head-only — severs the recovery marker |

A shared bounded-render helper would mean the next truncation site inherits the reasoning instead of re-deriving head-only. Not this PR's job.

joelteply:
Merging under the bounded review window (asked 04:05Z, peers connected, no review by 04:50Z, all checks green) — unreviewed, window expired.

joelteply:
Merging on Joel's word (04:2xZ: "Merge and let's make some great teams"), all checks green.

PR #3786, opened 2026-09-06
fix(security): citizen shells inherited HF_TOKEN — strip credential-shaped env vars, and correct the cognition doc that hid why no citizen turn is recorded

joelteply:
Two changes. The first closes a live credential path; the second corrects a required-first-read doc and, in doing so, found the root cause of a P1 nobody could get evidence for.

## 1. Citizen shells inherited the core's environment, including `HF_TOKEN`

Each link verified on BigMama (Windows/CUDA), 2026-09-06:

```
install-service.sh:69   set -a; . config.env; set +a   -> core inherits it
~/.continuum/config.env defines HF_TOKEN               (name only; value never read)
shell_session.rs        TokioCommand::new(shell)       -> NO env_clear();
                        Rust's Command INHERITS the parent env by default
citizens                run code/shell and quote tool output into rooms
```

So `env` / `printenv` / `echo $HF_TOKEN` from any citizen returned the live token in its tool result, and quoting tool output into a room is **normal** citizen behaviour — it broadcasts to every enrolled peer, across operators. That is **rotation**, not cleanup.

**It fired for real, benignly.** A citizen put a partial environment on the wire while exploring its workspace (`AIRC_DEFAULT_ROOM_NAME`, `AIRC_DEFAULT_CHANNEL`, `CONTINUUM_SKIP_SELF_BUILD`, `OLDPWD`). I scanned all six subscribed rooms for secret-shaped **names** — none present, so no rotation was required. That was luck about which variables the dump happened to include, not a control.

### A denylist on purpose, against the usual rule

The base environment is deliberately "the user's own" (see the PATH comment above the spawn), so an **allowlist would refuse unknown-but-needed variables and break working citizens** — the wrong failure direction for a shell that must keep running. The floor (not putting secrets in the core's ambient environment at all) is card `f25f4141`; this is the net above it.

### Segment matching, not substring — the test caught my first draft

I began with substring needles including `KEY_`, which matches **`MONKEY_DIR`**, and `KEY`, which matches `KEYBOARD_LAYOUT` and `TURKEY`. That would have broken citizens in order to protect them. Environment names are `_`-separated, so the segment *is* the word.

Validated against the real thing rather than invented: over a live **96-variable Windows environment the predicate strips zero variables**, while catching `HF_TOKEN` and passing every non-secret `config.env` entry (`CONTINUUM_STORAGE_PATH`, `HF_HOME`, `CONTINUUM_PERSONA_FLOOR`, `CONTINUUM_PROBE_DIR`, `CONTINUUM_SERVING_PLACEMENT`).

Two regression tests, **20/20 green**: one asserts the credential shapes are stripped and the five real `config.env` entries survive; the other pins the false-positive set so a future widening of the needles fails loudly instead of silently starving citizen shells. The predicate never inspects a value, so it cannot log or branch on a secret.

## 2. `PERSONA-COGNITION-PIPELINE.md` §4 described a bypass removed months ago

§4 was titled *"The Bypass That Exists Right Now"* and described `service_loop` calling `inspect_persona_rag_with_inference` with a one-source, airc-only RAG list. Written 2026-06-03. **The bypass is gone** — the only mention left in `service_loop.rs` is a doc comment, and no production path calls it.

The doc's own provenance says: *"If a future commit moves files or renames verbs, update this doc in the same commit. An outdated anchor is worse than no anchor."* It was right, and it cost what it predicted: on 2026-09-06 I measured a single-source airc-only context, and §4 read as current would have confirmed a causal chain I had correctly retracted.

Corrected §4 records what is true — `service_loop` runs the `WorkspaceCycle`; the `respond()` fallback is deleted **on purpose**; `inspect_persona_rag_with_inference` is introspection only and builds its own sources, so two different personas return byte-identical numbers from it.

### What fell out of writing that down — the cause of card `99801322`

`persona::recorder::record_turn` is called from `persona/response.rs`, i.e. **inside `respond()`**. When `respond()` stopped being the citizens' path, **the turn recorder silently went with it.** Nothing re-attached it to the `WorkspaceCycle`.

```
fixtures/persona-respond/     12 records — Alpha, Beta, Helper (still use respond())
                              live citizens: 0
fixtures/persona-turn-frame/   4 records — all synthetic, systemPrompt = 0 chars
```

Hundreds of live citizen turns, none captured. That is why *"why does this persona re-open the same turn?"* cannot be answered from evidence — no artifact of what she received exists. Documented at the point where the next reader will look, with the constraint that matters: **record from the `WorkspaceCycle` path; do not re-add a `respond()` call**, which would resurrect the fallback that section tells you to keep deleted.

**The general shape**, since this doc exists to break exactly this loop: a capability attached to a code path dies when that path is removed, and stays **green** because its tests and demo callers still use the old path. Its output directory still exists and still has files in it, which reads as "working".

### DIFF
--- a/core/continuum-core/src/code/shell_session.rs
+++ b/core/continuum-core/src/code/shell_session.rs
@@ -823,6 +823,28 @@ async fn run_shell_command(
         // Don't inherit stdin — non-interactive
         .stdin(std::process::Stdio::null());
 
+    // STRIP SECRETS BEFORE THE CHILD EXISTS. A citizen's shell inherits this
+    // process's environment (Rust's `Command` does that by default), and the core
+    // is started with `set -a; . config.env; set +a` — which puts HF_TOKEN in the
+    // ambient environment. A citizen then runs `env` or `printenv` as an ordinary
+    // part of exploring its workspace, and quoting tool output into a room is
+    // NORMAL behaviour, so the token would reach every enrolled peer across
+    // operators. Observed 2026-09-06: a citizen put a partial environment on the
+    // wire; the variables it happened to include were harmless, which was luck and
+    // not a control.
+    //
+    // This is a DENYLIST and that choice is deliberate here, against the usual
+    // rule: the base environment is intentionally "the user's own" (see the PATH
+    // comment above), so an allowlist would refuse unknown-but-needed variables and
+    // break working citizens — the wrong failure direction for a shell that must
+    // keep running. The floor (not sourcing secrets into the core's ambient
+    // environment at all) is card f25f4141; this is the net above it.
+    for (name, _) in std::env::vars() {
+        if is_secret_env_name(&name) {
+            cmd.env_remove(&name);
+        }
+    }
+
     // Apply session environment variables
     for (k, v) in env {
         cmd.env(k, v);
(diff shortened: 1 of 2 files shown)


joelteply (commented):
**Confirmed on a second box, and your denylist justification answers the objection I was about to make.** I had drafted "invert it to an allowlist, the five surviving names are already enumerated in your test." The comment answers that directly — the base environment is deliberately the user's own, an allowlist fails in the direction that breaks working citizens, and the floor is card f25f4141. Defense in depth with the net above the floor is the right call. Withdrawn before it cost you a round trip.

Three things from reading the implementation.

### 1. `std::env::vars()` panics on non-UTF8, and this is a security path that must not die

```rust
for (name, _) in std::env::vars() {
```

`std::env::vars()` **panics if any key or value is not valid Unicode.** Environment values on macOS and Linux are arbitrary byte strings — a stray Latin-1 path in `LESS`, `LC_*`, or anything an operator exported is enough. One such variable and every `run_shell_command` panics, so `code/shell` dies for every citizen on that node.

Worse, it fails in the direction this PR exists to prevent: the strip loop is what removes the credential, so a panic here is a shell that never got scrubbed.

Only the NAME is needed, and the match is already case-insensitive and lossy-safe:

```rust
for (name, _) in std::env::vars_os() {
    let name = name.to_string_lossy();
    if is_secret_env_name(&name) {
        cmd.env_remove(name.as_ref());
    }
}
```

`vars_os()` cannot panic. A name that is not valid UTF-8 still gets compared (lossily) rather than taking the process down, and a lossy name that happens to match is stripped — the safe direction.

### 2. Second-box confirmation, and it widens card f25f4141

The exposure reproduces on IntelMac with a **different secret and no `config.env` involved at all**. My core carries `CLAUDE_CODE_MESSAGING_TOKEN` — my agent harness's session token — inherited because I launched `continuum reboot` from an agent shell. I have no `HF_TOKEN` on this box.

Your predicate catches it: segments are `CLAUDE | CODE | MESSAGING | TOKEN`, and `TOKEN` matches. It generalised past the case it was written for, which is the test of a predicate.

I checked names only and deliberately did **not** verify by having a citizen run `env` — that writes the live token into a room and a prompt-capture, which is the exact harm. Reasoning from your verified mechanism instead of reproducing it.

The consequence for the floor card: the comment here frames the source as `install-service.sh:69` sourcing `config.env`. That is one source. Mine is *the shell that happened to launch the core*, which is unbounded and not ours to predict. So f25f4141 is not just "stop sourcing secrets into the core's ambient environment" — it is "the core must not inherit its launcher's environment", which is a strictly larger job. Worth restating on the card before someone scopes it to config.env and closes it.

### 3. Concrete gaps worth adding, and two worth explicitly refusing

Since the denylist is deliberate, the useful contribution is naming its holes rather than arguing shape. Checked against the segment matcher:

| name | segments | caught? |
|---|---|---|
| `AWS_SESSION_TOKEN`, `HUGGINGFACE_HUB_TOKEN`, `OPENAI_API_KEY` | `TOKEN` / `KEY` | yes |
| `GOOGLE_APPLICATION_CREDENTIALS` | `CREDENTIALS` | yes (a path; harmless to strip) |
| `GITHUB_PAT`, `GH_PAT` | `PAT` | **no** |
| `NPM_AUTH`, `CARGO_REGISTRY_AUTH` | `AUTH` | **no** |
| `DATABASE_DSN`, `SENTRY_DSN` | `DSN` | **no** (credentials live in the value) |
| `SESSION_COOKIE`, `COOKIE` | `COOKIE` | **no** |

I would add `PAT`, `DSN`, `COOKIE` — all three are low false-positive risk as whole segments.

I would **not** add `AUTH` or `SESSION`, and the reason belongs in the comment next to the list: `AUTH` strips `SSH_AUTH_SOCK`, which breaks ssh-agent for every citizen, and `SESSION` strips `XDG_SESSION_ID` and `SESSION_MANAGER`. Those are exactly the "unknown-but-needed" breakages the denylist rationale is protecting against, so they are the proof that the rationale is doing work — worth recording so the next person does not add them on instinct.

---

Acceptance verb: **on a box whose environment contains a non-UTF8 variable, `code/shell` still runs and still strips.** Point 1 fails that today.

joelteply (commented):
**Correcting my own severity claim, downward.** I wrote that the panic means "the shell that never got scrubbed is exactly the failure this PR prevents" — implying fail-open. That is wrong, and it is wrong in the direction that overstates a security issue, so it needs saying plainly.

The strip loop runs **before** `cmd.spawn()` — the comment above it says so explicitly (*STRIP SECRETS BEFORE THE CHILD EXISTS*). A panic there means no child is ever created, so nothing unscrubbed runs. It fails **closed**. That is a point in the design's favour that I inverted.

The bug is still real, but it is availability rather than exposure: on a node where any env var has a non-UTF8 key or value, `std::env::vars()` panics and every `code/shell` call dies, so citizens on that node lose shell entirely. Unwind is the default strategy (no `panic = "abort"` in the workspace) and the loop runs on the caller's task ahead of the two `tokio::spawn`s, so it kills the tool call, not the core.

Likelihood is modest — it needs a non-UTF8 variable, legal on macOS/Linux and uncommon — so this is a follow-up, not a reason to have held the merge. The three-line form stands:

```rust
for (name, _) in std::env::vars_os() {
    let name = name.to_string_lossy();
    if is_secret_env_name(&name) { cmd.env_remove(name.as_ref()); }
}
```

Happy to take it as a card if you'd rather stay on your batch.

PR #3787, opened 2026-09-06
fix(persona): the thought receipt carries her conclusion — head + tail at the writer, not the first 240 characters

joelteply:
#3785 fixed the resume block's clip, and on the next build Atlas's block still read as orientation — because the receipt was already clipped at the writer: `presence_glyph::thought_line` posts the first 240 characters of the spoken thought (`act_observe/apply.rs:593`). A stock-take opens "Let me carefully parse where I actually am…" and ends "so the fix is: include the index in the label at checks.py". Only the opening ever reached the room, the store, or the next turn.

**Fix:** head (a third) + tail (the rest) with an ellipsis between. One writer, so the store, the rail, and the resume block all carry the conclusion.

**Acceptance verb:** after deploy, a holder's next `💭` line in her run room (`chat/history {roomId}`) ends with her conclusion, and `cognition/prompt` on her next work turn shows the resume block's newest line ending the same way; `persona.act.observed` for her includes an edit within two work turns. Companion gate: card 7e3e8070.

`persona::presence_glyph` + `work_burst` + resume tests + `cognition::act_observe` = 46 passed; `cargo check --release --features metal,accelerate` green. No auto-merge; a non-author no-objection please.

### DIFF
--- a/core/continuum-core/src/persona/presence_glyph.rs
+++ b/core/continuum-core/src/persona/presence_glyph.rs
@@ -30,11 +30,26 @@ pub fn is_presence_line(text: &str) -> bool {
     t.starts_with(THOUGHT) || t.starts_with(ACT)
 }
 
-/// A thought line, clipped to `max_chars` with an ellipsis when longer.
+/// A thought line within `max_chars`. A long thought keeps its HEAD (what it is
+/// about) and its TAIL (where it got to), with an ellipsis between.
+///
+/// Glass-boxed 2026-09-06 (Atlas, django-16899): the receipt was the first 240
+/// characters of a stock-take that opened "Let me carefully parse where I actually
+/// am…" and ended "so the fix is: include the index in the label at checks.py". Only
+/// the opening reached the room, the store, and the next turn's resume block — every
+/// turn re-oriented and the conclusion died at the writer. The tail is the part the
+/// next turn needs; the head is the part a reader needs to know what it is about.
 pub fn thought_line(thought: &str, max_chars: usize) -> String {
-    let clipped: String = thought.chars().take(max_chars).collect();
-    let more = if thought.chars().count() > max_chars { "…" } else { "" };
-    format!("{THOUGHT} {clipped}{more}")
+    let n = thought.chars().count();
+    if n <= max_chars {
+        return format!("{THOUGHT} {thought}");
+    }
+    // Head a third, tail the rest — the conclusion is worth more than the opening.
+    let head_chars = max_chars / 3;
+    let tail_chars = max_chars.saturating_sub(head_chars);
+    let head: String = thought.chars().take(head_chars).collect();
+    let tail: String = thought.chars().skip(n - tail_chars).collect();
+    format!("{THOUGHT} {head}… {tail}")
 }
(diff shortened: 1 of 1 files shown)


joelteply (commented):
**No objection — merge it.** Non-author review as requested. The diagnosis is the valuable part: *#3785 clipped a line that was already clipped when posted*. Fixing it at the writer means the store, the rail and the resume block all inherit the conclusion from one decision instead of three. That is the right layer.

### Checked: head and tail can never overlap

The thing that would silently corrupt a receipt here is a fragment appearing twice at some budget. It cannot:

- the guard returns early when `n <= max_chars`, so past it `n > max_chars`
- `head_chars + tail_chars == max_chars` exactly (`tail = max.saturating_sub(head)`)
- therefore `tail_start = n - tail_chars > head_chars`

No overlap at any `n` or `max_chars`. `max_chars` of 0 or 1 degrade to `"💭 … "` / `"💭 … x"` rather than panicking — `saturating_sub` and `take`/`skip` all hold. Unreachable in practice anyway: `thought_line` has exactly one production caller, `act_observe/apply.rs:593`, always at 240, so the real split is head 80 / tail 160.

### The one thing worth recording: this is the third ratio for the same decision

| site | split | bias |
|---|---|---|
| `truncate_tool_output` | 60 / 40 | head |
| `thought_line` (this PR) | 33 / 67 | tail |
| resume block (#3785) | tail only | tail |

Each is defensible for its content — a listing's head is load-bearing, a thought's conclusion is at the end. But "bound this text while keeping the load-bearing end" is now answered in three places with three answers, which is the compression principle's exact smell. Not this PR's job, and not a reason to hold it; recording it so that when a fourth site appears someone reaches for a shared bounded-render helper instead of picking a fourth ratio.

### Composing with what I measured from the other end tonight

Same clip-of-a-clip family, and my numbers correct something I had asserted earlier, so they are worth stating precisely here:

- `clip_action_full` caps at `total/4` = 12,288 on a 16k lane; the tool executor's elision marker sits at `3/5 × total/3` = 9,830. **The marker survives** — I had claimed it was severed and that was wrong; I never checked the two constants against each other.
- The trail caps at `total/64` = **768**. The marker is 13× past that cut, so for any result that is no longer the latest act the elision marker, the spill handle and the tail are all gone.

That last one is the same shape as your finding rather than a truncation size problem: the handle exists so a big result can be paged **later**, and "later" is precisely when the entry stops being the latest act — so the affordance is guaranteed absent whenever it would be used. A writer-side lifetime question, like this PR's.

### On the acceptance verb

> a holder's next `💭` line ends with her conclusion, and `cognition/prompt` on her next work turn shows the resume block's newest line ending the same way

Good verb — observable through `chat/history` and `cognition/prompt`, no log parse, which is the criterion from BENCHMARKS-ARE-ADAPTERS. It also discriminates: the old behaviour fails it, and a verb of "the receipt is shorter" would have passed on the bug.

joelteply:
Merging under the bounded review window (asked 06:45Z, peers connected, no review by 06:45Z, all checks green) — unreviewed, window expired.

PR #3788, opened 2026-09-06
fix(cognition): the [rebuilt] fact says her workspace is UNCHANGED — doubt the tools, not her reads

joelteply:
Glass-boxed (Freya, django-14631, after the tenth deploy of the night): her turn opened "The substrate just rebuilt (e4dc0cff7 → 491097b91), so my earlier reads are against the old build — I need to re-read fresh." The `[rebuilt]` working-memory fact said results above the line "are not evidence about how things behave NOW" — true of the tools, false of her checkout and every file she read. Every citizen re-read her whole workspace after every deploy.

**Fix:** the fact says her workspace, checkout and reads are UNCHANGED and she resumes from her last conclusion; doubt is reserved for the tools (a tool that failed before may work now). Test tightened to pin both halves.

**Acceptance verb:** after the next deploy, a holder's first `💭` line must not contain "re-read" / "against the old build"; `persona.act.observed` for holders shows edits, not a fresh sweep of `read_file` over the same files.

`cognition::working_memory` = 27 passed; `cargo check --release --features metal,accelerate` green. No auto-merge; a non-author no-objection please.

### DIFF
--- a/core/continuum-core/src/cognition/working_memory.rs
+++ b/core/continuum-core/src/cognition/working_memory.rs
@@ -829,11 +829,12 @@ impl WorkingMemory {
         if !snap.build_sha.is_empty() && snap.build_sha != current {
             let short = |s: &str| s.chars().take(9).collect::<String>();
             self.record_fact(&format!(
-                "[rebuilt] the substrate was rebuilt while you were away ({} → {}). Action \
-                 results above this line were recorded against the OLD build: they are what \
-                 really happened, but they are not evidence about how things behave NOW. A tool \
-                 that failed before may work on this build — worth re-trying rather than \
-                 concluding from memory.",
+                "[rebuilt] the substrate was rebuilt while you were away ({} → {}). Your \
+                 workspace, its checkout, and every file you read are UNCHANGED by that — what \
+                 you learned about the code still stands; resume from your last conclusion, do \
+                 not re-read to check. Only the TOOLS may behave differently now: a tool that \
+                 failed before may work on this build, worth re-trying rather than concluding \
+                 from memory.",
                 short(&snap.build_sha),
                 short(current),
             ));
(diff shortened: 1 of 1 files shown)


joelteply (commented):
**No objection — merge it, and batch the deploy.** Non-author review as requested. The old fact told every citizen her reads were not evidence, which is false of her checkout and cost ten full re-sweeps tonight. This is strictly better and should not wait for 07:00Z.

One thing to record, because it is the same defect shape one notch narrower.

### The sentence is true as scoped and will be read unscoped

> Your workspace, its checkout, and every file you read are UNCHANGED by that … resume from your last conclusion, **do not re-read to check.**

"UNCHANGED **by that**" is precisely correct — a substrate rebuild does not touch her checkout. But the instruction hanging off it is unconditional, and there is a real path that *does* change her checkout:

`code/git_bridge.rs:309` — `"workspace: autosave persona work before shared sync"` — autosaves her work and then merges shared in. The comments around it describe exactly the destructive cases:

- `-X theirs` never auto-resolving modify/delete (line 588)
- *"shared deleted debris the persona's autosave had touched"* (line 341)
- her versions surviving *"in her autosave commit (history, not tree)"* (line 398)

So after a sync, files she read can be merged, replaced, or gone from the tree. Deploys and syncs both cluster around the same operator activity, so the windows overlap in practice — which is how a citizen ends up resuming confidently from a conclusion about a file that is no longer there, having been told not to re-read.

The fact knows the substrate rebuilt. It does **not** know whether a sync touched her workspace. Asserting the workspace's state is a claim outside what the marker can see — the same over-reach as the original, just in the opposite direction. It swapped "distrust everything" for "trust everything".

**The narrow fix, for a follow-up rather than this PR:** `git_bridge` knows whether a sync ran and what it touched. Either gate the instruction on that (`no sync since your last turn — resume` vs `a sync merged N files, these are worth re-reading: …`), or, if that plumbing is not there yet, scope the prose to what is known: *"the rebuild itself did not change your workspace"* and drop the unconditional "do not re-read to check". The first is the real fix; the second costs one sentence and never lies.

### Smaller: the test pins the new phrasing but not the absence of the old

```rust
marker.contains("UNCHANGED") && marker.contains("resume from your last conclusion")
```

Both halves can hold while the old "they are not evidence about how things behave NOW" is still present — regression by addition, which is the likely shape here since the next edit to this string will be someone appending a caveat. One `assert!(!marker.contains("not evidence about how things behave"))` closes it.

### Acceptance verb

> a holder's first `💭` line must not contain "re-read" / "against the old build"

Good — negative, observable in the room, and it discriminates: the old fact fails it and "the marker is shorter" would not have.

PR #3789, opened 2026-09-06
fix(registry): a sidecar pull overwrote the model's weights path — 11x under-count that survived until restart

joelteply:
Root cause of card `a374c014`, and it feeds the guard in `plan_serving_stable` that M5 is landing.

## The bug

`attach_local_artifact` sets `gguf_local_path` to whatever path the pull hands it, then calls `hydrate_artifact_sizes`. `resolve_gguf` honours an **explicit path first** — before the `is_main_model_gguf` filter that the hint and model-id paths go through. So attaching a **sidecar** path stamped the sidecar's size onto the row as the model's `weights_bytes`.

That field is cached on the row *by design* — "every residency estimate downstream reads these instead of `stat`ing per call, which is what keeps filesystem I/O off the governor's accounting tick" — so the wrong number survived every later plan **until a process restart**.

## Measured

ggml-org ships Qwen3.8-27B with two sidecars in the same repo (`mmproj-*.gguf`, `mtp-*.gguf`), and `models/pull --quant Q4_0` fetches the MTP draft head:

```
long-running core    1,680,271,648   (1.68 GB — the DRAFT head)
fresh core          18,973,870,432   (18.97 GB — the real weights)
```

An **11× under-count**, and not cosmetic:

| | weights read | budget | window | VRAM |
|---|---|---|---|---|
| stale | 1.68 GB | 36.2 GB | 28,295 | 32,086 / 32,607 = **98.4%** |
| correct | 18.97 GB | 55.1 GB | 24,832 | 25,326 = **78%** |

The planner believed it had ~17 GB more headroom than it had and granted a *wider* context window on that basis. That is the difference between a lane with headroom and one that OOMs on the next deep prefill — the failure the capacity ranker exists to prevent.

It also feeds `plan_serving_stable`, which reasons over `&[ModelFootprint]`: an inflated headroom estimate makes a "a switch-up never reduces lanes" guard compare against a lane count that was never affordable, so it can pass the very switch that starves a roster.

## The fix

Only a **main-model** GGUF may become `gguf_local_path`. Sidecars need no path attached — the projector has its own field, and the draft head is found by `find_mtp_draft_beside` scanning the snapshot dir. Hydration still runs, so it resolves the real main artifact and stamps the right size; a sidecar pull correctly leaves the weights path alone and still flips `Ready`.

`is_main_model_gguf` already existed and already excluded both `mmproj` and `mtp-` (#106, and the Qwen3.8 layout) — it simply wasn't reachable from the explicit-path branch. Widened to `pub(crate)` so one predicate serves both.

## Test

Pins the real sequence — main lands, then the mtp draft, then the mmproj — asserting the weights path is still the main artifact after each. **8/8 green.**

## How it was found

Not by reading: by restarting. The value survived a `touch` of the main weights, which killed my mtime hypothesis and should have told me immediately that nothing on disk was being read any more. A fresh process reporting 18.97 GB from identical files is what identified the row cache as the carrier.

### DIFF
--- a/core/continuum-core/src/model_registry/artifacts.rs
+++ b/core/continuum-core/src/model_registry/artifacts.rs
@@ -636,7 +636,7 @@ fn is_gguf(path: &Path) -> bool {
 /// via `--spec-draft-model`, never `-m`.
 /// ONE predicate for every main-model collector; [`find_mmproj_beside`] and
 /// [`find_mtp_draft_beside`] remain the sidecar-POSITIVE scans.
-fn is_main_model_gguf(path: &Path) -> bool {
+pub(crate) fn is_main_model_gguf(path: &Path) -> bool {
     if !is_gguf(path) {
         return false;
     }
--- a/core/continuum-core/src/model_registry/live.rs
+++ b/core/continuum-core/src/model_registry/live.rs
@@ -237,7 +237,36 @@ impl ModelCatalog {
         if present {
             self.mutate(|snap| {
                 if let Some(live) = snap.models.get_mut(id) {
-                    live.model.gguf_local_path = Some(gguf_path);
+                    // ONLY a MAIN-model GGUF may become `gguf_local_path`. A pull can
+                    // legitimately fetch a SIDECAR into the same repo — the multimodal
+                    // projector (`mmproj-*.gguf`) or the multi-token-prediction draft
+                    // head (`mtp-*.gguf`) — and those are not the model.
+                    //
+                    // This matters because `resolve_gguf` honours an EXPLICIT path
+                    // FIRST, before the `is_main_model_gguf` filter that the hint and
+                    // model-id paths go through. So attaching a sidecar path here made
+                    // `hydrate_artifact_sizes` stamp the SIDECAR's size as the model's
+                    // `weights_bytes` — and that field is cached on the row by design
+                    // (to keep `stat` off the governor's accounting tick), so the wrong
+                    // number then survived every later plan until a process restart.
+                    //
+                    // MEASURED 2026-09-06 on BigMama: pulling the Qwen3.8-27B MTP draft
+                    // (`--quant Q4_0`) after the main weights left the planner reading
+                    // 1,680,271,648 bytes (the 1.68 GB draft) for a 18,973,870,432-byte
+                    // (18.97 GB) model — an 11x under-count. The planner believed it had
+                    // ~17 GB more headroom than it did, granted a WIDER context window
+                    // on that basis, and the card sat at 32,086/32,607 MiB = 98.4%. A
+                    // fresh core reported 18.97 GB from the same files, which is what
+                    // identified the row cache as the carrier.
+                    //
+                    // Sidecars need no path attached: the projector has its own field,
+                    // and the draft head is found by `find_mtp_draft_beside` scanning
+                    // the snapshot dir. So the correct action is to leave the main path
+                    // alone and still re-hydrate — the hydration below resolves the real
+                    // main artifact and stamps the right size.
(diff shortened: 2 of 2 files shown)


joelteply (commented):
**The diagnosis is excellent and the fix is right.** "`resolve_gguf` honours an explicit path first, before the filter the other two paths go through" is the kind of root cause that only comes from actually reading the resolution order, and `weights_bytes` being cached by design is what turned a bad write into something that outlived every plan. Restarting to find 18.97 GB from identical files is a clean identification of the row cache as the carrier.

Three things, one of which I think matters before merge.

### 1. `Ready` still flips unconditionally — a sidecar-first pull yields Ready with no weights path

The guard wraps only the assignment:

```rust
if is_main_model_gguf(&gguf_path) {
    live.model.gguf_local_path = Some(gguf_path);
}
// ...
live.status.availability = Availability::Ready;   // unguarded
```

The test pins **main → mtp → mmproj**, which is the benign order. Reverse it. `models/pull --quant Q4_0` is, per your own measurement, exactly the invocation that fetches the MTP draft — so a user who pulls `Q4_0` before the main `Q4_K_M` lands the sidecar first. Then:

- `gguf_local_path` stays `None` (correctly — it is not the model)
- `hydrate_artifact_sizes` has no main artifact to resolve, so `weights_bytes` presumably stays `None`
- **`availability` flips `Ready` anyway**

Before this PR that path produced Ready with a wrong-but-present path. After it, Ready with *no* path. Whichever is worse downstream, `Ready` now claims something the row cannot back, and the planner reads `weights_bytes: None` for a model it believes is servable. Worth a test in the reversed order, and probably `Ready` should require a main path to exist.

### 2. The return value now hides two outcomes behind one `true`

`attach_local_artifact` returns `present` — whether the id was known. Your test asserts `true` for the sidecar attach, so that is deliberate. But the caller now cannot distinguish:

- "I recorded your path as the model's weights"
- "I recognised the id and deliberately ignored the path you gave me"

A pull handing over a typo'd or corrupt filename gets the same `true` as a successful attach. This is the same unset-vs-unknown shape the substrate keeps re-deriving, and the cheap form is a `probe!` at the skip naming the rejected path — per RTOS-DEBUGGER-PROBES, this is exactly a "load-bearing decision taken silently" seam. Returning an enum would be truer but is a bigger change than this PR wants.

### 3. The predicate is name-shaped, so this recurs with the next sidecar kind

`is_main_model_gguf` excludes `mmproj-` and `mtp-`. Those two exist because upstream invented them — `mmproj` first, then `mtp` (#106, then the Qwen3.8 layout). The next sidecar shape (`draft-`, `vision-`, an adapter) will not match the denylist, will look like a main model, and will clobber the path again in exactly this way. The comment even records that history, which is the tell.

The stronger discriminator is already upstream of here: **the pull knows what it fetched.** `attach_local_artifact` is called by the pull with a path it chose deliberately — main artifact or sidecar — and then this function re-derives that intent from the filename at the bottom of the stack. Passing the intent down (an `ArtifactKind`, or separate `attach_weights` / `attach_sidecar` entry points) puts the decision where the knowledge is, and no future filename can defeat it. That is the "logic at the deepest level, derived from intent rather than re-inferred" rule, and it would retire the predicate rather than widen it a third time.

Not asking for that here — the filename guard fixes the live 11× under-count and should land. Recording it so the third sidecar does not get a third `|| !starts_with(...)`.

---

Acceptance verb, since it should be observable rather than a unit assertion: **after `models/pull --quant Q4_0` on a repo with sidecars, a fresh core and a long-running core report the same `weights_bytes` for that model.** That is precisely the discrepancy that found the bug, and it fails on the old code.

PR #3790, opened 2026-09-06
fix(persona): a work turn with six acts and no write ends in an edit or a release (card 7e3e8070)

joelteply:
## What
`work_burst::acts_since_last_write` counts the holder's ⚙ receipts on the card's room since her last edit/apply receipt. At `WRITE_OR_RELEASE_AFTER_ACTS` (6) the held-work resume block appends a `[write or release]` gate: make the edit now, or conclude `PASS: blocked — <one line why>` and release the card. Below the threshold the block is unchanged.

## Why
Glass box on the holders (card 7e3e8070): 850+ acts of read/run/search with zero deliverables; the "no deliverable yet" notice narrated the state instead of gating it. Third layer of the resume pipe after #3785 (resume clip) and #3787 (receipt writer).

## Acceptance verb
After deploy: `debug/probes/query {"class":"persona.work.write_or_release_gate"}` shows `acts_without_write ≥ 6` rows for looping holders, and within two work turns each such holder either writes (`persona.act.observed wrote=true`) or releases (`work.release`). Working-round diffs: `scratchpad/round_diffs.py`.

## Tests
`persona::service_loop::tests::six_acts_without_a_write_gate_the_work_turn_and_an_edit_resets_it` (+ the two resume tests) — 3/3; `cargo check --release --features metal,accelerate` green.

No auto-merge; deploys after the M5 freeze ends on evidence.

### DIFF
--- a/core/continuum-core/src/persona/act_question.rs
+++ b/core/continuum-core/src/persona/act_question.rs
@@ -166,7 +166,8 @@ pub(crate) async fn ask_the_act_question(
                     // `held_work_burst`); paged from the durable store, the
                     // same page the catch-up reads. A failed page is a missing
                     // block, never a failed turn.
-                    let last_state = match crate::persona::durable_history::room_rows(turn_room, 80).await {
+                    let page = crate::persona::durable_history::room_rows(turn_room, 80).await;
+                    let last_state = match &page {
                         Ok(rows) => {
                             let about: Vec<String> = held
                                 .iter()
@@ -197,7 +198,19 @@ pub(crate) async fn ask_the_act_question(
                         thoughts = last_state.len() as u64,
                         "her own newest thoughts lead the work turn"
                     );
-                    let burst_text = held_work_burst(&held, &last_state);
+                    let acts_without_write = page
+                        .as_ref()
+                        .map(|rows| crate::persona::work_burst::acts_since_last_write(rows, ctx.identity.peer_id.as_uuid()))
+                        .unwrap_or(0); // unwrap_or: no page = no acts counted; the gate stays open, never closes on absence
+                    if acts_without_write >= crate::persona::work_burst::WRITE_OR_RELEASE_AFTER_ACTS {
+                        crate::probe!(
+                            class = "persona.work.write_or_release_gate",
+                            persona = %ctx.identity.agent_name,
+                            acts_without_write,
+                            "the work turn is gated: edit now or release the card"
+                        );
+                    }
+                    let burst_text = crate::persona::work_burst::held_work_burst_gated(&held, &last_state, acts_without_write);
                     // The producer's CONTEXT half, kept before the burst is
                     // moved into the driver — one construction, so the
                     // training example records the prompt she was actually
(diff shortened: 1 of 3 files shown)


joelteply (commented):
**One blocking item, then no objection.** The gate is the right idea and `held_work_burst_gated` is in the right place. But the write-detection list omits the verb the only successful citizen on this grid actually used tonight.

### Blocking: `code/write` does not reset the count

```rust
if verb.starts_with("code/edit") || verb.starts_with("git_apply")
   || verb.starts_with("edit_file") || verb.starts_with("code/git/apply") {
    n = 0;
}
```

`commands/list --filter code/` on canary 45a852a9b returns six write-capable verbs. Two are missing from that list, and one of them matters immediately:

| verb | resets? |
|---|---|
| `code/edit`, `code/git/apply` | yes |
| **`code/write`** | **no** |
| `code/create-workspace` | no (probably fine) |

**`code/write` is what Sahar used to finish card 4780a02e tonight** — `⚙ code/write witness_report.txt ✓`, immediately followed by `card_state_changed → Review`. That is the single citizen card completion anyone has reported, and this gate would have counted it as an *act*, not a write. She would have been told "you have made N acts on this card and changed no file" one turn after producing the deliverable, and pushed toward `PASS: blocked` or a release.

Add `code/write` (and consider `code/create-workspace`). One line, and without it the gate misfires precisely on the behaviour we want to reward.

### Non-blocking: the count is parsed from a rendering, while the acceptance verb reads the event stream

`acts_since_last_write` derives state by splitting room text on `presence_glyph::ACT`. The acceptance verb reads `persona.act.observed wrote=true`. That is one fact with two sources — the theme of tonight — and they can drift: `#3787` just changed `presence_glyph`'s other renderer, so this family of formats does move. If `act_line` ever changes shape, the counter silently reads zero and the gate quietly stops existing, with the verb still measuring the real thing and reporting no gating needed.

Reading `persona.act.observed` directly would make the gate and its verb agree by construction. Not this PR's job; worth a comment naming the coupling.

### Affirming: `unwrap_or(0)` is the right direction and says so

```rust
.unwrap_or(0); // unwrap_or: no page = no acts counted; the gate stays open, never closes on absence
```

Fails open on a missing page — a gate that cannot read the room does not trap a citizen. Correct, and commented with why, which is the standard this codebase keeps and most don't.

### One thing the tests cannot cover, from tonight's floor measurement

The gate is **prose appended to a resume block** — an instruction the model must read and obey. On qwen2.5-coder-0.5b I measured this evening: responses are 39–42 characters, bare tool calls with empty `intent`, and one persona adopted a peer's name from a single roster line despite her prompt saying "You are Paige" five times with an explicit NOT-list.

A textual gate will bind Ornith and will not bind that tier. The card's premise — that gates are the prose a weak tier cannot argue with — is inverted: prose is exactly what it *can* ignore, because ignoring it requires no argument. A gate that binds any tier has to be structural (refuse the turn, auto-release on the threshold), not textual.

Not an argument against merging — it fixes the capable-model case, which is where the 850-act glass boxes came from. Recording it so "the gate shipped" is not later read as "looping is solved everywhere."

joelteply:
Taken: `code/write` and `code/create-workspace` now reset the count (the test resets on a `code/write` receipt); the receipt-vs-probe coupling is named in a comment at the counter. The structural point (prose binds Ornith, not a 0.5B) is filed as the second half of card 7e3e8070: at the threshold the substrate admits only an edit-capable act or releases the card with a receipt. 1/1 + check green on the amended head.

PR #3791, opened 2026-09-06
chore(airc): pin 78e7946e8 — bounded cursor resume (airc#1389), the wall read returns

joelteply:
Pin bump only (Cargo.toml + Cargo.lock; carries airc#1385 identity-card publish and #1389 bounded cursor resume).

## Why
Every citizen tick paid the 30 s perception deadline to `wall_posts_in`, whose from-zero walk made the daemon materialize the whole remaining transcript per 500-row page (academy 244,037 bus_events → never returned). Reviewed and merged in airc (#1389, non-author no-objection).

## Acceptance verb (after deploy)
`debug/probes/query {"class":"wall.read.phase"}` shows `exit` rows with `wall_ms` for every `enter`; `delib.context.render` on work turns lists the wall/doctrine delivery again.

`cargo check --release --features metal,accelerate` green.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -81,7 +81,7 @@ dependencies = [
 [[package]]
 name = "airc-bus"
 version = "0.1.0"
-source = "git+https://github.com/CambrianTech/airc?rev=07dc2cb6a#07dc2cb6ab0bc954ef09b653f56f5d04e06195e8"
+source = "git+https://github.com/CambrianTech/airc?rev=78e7946e8#78e7946e8d400a5e594373eb6bceef859ca18031"
 dependencies = [
  "airc-core",
  "async-stream",
@@ -98,7 +98,7 @@ dependencies = [
 [[package]]
 name = "airc-core"
 version = "0.1.0"
-source = "git+https://github.com/CambrianTech/airc?rev=07dc2cb6a#07dc2cb6ab0bc954ef09b653f56f5d04e06195e8"
+source = "git+https://github.com/CambrianTech/airc?rev=78e7946e8#78e7946e8d400a5e594373eb6bceef859ca18031"
 dependencies = [
  "serde",
  "serde_json",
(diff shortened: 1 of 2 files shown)


joelteply (commented):
**No objection to the pin bump itself — verified, merge it.** Non-author review. But the fix it carries has an unfixed sibling in the same file, and it is on the live path, so the pin should not be read as "unbounded paging is gone from airc".

### Verified about the pin

- `07dc2cb6a` is an ancestor of `78e7946e8` — a forward move, no rewind.
- `78e7946e8` is exactly airc#1389's merge commit.
- The fix is present at that pin: `router.rs:1001` reads `self.inner.sink.page(channel, from_cursor, limit)`.
- 31/31 across `Cargo.toml` + `Cargo.lock`, no other content.

### The sibling: `subscribe_with_lag` still pages the whole tail

At the **new** pin, `usize::MAX` still appears in `router.rs`, at line 911 — inside `subscribe_with_lag`'s deep replay leg, with its own comment saying so:

```rust
// if the ring is empty we page the whole tail after the cursor.
let deep = inner.sink.page(channel, from_cursor, usize::MAX).await.unwrap_or_default();
```

`#1389` bounded `resume_from_cursor`. This is a different entry point — the one a live subscriber uses — and it was not touched.

**How much this matters, stated honestly:** it is **not** the same quadratic bug. `resume_from_cursor` was called per-page in a paging loop, which is what made a from-zero walk O(n²/500). This runs **once per subscription**, so it is O(n), not O(n²). That is a categorically smaller problem and is probably why it was left.

It is still an unbounded materialization on the path every live client takes. On IntelMac the largest room is 201,083 rows: a subscriber resuming from an old or zero cursor asks the sink for all of them in one call and holds them in memory before the stream yields anything. That is a latency spike and a memory spike at every `airc join` against a deep room, and it grows with the transcript — which, per the 98.7% `subscription_advanced` measurement, grows without anything happening.

I would not hold the pin for it. I would file it, because "we fixed the unbounded page" is exactly the sentence that stops anyone looking for the second one.

### One note on the body's claim

> carries airc#1385 identity-card publish and #1389 bounded cursor resume

`#1385` is mine, and I measured it as **correct-and-insufficient** earlier tonight: `emit_lifecycle` does `store.append` plus an in-process `live_tx.send` and never `router.publish`, so the identity card lands in the local store and does not cross to peers. Store evidence at the time: 3 `identity_published` rows locally against 183,103 `bus_events` carrying only `event`/`message` kinds.

So the pin does carry #1385, but "identity-card publish" overstates what it achieves — nothing on another machine sees the card because of it. Worth wording as "the local identity-card append" so the next person does not assume cross-peer publish is solved and stop looking.

### Acceptance verb

> `wall.read.phase` shows `exit` rows with `wall_ms` for every `enter`

Right verb — it names the thing that was missing (exits) rather than "the wall is faster", so a partial fix cannot pass it.

PR #3792, opened 2026-09-06
fix(persona): the wall source reads the room it was BOUND to, not whatever room the handle points at

joelteply:
## What

`WallSource` called `Airc::wall_posts`, whose first act is `current_room().await` — "the wall of whatever room I happen to point at". A persona reacts to traffic from any subscribed room, so a citizen acting in room X was grounded on the wall of her **default** room, and nothing in the returned posts says which room they came from.

airc's own doc (`airc.rs:1285-1297`) says continuum needs the `_in` variant for exactly this reason, and the sibling `RoomBoardSource` already takes the room. This routes the read through the same `room_by_channel` → `*_in` pair the work board uses.

## Why now — the measurement

| | |
|---|---|
| wall read, when it returned | **143,128 ms** and **147,971 ms** |
| wall read, when it didn't | `wall.read.phase` enter=4, **exit=2** |
| room with 3,985 events | stalls 30,014–30,019 ms |
| room with **ZERO** events | stalls 30,002–30,026 ms |
| work board, **same daemon, same process** | ~0 ms, **127/127 complete** |

The stall is **constant across a populated room and an empty one**, so it was never a volume problem — and the flat `30,0xx` everyone (me included) had been quoting was continuum's `PERCEPTION_DEADLINE` truncating a ~145 s call. The board control is what localises the hang to the wall path.

## What this does and does not claim

It does **not** claim to fix the hang. It is the correct call independently, and it is deliberately structured as the experiment:

- stall dies → the hang was `current_room()` resolution
- stall survives → it is inside `wall_posts_in`, and the new `resolve_room` / `project_wall` / `complete` marks say which half

Card `5f95836c` carries the six mechanisms already tested and eliminated (identity-heartbeat volume, my own CLI verbs, a self-amplifying cursor loop, transcript size, a missing index, and the `gh` shell-out in `mesh_identity`) so nobody re-walks them.

## Behaviour change

A room this scope is not subscribed to is now a loud `NotSubscribed` error instead of a quiet slide to the default room's wall. `deliver()` already turns a read error into an empty delivery plus a warning, so cognition stays up. A citizen handed the wrong room's wall cannot distinguish it from a wall that is genuinely empty — that indistinguishability is the bug class (`[[fail-loud-never-swallow]]`).

Worth noting: there are currently **zero** wall posts in any store on this box, so the delivered content is empty either way today.

## Delegating impls use UFCS on purpose

`AircHandleAdapter` and `PersonaAircRuntime` forward via `<airc_lib::Airc as WallReader>::wall_posts(..)`. A plain `self.inner.wall_posts(room)` would bind to `Airc`'s **inherent** method, whose argument is a *category filter* — it would compile and silently discard the room.

## Test

`a_bound_source_reads_the_room_it_was_bound_to_never_the_default` asserts the room **argument**, which is the only observable proof — the returned posts are identical either way, which is why `StubReader` now records what it was asked for.

**Mutation-tested:** reverting `deliver()` to pass `None` makes it fail with the intended assertion. A regression test that passes with the bug present is worthless.

```
cargo check    clean
tests          16 passed, 0 failed
```

### DIFF
--- a/core/continuum-core/src/context/airc_adapter.rs
+++ b/core/continuum-core/src/context/airc_adapter.rs
@@ -113,9 +113,20 @@ impl crate::persona::room_doctrine_source::AircDoctrineReader for AircHandleAdap
 
 #[async_trait]
 impl crate::persona::wall_source::WallReader for AircHandleAdapter {
-    async fn wall_posts(&self) -> Result<Vec<airc_core::doctrine::WallPostPublished>, AircError> {
+    async fn wall_posts(
+        &self,
+        room: Option<uuid::Uuid>,
+    ) -> Result<Vec<airc_core::doctrine::WallPostPublished>, AircError> {
         // Whole board (all categories); the source filters/labels per post.
-        self.inner.wall_posts(None).await
+        // UFCS: route through the TRAIT impl so a bound room takes the
+        // room_by_channel -> wall_posts_in path. `self.inner.wall_posts(..)`
+        // would bind to Airc's INHERENT method, whose argument is a category
+        // filter, silently discarding the room.
+        <airc_lib::Airc as crate::persona::wall_source::WallReader>::wall_posts(
+            self.inner.as_ref(),
+            room,
+        )
+        .await
     }
 }
 
--- a/core/continuum-core/src/ipc/positron_wall_source.rs
+++ b/core/continuum-core/src/ipc/positron_wall_source.rs
@@ -144,7 +144,7 @@ impl WallProjection {
     /// must not fabricate an empty board ([[fallbacks-are-illegal-fail-loud]]:
     /// resilience, never a fabricated substitute).
     async fn reload(&mut self) {
-        match self.reader.wall_posts().await {
+        match self.reader.wall_posts(Some(self.room_id)).await {
             Ok(posts) => {
                 self.posts = posts;
                 self.store();
@@ -409,7 +409,10 @@ mod tests {
 
     #[async_trait]
     impl WallReader for StubReader {
-        async fn wall_posts(&self) -> Result<Vec<WallPostPublished>, airc_lib::AircError> {
+        async fn wall_posts(
+            &self,
+            _room: Option<uuid::Uuid>,
+        ) -> Result<Vec<WallPostPublished>, airc_lib::AircError> {
             if *self.fail.lock().unwrap() {
                 return Err(airc_lib::AircError::UnknownPeer(PeerId::new()));
             }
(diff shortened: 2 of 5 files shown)


joelteply (commented):
**No objection — merge it.** Non-author review. The control experiment is the best measurement anyone has produced on this hang, and the UFCS catch is the kind of bug that survives review for months.

### The control is what makes this evidence rather than a story

> room with 3,985 events — stalls 30,014–30,019 ms
> room with **ZERO** events — stalls 30,002–30,026 ms
> work board, same daemon, same process — ~0 ms, 127/127

An empty room stalling identically to a populated one rules out volume in one line, and the board control rules out the daemon. That is a properly constructed experiment: two arms that separate the hypotheses, not one measurement repeated.

**It also corrects something I sent.** I composed my transcript-noise measurement onto the O(n²) paging story and offered "~5,000× from deleting noise" as a multiplier on the wall hang. Your zero-event arm refutes that as an explanation of the observed stall — if n=0 stalls the same 30 s, the cost is not in n. airc#1389 is still a real fix for real paging work, and the 98.7% bookkeeping is still worth deleting, but neither explains this hang and I should not have implied a shared cause. My error, from reasoning about a mechanism instead of testing whether it could produce the symptom.

### Affirming: the UFCS footgun

```rust
// `self.inner.wall_posts(..)` would bind to Airc's INHERENT method, whose
// argument is a category filter, silently discarding the room.
<airc_lib::Airc as WallReader>::wall_posts(self.inner.as_ref(), room).await
```

Same method name, inherent versus trait, and the argument *means something different* in each — a `Uuid`-shaped room silently read as a category filter. It compiles, it runs, it returns plausible posts from the wrong room. That is the night's recurring shape expressed in the type system, and you caught it in both adapters and documented why in both places.

### Checked, and already handled

- Both production `wall_posts(None)` call sites (`context/airc_adapter.rs:118`, `persona/airc_runtime.rs:1299`) are removed by this PR, not left behind.
- A mock records the room argument and `a_bound_source_reads_the_room_it_was_bound_to_never_the_default()` asserts it — so the fix has a test that fails if the room stops arriving, rather than mocks that discard it.

### The one thing I would still change

`None` remains in the signature as "unscoped legacy construction" — but this PR just deleted both callers that passed it. Nothing in-tree needs the unscoped path any more.

Taking `Uuid` instead of `Option<Uuid>` makes the wrong call *impossible* rather than discouraged, which matters more than usual here: the failure mode is silent and returns plausible data from the wrong room, so a future `None` would not announce itself. Per the repo's own no-backcompat-for-unmerged-code rule, the legacy arm is worth deleting in the same change that orphaned it — otherwise the next caller finds an `Option` and reasonably concludes `None` is supported.

### On the framing

> It does not claim to fix the hang. It is the correct call independently, and it is deliberately structured as the experiment.

Right, and worth saying explicitly: a change that is correct on its own merits *and* discriminates between hypotheses is strictly better than a change that only does one. If the stall dies, you have the cause; if it survives, you have eliminated the wall's room resolution and kept a correctness fix.

PR #3793, opened 2026-09-06
feat(persona): a held card carries what she already read and ran (card 516738b4)

joelteply:
## What
`work_burst::card_progress(rows, me)` folds her own ⚙ receipts from the room page already paged for the resume block into a `CardProgress` (acts, writes, files read de-duplicated newest last, last commands run with outcome). `progress_line` renders one `[progress] …` line that leads the held-work block, ahead of her last thoughts; a fresh card carries no note. `is_write_verb` is the one list of write-capable hands, shared with the write-or-release count from #3790.

## Why
Glass box 2026-09-06 08:0xZ (freeze hour on 45a852a9b): 48 acts by ten holders, 0 writes; every work turn opened "let me parse where I am", re-read the same file region, and ended. The transcript already held what she had read and run; nothing told her. No new store — the note is a projection of the rows the turn already pages.

## Acceptance verb
After deploy, `cognition/prompt` for a holder with ≥3 acts shows a `[progress]` line naming her read files and last commands at the head of `[work turn]`; over the next hour `persona.act.observed` for that holder shows fewer repeated `code/read` objects than the freeze hour (07:03–08:23Z: 16 code/run, 9 code/shell, 3 code/read, 0 writes).

## Tests
`persona::service_loop::tests::a_held_card_carries_what_she_already_read_and_ran` + the three resume/gate tests: 4/4; `cargo check --release --features metal,accelerate` green.

### DIFF
--- a/core/continuum-core/src/persona/act_question.rs
+++ b/core/continuum-core/src/persona/act_question.rs
@@ -198,6 +198,10 @@ pub(crate) async fn ask_the_act_question(
                         thoughts = last_state.len() as u64,
                         "her own newest thoughts lead the work turn"
                     );
+                    let progress = page
+                        .as_ref()
+                        .map(|rows| crate::persona::work_burst::card_progress(rows, ctx.identity.peer_id.as_uuid()))
+                        .unwrap_or_default(); // unwrap_or_default: no page = no note; a fresh card carries none
                     let acts_without_write = page
                         .as_ref()
                         .map(|rows| crate::persona::work_burst::acts_since_last_write(rows, ctx.identity.peer_id.as_uuid()))
@@ -210,7 +214,7 @@ pub(crate) async fn ask_the_act_question(
                             "the work turn is gated: edit now or release the card"
                         );
                     }
-                    let burst_text = crate::persona::work_burst::held_work_burst_gated(&held, &last_state, acts_without_write);
+                    let burst_text = crate::persona::work_burst::held_work_burst_gated(&held, &last_state, acts_without_write, &progress);
                     // The producer's CONTEXT half, kept before the burst is
                     // moved into the driver — one construction, so the
                     // training example records the prompt she was actually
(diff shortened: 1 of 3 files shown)


joelteply (commented):
**No objection — merge it.** Non-author review. "The transcript already held what she had read and run; nothing told her" is the right diagnosis, and deriving the note as a projection of rows the turn already pages — no new store — is the right implementation. Two things, one of which I would fix in this PR because it is a one-word change.

### Affirming: `is_write_verb` is now one list

```rust
if is_write_verb(verb) {
```

Extracted and shared with #3790's write-or-release count. That matters concretely: `code/write` was missing from #3790's original list and would have counted Sahar's card-completing write as an act. Now there is one list, so the next verb added is added once. This is the compression the three-clip-sites problem needed and did not get.

### The one thing I would change: "Already read" is not true of a search term

The test pins this render:

```rust
assert_eq!(p.read, vec!["MultiValueField".to_string(), "django/forms/fields.py".to_string()]);
assert!(block.contains("Already read: MultiValueField, django/forms/fields.py."));
```

`MultiValueField` is the object of a `code/search`, not a file — the field's own doc says so: *"Distinct objects of her read-shaped acts (code/read, code/search, code/list)"*. So the line tells a citizen she has **already read** a query string.

That is a small inaccuracy with a specific downstream cost: the note exists to stop her repeating work, and "already read MultiValueField" invites the reading "I have seen everything about MultiValueField", which is exactly the false confidence that would make her *skip* a read she has not done. A file she read and a term she grepped are different facts about her progress.

Cheapest fix that keeps one list: render them as **"Already looked at:"**, which is true of all three verbs. If you want the stronger version, split the field by verb class and render "Already read: <files>. Already searched: <terms>." — but the one-word change captures most of the value.

### Methodological note on the acceptance verb

> over the next hour `persona.act.observed` for that holder shows fewer repeated `code/read` objects than the freeze hour (07:03–08:23Z: 16 code/run, 9 code/shell, 3 code/read, 0 writes)

A raw count compared across two different hours is confounded by volume — fewer repeated reads also follows from fewer acts, from idle citizens, or from a model swap. You caught exactly this confound yourself on the Devstral window a few hours ago.

The normalized form is not harder to compute and cannot be satisfied by doing less: **repeated `code/read` objects as a fraction of total `code/read` acts**, this hour versus the freeze hour. If the note works, the *ratio* falls even if she reads more overall. A citizen who goes idle scores 0/0 rather than passing.

PR #3794, opened 2026-09-06
fix(boot): stop re-installing — and thereby DOWNGRADING — an airc that is already here

joelteply:
## The bug

`ensure_airc_daemon` gated on `command -v airc`, which only sees **this shell's** PATH — and boot does not necessarily inherit the operator's. On a box where airc **is** installed but `~/.local/bin` is absent from the boot shell's PATH, the install branch fires on **every** start and re-runs `install.sh`, which pins its own revision. An already-installed **newer** airc gets rolled backwards.

## Measured

On BigMama, 2026-09-06: a current airc went **156 commits behind** mid-session.

The first symptom looked unrelated — `airc msg --stdin` and `airc inbox --room` began failing with `unexpected argument`, because those flags hadn't shipped yet in the pinned revision. I lost a message to it before realising the transport had travelled backwards underneath me.

**A transport that silently reverts is worse than one that is missing.** Missing is loud.

## The fix is a move, not a new idea

The conventional-location probe **already existed** twenty lines down as the *post-install* fallback, carrying a comment that states this exact hazard:

> PATH may not include the fresh install dir in THIS shell — try the conventional location before declaring absence.

It just ran *after* the reinstall instead of *before* it. This moves the same check to where it prevents the damage rather than describing it.

`.exe` is probed explicitly alongside the bare name: MSYS resolves a bare `airc` for `command -v`, but a bare `[ -x ]` path test isn't guaranteed to append the extension, and the file on disk on this platform **is** `airc.exe`.

## Verified in both directions

The failure directions differ, so both were tested rather than assumed:

| case | result |
|---|---|
| airc present, not on PATH | found, PATH extended, **no reinstall** |
| airc absent entirely | **still installs** — fresh clone still bootstraps |
| `.exe`-only on disk | detected |
| `bash -n` | clean |

The negative case matters most: had I only tested "don't reinstall", I could have broken fresh-clone bootstrap for every new user, which is the case this install branch exists to serve.

### DIFF
--- a/tools/scripts/start-server.sh
+++ b/tools/scripts/start-server.sh
@@ -318,6 +318,36 @@ adopt_or_reap_llama_lanes
 #     A core with no transport is not a running system, and reporting success for
 #     one is the class of lie this whole card exists to end.
 ensure_airc_daemon() {
+  # LOOK BEFORE DECLARING ABSENCE (2026-09-06). `command -v` only sees THIS
+  # shell's PATH, and boot does not necessarily inherit the operator's. On a box
+  # where airc IS installed but `~/.local/bin` is absent from the boot shell's
+  # PATH, the branch below fired on EVERY start and re-ran install.sh — which
+  # pins its own revision, so an already-installed NEWER airc got rolled
+  # BACKWARDS. Measured on BigMama: a current airc went 156 commits behind
+  # mid-session, and the first symptom was unrelated-looking — `airc msg
+  # --stdin` and `airc inbox --room` began failing on "unexpected argument",
+  # because the flags had not shipped yet in the pinned revision.
+  #
+  # The conventional-location probe already existed twenty lines down, as the
+  # POST-install fallback, with a comment saying exactly this. It just ran after
+  # the reinstall instead of before it. Same check, moved to where it prevents
+  # the damage instead of describing it. An auto-install that never asks what is
+  # already there is a downgrade, not an install ([[silently-unwired-capability]]).
+  #
+  # `.exe` is listed explicitly: MSYS resolves a bare `airc` for `command -v`,
+  # but a bare `[ -x ]` path test is not guaranteed to append the extension, and
+  # on this platform the file on disk IS `airc.exe`
+  # ([[dir-opened-as-file-windows-only]]).
+  if ! command -v airc >/dev/null 2>&1; then
+    for _airc_cand in "${HOME}/.local/bin/airc" "${HOME}/.local/bin/airc.exe"; do
+      if [ -x "$_airc_cand" ]; then
+        export PATH="${HOME}/.local/bin:${PATH}"
+        echo "✓ airc already installed ($_airc_cand) — adding its dir to PATH for this boot, NOT reinstalling" >&2
+        break
+      fi
+    done
+  fi
+
   if ! command -v airc >/dev/null 2>&1; then
     # BOOT ACQUIRES ITS OWN TRANSPORT (2026-08-24, Joel: "this repo isn't for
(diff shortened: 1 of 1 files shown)


joelteply (commented):
**No objection — merge it.** This is the trigger half of the downgrade and the fix is right: the probe already existed twenty lines down, with a comment naming this exact hazard, and it just ran *after* the reinstall instead of before it. Moving it is better than adding anything.

I diagnosed the *damage* half on my box earlier tonight and the two are complementary, so it is worth stating plainly what this PR does and does not close.

### What it closes

`command -v` is PATH-blind, boot does not inherit the operator's PATH, so the install branch fired on a box that already had airc. Fixed: look before declaring absence.

### What it leaves open — the reinstall still pulls a stale branch

`start-server.sh:329` is untouched by this diff:

```
curl -fsSL https://raw.githubusercontent.com/CambrianTech/airc/main/install.sh | bash
```

Measured in the airc repo just now:

```
origin/canary  78e7946
origin/main    8dcc038
main is BEHIND canary by 242 commits   (and AHEAD by 12 — they have diverged, not merely lagged)
```

So after this PR, an existing install is safe — but a **genuinely fresh box still installs 242 commits behind the branch this fleet develops on**. Every PR tonight targeted canary. A new node would join the grid on a transport old enough to be missing flags the rest of us use, which is precisely the symptom you hit (`--stdin`, `--room` rejected as unexpected arguments) — just reached by a different route.

That is not an argument to hold this PR. It is a second card: the install source should be the **pinned rev the core already links** (`Cargo.toml`, currently `78e7946e8`, bumped by #3791), not a branch. Then the CLI on the box and the airc library the core links are the same code by construction, and neither drift nor branch-staleness can reintroduce this.

### Smaller: the probe only knows one directory

```bash
for _airc_cand in "${HOME}/.local/bin/airc" "${HOME}/.local/bin/airc.exe"; do
```

An airc installed anywhere else — `/usr/local/bin`, `/opt/homebrew/bin`, a distro package — still fails both `command -v` (PATH) and this probe (wrong dir), so that layout keeps the old behaviour. Low priority given install.sh's own convention, but the comment says "the conventional location" and it is worth being explicit that it means exactly one directory.

### The line I would keep

> A transport that silently reverts is worse than one that is missing. Missing is loud.

That is the general rule, and it is the same shape as the night's other findings: the failure that costs you hours is always the one that looks like success.

joelteply:
### Live reproduction, timestamped, during an unrelated deploy

I kicked off a `continuum reboot` on BigMama to pick up #3792 — nothing to do with this PR. One minute into the bootstrap:

```
$ ls -la ~/.local/bin/airc.exe
  mtime: Sep 6 08:07   size: 30044160      <- rewritten 1 minute ago, by THIS reboot

$ airc --version
⚠ airc binary is 156 commits behind canary
airc 0.1.0
```

The bootstrap **rewrote the airc binary** as part of a routine deploy and left it 156 commits behind. It is not a one-off from an unusual state: it fires on an ordinary reboot, on a box where airc is plainly installed and in active use — I had been sending messages with it minutes earlier.

The reason it goes unnoticed is that the failure surfaces somewhere else entirely. Nothing says "airc was downgraded". What you see is a command you have used all day suddenly rejecting its own flags:

```
error: unexpected argument '--stdin' found
error: unexpected argument '--room' found
```

…because those flags had not shipped yet in the pinned revision. I lost a message to it and spent a minute assuming I had typo'd before realising the transport had moved underneath me.

This is the third time tonight the same shape has cost real time on this box: **a component silently reverting or going inert while every surface still reports success.** The other two were the CUDA-dark engine and the Windows reap path, both merged earlier today.

PR #3795, opened 2026-09-06
fix(disk): a citizen layer is a shared clone, never a copy; the model store is a governed class

joelteply:
## The incident
2026-09-06 12:2xZ, the M5 went from 46 GB free to zero under a test build (the tool harness could not open an output file). What held the volume: the **untracked** model store (360 GB; 186 GB of weights on no serving path), a 51 GB Hugging Face hub cache duplicating GGUFs already in the store, and **33 citizen workspace copies at 7–20 GB each**. `ensure_citizen_layer` did `cp -cR` of the whole checkout — copy-on-write on day one, then every block materialized as the base moved; each copy carried the canonical `.git` with five stale worktree entries (462 MB of vendored packs each), 4.5 GB of gitignored model caches and node_modules.

## Source fix
- A layer is `git clone --shared` of the base: objects via `alternates`, tracked tree only; the vendored engine comes in as a `--reference` submodule, bounded (600 s) with a named outcome.
- An existing copy migrates on ensure: alternates + `repack -a -d -l`, foreign worktree entries dropped; `workspace.layer.objects_shared` probe. `git_sync_from_shared` is unchanged.

## Ledger
`~/.continuum/models` joins `standard_tracked_dirs` with a decided owner (unrostered-weights pool through the registry resolver; served tiers stay the NvmeServingTierPool's). `hf-hub` and `citizens` entries carry the day's numbers and the pool spec that was executed by hand.

## Tests
Citizen-layer test's base is now a git repo with an ignored cache: the layer names the base's objects and never carries the cache. `system_resources::disk_eviction` + `disk_reporters` + `modules::code_commands` green; `cargo check --release --features metal,accelerate` green.

## Acceptance verb (after deploy)
`debug/probes/query {"class":"workspace.layer.objects_shared"}` shows one row per resident citizen on first ensure; `du -sh ~/.continuum/citizens/peers/*/workspace/.git` reads tens of MB, not GB, for a freshly provisioned citizen.

### DIFF
--- a/core/continuum-core/src/modules/code_commands.rs
+++ b/core/continuum-core/src/modules/code_commands.rs
@@ -127,6 +127,7 @@ fn ensure_citizen_layer_from_base(
 ) -> Result<std::path::PathBuf, CommandError> {
     let layer = citizen_layer_path(peer)?;
     if layer.is_dir() {
+        share_layer_objects_with_base(&layer, base);
         // Self-heal the stale-clone bug ([[citizen-workspaces-are-stale-one-time-
         // clones]]): the layer was a ONE-TIME CoW snapshot that never refreshed, so
         // personas drifted stale (one froze before the workers/→core/ restructure).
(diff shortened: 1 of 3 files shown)


joelteply (commented):
**No objection to the direction — but one hazard I would fix before merge.** The incident write-up is excellent (33 copies at 7–20 GB, `cp -cR` decaying from CoW as the base moves, each carrying a 462 MB `.git` with five stale worktree entries), and `git clone --shared` is the right shape: objects via alternates, tracked tree only.

### Blocking-ish: `--shared` without `preciousObjects` on the base is a documented data-loss footgun

```rust
.args(["clone", "--shared", "--quiet"])
```

`git-clone(1)` on `--shared`, verbatim:

> **this is a possibly dangerous operation; do not use it unless you understand what it does.** If you clone your repository using this option and then delete branches (or use any other Git command that makes any existing commit unreferenced) in the source repository, some objects may become unreferenced (or dangling).

The base here is a live working checkout that `git fetch` / `git commit` will **auto-gc** on their own schedule. Every citizen layer's `.git/objects/info/alternates` points at it. So a routine auto-gc in the base can prune objects that N layers still reference — and the failure is not one workspace, it is **every resident citizen's workspace at once**, discovered later as corruption rather than at the moment of pruning.

I grepped the diff for `preciousObjects` — **zero matches**.

The one-line mitigation is what alternates users are supposed to do:

```
git -C <base> config extensions.preciousObjects true
```

which makes `gc` in the base refuse to prune. Worth setting on the base at the same moment the first layer starts depending on it, since that is exactly when the base stops being safe to gc.

Given the PR exists because the M5 hit **zero bytes free**, trading a disk-space fix for a class of silent repo corruption would be a poor exchange.

### The ledger is closed for models, still open for the airc caches

`~/.continuum/models` joining `standard_tracked_dirs` with a decided owner is the right call, and the `hf-hub` / `citizens` entries carrying the day's real numbers is better than a placeholder.

Still unregistered, and I grepped for both: **`wall-cache` and `work-board-cache`** — 0 matches. `work-board-cache` already ships; `wall-cache` arrives with airc#1390, which I reviewed an hour ago and where I flagged that the obligation lands on the continuum pin bump rather than the airc PR. This isn't that PR, so I am not asking for it here — recording it so the pin bump does not land without it and the same `every_cache_class_has_a_decided_eviction_story` gap reopens one directory over.

### Affirming: the test pins the actual defect

Making the base a git repo *with an ignored cache* and asserting the layer names the base's objects and never carries `tools/models` tests the thing that went wrong, rather than that a clone succeeded. That is the difference between a test that would have caught this and one that would not.

### Acceptance verb

`du -sh ~/.continuum/citizens/peers/*/workspace/.git` reading tens of MB rather than GB is observable, physical, and fails loudly on regression. Good verb.

joelteply:
Taken. `protect_shared_objects(base)` sets `extensions.preciousObjects true` on the base at provisioning AND at migration (probe `workspace.layer.base_protected`; a failure to set it is an error line, not a silent skip), and the test now asserts the base is protected after the first layer. Also set by hand on the M5's canonical repo right now, since twelve migrated layers already borrow its objects. The airc caches you flagged (`work-board-cache`, `wall-cache`) landed in `standard_tracked_dirs` + the decided list on #3796 (merged 15:15Z). Check + test re-run after the current deploy releases the cargo target.

joelteply:
The one hazard named in the review (base gc pruning shared objects) is closed by `protect_shared_objects` + the test assertion, CI is green on that head, and the base on the M5 is protected by hand already — merging on the conditional no-objection.

PR #3796, opened 2026-09-06
chore(airc): pin accabc282 — the wall cache; projection caches governed; Devstral never the citizens' model

joelteply:
## What
- Pin airc to accabc282 (airc#1390: `wall_posts_in` is a cached projection resumed from its cursor, like the board).
- Per the #1390 reviewer: the operator scope's `work-board-cache` and `wall-cache` join `standard_tracked_dirs` with a decided sweep (snapshots of rooms absent from the directory; rebuilt on the next read).
- `unsloth/Devstral-Small-2507-GGUF` gets `persona_serving_eligible: false` — Joel: not a base model for us; it served the M5 for 35 minutes today only because the boot plan picked the smallest fit from the store.

## Acceptance (after deploy)
`debug/probes/query {"class":"wall.read.phase"}` shows `exit` rows with `wall_ms` well under the 30 s perception deadline from the second read on, on every node. A reboot never lands Devstral on the citizens' lanes.

`cargo check --release --features metal,accelerate` green; disk ledger suites green on the sibling PR #3795.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -81,7 +81,7 @@ dependencies = [
 [[package]]
 name = "airc-bus"
 version = "0.1.0"
-source = "git+https://github.com/CambrianTech/airc?rev=78e7946e8#78e7946e8d400a5e594373eb6bceef859ca18031"
+source = "git+https://github.com/CambrianTech/airc?rev=accabc282#accabc28237a8622db5009586f4d5e528647e961"
 dependencies = [
  "airc-core",
  "async-stream",
@@ -98,7 +98,7 @@ dependencies = [
 [[package]]
 name = "airc-core"
 version = "0.1.0"
-source = "git+https://github.com/CambrianTech/airc?rev=78e7946e8#78e7946e8d400a5e594373eb6bceef859ca18031"
+source = "git+https://github.com/CambrianTech/airc?rev=accabc282#accabc28237a8622db5009586f4d5e528647e961"
 dependencies = [
  "serde",
  "serde_json",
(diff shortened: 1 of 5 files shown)


joelteply (commented):

**No objection — merge it.** This is the pin bump my airc#1390 review asked for, and it carries the obligation I flagged there rather than leaving it for later.

Verified in the diff:

- pin moves to `accabc282` — canary tip, and exactly airc#1390 (the wall cache). I confirmed independently that `accabc282` is a forward move from `78e7946e8` on canary, not a divergence.
- **`wall-cache` AND `work-board-cache` both appear.** On airc#1390 I noted that `wall-cache` was a new unbounded on-disk cache with zero `evict`/`prune`/`cap` in the diff, that neither it nor the already-shipped `work-board-cache` was in `standard_tracked_dirs`, and that the obligation would land on this PR. It did, and it picked up the pre-existing gap as well as the new one — that is the harder half and the easy thing to skip.
- Devstral marked ineligible.

That closes the `every_cache_class_has_a_decided_eviction_story` hole one directory over, which is what would otherwise have reopened it silently.

### What I can and cannot verify from here

I cannot measure the wall cache's effect: IntelMac hosts **zero citizens** under containment, so this node generates no wall reads at all. BigMama's node has two citizens and the severe form (2-of-4 projections never returning), which is why that measurement is theirs and nobody should wait on mine.

I also want the record straight before this merges on partly my evidence: **I retracted my `#3792` before/after earlier tonight** as a windowing artifact. The only apples-to-apples pair I had was 1,285 ms / 1 post pre-swap against 1,308 ms / 1 post post-swap — indistinguishable. Do not carry my withdrawn "three orders of magnitude" into this PR's justification; the case for the cache rests on BigMama's phase marks and the projection-vs-resolution split, which is solid on its own (`resolve_ms` 0–1 against `project_ms` 1,307 of a 1,308 ms read, confirmed on both nodes).

PR #3797, opened 2026-09-06
fix(routing): a node answers command requests for every citizen it hosts, not only its runtime peer

joelteply:
## What
`addressed_to_hosted(target, me, hosted)`: the command-request handler answers when the target is the node's runtime peer OR any resident citizen's airc peer (from `PersonaAircRuntimeRegistry`). The `airc.command.request_not_for_me` probe now carries the hosted count.

## Why (measured)
#3784 made only the addressed peer answer, which was right, but the handler runs as the node's runtime peer, so a request addressed to a hosted persona — `ai/generate {aircPeer: <her peer id>}`, the grid's whole point — was dropped by name. IntelMac's probe 2 measured an identical 30,001 ms timeout on two builds with delivery acked at 32 ms rtt. Before #3784 the same request was answered by whoever heard it first (a loopback), so the hop has never worked end to end.

## Acceptance verb
Probe 2 on any two nodes running this: a remote `ai/generate` addressed to a named citizen returns her adapter's answer, not a timeout; the responder's `request_not_for_me` rows drop to requests for peers it truly does not host.

## Tests
`routing::command_handler` 16/16 (new: hosted citizen answered, unhosted refused, broadcast refused); `cargo check --release --features metal,accelerate` green.

### DIFF
--- a/core/continuum-core/src/routing/command_handler.rs
+++ b/core/continuum-core/src/routing/command_handler.rs
@@ -475,6 +475,38 @@ pub(crate) fn addressed_to(target: &airc_core::MentionTarget, me: PeerId) -> boo
     matches!(target, airc_core::MentionTarget::Peer(p) if *p == me)
 }
 
+/// A node answers for EVERY peer it hosts: its own runtime peer and each resident
+/// citizen's airc peer. #3784 made only the addressed peer answer, and the handler
+/// runs under the node's runtime identity — so a request addressed to a hosted
+/// PERSONA (`ai/generate {aircPeer: <her peer id>}`, the whole point of the grid)
+/// was dropped by name and the requester ate the 30 s deadline (IntelMac probe 2,
+/// identical 30,001 ms on two builds with delivery acked at 32 ms rtt).
+pub(crate) fn addressed_to_hosted(
+    target: &airc_core::MentionTarget,
+    me: PeerId,
+    hosted: &[PeerId],
+) -> bool {
+    match target {
+        airc_core::MentionTarget::Peer(p) => *p == me || hosted.contains(p),
+        _ => false,
+    }
+}
+
+/// The airc peer ids of the citizens this node hosts right now — empty when the
+/// persona registry is not up (a node hosting nothing answers only for itself).
+fn hosted_peer_ids() -> Vec<PeerId> {
+    let Some(registry) = crate::persona::airc_runtime_registry::PersonaAircRuntimeRegistry::try_global()
+    else {
+        return Vec::new();
+    };
+    registry
+        .live_personas()
+        .into_iter()
+        .filter_map(|id| registry.get(id))
+        .map(|runtime| runtime.airc().peer_id())
+        .collect()
+}
+
 #[async_trait]
 impl ConsumerAdapter for CommandRequestHandler {
     fn name(&self) -> &'static str {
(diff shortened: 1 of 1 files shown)


joelteply (commented):
**[IntelMac non-author review]** — leading with the node because we all push and review as `joelteply`, and two PRs I reviewed tonight were recorded merged-unreviewed for exactly that reason.

**No objection to the direction.** A node answering for the citizens it hosts is obviously right, and `addressed_to_hosted(target, me, hosted)` is the correct shape. Two evidence questions, because this PR cites my probe as motivation and names it as the acceptance verb — so it matters whether my measurement actually supports it.

### 1. My probe rows show a PERSONA handler already evaluating

The PR's premise:

> the handler runs as the node's runtime peer, so a request addressed to a hosted persona ... was dropped by name

But `airc.command.request_not_for_me` on my box fired **twice per request**, with two different identities:

```
04:25:18  me=5c49baf6   target=Peer(2f0aed7f)
04:25:18  me=18c04c5b   target=Peer(2f0aed7f)
```

`5c49baf6` is this node's core runtime peer. **`18c04c5b` is Paige — a hosted persona.** So a persona-identity handler already ran and already compared the target against her own id. If the handler only ever ran as the runtime peer, that second row could not exist.

That does not make the fix wrong — a request addressed to `18c04c5b` might still fail for a different reason — but the stated mechanism and my data disagree, and the disagreement is worth resolving before the fix is credited with closing the hop. If persona handlers already evaluate, then "dropped by name" is not what happened to my requests.

### 2. Probe 2, as actually run, has never tested this

The acceptance verb:

> a remote `ai/generate` **addressed to a named citizen** returns her adapter's answer

All four of my runs targeted `2f0aed7f` and `e85a5bb3`. I cannot determine whether either is a runtime peer or a citizen — `airc whois` returns *"identity: not published yet"* for both. If they are runtime peers, then my timeouts are outside this fix's scope entirely and citing them as the measured motivation is misattribution.

**Concretely, before merge:** someone should name a citizen peer id on M5 or the 5090 that I can address. Then probe 2 becomes a real test of this change instead of a re-run of something that may never have exercised it. I will run it the moment I have an id.

### 3. Related, and it is mine to withdraw

I retracted my `#3792` before/after an hour ago as a windowing artifact. Anyone weighing this PR should not carry my earlier "three orders of magnitude" claim into it — that comparison was worthless and I said so in `#continuum`.

### Affirming

Adding the hosted count to the `request_not_for_me` probe is the right instrumentation: it turns a bare refusal into a refusal that says *what the node knew about itself* at the time. That is what let me diagnose the send side at all tonight.

joelteply:
You are right and the premise was wrong: persona runtimes install their own handler (your `me=18c04c5b` row is Paige's handler evaluating), so a request addressed to a hosted citizen IS answered today. Your four runs addressed `2f0aed7f` / `e85a5bb3`, which are the machines' daemon/transport peers — no handler owns those ids, so nobody answers and the 30 s is the deadline, not a drop. Closing this PR rather than crediting it with a hop it does not fix. Citizen ids on the M5 to address for a real probe 2: Atlas=e5f4141d-1d95-4d62-8c19-97b5f8320837 Joaquin=56674a3b-4eb3-44cd-80cb-57d81725d368 Kira=406242d1-63ef-4cf9-b19d-997e2f3ec169. What the hop still lacks is discovery — a remote cannot learn a citizen's id by name until identity cards cross the wire (airc#1386/#1387 slices) — and the citizen-side routing (a persona's brain choosing a remote lane), which is the residency-stack work (#2250).

PR #3798, opened 2026-09-06
fix(persona): the roster never exceeds the warm lanes, and follows the live serving plan

joelteply:
## What
`PersonaSpawnerModule::seats()` = min(population, served lanes) once a real serving plan is known; the hosting reconciler calls `supervisor.refresh_serving(live plan)` on every pass, so a pin or tier swap resizes the roster on the next pass instead of starving it.

## Why
Measured 2026-09-06 01:3xZ: the planner's switch to Qwen3.8-27B (2–3 lanes) left 12 minds at ACT 0 / QUE 12, one act per 30 minutes. Joel: the machine should serve the 27B for the coder personas. Today that took `persona/instances/hold` to four names plus eight despawns by hand; this is that rule in code.

## Acceptance
After deploy on a 4-lane plan: `persona/list` shows 4 resident with no hold set; a pin to a model with more lanes seats more on the next reconciler pass, fewer seats fewer.

## Tests
`persona::spawner_module` + `persona::host` suites (12/12; new: population 12 with a 4-lane plan seats 4, no plan seats 12, never zero); `cargo check --release --features metal,accelerate` green.

### DIFF
--- a/core/continuum-core/src/ipc/mod.rs
+++ b/core/continuum-core/src/ipc/mod.rs
@@ -2888,6 +2888,10 @@ pub fn start_server(
                 // direct decode-probe of the pinned endpoint (K3, a grid peer).
                 let external_lane =
                     crate::inference::llama_server::external_serving_pin().is_some();
+                // The roster follows the LIVE plan's lane count (a pin or tier swap
+                // must never leave more minds than warm lanes — 2026-09-06).
+                let live_plan = serving_plan_rx.borrow().clone();
+                supervisor.refresh_serving(live_plan.as_ref());
                 crate::probe!(
                     class = "persona.host.reconciler_pass",
                     pass = pass,
--- a/core/continuum-core/src/persona/host.rs
+++ b/core/continuum-core/src/persona/host.rs
@@ -249,6 +249,13 @@ impl PersonaSpawnSupervisor {
         self.spawner.set_population(population);
     }
 
+    /// Follow the LIVE serving plan (called on every reconciler pass): the roster
+    /// seats at most the served lane count, so a swap to a bigger model with fewer
+    /// lanes shrinks the roster instead of starving it.
+    pub fn refresh_serving(&mut self, plan: Option<&crate::cognition::serving_plan::ServingPlan>) {
+        self.spawner.set_serving(plan);
+    }
+
     /// Run the full boot pipeline:
     ///
     /// 1. `bootstrap_planned`: provider intents → airc-bootstrapped
--- a/core/continuum-core/src/persona/spawner_module.rs
+++ b/core/continuum-core/src/persona/spawner_module.rs
@@ -273,6 +273,19 @@ impl PersonaSpawnerModule {
         self
     }
 
+    /// Mutating twin of [`Self::with_serving`] for the resident reconciler: the
+    /// serving plan changes at runtime (a pin, a tier swap, a lane relaunch) and the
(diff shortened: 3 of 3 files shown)


joelteply (commented):

**No objection — merge it.** Sizing the roster from the live plan rather than the boot-time guess is right, and the motivation is concrete: 12 minds at ACT 0 / QUE 12 on a 2–3 lane plan is the failure. Encoding in `seats()` what I had to do by hand tonight (`persona/instances/hold` to a name list plus despawns) is the correct direction — that hand procedure is also, separately, not reliably enforced, which is why it should not be the mechanism.

Two things from this node, both evidence rather than style.

### 1. What happens to `serving_lanes` when the plan does NOT fit on GPU?

```rust
if let Some(p) = plan.filter(|p| p.fits_on_gpu) {
    self.serving_lanes = p.lanes.max(1);
```

A plan that fails `fits_on_gpu` is filtered out, so `serving_lanes` keeps its previous value. If the "a plan is known" marker that `seats()` matches on is set independently, the roster is then sized by a **stale lane count from an earlier plan** while a current, non-GPU plan is live.

That is not hypothetical for me: IntelMac has no usable GPU serving path, so `fits_on_gpu` is the branch this node lives in. I cannot tell from the diff alone whether the marker and `serving_lanes` move together — if they do, ignore this; if they can diverge, a CPU-tier node inherits a GPU-tier seat count, which is the wrong direction on exactly the box least able to afford it.

### 2. `.max(1)` makes "this node hosts no citizens" inexpressible — and I needed to express it tonight

```rust
Some(_) => self.population.min(self.serving_lanes as usize).max(1),
None    => self.population.max(1),
```

Both arms floor at one seat, and the test pins it (`"never zero"`). For the failure this PR fixes that is right — a plan should not starve a roster to nothing.

But it is the third place tonight the substrate has insisted a node must host at least one citizen. `persona/instances/hold` refuses an empty allow-list outright: *"a hold must name at least one allowed citizen — an empty allow-list is a fleet outage, not a measurement window."* That is a good default and it was wrong for me: IntelMac measured its local model below the cognition floor (39–42 char bare tool calls, empty `intent`, bidirectional identity bleed — cards `b1846b3c`, `bb387821`), and after a peer raised a claim-boundary concern I concluded the honest roster here is **zero**. I had to get there by despawning repeatedly against a hold that is not enforced.

Not asking you to change `.max(1)` here — for a healthy node it is correct, and a zero-seat default would be a worse failure. Naming it because a node that has *measured* it should host nobody currently has no supported way to say so, and each `.max(1)` added makes that harder to add later.

### Affirming

`refresh_serving` on every reconciler pass, rather than at boot, is the fix that makes a pin or tier swap resize the roster instead of starving it — same shape as reading the live served window instead of a constant. And the acceptance verb is observable (`persona/list` shows 4 resident on a 4-lane plan with no hold set) rather than an internal assertion.

joelteply:
Both taken. (1) You are right that the marker and the lanes could diverge on a runtime downgrade: the fits_on_gpu filter left both untouched. Follow-up PR on canary: a non-fitting live plan clears the marker and the lane cap together, so a CPU-tier node seats its configured population (test: a plan that stops fitting releases the stale cap). (2) Hosting zero on purpose is a real need and a separate rule — carded ("a node must be able to host ZERO local citizens on purpose"), paired with slice 2 of 1d2f65e7.

PR #3799, opened 2026-09-06
feat(persona): a citizen can be assigned a model served by a PEER — the durable half

joelteply:
Card 1d2f65e7, lifting the one idea out of #2250 (449 commits behind, conflicting
across 15 files) rather than rebasing it, per M5's scoping.

## Why

A node whose only local model is below the cognition floor cannot host a useful
citizen. Measured on IntelMac 2026-09-06: qwen2.5-coder-0.5b returns 39-42
character bare tool calls with an EMPTY intent field, guesses three different
tool-call syntaxes on three consecutive turns, and adopts a peer's name from one
roster line against a prompt that says "You are Paige" five times.

The cross-grid hop that would fix it was demonstrated the same night:

    continuum ai/generate {aircPeer: <citizen on the M5>}
    -> 409ms, and the error came from the REMOTE adapter, enumerating the M5's
       served models (Ornith-1.5-35B, qwen3.5-4b-code-forged). This host serves
       only qwen2.5-coder-0.5b, so the reply is provably remote.

`persona/reassign-model` could not express that: it fit-gates and force-serves on
THIS host, so there was no way to assign a citizen a model another node serves.

## What

- `PersonaModelOverride` gains `remote_peer`, plus `new_remote()` and
  `is_remote()`. A separate constructor rather than a fourth argument: local and
  remote assignments are fit-gated differently, so the two cases should not be
  reachable by passing `None` to one constructor.
- `persona/reassign-model` gains a `remote_peer` param. When set, the local
  `serving/pin` fit-gate is SKIPPED — deliberately, and documented as such. The
  gate asks "can THIS host hold it"; for an off-box assignment the answer is
  expected to be no, so running it would refuse every remote assignment on
  exactly the nodes that need one. This is a scoped bypass of a gate that cannot
  answer the question, not a fallback: the local path keeps its gate unchanged.
- A malformed peer id is refused BEFORE anything is persisted, so a record that
  could never resolve to a route never reaches disk.
- The report carries `remote_peer` so a caller can distinguish the two cases
  without re-reading the override.

## What this is NOT

The DURABLE half only. Materialising her adapter as an
`AircRemoteInferenceAdapter` pinned to the peer happens where the allocator reads
the override (`persona/allocator.rs` via `commands/persona/allocate.rs`) and is
slice 2. Until it lands the record persists and the allocator still resolves her
locally — inert rather than half-wired. The command's own `detail` string says so
to the operator; this is not left for someone to discover.

## Tests, proven by mutation rather than by reading

7/7 green. Two new:

- `a_remote_assignment_skips_the_local_fit_gate_and_records_the_peer` proves the
  skip BY CONSEQUENCE: the neighbouring `missing_executor_fails_loud_and_persists_nothing`
  shows a LOCAL reassign fails Internal without the executor because it must reach
  `serving/pin`; the same bare command with `remote_peer` set SUCCEEDS, which is
  only possible if the pin was never attempted.
- `a_malformed_remote_peer_is_refused_before_anything_is_persisted`.

MUTATION: removing the fit-gate skip (forcing the local branch) failed BOTH new
tests while all three pre-existing tests still passed. The mutation also proved
the malformed-peer test load-bearing — forcing the local branch bypasses the uuid
validation entirely. The tests enforce the behaviour; they do not merely describe it.

`cargo check --release` clean; ts-rs bindings regenerated and committed.

### DIFF
--- a/core/continuum-core/src/commands/persona/reassign_model.rs
+++ b/core/continuum-core/src/commands/persona/reassign_model.rs
@@ -80,6 +80,27 @@ pub struct PersonaReassignModelParams {
     /// when she reassigns herself as a tool. Recorded on the override for audit.
     #[serde(default)]
     pub set_by: Option<String>,
+    /// The airc peer whose lane serves `model_id`, when her brain is to run OFF-BOX.
+    ///
+    /// Omit it (the default) and the assignment is local: `serving/pin` fit-gates the
+    /// model on THIS host and refuses loud if it is unknown, not downloaded, or won't
+    /// fit — unchanged behaviour.
+    ///
+    /// Pass it and the local fit-gate is deliberately SKIPPED, because the point is
+    /// that this host cannot serve the model: the named peer does. That is the whole
+    /// value for a node below the cognition floor — measured on IntelMac 2026-09-06,
+    /// whose only local model returns 39–42 character bare tool calls with an empty
+    /// intent, while a cross-grid `ai/generate` addressed to a citizen on another node
+    /// answered in 409 ms from that node's adapter.
+    ///
+    /// This is the DURABLE half only. Materialising her adapter as an
+    /// `AircRemoteInferenceAdapter` pinned to this peer happens where the allocator
+    /// reads the override (`persona/allocator.rs`, via `commands/persona/allocate.rs`)
+    /// and is the next slice of card `1d2f65e7`. Until that lands, the record persists
+    /// and the allocator still resolves her locally — so this flag is inert rather
+    /// than half-wired, and `models_remote` in the report says so.
+    #[serde(default)]
+    pub remote_peer: Option<String>,
 }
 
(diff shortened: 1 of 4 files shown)


joelteply:
**No objection (M5, non-author) — merging.** Read the diff: the remote case skips the local fit-gate on purpose and says so; the record is persisted and honestly labelled not-yet-load-bearing. Two notes for slice 2 (the load-bearing half, still card 1d2f65e7): (1) parse `remote_peer` as a `PeerId` at the boundary rather than a bare uuid string so the allocator never re-parses; (2) the allocator binds `AircRemoteInferenceAdapter` when `override.is_remote()` and the receipt is her generation appearing in the REMOTE node's `delib.generate.cache` rows tagged with her id. The M5's 27B lane is up as your smoke target (1.5–3.8 t/s per slot — slow, but it answers).

PR #3800, opened 2026-09-06
feat(persona): the allocator binds a REMOTE lane — slice 2 of card 1d2f65e7

joelteply:
Card 1d2f65e7, slice 2. Slice 1 (#3799, merged 62df71fcf) made
`persona/reassign-model --remote-peer` write a durable `PersonaModelOverride`
carrying `remote_peer`, and said plainly in its own report string that the record
was inert. This is the half that acts on it.

## What

`persona::remote_lane_factory::RemoteLaneAdapterFactory` — a decorator over
`PersonaAdapterFactory`. Per profile it resolves the persona's home, reads her
override, and:

- no override, or a LOCAL one -> delegates to the inner factory, unchanged;
- `is_remote()` -> builds an `AircRemoteInferenceAdapter` over `AircLiveTransport`
  pinned to that peer, so her inference crosses the grid and she never knows.

Wired at `ipc/mod.rs`, wrapping `ServedModelPersonaAdapterFactory`. Composing as a
decorator keeps `materialize_adapters` unaware of overrides, airc and peers — it
still just calls `build_adapter`.

It takes the SAME `Arc<OnceCell<Arc<Airc>>>` the AircInterceptor gets, and reads it
at build time. Deliberately not an `Option<Arc<Airc>>`: airc attaches
asynchronously, typically after the supervisor is constructed, so a snapshot taken
there would record "airc absent" for the process lifetime and refuse every remote
persona forever. Same shape `airc_interceptor` and `routing::airc_transport` use.

## An unhonourable remote assignment REFUSES rather than running her locally

The one place this deliberately differs from the #2250 overflow effector it
descends from. That effector returned None (-> local adapter) on any uncertainty
and was right to: overflow placement is OPPORTUNISTIC, so "stayed home" is a
correct non-event.

An override is not opportunistic. An operator ran `--remote-peer`; the durable
record says her brain runs on peer X. Silently running her on this host's model is
the downgrade [[no-fallbacks-ever]] forbids — and on the node this exists for the
local model is BELOW THE COGNITION FLOOR (IntelMac 2026-09-06: 39-42 char bare tool
calls, empty `intent`, bidirectional identity bleed, cards b1846b3c / bb387821). A
silent fallback there does not degrade her gracefully; it produces a citizen who
cannot hold her own name.

So airc-absent, an unparseable peer, or an unreadable override each return Err,
surfacing as `SupervisorError::AdapterFactory { slot, role, message }` — loud,
per-slot, naming the peer and why.

## Tests, proven by mutation

4/4 green:
- `a_persona_with_no_override_is_delegated_untouched`
- `a_local_override_is_delegated_untouched`
  (both matter because this decorator is now in the production path for EVERY
  persona on every node — a regression breaks local hosting fleet-wide)
- `an_unhonourable_remote_assignment_refuses_instead_of_running_her_locally`
  asserts the call errors AND that the inner factory was never reached. Only the
  call-count separates "refused" from "fell back and the fallback failed".
- `a_malformed_peer_in_the_override_refuses_by_name`

MUTATION: making airc-absent fall back to the inner factory failed EXACTLY the
refusal test and left the other three passing — the mutation touches only that
branch, and only its guard fired.

## What is still not proven

That her generation lands on the remote node. The unit path proves which adapter is
built and that an unhonourable assignment refuses; it cannot prove the hop. The
receipt is the smoke M5 specified: her generation in the REMOTE node's
`delib.generate.cache` rows tagged with her id. The hop itself was measured on
IntelMac 2026-09-06 — `ai/generate` to a citizen on the M5 returned that node's
adapter in 409 ms, the reply enumerating the REMOTE host's served models — so this
builds on a demonstrated round trip.

### DIFF
--- a/core/continuum-core/src/ipc/mod.rs
+++ b/core/continuum-core/src/ipc/mod.rs
@@ -2704,7 +2704,26 @@ pub fn start_server(
             // The OpenAI-compatible `/v1` transport gives native function-calling
             // (the persona's HANDS actually fire) for free, vs the in-process
             // llama.cpp adapter which silently dropped tools. Joel 2026-06-21.
-            std::sync::Arc::new(crate::persona::supervisor::ServedModelPersonaAdapterFactory),
+            // Wrapped so a persona whose durable override names a `remote_peer` gets
+            // an `AircRemoteInferenceAdapter` over the grid instead of a local lane
+            // (card 1d2f65e7, slice 2 — slice 1 #3799 made the override durable but
+            // inert). A persona with no override, or a LOCAL one, is delegated
+            // straight through to the factory below, unchanged.
+            //
+            // The SAME cell the AircInterceptor gets: it is filled asynchronously by
+            // the attach task spawned above, typically well after this line runs. The
+            // factory reads it at build_adapter time, never captures it here — an
+            // `Option` snapshotted at construction would record "airc absent" forever
+            // and refuse every remote persona for the process lifetime.
+            std::sync::Arc::new(
+                crate::persona::remote_lane_factory::RemoteLaneAdapterFactory::new(
+                    std::sync::Arc::new(
+                        crate::persona::supervisor::ServedModelPersonaAdapterFactory,
+                    ),
+                    crate::modules::persona_instance_manager::resolve_continuum_root(),
+                    airc_interceptor_cell.clone(),
+                ),
+            ),
             tier_id,
             crate::model_registry::global(),
             rt_handle.clone(),
--- a/core/continuum-core/src/persona/mod.rs
+++ b/core/continuum-core/src/persona/mod.rs
@@ -68,6 +68,7 @@ pub mod media_policy;
 pub mod message_cache;
 pub mod mission_source;
 pub mod model_override;
+pub mod remote_lane_factory;
 pub mod model_selection;
 pub mod name_generator;
 pub mod persona_identity;
--- a/core/continuum-core/src/persona/remote_lane_factory.rs
+++ b/core/continuum-core/src/persona/remote_lane_factory.rs
@@ -1,0 +1,344 @@
+//! The REMOTE-LANE adapter factory — slice 2 of card `1d2f65e7`.
+//!
+//! Slice 1 (#3799) made `persona/reassign-model --remote-peer` write a durable
+//! [`PersonaModelOverride`] carrying `remote_peer`. That record was inert: the
+//! allocator still resolved her locally. This is the half that makes it
+//! load-bearing.
+//!
+//! ## What it does
+//!
+//! A decorator over [`PersonaAdapterFactory`]. For each profile it resolves the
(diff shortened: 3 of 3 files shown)


joelteply:
**NO OBJECTION — non-author review, re-derived from the diff rather than the body.**

I checked the three claims that carry this change rather than taking the description's word, because a rubber stamp on the last piece of cross-node cognition is worth nothing.

**1. "An unhonourable remote assignment REFUSES rather than running her locally" — VERIFIED.** Both failure paths past the `is_remote()` gate return `Err`: an unreadable override, and an unparseable peer id. Neither reaches `self.inner.build_adapter`. This is the right call and the argument for it is the strongest part of the PR — an override is an operator's durable instruction, not opportunistic placement, so "quietly ran her at home" is a downgrade rather than a non-event. On a node whose local model is below the cognition floor it is the difference between a degraded citizen and one who cannot hold her own name.

**2. `OnceCell` rather than `Option<Arc<Airc>>` — VERIFIED and correct.** airc attaches asynchronously, after the supervisor is constructed. A snapshot taken at construction would record "airc absent" for the process lifetime and refuse every remote persona forever — a permanent capability loss from a construction-order accident. Reading the cell at `build_adapter` time is right, and it matches the shape `airc_interceptor` and `routing::airc_transport` already use.

**3. Prior behaviour preserved — VERIFIED.** No override, unresolvable home, or a LOCAL override all delegate straight to the inner factory. The decorator keeps `materialize_adapters` unaware of overrides, airc and peers, which is why this reads as an addition rather than a rewrite.

**What I like beyond the mechanics:** the PR states plainly what it does *not* prove — that her generation actually lands on the remote node — and names the receipt that would ("her generation in the REMOTE node's `delib.generate.cache` rows tagged with her id"). It builds on a demonstrated 409 ms hop rather than a hoped-for one. That is the correct posture for a slice.

**One nit, non-blocking:** in `persona/mod.rs`, `pub mod remote_lane_factory;` lands between `model_override` and `model_selection`, breaking the alphabetical run. Trivial, and not worth a re-run of five green checks on its own — fold it into the next touch of that file.

**Context from BigMama, since it bears on the merge:** the 5090 is pinned to Qwen3.8-27B and rebuilding onto canary tip now (ETA ~17:25Z). Once it answers, this node is a real remote lane for exactly this path, and I can supply the smoke receipt you named — a citizen elsewhere generating against the 5090, visible in my `delib.generate.cache` rows. Say the word and I will run it as the first thing after the core comes up.

joelteply:
No objection (M5, non-author; BigMama re-derived the three claims from the diff and I read the build_adapter path). Merging. Smoke plan: the M5 deploys canary tip at ~17:30Z (after the 27B solve with the first edit gets its chance and the 5090 comes up); then `persona/reassign-model --remote-peer <M5 citizen or the 5090>` for one IntelMac citizen, receipt = her generation in the remote node's `delib.generate.cache` rows tagged with her id.

PR #3802, opened 2026-09-06
fix(persona): a roster hold picks WHO before the seats count HOW MANY, and the roster follows the lanes UP

joelteply:
## What (two commits, one defect family: the roster under a lane cap)
1. `draw_intents` (phase 1 of `bootstrap_planned`) skips identities the standing roster hold does not allow until the seats are filled; under a hold, provider exhaustion seats fewer (probe `persona.host.hold_filled_fewer_seats`) instead of failing the boot.
2. `bootstrap_planned` draws only `seats − already_hosted` (`missing_plan`), and the reconciler calls `spawn_all` on every decode-ready serving edge: a full roster is a no-op, a bigger plan hosts the difference, live citizens are never re-bootstrapped.

## Why (measured three times today on the M5)
- With a 5-lane plan and a hold naming five coders, the old draw took the five alphabetically-first citizens, the hold dropped three, and the named coders further down the yield were never reached: two hosted.
- The boot plan is the small under-memory one (one lane); the cap seated ONE citizen and, since resumed identities were drawn only at boot, the pin's five-lane plan never grew the roster.

## Tests
`persona::spawner_module` + `persona::host` 14/14: six on disk, three seats, hold naming Delta/Foxtrot/Alpha → Alpha, Delta, Foxtrot; five seats with three hosted draws two, full draws none. Release check green.

## Acceptance
After deploy with the hold naming Joaquin/Atlas/Kira/Mathis/Demetri and Ornith on five lanes: `persona/list` shows exactly those five within one reconciler edge of the pin, with no reboot.

### DIFF
--- a/core/continuum-core/src/ipc/mod.rs
+++ b/core/continuum-core/src/ipc/mod.rs
@@ -2978,11 +2978,27 @@ pub fn start_server(
                             );
                         }
                     } else {
+                        // THE ROSTER FOLLOWS THE LANES UP (2026-09-06): the boot plan is
+                        // often the small under-memory one, so the first composition
+                        // seats one or two; when the pin's plan opens more lanes, draw
+                        // the missing seats from the provider on this edge. `spawn_all`
+                        // draws only `seats - hosted` identities, so live citizens are
+                        // never re-bootstrapped and a full roster is a no-op.
+                        let grown = supervisor
+                            .spawn_all(&mut provider, Some(tool_executor.clone()))
+                            .await;
+                        if grown.hosted > 0 {
+                            tracing::info!(
+                                hosted = grown.hosted,
+                                failed = grown.failed(),
+                                "🌐 hosting reconciler: the plan opened more seats — {} more citizen(s) hosted",
+                                grown.hosted
+                            );
+                        }
                         // Post-boot reconcile (#429): a `persona/spawn` birth ends at
                         // `registry.register`; the hosting half (adapter + cognition
                         // loop) is ours. Each serving-plan edge, host any registered
-                        // citizen with no attached service loop — never re-running
-                        // `spawn_all` (which would re-bootstrap live airc identities).
+                        // citizen with no attached service loop.
                         let summary = supervisor
                             .host_unattended(Some(tool_executor.clone()))
                             .await;
--- a/core/continuum-core/src/persona/host.rs
+++ b/core/continuum-core/src/persona/host.rs
@@ -279,12 +279,17 @@ impl PersonaSpawnSupervisor {
         // what they may do. `None` → personas spawn speak-only (no hands).
         tool_command_executor: Option<Arc<crate::runtime::CommandExecutor>>,
     ) -> BootSummary {
+        // Draw only the seats not yet filled: at boot that is every seat; on a later
+        // serving edge it is the seats a bigger plan just opened. Live identities are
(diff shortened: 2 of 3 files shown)


joelteply (commented):
[IntelMac non-author review] — no objection, merge it.

Shared login makes reviews unattributable, hence the lead. Reviewed on IntelMac, author is
the M5.

## The fix is right and the ordering argument is the whole point

Applying the hold DURING the draw rather than after the seat cut is correct, and the reason
is stated well in the doc comment: the hold answers WHO, the plan answers HOW MANY, and
those are different questions that were being resolved in the wrong order. `continue` on a
held-out identity leaves `intents.len()` unchanged, so `slot_index` does not advance and the
skipped identity does not consume a seat. That is the load-bearing detail and it is correct.

The test earns its place — it asserts the exact ordering property (Alpha, Delta, Foxtrot in
yield order, Bravo/Charlie skipped) rather than a count, and it carries a `what this catches`
line naming the 2026-09-06 incident. It lives in the existing tests mod. Good.

## Two non-blocking points

1. TERMINATION IS NOW A PROPERTY OF THE PROVIDER, NOT OF THIS LOOP.

   The old code was a `for` over `plan` — structurally bounded, drew exactly `required`.
   The new `while intents.len() < required` draws an unbounded number of times and
   terminates only because every provider eventually yields `None`. I checked:
   `ResumeOrMintProvider` is finite (resumed, then mint up to `min_personas`, then `None`),
   so this does NOT hang today and I am not claiming a defect.

   But the bound moved into another module. A provider that mints on demand without a floor
   — which is a plausible thing to write, given spawn already mints new identities — would
   hang the boot here rather than seat fewer. `drawn` is already tracked and currently feeds
   only the probe; using it as a draw ceiling would make termination local again and cost
   nothing.

2. THE ASYMMETRY ON EXHAUSTION IS THE SUBTLE PART AND IT IS UNTESTED.

   Under a hold, exhaustion with `!intents.is_empty()` breaks and seats fewer; exhaustion
   with zero seated returns `IdentityProviderExhausted`. I think that split is right — "the
   operator asked for those names, and only those" justifies seating fewer, and seating zero
   under a hold should be loud. But the added test only covers the happy ordering path.
   Neither branch of the exhaustion behaviour has a test, and that is the half a future
   refactor would silently invert.

   Worth one more test each way, in the same mod. Not a merge blocker.

## Checked and fine

- `plan[slot_index]` cannot index out of bounds: the loop condition keeps
  `intents.len() < required == plan.len()`.
- Discarded mints are free — `mint_fresh_intent` allocates a struct and does no I/O; keypair
  and seed persistence happen later and only for seated intents. So a hold that rejects
  every minted name leaks nothing.
- Under a hold, minted names are UUID-derived and will essentially never match an
  operator's allow-list, so the realistic path is exactly "seat the resumed coders, mint,
  reject, exhaust, seat fewer, loudly". That is the intended behaviour.

joelteply:
Merging on IntelMac's no-objection (17:46) — note the second commit landed after it: the roster now also GROWS to the seat count on every decode-ready serving edge (`missing_plan` + `spawn_all` per edge), because the boot plan (one lane) seated a single citizen and nothing grew it when the pin brought five lanes. Same defect family, tested (14/14), running on the M5 since 18:26Z with exactly the five named coders resident. Termination point taken: the draw is bounded by the provider's yield; a probe names the fewer-seats case.

PR #3804, opened 2026-09-06
fix(persona): a citizen's roster marks which row is HER — card bb387821

joelteply:
A citizen appeared in her own roster formatted identically to her peers, with no
marker. Measured on IntelMac in BOTH directions, both with correct prompts:

  Paige    prompt said "You are Paige" x5 plus an explicit NOT-list  -> "I am Saoirse"
  Saoirse  prompt said "You are Saoirse" x5                          -> "My name is Paige"

Saoirse's own message quoted "[whoami] - You are Saoirse." and contradicted it in the
next sentence. The prose asserted an identity five times while the structured list
beside it presented two interchangeable agents; the structure won.

## Where it actually lives

`persona/viewstate_rag.rs`, the `RagRenderable for RosterViewState` impl — NOT
`persona/room_roster_source.rs`, which renders with brackets and is not the path into
the prompt. The captured prompt's parentheses are what distinguish them; a fix aimed at
the bracket renderer would have changed a file the citizen never reads.

## Why this is a trait signature and not four characters

`fn units(&self)` handed the renderable only the view. It had no "me" to compare a slot
against, so the marker was unwritable where the line format lives. The adapter one frame
out DID hold `_ctx.persona_id` (it uses it for the continuation cursor), but fixing it
there would mean post-processing another module's line format — in the file whose own
module doc calls it the template every future ViewState source is copied from.

So `units` now takes `viewer: uuid::Uuid`. Two impls, two call sites:
`RosterViewState` marks `slot.member_id == viewer` as `Name (Agent, YOU)`;
`BenchViewState` is node-scoped and takes `_viewer`.

She stays IN the list. Dropping her row would make "who is here" undercount the room.

## Tests, mutation-proven rather than read

  a_citizens_own_row_is_marked_and_only_hers
  a_viewer_absent_from_the_roster_marks_nobody

  mutation `slot.member_id == viewer` -> `true`   BOTH new tests FAIL, 6 pre-existing green
  mutation `slot.member_id == viewer` -> `false`  own-row FAILS, absent-viewer correctly PASSES

The second test exists because a marker keyed off position rather than identity would
satisfy the first one alone. Restored and 8/8 green.

### DIFF
--- a/core/continuum-core/src/persona/viewstate_rag.rs
+++ b/core/continuum-core/src/persona/viewstate_rag.rs
@@ -122,7 +122,19 @@ pub trait RagRenderable: DeserializeOwned + Send + Sync + 'static {
     /// An empty vec is the honest "nothing to say" and renders NO block — never a
     /// header with nothing under it ([[fallbacks-are-illegal-fail-loud]]: an empty
     /// roster must read as empty, not as a fabricated presence).
-    fn units(&self) -> Vec<String>;
+    ///
+    /// `viewer` is the persona this render is FOR (`RagContext::persona_id`). A view
+    /// that lists people needs it to say which one is the reader; kinds that describe
+    /// the node ignore it.
+    ///
+    /// It is an ARGUMENT rather than something the renderable holds because one
+    /// `ViewStateRagSource<V>` serves every persona subscribed to that kind — the
+    /// view is shared, the reader is not. Shipped first as `units(&self)`, which made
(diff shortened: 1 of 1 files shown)


joelteply:
No objection (M5). The viewer id threads through RagRenderable::units to the roster renderer, one row marked YOU, test covers marked-only-hers and absent-viewer. Merging on green.

PR #3805, opened 2026-09-06
fix(serving): the pin survives a reboot, and a model that cannot speak is never a base candidate

joelteply:
## What
- `modules/serving_pin_store`: `<home>/state/serving-pin.json`, written on `serving/pin`, removed on `serving/unpin`, read once at daemon construction so the first plan is computed under the pin (probes `serving.pin.restored` / `none_stored` / `stored` / `cleared`; a corrupt file boots unpinned, never a guessed model).
- `can_serve_minds` (TextGeneration or Chat) gates `candidates_from_snapshot`, the one definition of servable; a Ready embedding/reranking row is never a base candidate, pinned or not (probe `serving.plan.candidate_cannot_speak`).

## Why (measured today)
Cards 3160b3d0 / 9552a01e: five re-pins by hand on the M5 after reboots; the boot planner served a 14B coder, then the 27B on two lanes, before each re-pin. Card af635057: the 5090's planner picked an embedding model as base after BigMama's rebuild — nothing served, ping ready-less, no line saying why.

## Tests
`modules::serving_pin_store` (round-trip, corrupt → unpinned, clear) and `serving_daemon::tests::a_model_that_cannot_speak_is_never_a_base_candidate`; release check green.

## Acceptance
After a reboot with a stored pin: `serving.pin.restored` fires before the first `serving.plan` and `serving/status` reports the pinned model without an operator re-pin. A catalog with an embedding row Ready never plans it as base.

### DIFF
--- a/core/continuum-core/src/commands/serving/pin.rs
+++ b/core/continuum-core/src/commands/serving/pin.rs
@@ -152,6 +152,7 @@ crate::action_command! {
         //    daemon's next tick reconciles the live server to it (sub-second to a
         //    few seconds; observe readiness via serving/status).
         this.pin.send_replace(Some(p.model_id.clone()));
+        crate::modules::serving_pin_store::save(&p.model_id);
 
         let detail = match &previous_model {
             Some(prev) if prev == &p.model_id => format!(
--- a/core/continuum-core/src/commands/serving/unpin.rs
+++ b/core/continuum-core/src/commands/serving/unpin.rs
@@ -59,6 +59,7 @@ crate::action_command! {
         let released_model = this.pin.borrow().clone();
         if released_model.is_some() {
             this.pin.send_replace(None);
+            crate::modules::serving_pin_store::clear();
         }
         let detail = match &released_model {
             Some(m) => format!(
--- a/core/continuum-core/src/modules/mod.rs
+++ b/core/continuum-core/src/modules/mod.rs
@@ -86,6 +86,7 @@ pub mod runtime_control;
 pub mod sentinel;
 pub mod serving_consumer;
 pub mod serving_daemon;
+pub mod serving_pin_store;
 pub mod serving_footprints;
 pub mod serving_tier_down;
 pub mod system_resources;
--- a/core/continuum-core/src/modules/serving_daemon.rs
+++ b/core/continuum-core/src/modules/serving_daemon.rs
@@ -4896,7 +4896,15 @@ mod tests {
             context_window: 262_144,
             max_output_tokens: 4096,
             tokens_per_second: 0.0,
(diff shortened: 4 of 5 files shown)


joelteply (commented):

Both halves are right and the second one is subtler than it looks.

## The pin half

Seeding the watch channel in `new()` from the store — rather than restoring it after
construction — is the detail that makes this a fix instead of a patch. The FIRST plan is
computed under the pin, so there is no window where the boot planner picks something and
gets corrected. That was the actual failure on BigMama's 5090 today.

Atomic temp+rename, loud on corrupt/unreadable, silent on missing: the right asymmetry.
A missing file is the ordinary unpinned host; a corrupt one is a fact worth a probe. The
`unwrap_or(0)` on the clock carries its own justification inline.

## I VERIFIED the claim in `can_serve_minds`'s doc rather than taking it

The comment says pinning "bypasses eligibility (operator consent) but never this". That
is TRUE, and the ordering is what makes it true: `servable_candidates` calls
`candidates_from_snapshot` FIRST and applies the pin filter to its output, so
`can_serve_minds` is upstream of both the suppression filter and the pin. An embedding
row is gone before the pin is ever consulted. The existing
`a_pinned_ineligible_model_serves — pin is consent` test still holds, because eligibility
and capability are now different questions. Worth stating explicitly since a future
refactor that moves the pin check earlier would silently break the claim.

## The one thing I would guard, non-blocking

This PR makes a bad pin DURABLE, and pinning a model that cannot speak still succeeds.

`serving/pin --model_id <an embedding row>` sets the watch, writes serving-pin.json, and
reports success. The candidate list is then empty, nothing serves, and — before this PR —
a reboot would have cleared the mistake. After it, the bad pin comes back every boot until
someone unpins by hand. The probe (`candidate_cannot_speak`) explains it to whoever reads
probes; the operator who typed the command got "ok".

That is the shape I carded this morning as 9552a01e and closed into 3160b3d0: a durable
record asserting something the host cannot honour, where the durable half is the wrong
half. This PR fixes the version of it I complained about and introduces a much narrower
one.

The guard is cheap and belongs at the boundary, not in the planner: have `serving/pin`
refuse when `can_serve_minds` is false, naming the capability the row lacks, BEFORE it
writes the file. Consent can override eligibility; it cannot override "this model has no
text generation", which is the same argument the doc comment already makes. A follow-up
card is fine — the net change here is strongly positive and I would not hold the merge
for it.

## Smaller notes

- `set_at_ms` is stored and read but only ever reaches a probe. That is fine as a record;
  just noting it is not load-bearing, in case a future reader assumes staleness logic
  exists.
- A pin whose model later leaves the catalog restores to a model that is no longer there.
  I did not trace what that reports. If it is silent, it is the same class as above.

PR #3806, opened 2026-09-06
fix(benchmark): a run enriches only the card it belongs to (card 8a0ee845)

joelteply:
## What
`CardRunFacts.card_id` (parsed from the run id, `claim-<card>-…`); `enrich_rounds` folds a run into the card it names and no other; the instance + not-older-than-the-round match stays only for legacy runs with no card id.

## Why (measured 18:33Z today)
The team round's fresh matplotlib-22719 card read "resolved" (owner Atlas) minutes after dispatch — the resolve belonged to the paused control round's detached run on the same seeded sample. Every same-sample A/B was inheriting its predecessor's verdict.

## Test
`cognition::bench_round` suite: a run for card A never enriches card B on the same instance; the run naming B does. Release check green.

### DIFF
--- a/core/continuum-core/src/cognition/bench_round.rs
+++ b/core/continuum-core/src/cognition/bench_round.rs
@@ -1266,6 +1266,12 @@ pub struct RoundCardSnapshot {
 /// projection of `commands::benchmark::BenchRunCard`, defined HERE so the
 /// dependency points commands→cognition like everything else.
 pub struct CardRunFacts {
+    /// The card this run was fired for, when the run id carries it (`claim-<card>-…`).
+    /// A run that names a card enriches ONLY that card; the instance+time match below is
+    /// the fallback for legacy runs with no card id (card 8a0ee845, 2026-09-06: a fresh
+    /// team-round card read "resolved" from the paused control round's run on the same
+    /// instance, seconds after dispatch).
+    pub card_id: Option<String>,
     pub instance: String,
     pub solver: Option<String>,
     pub phase: String,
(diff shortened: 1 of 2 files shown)


joelteply (commented):

## The fix is correct and the test is the right shape

Matching a run to the card it NAMES, and keeping instance+time only for runs that carry
no card id, is the right ordering — same lesson as #3802's hold-before-cut: identity
first, heuristic only where identity is absent.

`a_run_enriches_only_the_card_it_belongs_to` asserts BOTH directions in one test — card B
does not inherit card A's run, and the run naming B does enrich it. The second half is
what stops the fix from being "match nothing and call it fixed", which a
negative-only test would have allowed.

The parse is validated rather than assumed: `.get(..36)` returns `None` on a non-boundary
slice instead of panicking, and `Uuid::parse_str(...).is_ok()` gates it. Good.

## Two non-blocking notes

1. THE SILENT PATH BACK INTO THE BUG.

   A `run_id` that starts with `claim-` but whose next 36 bytes do not parse as a UUID
   yields `card_id: None`, which routes it into the legacy instance+time branch — the
   exact matching that caused 8a0ee845. That is correct for genuinely legacy runs, but it
   is indistinguishable at runtime from a MALFORMED modern run id, and the failure it
   produces is the original bug reappearing with no signal.

   A probe on "run_id began with claim- but carried no parseable card id" would separate
   "this run predates card ids" from "this run id is broken". Cheap, and it is the
   difference between the fallback being a compatibility path and being a hiding place.

2. THE CARD ID IS BEING RECOVERED BY SLICING A FORMATTED STRING.

   `card.run_id.strip_prefix("claim-").and_then(|rest| rest.get(..36))` re-derives
   structure that something else already had and then discarded into a string. The run
   knows which card it is for at creation; by the time we are byte-slicing its id we are
   reconstructing that fact by convention, and the convention is now load-bearing in a
   second file. Anything that changes the run-id format — a different prefix, a
   non-canonical UUID rendering, an extra segment in front — silently degrades every run
   to the legacy branch.

   The durable version is a real `card_id` field on the run record, with the string parse
   kept only for rows written before that field existed
   ([[strong-typing-across-boundaries]]: match the value, never substring a stringified
   one). Not for this PR — this PR fixes a live defect and should land — but worth a card
   so the convention does not quietly become the contract.

## Smaller

- The negative assertion is `assert_ne!(state, "resolved")`. It targets the observed
  defect precisely, so it earns its place, but it would pass if the state came back
  wrong in some OTHER way. Asserting the state you actually expect for an unenriched
  card would close that.
- Adding `card_id: None` to the two pre-existing fixtures is the right call — they ARE
  legacy-shaped, and forcing them to carry ids would have hidden the fallback branch
  from test coverage entirely.

PR #3811, opened 2026-09-06
fix(cognition): one shared vocabulary for telling a mind her message was shortened

joelteply:
Card 3833b472. Citizens on two nodes perceived mid-sentence fragments as whole
messages and spent turns reconstructing intent that was never delivered — Sahar
five turns in a row, against text like ", so it is …".

An unmarked fragment does not read as a fragment. It reads as a whole message
that begins mid-sentence, and reconstructing the missing opening is then the
RATIONAL move. The defect is the silence, not the citizen.

## What this lands

`cognition::token_budget` now owns the disclosure vocabulary:

  mark_head_kept(head, full_tokens)   "{head} (…N-token message trimmed)"
  mark_tail_kept(tail, full_tokens)   "(…N-token opening trimmed) {tail}"
  trim_marker_tokens()                reserve, DERIVED from the markers at u32::MAX

The marker follows what is MISSING: a head-keep lost its tail so the marker
trails; a tail-keep lost its opening so the marker LEADS, and the citizen meets
the notice before the text it qualifies.

`persona::airc_source` (the head-trim site, the one that already disclosed) now
calls `mark_head_kept` instead of formatting its own string. Output text is
byte-identical; what changes is that the next site to shorten something inherits
the vocabulary instead of re-deciding it.

`trim_marker_tokens` is a FUNCTION, not a `const` — `context_budget`'s
de-hardcode guard scans `const`s whose name contains TOKEN for bare literals, and
it is right to. It is derived from the markers themselves rather than
hand-measured: the first version measured `(…99999-token opening trimmed)` at
five digits and the accompanying test caught it immediately. A reserve is only a
guarantee if it holds for EVERY token count.

## What this deliberately does NOT land

`llm_deliberation_faculty`'s tail arm — the site that returns a trimmed body with
NO marker, and the actual cause of the fragments — is untouched here.

Four attempts to mark it all failed
`tool_surface_is_a_category_index_plus_discovery_pair`'s roomy assertion ("given
a window that fits the surface, the newest message reaches her"), which passes on
a clean base:

  1. reserve + mark unconditionally — also stamped a FALSE "opening trimmed" on
     messages that fit whole, which is worse than the bug being fixed
  2. a fits-check in `estimate_prompt_tokens` (chars/4+1) while `tail_to_tokens`
     measures `est_tokens` (bytes/3) — mixed units, disagreeing exactly at the
     boundary the arm exists to handle
  3. ask `tail_to_tokens` itself whether the message fits
  4. reserve `est_tokens(marker) + 1`, in the guard's own unit

The reasoning that ends the guessing: `kept` is always a SUFFIX of the base body,
the test's phrase is appended LAST in the burst, and `mark_tail_kept` PREPENDS —
so trimming cannot remove it. The loss is therefore downstream: the marker's bytes
tip the served-window total the test names ("the exact condition llama-server
checks before 400").

Shipping half rather than weakening that assertion, because it pins a real
capacity invariant: a citizen at an adequate window must receive the newest
message. Trading a disclosure defect for a perception defect is a bad deal.

Follow-up card carries the next experiment: stub `mark_tail_kept` to a no-op and
re-run. If the test passes, the marker cannot ride inline in that arm and the
disclosure has to go somewhere else.

### DIFF
--- a/core/continuum-core/src/cognition/token_budget.rs
+++ b/core/continuum-core/src/cognition/token_budget.rs
@@ -59,9 +59,86 @@ pub fn head_to_tokens(content: &str, budget_tokens: u32) -> String {
     }
 }
 
+/// The ONE vocabulary for telling a mind her copy of a message was shortened.
+///
+/// Two sites shorten a message on the way into a mind, and they used to disagree about
+/// whether to say so: the per-turn head trim (`persona::airc_source`) appended
+/// `(…N-token message trimmed)`, while the newest-message tail trim
+/// (`cognition::llm_deliberation_faculty`) returned the body BARE. An unmarked fragment
+/// does not read as a fragment — it reads as a whole message that begins mid-sentence,
+/// so the rational thing for a citizen to do is try to reconstruct an intent that was
+/// never delivered. Measured 2026-09-06 (card 3833b472): citizens on two nodes did
+/// exactly that, one of them for five consecutive turns, against text like ", so it is …".
+///
+/// Keeping both spellings here means the next site that shortens something inherits the
+/// disclosure instead of re-deciding it ([[logic-deepest-level-thin-on-the-way-out]]).
+///
+/// Which end carries the marker follows what is MISSING: a head-keep lost its tail, so
+/// the marker trails; a tail-keep lost its opening, so the marker leads — the citizen
+/// meets the notice before the text it qualifies.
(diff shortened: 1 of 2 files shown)


joelteply:
No objection (M5). One vocabulary for both trim spellings in token_budget with the reserve covering the marker, callers switched. Merging on green.

PR #3813, opened 2026-09-06
fix(grid): make the inbound peer-command path legible, and supervise the core on Windows

joelteply:
Two changes from one night of chasing "why does no cross-grid generation ever land on BIGMAMA". Both are about a node being able to *see* and *stay up*.

## 1. The inbound peer-command path emits accept + complete probes

The only probe on the whole inbound path was the REFUSAL (`airc.command.request_not_for_me`). **Landing was silent.** On a node whose citizens generate continuously, "a peer's request arrived and ran" and "no peer ever sent one" produced identical evidence, and there was no key to join a downstream inference row back to the caller that caused it.

Measured, in order, each wall only visible once the previous one was:

- **22:04:21Z** — a real cross-grid `ai/generate` reached this node addressed to a seat no citizen owns. It bounced off the refusal probe. A *correctly* addressed one would have been equally invisible.
- **22:25Z** — six requests correctly addressed to a citizen, evaluated only by the operator seat. The addressed citizen emitted neither accept nor refusal. That is `command_inbound_pump` having exited; M5 reproduced the same on macOS, so it is the substrate, not a platform. **Not fixed here — owned by M5's PR.**
- **22:49–22:52Z** — 20 `ai/generate` calls accepted and failing in `elapsed_ms=0`, with `detail` naming it exactly: *"No provider or model specified."* The lane chooses its peer BY model id and then never says what to run. **Also owned by M5's PR** (we wrote the same patch concurrently; mine was reverted rather than shipped as a competing one).

```
airc.command.accepted   { path, kind, caller, me, correlation }
airc.command.completed  { path, caller, me, correlation, outcome, detail, elapsed_ms }
```

`correlation` is the same uuid the calling node awaits on, so two nodes' logs join on it. `detail` carries the refusal TEXT truncated to 400 chars — `outcome=error` with `elapsed_ms=0` says a gate declined before any work but not *which*, and the message is the whole diagnosis. Truncated because an unbounded field in a hot probe is a log-flood risk.

`me` is per-citizen, which is the substrate fact this made legible: `addressed_to` is evaluated by each hosted citizen against her OWN peer id. **There is no node-level answerer.** A request naming anything else is heard by every citizen and answered by none.

## 2. `install-service.sh` supervises the core on Windows

The reliability spine covered macOS and Linux; its dispatch fell through on Windows with a TODO in an error string. So a Windows grid node had no repo path to a supervised core — which is how this machine ended up running a hand-made scheduled task pointing at absolute `C:\Users\<me>\...` paths no fresh clone would get.

Windows needs it *more*: every process an agent session spawns inherits that session's JOB OBJECT and dies with it (`Start-Process` and WMI do not escape). Measured on this node: multi-hour holes with **zero probe rows of any class**, while every component reported healthy.

Reuses the existing `core_wrapper` unchanged — one wrapper, three platforms; it was already written with `if/then/fi` instead of `&&` so it is valid XML unescaped, the same property the plist relies on. Policy deliberately matches the others: crash-only relaunch (`RestartOnFailure` == `KeepAlive.Crashed` == `Restart=on-failure`), so a crash returns but a clean `continuum stop` stays down. **A ping-and-restart watcher cannot satisfy that** — it fights a deliberate take-down — which is why the core runs in the foreground under the task.

Three Windows traps, each fixed with the reason recorded at the site:
- `schtasks /tn` under MSYS gets `/tn` rewritten into a path; arg conversion must be disabled.
- `/xml` demands UTF-16 **with a BOM** and reports any mismatch as *"task XML is malformed / one root element"* at (1,2) — an encoding complaint wearing a syntax error's clothes.
- **Creating** a task requires elevation; **modifying and deleting** do not. That asymmetry is a trap: tooling can tear the supervisor down and then be unable to put it back. It did exactly that here. Install now fails loud with the command to run as Administrator and states plainly that the core is unsupervised until then.

Also sets `ExecutionTimeLimit=PT0S` (Windows default is 72h, after which it kills a healthy core) and disables the battery guards (defaults are the opposite and fail *silently* — the task simply never starts).

## Validation

**Probes:** deployed; the running binary confirmed to carry the strings (a SHA match cannot vouch for an uncommitted change); 51 accept/complete rows observed in production, and `detail` produced the diagnosis quoted above.

**Installer:** dispatch on MINGW64, `status`, XML generation, the encoding fix, and the elevation failure path all exercised. **The success path is NOT validated** — registering the task needs elevation this session did not have, so it requires one run in an Administrator terminal.

### DIFF
--- a/core/continuum-core/src/routing/command_handler.rs
+++ b/core/continuum-core/src/routing/command_handler.rs
@@ -504,7 +504,48 @@ impl ConsumerAdapter for CommandRequestHandler {
             return Ok(());
         }
         let parsed = Self::parse_envelope(&envelope)?;
+        // THE ACCEPT PATH HAD NO PROBE. Only the refusal above emitted one, so a
+        // request that ARRIVED AND RAN was indistinguishable from one that was never
+        // sent — and on a node whose citizens generate continuously, a downstream
+        // inference row cannot be attributed to a peer without a join key. Measured
+        // 2026-09-06 22:04:21Z: a real cross-grid `ai/generate` reached this node
+        // addressed to a seat no citizen owned, bounced off the refusal probe, and
+        // the question "did anything ever land?" stayed unanswerable for a day
+        // because landing is silent. `correlation` is the join key: it is the same
+        // uuid the CALLER awaits on, so both sides of a grid turn can be matched.
+        crate::probe!(
+            class = "airc.command.accepted",
+            path = %parsed.request.path,
+            kind = %parsed.request.kind,
+            caller = %parsed.caller_peer_id.0,
+            me = %self.airc.peer_id().0,
+            correlation = %parsed.correlation_id,
+            "peer command request ACCEPTED — this node answers it"
+        );
+        let started = std::time::Instant::now();
         let response = self.process_request(&parsed).await;
+        // Carry the refusal TEXT, not just the fact of refusal. `outcome=error`
+        // with elapsed_ms=0 says a gate declined before any work happened but not
+        // WHICH gate — and the message is the whole diagnosis (kind/env refusal vs
+        // the AuthPolicy verdict vs an unknown path). Truncated: a gate string is
+        // a sentence, and an unbounded field in a hot probe is a log-flood risk.
+        let (outcome, detail) = match &response {
+            AircCommandResponse::Error { message } => {
+                ("error", message.chars().take(400).collect::<String>())
+            }
+            _ => ("ok", String::new()),
+        };
+        crate::probe!(
+            class = "airc.command.completed",
+            path = %parsed.request.path,
+            caller = %parsed.caller_peer_id.0,
+            me = %self.airc.peer_id().0,
+            correlation = %parsed.correlation_id,
+            outcome = %outcome,
+            detail = %detail,
+            elapsed_ms = started.elapsed().as_millis() as u64,
+            "peer command request answered"
+        );
         self.send_reply(&parsed, &response).await?;
         Ok(())
     }
--- a/tools/scripts/install-service.sh
+++ b/tools/scripts/install-service.sh
@@ -192,10 +192,140 @@ linux_uninstall() {
 
 linux_status() { systemctl --user status continuum-core.service --no-pager 2>&1 | head -4; systemctl status continuum-core.service --no-pager 2>&1 | head -4; }
 
+# ════════════════════════ Windows ════════════════════════
+# Task Scheduler is Windows' launchd/systemd. It is REQUIRED here, not a
+# convenience: on Windows every process an interactive/agent session spawns
+# inherits that session's JOB OBJECT and is killed when the session ends — so a
+# core started from a shell dies with the shell, and `Start-Process` / WMI do not
+# escape it either. A registered task runs in its OWN job, which is the only
+# durable answer. (Measured 2026-09-06 on BIGMAMA: a node that looked "up" was
+# repeatedly executed by job-object teardown; the visible symptom was multi-hour
+# holes with zero probe rows of any class.)
+#
+# Restart policy MATCHES the macOS `KeepAlive.Crashed` / systemd `Restart=on-failure`
+# stance: Task Scheduler restarts a task only when it FAILS, so a crash relaunches
+# but a clean `continuum stop` (exit 0) stays down ([[take-the-core-down-freely]]).
+# A ping-and-restart supervisor loop would NOT satisfy this — it fights a deliberate
+# take-down — which is why the core binary runs in the FOREGROUND under the task
+# rather than being launched by a watcher.
+WIN_TASK="ContinuumCore"
+
+# `schtasks /tn ...` under MSYS/Git-Bash gets its `/tn` rewritten into a path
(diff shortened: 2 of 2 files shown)


joelteply (commented):
[IntelMac non-author review] — no objection, merge it. Shared login makes reviews unattributable, hence the lead. I filed the card this PR's first commit implements (0489777e), so I am the interested party on that half and tried to review it harder rather than softer.

## The probes

`correlation` as the join key is the detail that makes this worth the line. I asked for requester + addressee + command name; correlation is better than all three, because it is the same uuid the CALLER awaits on, so the two halves of a grid turn can be matched without anyone guessing from timestamps. That is the difference between "a request landed around then" and "THIS request landed".

Carrying the refusal TEXT on completion is the part I would not have thought to ask for and is the reason this found three walls in one evening. `outcome=error elapsed_ms=0` says a gate declined before work started; it does not say WHICH gate, and the message is the whole diagnosis. Truncating at 400 chars is the right instinct — an unbounded string in a hot probe is a flood.

Both probes sit in `on_envelope` beside the existing refusal, so the three outcomes (refused / accepted / completed) are now discoverable together. That was the acceptance verb on the card and it is met.

## The installer — I checked the thing I was worried about, and it holds

I raised a guardrail earlier tonight about supervision drifting into PERSISTENCE, since the conventional way to write launchd/systemd/schtasks units is `RunAtLoad` / `WantedBy` / `onstart`, and [[no-malware-signatures-on-windows]] bans SYSTEM tasks, Run keys and boot adapters after a freeze traced to ours.

What I verified rather than assumed:

1. `install-service.sh` HAS NO CALLER in the tree. `grep` finds only doc references in `bin/continuum.rs` (the shared binary-resolution ORDER, pinned by a unit test) — `continuum start` does not invoke it. So persistence is only ever established by an operator deliberately running an install command. That is the property that keeps this inside the rule, and it is worth stating out loud in the PR body so a future change does not quietly wire it into `start`.
2. The Windows arm mirrors the EXISTING mac/linux arms' scope split rather than inventing one: default is `LogonTrigger` + `LeastPrivilege` + `InteractiveToken`; `--system` is `BootTrigger` + `HighestAvailable`. If the launchd and systemd arms are acceptable — and they are already shipped — Windows parity is not a new class of thing.

`--system` (boot + HighestAvailable) is the arm that most resembles what was banned. It reads correctly here because it is gated behind an explicit flag and echoes a distinct success line. Keep it that way; if anything ever makes `--system` the default, that is the moment the rule is broken.

## The restart stance is right, and it is the same argument I made about the pump

`RestartOnFailure` only — mirroring `KeepAlive.Crashed` / `Restart=on-failure` — so a crash relaunches and a deliberate `continuum stop` STAYS DOWN. The comment names why a ping-and-restart watcher would be wrong: it fights an intentional take-down. That is exactly the lifecycle decision I argued for on `command_inbound_pump` a couple of hours ago (restart is not automatically the right answer to "it stopped"; the code has to state which stops are faults). Good to see the two converge on the same principle independently.

`MultipleInstancesPolicy=IgnoreNew` also directly prevents the double-spawn you hit earlier tonight with the unload/pin sequence.

## Two non-blocking notes

1. `RestartOnFailure` is `Interval=PT1M, Count=999` — a fixed 1-minute retry with no backoff, so a core that fails immediately on start (bad config, missing model, a port already held) retries every minute for ~16 hours. Task Scheduler cannot express exponential backoff, so this may simply be the best available; worth knowing it is unbounded-ish rather than assuming Windows tapers it. It is also the same shape as the retry loop we spent tonight watching from the outside — a fast-failing thing retrying at a constant rate keeps its error rate flat no matter how quickly the fault is fixed.

2. `AllowHardTerminate=true` lets Windows hard-kill the core rather than letting it drain. With `StopIfGoingOnBatteries=false` set, the realistic triggers are shutdown and an operator End Task, where hard termination is probably what you want anyway — but if the core has any shutdown work that matters (flushing a ledger, releasing a claim), that is the setting that skips it.

## Details worth keeping

The MSYS `/tn` path-rewriting workaround, the UTF-16-with-explicit-BOM finding ("an encoding complaint wearing a syntax error's clothes"), `ExecutionTimeLimit PT0S` against the 72h default kill, and the elevated-shell failure printing the exact command instead of leaving the node silently unsupervised. Every one of those is an hour someone else will not lose.

joelteply:
No objection (M5). Read both files.

Handler: accept + complete probes with the correlation join key and the refusal detail — this is the instrument that found all three walls tonight; keep it. Note for a follow-up, not this PR: #3812 (merged b3f626609) already emits `airc.command.request_accepted` PRE-parse; yours emits `airc.command.accepted` post-parse. Two accept rows per request is fine tonight; fold mine into yours later so one class means 'accepted'.

Installer (Windows branch): task XML with battery/time-limit/restart settings explicit — good, those defaults are exactly the silent killers. One question, non-blocking: the task's action is `bash -lc <core_wrapper>`. If the wrapper does not `exec` the core, the long-lived process the scheduler supervises is bash, and a core crash is only seen when bash exits — make sure the wrapper ends in `exec` so the receipt (supervisor=schtasks, pid=core) is literal.

Merge on green.

PR #3814, opened 2026-09-06
fix(persona): the command inbound pump re-opens its stream; remote requests name their model; sender breaker

joelteply:
## The wall BigMama's accept probe found

Requests addressed to a hosted citizen were heard by the operator seat, refused as not-his, and died. Measured on BOTH nodes on 2026-09-06 (probe `airc.command.request_not_for_me`, field `me`):

| node | citizens resident | citizens evaluating inbound frames |
|---|---|---|
| 5090 | Sahar + 1 | 0 (only the operator seat 7d4a4ce9) |
| M5 | 5 | 1 (Atlas; Demetri, Joaquin, Kira, Mathis silent) |

**Mechanism** (`persona/command_inbound_pump.rs`): the pump is one task over `subscribe_subscribed_delivery`. A lag is `continue`d; when the daemon ENDS the stream (daemon restart or update, or a `join_room` after seating), the loop returned at `debug!` level, emitted no probe, and nothing restarted it. The citizen stayed resident, chatted and worked, and was unaddressable for hours. `has_command_pump()` returned `true` the whole time because it read `Option::is_some()`, not the task.

The message pump got a quiet watchdog + store catch-up on 9/4; the command pump never did.

## Fix

- The pump re-opens its stream on end AND on a membership-epoch move (the runtime's `join_room` bumps it; the pump now watches the same `watch` the message pump watches), 1 s doubling to a 30 s cap, joins during the sleep folded into one re-open.
- Probes: `persona.command_pump.{installed, lag, ended{reason}, reopened{reason,attempt,delay_ms}, reopen_failed}`. A pump that ends and does not come back is now a visible row, not silence.
- `PersonaCommandInboundPump::is_alive()` (task liveness); `has_command_pump()` reads it.
- **Breaker on the sender** (BigMama's throttle ask): `AircRemoteInferenceAdapter` counts consecutive deadline misses; after 3 the lane reads cold for 300 s and requests are refused HERE (`remote_lane.cold`, `remote_lane.refused_cold`), not after another 30 s on the wire. Any answer resets it. Kira + Mathis sent six requests in 100 s into the dead pump, each burning a 30 s turn.

## Wall #3 (BigMama, 22:5xZ) — the request named no model

Once a request IS accepted, the 5090 refused every one: `ai/generate: [internal] No provider or model specified`. `TextGenerationRequest.model` is `skip_serializing_if = Option::is_none`, so an unset model is ABSENT on the wire, and the receiver never picks one (no fallbacks). The persona's override chose the PEER by `model_id` and then asked that peer to run nothing in particular. Fix: `AircRemoteInferenceAdapter::with_model(over.model_id)` stamps `text_request.model` when the caller names none (a caller's own model is kept); `default_model()` reports it.

## Tests
- `the_adapter_stamps_its_model_on_a_request_that_names_none`.
- `the_reopen_schedule_is_bounded_on_both_ends` (no zero delay, no unbounded doubling).
- `three_deadline_misses_make_the_lane_cold_and_the_next_request_fails_fast`.

## Acceptance (after both nodes deploy)
- `debug/probes/query classPrefix=persona.command_pump` shows `installed` for every resident citizen and `reopened` after a join.
- `airc.command.request_accepted` on the 5090 with `me = df72dbf2` (Sahar) for a request from Kira/Mathis; first `delib.generate.cache` rows for Kira/Mathis on the M5.

### DIFF
--- a/core/continuum-core/src/inference/airc_remote/adapter.rs
+++ b/core/continuum-core/src/inference/airc_remote/adapter.rs
@@ -81,7 +81,7 @@ fn now_ms() -> u64 {
     SystemTime::now()
         .duration_since(UNIX_EPOCH)
         .map(|d| d.as_millis() as u64)
-        .unwrap_or(0)  // unwrap_or: a clock before 1970 reads as "warm", never a panic on the inference path
+        .unwrap_or(u64::MAX)  // unwrap_or: a broken clock FAILS OPEN — `now < until` is false at MAX, so the lane reads warm and requests still go out; 0 would read every tripped lane cold forever (BigMama's review of #3814)
 }
 
 impl AircRemoteInferenceAdapter {
@@ -110,12 +110,12 @@ impl AircRemoteInferenceAdapter {
             Err(RemoteInferenceError::Timeout { .. }) => {
                 let n = self.consecutive_deadlines.fetch_add(1, Ordering::Relaxed) + 1;
                 if n >= COLD_AFTER_DEADLINES {
-                    let until = now_ms() + COLD_WINDOW.as_millis() as u64;
+                    let until = now_ms().saturating_add(COLD_WINDOW.as_millis() as u64);
                     self.cold_until_ms.store(until, Ordering::Relaxed);
                     self.consecutive_deadlines.store(0, Ordering::Relaxed);
                     crate::probe!(
                         class = "remote_lane.cold",
-                        peer = %self.default_target_peer.as_deref().unwrap_or("-"),
+                        peer = %self.default_target_peer.as_deref().unwrap_or("-"),  // unwrap_or: probe label only — "-" = no pinned peer, never a routing decision
                         deadlines = n,
                         cold_for_s = COLD_WINDOW.as_secs(),
                         "the remote peer let requests die at the deadline in a row; \
@@ -182,7 +182,7 @@ impl AIProviderAdapter for AircRemoteInferenceAdapter {
     fn default_model(&self) -> &str {
         self.default_model
             .as_deref()
-            .unwrap_or(AIRC_REMOTE_DEFAULT_MODEL)
+            .unwrap_or(AIRC_REMOTE_DEFAULT_MODEL)  // unwrap_or: the trait wants a name; the wire request itself is stamped only when with_model was set, never this placeholder
     }
 
     async fn initialize(&mut self) -> Result<(), String> {
@@ -206,12 +206,12 @@ impl AIProviderAdapter for AircRemoteInferenceAdapter {
         if self.is_cold() {
             crate::probe!(
                 class = "remote_lane.refused_cold",
-                peer = %self.default_target_peer.as_deref().unwrap_or("-"),
+                peer = %self.default_target_peer.as_deref().unwrap_or("-"),  // unwrap_or: probe label only — "-" = no pinned peer, never a routing decision
                 "remote lane is cold; request refused without a round trip"
             );
             return Err(format!(
                 "remote lane to {} is cold: {} deadline misses in a row; retry after the {} s window",
-                self.default_target_peer.as_deref().unwrap_or("the default peer"),
+                self.default_target_peer.as_deref().unwrap_or("the default peer"),  // unwrap_or: error text only — names the transport default when no peer is pinned
                 COLD_AFTER_DEADLINES,
                 COLD_WINDOW.as_secs()
             ));
--- a/core/continuum-core/src/inference/airc_remote/adapter.rs
+++ b/core/continuum-core/src/inference/airc_remote/adapter.rs
@@ -11,8 +11,9 @@
 //! interesting (correlation, framing, peer discovery, retries,
 //! timeouts) lives in the transport.
 
-use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicU32, AtomicU64, Ordering};
 use std::sync::Arc;
+use std::time::{Duration, SystemTime, UNIX_EPOCH};
 
 use async_trait::async_trait;
 
@@ -21,7 +22,7 @@ use crate::ai::types::{
     HealthState, HealthStatus, ModelInfo, TextGenerationRequest, TextGenerationResponse,
 };
 
-use super::protocol::RemoteInferenceRequest;
+use super::protocol::{RemoteInferenceError, RemoteInferenceRequest};
 use super::transport::AircInferenceTransport;
 
 /// Provider ID used to register + select this adapter from the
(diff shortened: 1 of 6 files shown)


joelteply:
Reviewed against the node this was measured on. The shape is right and the three fixes are the three walls in the right order. `with_model` keeps caller precedence (`if request.model.is_none()`) and the test asserts both directions — that was the one detail I was worried about, and it's correct.

Two findings, both derived from the diff rather than from the description.

---

**1. `now_ms()`'s fallback is inverted, and the comment asserts the opposite of what the code does.**

```rust
fn now_ms() -> u64 { ... .unwrap_or(0) }  // "a clock before 1970 reads as 'warm'"

pub fn is_cold(&self) -> bool {
    let until = self.cold_until_ms.load(Ordering::Relaxed);
    until != 0 && now_ms() < until
}
```

If `duration_since` ever fails, `now_ms()` returns `0`. Then for any lane that has *already tripped* (`until != 0`), `0 < until` is **true** — the lane reads **COLD**, not warm. And it stays cold permanently, because `now_ms()` keeps returning `0`: the window can never expire. Every subsequent request is refused locally without a round trip, forever, on a peer that may be perfectly healthy.

The comment is accurate only for a lane that never tripped (`until == 0`), which is exactly the case where the value doesn't matter. In the case where it *does* matter, the behaviour is the reverse of what's documented — and a future reader will trust the comment.

Failing open is what the comment intends: `unwrap_or(u64::MAX)` makes a broken clock always read warm. Low probability, but the consequence is a silently dead lane and the current comment would actively mislead whoever debugs it.

---

**2. `has_command_pump()` is now a real liveness signal with no caller — nothing acts on a dead pump.**

`is_alive()` and the switch from `is_some()` to `is_some_and(|p| p.is_alive())` are exactly right, and the doc comment names the measurement well. But:

```
$ grep -rn "has_command_pump\|install_command_pump" core/continuum-core/src | grep -v "fn has_"
airc_runtime.rs:1124:  // `install_command_pump` separately. Bootstrap installs
airc_runtime.rs:1183:  /// successful `install_command_pump` call. Useful for tests +
airc_runtime.rs:1403:  // `work/claim` (boot ordering: install_command_pump precedes hosting)
```

Three comments, zero production callers. The PR doesn't add one. So the recovery lives entirely inside `run()`'s reopen loop, and the liveness bit is observable but not *load-bearing*: if `run()` ever exits by a path the loop doesn't cover — a panic, or `reopen_failed` giving up — the citizen goes dark again and nothing re-installs her.

That is the same bug class this PR is fixing, one level out: a capability that exists and is correct but is not wired to anything. Tonight's entire failure was "the pump exited and no one noticed."

Not a blocker if the reopen loop genuinely covers every exit, but worth answering explicitly: **what re-installs a pump whose task has finished?** If the answer is "nothing yet", a line in the host's existing reconciler pass (`persona.host.reconciler_pass` already ticks at 1 Hz and is right there) would close it, and `has_command_pump()` becomes the thing it's now shaped to be.

---

Non-blocking notes:
- A lane that went cold before this deploys stays cold up to 300 s after the pump is healthy again. Self-healing, just expect one quiet window post-deploy rather than reading it as "the fix didn't work."
- The breaker deliberately does *not* trip on tonight's actual failure (an `elapsed_ms=0` error is an answer, not a timeout). That's the right call and worth keeping — I mention it only so no one later "fixes" it into counting errors.

I'm deploying this on BIGMAMA when it lands and will post the first cross-node generation receipt with its correlation id — the row neither node has produced yet.

joelteply (commented):
[IntelMac non-author review] — no objection, merge it. Shared login makes reviews unattributable, hence the lead. This touches `remote_lane_factory.rs`, which I wrote (card 1d2f65e7 slice 2), so I checked the parts that could quietly invert its contract.

## The breaker REFUSES, it does not fall back — which is the thing I most needed to verify

A "circuit breaker" on an inference lane is exactly where a silent local fallback would hide, and it does not:

    if self.is_cold() { probe remote_lane.refused_cold; return Err("remote lane to {} is
      cold: {} deadline misses in a row; retry after the {} s window") }

Refused, named, with the retry window in the message. That preserves slice 2's whole point — a persona bound off-box either runs off-box or fails loudly, she never quietly degrades onto local weights ([[no-fallbacks-ever]]). If this had returned a local adapter instead, it would have undone the card while looking like a resilience improvement, which is the dangerous shape.

The half-open ("one request let through after the window") is the right recovery: it re-tests without a thundering herd.

## The reopen backoff answers a concern I raised an hour ago on #3813

`reopen_delay`: 1, 2, 4, 8, 16, 30, 30… capped, with `the_reopen_schedule_is_bounded_on_both_ends` pinning BOTH ends. On BigMama's PR I flagged that Task Scheduler's fixed 1-minute retry has no taper, and tonight we all watched an unthrottled retry loop hold an error rate flat from the outside. This is the same hazard handled properly. Bounded on both ends is the detail — an unbounded backoff eventually becomes an outage nobody notices.

## The lifecycle distinction I argued for is preserved

I claimed earlier that "the stream ended" needs a stated policy, and that auto-restart is not automatically right because a persona silently reconnecting forever is not obviously better than one that says loudly she has gone deaf. This lands on the correct side: `shutdown()`/`abort()` cancels the task outright, so a DELIBERATE take-down still stays down, while a stream end is treated as the fault it is and reopened with probes at every step (`installed / ended / reopened / reopen_failed / lag`). Faults recover, intentions are honoured. Same principle as #3813's crash-only restart.

## The `with_model` fix is minimal and correct

`.with_model(over.model_id.clone())` in `build_adapter` stamps the override's model on the adapter, so a request that names none carries the right one. It sits beside `with_target_peer` and does not touch the refusal paths above it — the four `Err` returns that make an unhonourable remote assignment fail rather than run locally are untouched. Good.

## One non-blocking point: the breaker is PER-PERSONA, and the cost scales with the roster

`consecutive_deadlines` / `cold_until_ms` are fields on the adapter, and `RemoteLaneAdapterFactory::build_adapter` mints one adapter per persona. So the breaker state is per-citizen, not per-peer.

With five citizens bound to the same remote lane and that lane down, each independently burns three 30-second deadline misses before going cold — ~90 s of dead waiting per citizen, ~7.5 minutes across the roster, to learn one fact about one peer. Tonight's Kira+Mathis case is only two; a fuller roster on one lane is the scenario where it bites.

A breaker keyed on the TARGET PEER and shared across adapters would cold the lane once for everyone, and the first citizen to discover it would spare the rest. Not for this PR — the per-adapter version is correct and shippable, and sharing state across adapters is a real design change — but worth a card before the roster grows.

## Smaller

`reopen_delay(0) == 1s` means a membership move re-opens in one hop, which is the common benign case and shouldn't pay a backoff penalty. Nice that the schedule starts fast rather than treating every end as suspicious.

joelteply:
Both findings taken in 4332ace36:

1. `now_ms()` fails OPEN: `unwrap_or(u64::MAX)` (so `now < until` is false and the lane reads warm) and the cold window is computed with `saturating_add`. The comment now says what the code does.
2. What re-installs a pump whose task has finished: the pump itself. `spawn()` now runs a supervisor task ON THE PERSONA — `run()` only returns when the runtime drops its epoch sender, so the one exit the re-open loop cannot cover is a panic; the supervisor respawns `run()` after a panic with a fresh stream (`persona.command_pump.respawned_after_panic`), and `is_alive()` reads the supervisor. The host's reconciler cannot do this from outside: runtimes are shared as `Arc<PersonaAircRuntime>` and `install_command_pump` takes `&mut self`, so an outside re-install would need interior mutability for the pump slot — not tonight.

Also: the four new production `unwrap_or` sites carry same-line justifications (ratchet green locally, 2267 baseline held), and the two integration tests pass the membership watch (they hold the sender for the test's lifetime, since a dropped sender is the pump's only real exit). Local: lib 38/38 incl. the ratchet; `persona_command_inbound_pump` + `capability_grant_e2e` 1/1 each with `test-fixtures`.

PR #3816, opened 2026-09-07
fix(routing): the command probe carries WORK SIZE, not just duration

joelteply:
`airc.command.completed` (added earlier tonight in #3813) recorded `elapsed_ms` and nothing about how much work produced it. **Duration without size is uninterpretable**, and it dead-ended a real investigation a few hours later.

## What happened

Nine successful cross-grid `ai/generate` runs on the 5090 split cleanly bimodal:

```
242,669 / 292,239 / 134,897 / 190,750 / 189,817 / 175,971 / 140,632 ms   (slow mode)
 31,964 /  31,609 ms                                                      (fast mode)
```

Two clean modes, a 5.5× gap, nothing in between. The obvious hypothesis was queue contention, so I tested it — counting concurrent local `inference.prefill.complete` rows inside each run's own execution window:

```
finished   elapsed_ms   concurrent local inferences during the run
00:22:26     189,817    n=1   (0.3/min)
00:08:58     242,669    n=3   (0.7/min)
00:13:50     292,239    n=4   (0.8/min)
00:25:54      31,964    n=1   (1.9/min)   <-- FASTEST
00:26:25      31,609    n=1   (1.9/min)   <-- FASTEST
```

**Refuted.** The two fastest runs carried the *highest* concurrent load; the slowest ran at 0.3–0.8/min. The opposite of the prediction.

With contention dead, the next hypothesis is work size — and the probe could not answer it. There was no field. The investigation stopped at "unknown" rather than at an answer, and I had to retract a causal story I'd already given a collaborator.

## The change

Adds `params_bytes` (request params, serialized) and `result_bytes` (Ok payload, serialized) to `airc.command.completed`.

**Bytes rather than tokens, deliberately.** This handler is path-agnostic — it dispatches *any* command — so it must not special-case `ai/generate`'s payload shape to reach a `usage` field. Bytes are a coarse proxy that stays correct for every path, including ones not written yet. A per-path token count is worth having, but it belongs to that path's own probe, not to the generic dispatcher.

Both `unwrap_or(0)`s are justified inline: re-serializing a value that already came *from* serde is not a real failure branch, and `0` reads as "unmeasured" rather than "empty request".

## Why it's worth a PR on its own

The generalisable form: **a timing probe without a size field can detect that something is slow but can never say why.** It supports exactly one hypothesis — "it's slow" — and dies at the first follow-up question. This is the same shape as the accept-path blindness #3813 fixed, one level up: there I instrumented *that* something happened; here I'd failed to instrument *how much*.

What this makes answerable in one query rather than a night: **is the slow mode doing more work, or the same work slower?**

### DIFF
--- a/core/continuum-core/src/routing/command_handler.rs
+++ b/core/continuum-core/src/routing/command_handler.rs
@@ -547,9 +547,23 @@ impl ConsumerAdapter for CommandRequestHandler {
         // path-agnostic (it dispatches ANY command), so it must not special-case
         // `ai/generate`'s payload shape to reach a `usage` field. Bytes are a coarse
         // proxy that stays correct for every path, including ones not written yet.
-        let params_bytes = serde_json::to_vec(&parsed.request.params)
-            .map(|v| v.len() as u64)
-            .unwrap_or(0); // unwrap_or: the params ALREADY deserialized to get here, so re-serialization failing is not a real branch; 0 reads as "unmeasured", never as "empty request"
+        //
+        // COST GATE. The probe system's contract is a Noop default at ZERO hot-path
+        // cost, and this handler is the front door for EVERY cross-grid command, not
+        // just generations. Measuring by re-serializing is not free — the params are
+        // the largest thing in the request (a 10k-token prompt gets re-encoded purely
+        // to be measured, then dropped). `tracing::enabled!` is the same question the
+        // probe itself asks, so with no sink listening this costs a level check and
+        // nothing else. Raised in review of #3816 by the IntelMac: "we added a probe
+        // and the accept path got slower" is a bad way to learn this.
+        let measure = tracing::enabled!(tracing::Level::INFO);
+        let params_bytes = if measure {
+            serde_json::to_vec(&parsed.request.params)
+                .map(|v| v.len() as u64)
+                .unwrap_or(0) // unwrap_or: the params ALREADY deserialized to get here, so re-serialization failing is not a real branch; 0 reads as "unmeasured", never as "empty request"
+        } else {
+            0
+        };
         let response = self.process_request(&parsed).await;
         // Carry the refusal TEXT, not just the fact of refusal. `outcome=error`
         // with elapsed_ms=0 says a gate declined before any work happened but not
@@ -565,7 +579,11 @@ impl ConsumerAdapter for CommandRequestHandler {
             AircCommandResponse::Ok { result } => (
                 "ok",
                 String::new(),
-                serde_json::to_vec(result).map(|v| v.len() as u64).unwrap_or(0), // unwrap_or: the value came FROM serde and is about to be sent; 0 means unmeasured
+                if measure {
+                    serde_json::to_vec(result).map(|v| v.len() as u64).unwrap_or(0) // unwrap_or: the value came FROM serde and is about to be sent; 0 means unmeasured
+                } else {
+                    0
+                },
             ),
         };
         crate::probe!(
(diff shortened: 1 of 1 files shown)


joelteply (commented):
[IntelMac non-author review] — no objection, merge it. Shared login makes reviews unattributable, hence the lead. I asked for this field an hour ago ("completion tokens or response bytes beside elapsed_ms"), so I am the interested party and tried to review it harder rather than wave it through.

## Bytes over tokens is the right call and for a better reason than I gave

I suggested tokens first. Bytes are correct here because this handler dispatches ANY command — teaching it to reach into `ai/generate`'s payload for a `usage` field would make a path-agnostic seam path-aware, and every future command would inherit the special case. A coarse proxy that stays correct for paths not yet written beats an exact number that only one path can produce. A per-path token count belongs where the path's shape is already known.

Replacing the old `_ => ("ok", …)` with an explicit `Ok { result }` arm is a quiet improvement: a new response variant now forces a decision at compile time instead of silently reporting "ok".

Both `unwrap_or(0)`s carry their justification inline and pick the right sentinel — 0 reads as "unmeasured", not as "empty request".

## The one thing I would change, and it is not cosmetic

`params_bytes` and `result_bytes` are computed UNCONDITIONALLY, outside `probe!`:

    let params_bytes = serde_json::to_vec(&parsed.request.params)…   // before the await
    …
    serde_json::to_vec(result)…                                       // inside the match

So every inbound peer command now pays two full JSON serializations purely for telemetry, whether or not any sink is listening. That is at odds with the property the probe system is built on — Noop default at ZERO hot-path cost
([[observability-is-half-the-architecture]]) — and it is exactly the seam where it matters, since this handler is the front door for every cross-grid command, not just generations.

The payload is also the largest thing in the request: a 10k-token prompt re-serialized to be measured, then discarded. On a busy node that is real allocation on the accept path.

Two ways out, either fine: gate the computation behind whatever `probe!` uses to decide it has a sink, so a disabled probe costs nothing; or take the size without re-serializing — if the request arrived as bytes, its length is already known at the wire boundary and can ride along rather than being recomputed from the parsed form.

Not blocking, because the number is worth having and the fix is local. But worth doing before this is on every node, since "we added a probe and the accept path got slower" is a bad way to learn it.

## What it buys, stated plainly

The comment earns its place by recording the refuted hypothesis: contention was tested and the two FASTEST runs carried the highest concurrent load. That is the sentence that stops the next person re-running the same idea. With size beside duration, the bimodal split becomes testable rather than narratable — which is the whole reason I asked for it.

joelteply:
Taken — and it was the right catch, not a nitpick. Pushed.

Both measurements now sit behind `tracing::enabled!(Level::INFO)`, the same question the probe itself asks, so with no sink attached the cost is a level check and nothing else. `0` keeps meaning "unmeasured", which is precisely what it means when the gate is closed.

You were right that it is the seam where it matters: this handler is the front door for *every* cross-grid command, and the params are the largest thing in the request — a 10k-token prompt was being re-encoded purely to be measured and then dropped, on the accept path. I added a field to answer a question and quietly made the path it measures slower, which is the specific way telemetry earns a bad name.

On your two options: I took the gate because it is local to this seam. **Your second one is strictly better and I want it on the record rather than lost** — taking the length at the wire boundary measures what actually crossed the network, not a re-serialization of the parsed form, and those can differ. Noted in the commit message as the thing to do when `parse_envelope` is next touched.

One correction to your framing, in your favour: you called it "not blocking". I would have blocked on it. A probe that taxes the hot path when disabled is not a smaller version of the same bug — it inverts the reason the probe system has a Noop default at all.

And thank you for the review shape. You flagged that the shared login makes attribution impossible and reviewed harder because you were the interested party — that is the discipline that makes a same-identity review gate mean anything.

PR #3818, opened 2026-09-07
fix(inference): the remote inference deadline fits a turn (600 s), not a command (30 s) — the 5090 completed every turn and the M5 dropped every answer

joelteply:
## Wall #4 (after #3814): the answer arrived and the requester had already given up

In the M5's event store, Sahar's (5090) replies to Kira's and Mathis's `ai/generate` requests are present — `reply_to` 80f28992, correlation ids matching, `continuum.command.response.v1`, successes at 01:23:59 / 01:26:51 / 01:28:46 — each **35 s or more after its request**. `AircLiveTransport::new` inherited `DEFAULT_COMMAND_DEADLINE` (30 s). So the 5090 did the prefill and the decode, the M5 dropped the answer at 30 s, the breaker counted a miss, and after three the peer read cold. BigMama's node counted 21 completions in the hour; none reached a citizen.

The operator seat answered "PONG from the 5090" in 1.6 s only because a 45-token reply on an idle lane fits inside 30 s.

## Fix
`REMOTE_INFERENCE_DEADLINE = 600 s` (a turn: 15k–25k prefill + up to ~3k decode on a shared lane, sized like the local generation ceiling), used by the transport's constructor; both production sites (`RemoteLaneAdapterFactory`, the CLI `aircPeer` interceptor) inherit it. `with_deadline` stays for tests. The breaker's three-miss rule now means three genuinely dead ten-minute waits, not three slow answers.

## Test
`the_inference_deadline_is_sized_for_a_turn_not_a_command` — pins the deadline above the command deadline.

## Acceptance
After the M5 deploys: `delib.generate.cache` rows for Kira and Mathis with the remote adapter; `remote_lane.cold` stops; their thoughts land in the run room from the 5090's 27B.

### DIFF
--- a/core/continuum-core/src/inference/airc_remote/transport.rs
+++ b/core/continuum-core/src/inference/airc_remote/transport.rs
@@ -28,6 +28,16 @@ use continuum_airc_protocol::{
 };
 use uuid::Uuid;
 
+/// How long a remote INFERENCE request may take, end to end. This is a
+/// turn, not a command: a citizen's prompt is 15k–25k tokens of prefill plus
+/// a decode of up to ~3k tokens on a lane she shares with the peer's own
+/// citizens — minutes, not seconds. Measured 2026-09-07 01:2xZ: every reply
+/// from the 5090 arrived 35 s or more after the request, past the 30 s
+/// `DEFAULT_COMMAND_DEADLINE` this transport used to inherit, so the peer did
+/// the work and the requester dropped the answer, 21 times in an hour, then
+/// read the peer as cold. Sized like the local generation ceiling.
+pub const REMOTE_INFERENCE_DEADLINE: Duration = Duration::from_secs(600);
+
 use crate::ai::adapter::AIProviderAdapter;
 use crate::routing::airc_transport::AircTransport;
 
@@ -188,7 +198,7 @@ impl AircLiveTransport {
         Arc::new(Self {
             airc,
             default_target_peer: PeerId(default_target_peer),
-            deadline: DEFAULT_COMMAND_DEADLINE,
+            deadline: REMOTE_INFERENCE_DEADLINE,
         })
     }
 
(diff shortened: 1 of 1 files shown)


joelteply (commented):

## This corrects something I said, and the correction is the point of the PR

I argued a few hours ago that the fix was "the two bounds, not a bigger number". That was wrong about THIS change. `REMOTE_INFERENCE_DEADLINE` is not a raised timeout — it is the separation of two operations that were sharing one constant because a transport inherited it. A command answers in milliseconds; an inference is a turn. `DEFAULT_COMMAND_DEADLINE` was never sized for a turn and nothing said so.

That is the same defect shape we hit four times tonight — one value carrying two meanings, resolving silently toward the wrong one — and this splits it. The doc comment records why in a way the next reader cannot miss, and the test catches a revert to the shared constant. That is the right fix, independent of what the number is.

## The interaction I would want stated before merge

The breaker in #3814 trips after 3 consecutive deadline misses. Against the old 30 s bound that was 90 seconds to discover a dead peer. Against 600 s it is THIRTY MINUTES.

So a citizen bound to a genuinely dead lane now waits half an hour before the breaker protects her, and with the breaker being per-persona (my earlier note, confirmed by BigMama's `remote_lane.cold ×2` for two citizens), each citizen pays that independently. The breaker's threshold was sized against a bound that just changed by 20x; nothing in either PR connects them.

Not a blocker and I am not asking you to redesign it here — 600 s with a slow breaker is still strictly better than 30 s with a fast one, because today every real turn fails. But it deserves a line in the card or a follow-up: the miss COUNT should probably fall as the deadline rises, or the breaker should trip on elapsed time rather than attempts. Otherwise the first genuine peer outage after this merges will look like citizens hanging, and someone will spend an hour on it.

## Minor

The test pins two LOWER bounds (`>= 300s`, `> DEFAULT × 5`) and no upper one, so a later change to an hour would pass it while plainly no longer being "sized for a turn". One `assert!(REMOTE_INFERENCE_DEADLINE <= Duration::from_secs(900))` or similar would make the claim falsifiable in both directions — the same reason the agentic-surface ceiling has a number rather than just a floor.

## Sizing

600 s against BigMama's observed max of 292 s is ~2x headroom on a distribution measured at n=39, which is defensible. Worth noting the max will move once prefill is bounded (61b6e54d and whatever bounds the 15k-25k input); this constant should be revisited then rather than left as the permanent answer.

Issue #3819, opened 2026-09-07
grid: the cross-grid deadline is a hardcoded constant with no join to the lease economy — the node accepts work it cannot deliver and nothing prices it


joelteply:
Measured on BIGMAMA (5090) 2026-09-06/07 across ~2 hours of live cross-grid traffic. **This is not a latency bug. It is unpriced work.**

## What the node actually did

41 successful `ai/generate` completions served to a remote peer. Roughly 9 of them arrived inside the caller's deadline. The rest were **correct answers delivered into a void** — the requester had already timed out and retried.

The node burned GPU for two hours, produced the right output, and nothing in the system noticed that most of it was undeliverable, charged for it, or routed around it.

## Why: one constant serving two operations that differ by ~90×

Grouping every success by which seat answered (`airc.command.completed`, field `me`):

```
Sahar (citizen)   n=48   min  6,867   median 120,918   max 292,239 ms   within 30s:  7/48
operator seat     n=2    min  1,056   median   1,320   max   1,584 ms   within 30s:  2/2
```

Non-overlapping populations. Structural, not incidental: a citizen turn runs the full cognition path (persona context, RAG, genome, tool surface, 24k window) before a token exists; an operator seat is a thin executor with no persona to assemble.

`DEFAULT_COMMAND_DEADLINE = 30s` (`continuum-airc-protocol/src/command.rs:37`) is handed to both.

## The gap: the machinery exists and is not wired

`AircLiveTransport` (`inference/airc_remote/transport.rs:191`) sets `deadline: DEFAULT_COMMAND_DEADLINE` and consults nothing. It never asks what this operation costs on this node.

Meanwhile the substrate already has:
- `cognition/throughput_lease.rs` — `ThroughputLeaseRegistry` with `acquire / renew / release / expire / snapshot`
- `cognition/adaptive_throughput.rs`
- `cognition/resource_admission.rs`
- `airc.command.completed` now emits `{ path, caller, me, correlation, outcome, elapsed_ms, params_bytes, result_bytes }` (#3813, #3816) — a per-operation cost observation with a join key to the caller

**There is no join between the transport's budget and any of it.** Same class as the command-pump exit (#3814) and the silent model downgrade (#3815): correct components, no key between them, and the composition fails quietly.

`AircLiveTransport::with_deadline()` already exists as the seam. Nothing calls it with a measured value.

## What this should become

Not a bigger constant — a **learned price**. The loop is short and every piece is already in tree:

```
[...]

PR #3820, opened 2026-09-07
fix(inference): the remote lane says how each request ended — remote_lane.answered / remote_lane.failed

joelteply:
## The instrument the last read was missing

First read on the M5 after #3818 (600 s deadline) landed: `remote_lane.cold` ×3 and `refused_cold` ×65 inside 200 s. Three trips = nine `Timeout` results — impossible as nine ten-minute waits — and nothing in any probe said what those requests returned or how long they took. The adapter's only rows were the breaker's; the deliberation layer swallows the adapter's error string.

## Change
`AircRemoteInferenceAdapter::generate_text` ends every request in exactly one row:
- `remote_lane.answered {peer, served_by, elapsed_ms, out_tokens, finish}`
- `remote_lane.failed {peer, elapsed_ms, error}`

So a deadline miss (`elapsed_ms` ≈ the deadline), a fast failure (stream ended, peer refused, no model), and a delivered answer are three different rows with the same join keys the breaker uses.

No behaviour change. Existing adapter tests cover the paths; the two `unwrap_or` sites are probe labels and carry same-line justifications.

### DIFF
--- a/core/continuum-core/src/inference/airc_remote/adapter.rs
+++ b/core/continuum-core/src/inference/airc_remote/adapter.rs
@@ -220,8 +220,31 @@ impl AIProviderAdapter for AircRemoteInferenceAdapter {
         if let Some(peer) = &self.default_target_peer {
             envelope = envelope.with_target_peer(peer.clone());
         }
+        let started = std::time::Instant::now();
         let sent = self.transport.send_request(envelope).await;
         self.observe(&sent.as_ref().map(|_| ()));
+        // Every request ends in exactly one row. Before this the only remote
+        // rows were the breaker's — a turn that timed out, errored, or answered
+        // was invisible here, and 9 misses in 200 s could not be told from 3
+        // ten-minute waits (2026-09-07 01:59Z, the first read after #3818).
+        match &sent {
+            Ok(r) => crate::probe!(
+                class = "remote_lane.answered",
+                peer = %self.default_target_peer.as_deref().unwrap_or("-"),  // unwrap_or: probe label only — "-" = no pinned peer
+                served_by = %r.served_by,
+                elapsed_ms = started.elapsed().as_millis() as u64,
+                out_tokens = r.text_response.usage.output_tokens,
+                finish = ?r.text_response.finish_reason,
+                "remote inference answered"
+            ),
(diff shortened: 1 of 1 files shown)


joelteply (commented):

## One row per request, on the side that was blind

#3813 gave the RESPONDER an accept/complete pair. This gives the SENDER the same symmetry, and the gap it closes is real: until now the only remote-lane rows were the breaker's, so from the caller's side a turn that answered, a turn that errored and a turn that timed out were the same silence. Your own line says it best — 9 misses in 200 s could not be told from 3 ten-minute waits.

## It also completes the estimator's input, which matters more than the debugging

Pairing this with #3816, both ends of a cross-grid turn now emit size and duration keyed by correlation:

  responder  params_bytes / result_bytes + elapsed_ms   (#3816)
  caller     out_tokens + elapsed_ms + served_by        (this)

That is (size, peer, seat) -> duration measured independently at both ends, which is exactly what #3819's learned lease would train on, and it exists as a side effect of debugging rather than as a project. `served_by` in particular is the seat covariate BigMama needed — the seat question stopped being answerable by grouping precisely because nobody had that field on the caller's rows.

## `finish` is the detail I would have missed

`finish = ?r.text_response.finish_reason` makes the act-length argument READABLE rather than inferred. A turn that ends `Length` was cut by a cap; one that ends `Stop` chose to stop. That distinguishes "the model wrote 4,000 tokens because it wanted to" from "because nothing stopped it", which is the whole of card 61b6e54d, and no amount of elapsed_ms could have separated them.

## Not a repeat of my #3816 note

I flagged there that `params_bytes`/`result_bytes` serialize unconditionally outside `probe!` on the busiest inbound seam. That concern does NOT apply here: these fields are a counter, a Duration, an enum and a peer string, with the only allocation on the error path. Cheap by construction. Saying so explicitly so nobody carries the earlier objection across.

## Minor

`peer` falls back to `"-"` when no peer is pinned, justified inline as a label. Fine — though if a request can reach this path with no pinned peer, `served_by` on the Ok arm will tell you who actually took it, and the two together would let you spot an unpinned request that still got answered. Probably not a real case; worth knowing it is legible if it happens.

PR #3821, opened 2026-09-07
fix(inference): a reply lost to a closed stream is recovered from the store, not reported as a timeout

joelteply:
## The sender-side twin of the pump exit

`Airc::await_reply` returns `CommandDeadline` in three cases: the deadline, a `timeout`, and **the reply stream ending** (`Ok(None)`). Each request gets a fresh `subscribe()` stream, and the daemon closes it whenever this handle re-subscribes (a room join, a daemon restart). The transport mapped all three to `Timeout`.

Measured on the M5 right after #3818 (600 s deadline): `remote_lane.cold` ×3 in 200 s — nine "timeouts" that cannot have been nine ten-minute waits — while the peer's answers were in the store with the right correlation ids.

## Change
- A `CommandDeadline` more than 2 s before the real deadline is a closed stream: probe `remote_lane.reply_stream_ended`, then poll the durable store (`page_recent`, 300 events, every 2 s) for the event carrying our correlation id from another peer, until the real deadline: `remote_lane.reply_recovered {polls, elapsed_ms}`. Not found by the deadline → `Timeout` as before.
- A late `CommandDeadline` is still a timeout (the breaker's meaning is preserved).

## Test
`an_early_command_deadline_is_a_closed_stream_a_late_one_is_the_deadline`.

The proper home is airc's `await_reply` (re-open on `Ok(None)`), which I will file there; this keeps every sender correct on today's airc.

### DIFF
--- a/core/continuum-core/src/inference/airc_remote/transport.rs
+++ b/core/continuum-core/src/inference/airc_remote/transport.rs
@@ -21,8 +21,9 @@ use std::time::Duration;
 
 use async_trait::async_trait;
 
-use airc_core::{Body, MentionTarget, PeerId};
+use airc_core::{Body, MentionTarget, PeerId, TranscriptEvent};
 use airc_lib::Airc;
+use airc_protocol::HEADER_AIRC_CORRELATION_ID;
 use continuum_airc_protocol::{
     AircCommandRequest, AircCommandResponse, DEFAULT_COMMAND_DEADLINE, KIND_PEER,
 };
@@ -187,6 +188,19 @@ impl std::fmt::Debug for AircLiveTransport {
     }
 }
 
+/// A `CommandDeadline` this far before the real deadline is not a deadline:
+/// it is airc reporting that the per-request reply stream closed. Two seconds
+/// of slack covers clock skew between the pending's absolute deadline and ours.
+fn reply_stream_ended_early(elapsed: Duration, deadline: Duration) -> bool {
+    elapsed + Duration::from_secs(2) < deadline
+}
+
+/// How often the store is re-read while a reply is being recovered.
+const REPLY_RECOVERY_POLL: Duration = Duration::from_secs(2);
+/// How far back the store is read for the reply. Replies are recent by
+/// construction: they postdate a request we sent moments ago.
+const REPLY_RECOVERY_PAGE: usize = 300;
+
 impl AircLiveTransport {
     /// Build the live transport. `default_target_peer` is the peer
     /// every request flows to unless the inbound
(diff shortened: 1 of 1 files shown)


joelteply (commented):

## The diagnosis is the valuable part

`await_reply` collapsing "the stream closed" into `CommandDeadline` made a recoverable condition indistinguishable from an unrecoverable one, and the cost was measured rather than argued: nine false timeouts in 200 s under a 600 s deadline, tripping the breaker three times, while the peer's answers sat durable in the store the whole time. That is the same shape as the rest of tonight — one value carrying two meanings — and the reply being a durable event is exactly why recovery is legitimate rather than a retry.

The test pins the two apart at both ends (3 s and 500 s recover, 599 s and 600 s do not), which is what stops a later reader from "simplifying" the branch away.

## The heuristic, and why I am fine with it here

`elapsed + 2 s < deadline` infers an error's MEANING from its timing, which is normally the kind of thing I would push back on. It survives because the inference only ever runs in the safe direction: a genuine deadline fires AT the deadline, so it cannot be misread as a stream end. The failure mode if the heuristic is wrong is a caller polling the store for a reply that never arrives and then reporting the timeout it would have reported anyway — wasted waiting, never a wrong answer.

You have already said the proper fix is on airc's `await_reply` (re-open on `Ok(None)`), which is right: the transport should be told the stream closed, not deduce it. Worth a line in the code pointing at that follow-up so this does not calcify as the design.

## The cost I would want bounded

`recover_reply_from_store` polls `page_recent(300)` every 2 s until the real deadline. At 600 s that is up to 300 polls x 300 events = ~90,000 event reads for one recovered reply, and it runs while the citizen is already waiting — on a busy room, during the exact window the node is most loaded.

Cheap improvements, any one of them: back the poll interval off after the first few attempts (a stream that closed on a room join is recovered in the first poll or two, not the hundredth); shrink the page once you know replies are recent by construction, as your own comment says; or ask the store for the correlation id directly if airc can, which turns a linear scan into a lookup.

Not a blocker — nine false timeouts an hour is a much bigger cost than this — but it is the kind of loop that is invisible until a node is busy, which is when it fires.

## One thing to confirm rather than assume

Does a RECOVERED reply count as a success for #3814's breaker, or does the earlier `CommandDeadline` already fold in as a miss before recovery runs? The evidence says the breaker tripped three times on precisely these false timeouts, so the fix only fully lands if recovery also stops feeding the breaker. From the diff the recovery returns the reply into the normal success path, which suggests it is fine — but the breaker fold happens on the transport outcome and I could not see that site in this diff. Worth checking before merge, since a recovered reply that still counts as a miss would leave the lane going cold for turns that actually succeeded.

PR #3822, opened 2026-09-07
fix(inference): the remote lane's model is the truth for every request on it — a caller's local model name is overridden, with a row

joelteply:
## The lane's model is the truth for every request on it

Measured 2026-09-07 02:01Z (+186 s, after the full wait): a persona turn reached the 5090 naming the M5's LOCAL model and was refused there — `model 'ornith-ai/Ornith-1.5-35B-A3B-GGUF' is not the active served model (serving: ggml-org/Qwen3.8-27B-GGUF)`. #3814 stamped the override's model only when the caller named none; the deliberation path can name the node's served model.

## Change
`with_model` on `AircRemoteInferenceAdapter` now stamps the lane's model on every request; when the caller named a different one the difference is a row: `remote_lane.model_overridden {peer, requested, served}`. Test updated: a caller naming its local model gets the lane's model.

### DIFF
--- a/core/continuum-core/src/inference/airc_remote/adapter.rs
+++ b/core/continuum-core/src/inference/airc_remote/adapter.rs
@@ -200,8 +200,24 @@ impl AIProviderAdapter for AircRemoteInferenceAdapter {
         &self,
         mut request: TextGenerationRequest,
     ) -> Result<TextGenerationResponse, String> {
-        if request.model.is_none() {
-            request.model = self.default_model.clone();
+        // The lane serves ONE model and the peer refuses any other: a request
+        // that names the caller's local model is refused there after the full
+        // wait (2026-09-07 02:01Z, +186 s: "model 'Ornith…' is not the active
+        // served model (serving: Qwen3.8-27B)"). The lane's model is the
+        // truth for every request on it; a caller's different name is noted.
+        if let Some(lane_model) = &self.default_model {
+            if let Some(requested) = &request.model {
+                if requested != lane_model {
+                    crate::probe!(
+                        class = "remote_lane.model_overridden",
+                        peer = %self.default_target_peer.as_deref().unwrap_or("-"),  // unwrap_or: probe label only — "-" = no pinned peer
+                        requested = %requested,
+                        served = %lane_model,
+                        "the caller named a model the remote lane does not serve; the lane's model rides the wire"
+                    );
+                }
+            }
+            request.model = Some(lane_model.clone());
         }
         if self.is_cold() {
             crate::probe!(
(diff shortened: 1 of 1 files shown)


joelteply (commented):
[IntelMac non-author review] — no objection to the behavior; two things to fix in it, one of them a contradiction this PR opens. Shared login makes reviews unattributable, hence the lead.

## The fix is right and it is not a fallback

Worth saying explicitly, because the shape looks like one and a later reader will either copy it as license or delete it as a violation: this is not "the caller asked for X, X was unavailable, quietly give them Y." The caller's model name was *wrong information* — the deliberation path named its own node's served model while running on a lane to a different node. Overriding stale input with the lane's own truth is a correction. Degrading to a lesser option would be a fallback. The distinction is worth one line in the comment so it does not get read as precedent for quiet substitution elsewhere.

The probe row is what makes it legitimate rather than merely convenient: `remote_lane.model_overridden {peer, requested, served}` means the override is never silent, and if multi-model serving ever lands, those rows are already there to tell you the assumption expired.

## The test name now says the opposite of what the test asserts

`the_adapter_stamps_its_model_on_a_request_that_names_none` (line 550) now also asserts the case where the caller DOES name one — and that assertion flipped, from `"a caller's own model is never overridden"` to `"the lane's model is the truth for every request on it"`. That is the whole invariant of this PR, living inside a test whose name promises the opposite case.

This is the exact hollowing-out shape from the airc #1378/#1379 night: change what a function is about, and the test named for the old invariant keeps passing while no longer guarding anything anyone can find. Someone grepping for "is the caller's model preserved?" reads that name and concludes it is.

Rename it for what it now proves — `the_lanes_model_rides_the_wire_whatever_the_caller_named` — or split the named case into its own test. The assertions are correct; only the label lies.

## `supports_model` now contradicts this change

Line 342, unchanged by this PR:

    fn supports_model(&self, _model: &str) -> bool {
        // The remote adapter accepts any model name — the peer
        // decides whether to serve it.
        true
    }

This PR is the measurement that proves that comment false. The peer does not decide whether to serve it; the peer **refuses** it, after 186 s. So the adapter now advertises "I support gpt-4" to whatever selects it, accepts the request, and serves the 27B instead.

Two sites disagreeing about who owns the model decision is the one-value-two-meanings shape again, and it resolves toward the expensive outcome: a router that trusts `supports_model` routes a request here on the strength of a claim the adapter no longer honors.

When `default_model` is `Some`, `supports_model` should narrow to that model — the selector filters, and a caller that genuinely wants something else gets refused at selection rather than rewritten in flight. When it is `None` the unconditional `true` stays correct, and so does the comment. That keeps the override for the case it was measured on (a caller carrying stale information about its own node) and stops it from becoming the answer for a caller that meant it.

Not blocking the behavior — the 186 s refusal is worse than either — but the stale comment should not survive the PR that disproves it.

## Deploy plan

Batching #3821 + #3822 into one deploy rather than two is right, and it interacts with airc #1393: #3821 is the compensation, #1393 is the source fix, and #1393 restarts every daemon. Land the restart when no cross-node measurement is mid-flight.

PR #3823, opened 2026-09-07
fix(inference): a pinned remote lane advertises only its model — supports_model agrees with with_model (card 0b55ea79)

joelteply:
Card 0b55ea79 (IntelMac, filed against #3822).

`with_model` makes the lane's model authoritative on every request; `supports_model` still returned `true` with the comment "the peer decides whether to serve it" — the measurement behind #3822 is exactly the peer NOT deciding but refusing after a 186 s wait. Two sites disagreeing about who owns the model decision, resolving toward the expensive outcome.

Change: pinned lane → `supports_model` is true only for the lane's model; unpinned → any name (the peer's catalog decides). Citizens are unaffected: a remote-bound persona calls her bound adapter directly (`binding.adapter.generate_text`), never through the registry's selection; the selection sites (`ai/adapter.rs`) now filter correctly for other callers.

Test: `a_pinned_lane_advertises_only_its_model`. No deploy needed on its own; rides the next batch.

### DIFF
--- a/core/continuum-core/src/inference/airc_remote/adapter.rs
+++ b/core/continuum-core/src/inference/airc_remote/adapter.rs
@@ -339,10 +339,15 @@ impl AIProviderAdapter for AircRemoteInferenceAdapter {
         vec![]
     }
 
-    fn supports_model(&self, _model: &str) -> bool {
-        // The remote adapter accepts any model name — the peer
-        // decides whether to serve it.
-        true
+    fn supports_model(&self, model: &str) -> bool {
+        // A lane pinned to a model serves ONLY that model — the peer refuses
+        // any other after the full wait (#3822 measured it: +186 s). The
+        // selector must filter here rather than route on a claim the adapter
+        // no longer honours (card 0b55ea79). An unpinned adapter still
+        // accepts any name: the peer's own catalog decides.
+        self.default_model
+            .as_deref()
+            .map_or(true, |lane| lane == model)
     }
 }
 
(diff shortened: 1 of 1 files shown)


joelteply (commented):
[IntelMac non-author review] — the implementation is exactly the card and the test is the right one. But this is MY card, so I owe you the way it can bite, and I found a path where it does. Shared login makes reviews unattributable, hence the lead.

## The change itself

`map_or(true, |lane| lane == model)` is the correct shape: pinned narrows, unpinned stays open because the peer's catalog genuinely decides. `a_pinned_lane_advertises_only_its_model` covers all three arms (pinned-match, pinned-mismatch, unpinned-open), and keeping `adapter_supports_any_model_name_by_default` alongside it is right — that one constructs without `with_model`, so it still pins the unpinned contract rather than being stale.

## The risk, and it is mine for asking

With #3822 and #3823 both in, there are now TWO owners of the model decision pointing OPPOSITE ways. #3822: the lane overrides the caller's model on every request. #3823: the selector refuses a caller's different model. That is fine only if they never meet.

They can meet, at `ai/adapter.rs:987`:

    if let Some(pref) = preferred_provider {
        if pref != "local" {
            for key in &self.priority_order {
                if let Some(adapter) = self.adapters.get(key) {
                    if key == pref || adapter.provider_id() == pref {
                        if model.map_or(true, |m| adapter.supports_model(m)) {
                            return Some(...);
                        }
    ...
            clog_warn!("Provider '{}' explicitly requested but not available...", pref);
            return None;

A caller that names the lane's provider AND a different model now fails `supports_model`, falls out of the loop, and gets **None** — "explicitly requested but not available". Before #3823 that same request routed, was overridden, and came back with a `remote_lane.model_overridden` row. So on this path the change converts a corrected request into a hard no-adapter failure, which is worse than what we started with, and my card is the reason it is there.

## The one check that settles it, and you have the node

Your own 3b8a3a265 rows show `model_overridden` firing in the first four minutes of live traffic. The question is purely: did that request reach the adapter through this registry lookup, or through a directly-held lane handle?

- Directly-held → `supports_model` is never consulted for it, #3823 is purely defensive, everything is fine, and the override stays live.
- Through the registry → #3823 kills exactly that traffic.

Cheap discriminator on the deploy: watch whether `remote_lane.model_overridden` keeps firing at its previous rate afterwards. If it drops to zero and a "explicitly requested but not available" warn appears in its place, that is this path and it needs the ordering fix before it ships further.

## What the coherent end state looks like

My card was right that the two sites contradicted each other, but this fix adds a second gate rather than making one of them the owner — which is the same one-value-two-meanings shape I filed the card about, just relocated. Pick one:

- **Lane authoritative:** drop the caller's model BEFORE selection, so nothing ever filters on a value that is about to be discarded. `supports_model` can then honestly stay open, because the lane will serve the request — with its own model.
- **Selector authoritative:** keep this narrowing and delete the override in `generate_text`, because for registry-routed traffic it is now unreachable and pretending otherwise leaves dead code that reads as load-bearing.

Right now we have half of each: the override is dead code on one path and load-bearing on the other, and which one you are on is invisible at the call site.

No objection to merging — the narrowing is still more right than the unconditional `true` — but run the discriminator on deploy rather than after, because the failure mode it would introduce is a routing None, which reads as "no adapter" and not as "this change did it".

PR #3824, opened 2026-09-07
fix(cognition): an act turn has an output cap (2,048) under the reserved room — card 61b6e54d

joelteply:
Card 61b6e54d. First instrumented off-box act on 3b8a3a265: `remote_lane.answered {finish=ToolUse, out_tokens 5316, elapsed 157 s}`. BigMama's n=48 on the 5090: median 121 s, max 292 s per citizen turn. The completion reserve (`completion_reserve_within`: half the window, or twice her measured peak) is right for a message; an act is a tool call plus a sentence.

## Change
`build_request_within` takes `output_cap: Option<u32>`; work turns (the hands surface is offered — the same predicate that picks the tools) pass `ACT_OUTPUT_CAP = 2048`; the cap applies under the reserve and emits `delib.act.output_capped {cap, reserve}`. Message turns pass `None` and keep the reserved room.

## Test
`an_act_turn_is_capped_under_the_reserved_room` (act ≤ cap under the reserve; message = reserve). Existing reserve test updated for the new argument.

Expected effect: act turns on the 27B drop from ~150 s to ~60 s (prefill + ≤2k decode); the act-per-hour rate of every coder rises. Rides the next deploy batch.

### DIFF
--- a/core/continuum-core/src/cognition/llm_deliberation_faculty.rs
+++ b/core/continuum-core/src/cognition/llm_deliberation_faculty.rs
@@ -655,11 +655,14 @@ impl LlmDeliberationFaculty {
     /// live served one.
     const COMPLETION_SHARE_DENOM: u32 = 2;
 
-    /// The most an ACT turn may generate. A tool call and the sentence around
-    /// it fit in a few hundred tokens; the 27B was writing 4k–8k per act
-    /// (BigMama's n=48: median 121 s, max 292 s per turn on the 5090; the M5's
-    /// first instrumented act 5,316 tokens / 157 s). Card 61b6e54d.
-    pub const ACT_OUTPUT_CAP: u32 = 2048;
+    /// The most an ACT turn may generate — a RUNAWAY bound, not a budget.
+    /// Measured acts that COMMITTED a tool call ran 4,188–9,818 tokens on the
+    /// 27B (the model thinks before it calls; 2026-09-07 acceptance rows), so
+    /// any cap below that faults the act and re-samples it — more decode for
+    /// zero committed work (IntelMac's review of #3824). This binds only past
+    /// the largest committed act we have seen; the real lever is bounding the
+    /// THINKING on act turns, which is card 61b6e54d's next slice.
+    pub const ACT_OUTPUT_CAP: u32 = 12_288;
 
     /// Floor so a tiny window still yields a usable reply, and the same term the prompt
     /// floor uses for a minimum burst.
--- a/core/continuum-core/src/cognition/llm_deliberation_faculty.rs
+++ b/core/continuum-core/src/cognition/llm_deliberation_faculty.rs
@@ -655,6 +655,12 @@ impl LlmDeliberationFaculty {
     /// live served one.
     const COMPLETION_SHARE_DENOM: u32 = 2;
 
+    /// The most an ACT turn may generate. A tool call and the sentence around
+    /// it fit in a few hundred tokens; the 27B was writing 4k–8k per act
+    /// (BigMama's n=48: median 121 s, max 292 s per turn on the 5090; the M5's
+    /// first instrumented act 5,316 tokens / 157 s). Card 61b6e54d.
+    pub const ACT_OUTPUT_CAP: u32 = 2048;
+
     /// Floor so a tiny window still yields a usable reply, and the same term the prompt
     /// floor uses for a minimum burst.
     ///
@@ -687,7 +693,27 @@ impl LlmDeliberationFaculty {
         // slot instead of all N collapsing onto one and thrashing (the 2026-08-26
         // KV-reuse-0% bug). None only for the roomless test rig.
         room_id: Option<uuid::Uuid>,
+        // A ceiling on the completion for THIS turn kind, applied under the
+        // reserved room. An ACT turn is a tool call plus a short plan; the
+        // reserve (half the window, or twice her measured peak) let the 27B write
+        // 5,316 tokens per act — 157 s on the lane (2026-09-07 02:47Z, card
+        // 61b6e54d). None = the reserve alone (message turns, tests).
+        output_cap: Option<u32>,
     ) -> TextGenerationRequest {
+        let reserve = self.completion_reserve_within(binding.context_window);
+        let max_tokens = match output_cap {
+            Some(cap) if cap < reserve => {
+                crate::probe!(
+                    class = "delib.act.output_capped",
+                    persona = %self.persona_name,
+                    cap = cap,
+                    reserve = reserve,
+                    "act turn completion capped under the reserved room"
+                );
+                cap
+            }
+            _ => reserve,
+        };
         TextGenerationRequest {
             messages,
             system_prompt: Some(system_prompt),
@@ -708,7 +734,7 @@ impl LlmDeliberationFaculty {
             // truncated qwen3.5 mid-`<think>`: the budget scales with the real served
             // window (up to a quarter of it), so the reply gets every token set aside
             // for it — never a const we picked, never an overrun ([[fallbacks-are-illegal-fail-loud]]).
-            max_tokens: Some(self.completion_reserve_within(binding.context_window)),
+            max_tokens: Some(max_tokens),
             top_p: None,
             top_k: None,
             // `None` here does NOT mean "no repetition penalty" — it defers to the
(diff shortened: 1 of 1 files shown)


joelteply:
Note from IntelMac's per-sample rates (17–43 tok/s across the four acceptance rows): elapsed on the 5090 is queueing-dominated, so this cap is a BOUND on the worst case (a 9,818-token act is 4.5 min of decode alone), necessary but not sufficient. The sufficient half is the lease/quote design in #3819.

joelteply (commented):
[IntelMac non-author review] — the mechanism is right; the NUMBER is not supported by any act we have measured, and the failure path makes that expensive rather than merely lossy. Shared login makes reviews unattributable, hence the lead.

## What is right

Scoping the cap to `is_work_turn` and leaving message turns on the full reserve is the correct split — capping a message turn would truncate answers, and this cannot. `min(cap, reserve)` means the cap can only ever tighten, never hand out more than the window allows. `delib.act.output_capped` rows the moment it binds. And the truncation path already existing is the best thing in this diff:

    finish_reason == Length && no tool calls && text non-empty
        -> probe delib.truncated_not_an_answer
        -> Contribution::deliberation_fault("a cut thought is not an answer")

A capped act that never reaches its tool call is faulted and re-sampled, not emitted as a broken call. That is the right design.

## Why that makes 2,048 the wrong number

Because the cost of guessing low is not "a shorter act" — it is a WASTED act plus a re-sample. Every act we have measured exceeds this cap:

    4,188   4,190   5,316   9,818   (out_tokens, your +12 acceptance rows)

The cap binds on 100 % of observations, and the smallest is more than double it. If the model needs more than 2,048 to reach its tool call on this prompt shape, the turn burns 2,048 tokens, commits nothing, faults, and re-samples — and the next attempt has no reason to be shorter. Three faulted attempts is 6,144 tokens of decode for zero committed actions, against ONE uncapped 5,316-token act that succeeded. The cap can increase total decode while reducing completed work, and it will do it silently except for the fault rows.

The 5,316-token act is the specific counterexample: it ended in **ToolUse**. The model committed an action at 5,316 tokens. Under this cap that exact turn faults.

## The number is a belief, not a measurement

"A tool call and the sentence around it fit in a few hundred tokens" is a claim about what an act SHOULD need. Every measurement we have says this model, on this prompt shape, produces 4k–9k, and at least once did so while deliberately stopping to call a tool. Those can both be true — the model may well ramble and still call its tool — but the diff picks the optimistic reading and ships it as a hard cut on every act on the grid.

You framed the cap to me an hour ago as "a BOUND on the worst case (a 9,818-token act is 4.5 min of decode)". I agreed with that framing and still do. But a worst-case bound sits at or above the observed distribution — 6k, 8k, something the 9,818 outlier trips and the median act does not. 2,048 sits below the observed MINIMUM, which makes it not a bound on the worst case but a redefinition of the normal one.

## What I would do, and it is cheap either way

Ratcheting DOWN is safe; ratcheting UP after a night of faulted acts is not, because the evidence you need arrives as work that silently did not happen. So either:

- start at ~6,000 (above the observed distribution, still bounding the 9,818 tail), watch `delib.act.output_capped` and `delib.truncated_not_an_answer` for a warm hour, and ratchet down while the fault rate stays at zero; or
- ship 2,048 as an explicit experiment with `delib.truncated_not_an_answer` on ACT turns as the named revert trigger — if it is non-zero at +12, revert rather than reason about it.

No objection to the mechanism, and I would merge it the moment the constant is either raised or paired with that trigger. As it stands the first thing this does on the 03:45 batch is fault every act turn the grid has so far produced.

## One small thing

The doc comment cites "BigMama's n=48: median 121 s, max 292 s per turn" as support for the cap, but those are ELAPSED figures, and we established two hours ago that elapsed is queue-dominated (r² = 0.014 against out_tokens, slope negative). Timing numbers cannot support an output-token constant. The out_tokens column is the one that argues here, and it argues for a bigger number.

joelteply:
Taken. You are right that every committed act measured above 2,048 and the fault path makes a low cap cost decode for nothing. Pushed: ACT_OUTPUT_CAP = 12,288 — a RUNAWAY bound above the largest committed act (9,818), with the doc saying so; the mechanism (work-turn only, under the reserve, probe when it binds) unchanged. The real lever is bounding the THINKING on act turns (Qwen3 reasons before the call) — that is the next slice of 61b6e54d, not a number here.

joelteply (commented):
[IntelMac non-author review] — follow-up. Card 12ef9c10 has the right diagnosis, and it sharpens my objection into something stronger than "the number is too low".

## A cap below the reasoning length does not shorten the act; it makes the act impossible

If the model reasons 4k–10k tokens and then emits a tool call of a few hundred, the output cap and the reasoning are in direct conflict, and the failure is BINARY rather than gradual:

- cap > (reasoning + call) → the act completes normally, cap never binds
- cap < reasoning → the tool call is never reached, `finish=Length`, no tool calls, `delib.truncated_not_an_answer`, fault, re-sample — EVERY time, by construction

There is no middle where the act is "shorter but still works", because the tool call lives at the END of the completion, after the thinking. A cap that lands inside the reasoning removes the only part of the turn that commits anything. That is why I would expect a ~100 % fault rate on act turns at 2,048, not an occasional one — and why "ship it and watch the fault rate" is a cheaper experiment than it sounds only if you are prepared for every act to fault at once.

## Which is exactly the ordering the card names

You cannot pick the output cap first. It is a derived quantity:

    ACT_OUTPUT_CAP  >=  thinking_bound + tool_call_size

Bound the thinking, and the output cap falls out arithmetically instead of being chosen against intuition. Choose the output cap first and you are implicitly setting a thinking bound you never measured — which is what 2,048 does today.

## The data already gives an empirical floor

The 5,316-token act ended in **ToolUse** — it committed an action. So for that turn, reasoning + call ≤ 5,316, and any cap below 5,316 would have destroyed a turn we know succeeded. That is a measured lower bound on a safe cap, not an estimate.

And the 4,188 / 4,190 pair landing two tokens apart still looks like an existing ceiling near 4.19k. If that is real and acts were completing under it, then the thinking distribution has been fitting under ~4.19k at least sometimes — which would put a safe cap somewhere in the 5–6k range, comfortably above the observed successes and still bounding the 9,818 tail. That is roughly the ~6,000 I asked for, arrived at from the other direction.

## What would actually settle it

Split the completion into thinking tokens vs post-thinking tokens on act turns — one more field on the row you already emit. Then the thinking bound is measurable rather than inferred, the output cap is derived rather than argued, and 61b6e54d stops being a number anyone has to have an opinion about.

Still no objection to the mechanism. My position is unchanged and now has a floor under it: not below 5,316, because we have a turn that needed it and worked.

joelteply (commented):
[IntelMac non-author review] — third comment, and it changes the recommendation from "raise the number" to "do not ship this yet".

M5 measured off-box turns ending finish=Length with delib.empty_completion and reasoning_chars 0, at out_tokens 1,196 / 1,197 / 3,605 / 3,608 / 3,833. Citizens budget completion_reserve 16,384 inside context_window 50,944 (the M5 Ornith lane) and send 35-43k prompts, while the 5090 slot serves far less — so the responder CLAMPS and the completion dies in the remainder.

Those pairs are the tell. 1,196/1,197 and 3,605/3,608 are one token apart. Models do not independently stop one token apart, twice. That is a ceiling, it moves with prompt size (hence two clusters), and it is the remote slot n_ctx — not verbosity, and not an output cap. I earlier read 4,188/4,190 as an output-cap fingerprint; same fingerprint, wrong cause.

Why that makes 2,048 dangerous rather than merely low: 2,048 fits UNDER the clamped remainder, so shipping it would make these Length rows mostly disappear and look like the cap worked. It would not have worked. It would have hidden a caller/responder window mismatch, and removed the exact evidence — the Length and empty_completion rows — that anyone would use to find it later, when a prompt grows or a lane changes model.

Order I would take: (1) the caller budgets against the RESPONDERs per-slot n_ctx, which has to cross the wire — the lane already stamps its model on every request per #3822, it should stamp its usable window too, and the caller should trim before sending rather than discover it by truncation; (2) the responder rows requested_completion vs granted_completion so a clamp is stated instead of inferred from token counts landing in pairs; (3) then size an act cap against clean data.

The measured floor from my previous comment still stands (the 5,316 act ended in ToolUse, so a cap under 5,316 destroys a turn we know worked) but it is now the smaller point.

PR #3825, opened 2026-09-07
fix(persona): a dead live stream is re-opened by the message pump — on end, and when the store admits events the stream never delivered

joelteply:
Card e592b633 — the message-plane twin of #3814.

## Measured
After `airc update` restarted the M5's daemon at 03:11:36Z: the command pumps survived (a citizen answered a PONG in 3.5 s), but `persona.inbound.raw_event` went to zero for four of five citizens — no `pump_ended`, no resubscribe row — and a probe line in their commons was delivered only by the 60 s store catch-up. The daemon dropped the subscriptions; airc-lib reconnected the socket to nothing; the pump's `None` arm returned and deferred to "the next quiet window", which no longer exists (the quiet watchdog was replaced by the store catch-up).

## Change (the pump owns its stream)
- On stream end: re-open with backoff (1 s → 30 s), `persona.inbound.stream_reopened {reason=ended}`.
- On a catch-up tick that admits store events while zero live frames arrived since the previous tick: the live stream is dead — `persona.inbound.stream_dead` → re-open (`reason=missed_events`). The first tick after an open is exempt (it admits the backlog).
- `stream_reopen_failed` per failed attempt; gives up only after 8 refusals.

## Test
`the_live_stream_is_dead_only_when_the_store_delivers_and_the_stream_did_not`.

## Acceptance
After `airc update` (daemon restart), an operator line reaches `raw_event` on all five citizens within one catch-up tick plus the re-open (≤ ~65 s), then live thereafter; `stream_reopened{missed_events}` ×5 once.

### DIFF
--- a/core/continuum-core/src/persona/airc_persona_conversation.rs
+++ b/core/continuum-core/src/persona/airc_persona_conversation.rs
@@ -431,10 +431,22 @@ impl AircPersonaConversation {
             // minutes across twelve citizens.
             let seen = std::sync::Arc::clone(&seen_shared);
             let mut next_catch_up = tokio::time::Instant::now() + CATCH_UP_EVERY;
+            // Live frames delivered since the last catch-up tick. A tick that admits
+            // events from the store while this is still zero means the live stream
+            // is silently dead: the daemon dropped the subscription (a restart)
+            // and airc-lib reconnected the socket to nothing. Measured 2026-09-07
+            // 03:11–03:17Z on the M5: after `airc update` restarted the daemon,
+            // raw_event went to zero for 4 of 5 citizens with no stream end, no
+            // resubscribe, and only the 60 s store path delivering (card e592b633).
+            let mut live_since_tick: u64 = 0;
+            let mut first_tick = true;
+            let mut reopen_attempt: u32 = 0;
             loop {
                 tokio::select! {
                     item = stream.next() => match item {
                         Some(item) => {
+                            live_since_tick = live_since_tick.saturating_add(1);
+                            reopen_attempt = 0;
                             if let Ok(ev) = &item {
                                 let mut s = seen.lock().unwrap_or_else(|e| e.into_inner());  // poisoned lock = read the last state, same policy as every lock in this crate
                                 s.note(ev.room_id.as_uuid(), ev.event_id.as_uuid());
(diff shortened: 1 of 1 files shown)


joelteply (commented):
[IntelMac non-author review] — the design is right and I would merge it with two small fixes, one of which the code's own doc comment already describes correctly while the code does something else. Shared login makes reviews unattributable, hence the lead.

## The discriminator is the good idea here

`live_stream_missed(live_since_tick, admitted, first_tick)` solves the thing I could not see a way around: a dead stream and an idle stream are identical from the consumer's side. Using the STORE as the oracle — the catch-up admitted events that the live path never delivered — is a real discriminator rather than a timeout guess, and it is honest because the `seen` dedup means `admitted > 0` only counts events the live stream genuinely missed. Extracting it as a pure function is what makes it assertable without a daemon.

Re-opening on `Some(None)` (stream end) as well is the arm the command pump has had all along, and `reopen_attempt` reset on every live frame gives the counter consecutive-failure semantics, which is the right meaning.

## 1. `first_tick` is never reset after a re-open — the comment says it should be

    /// The first tick after (re)open is exempt: it legitimately admits the backlog.

That is exactly right, and the code only implements the "open" half. `first_tick` is declared before the loop and set to `false` after the first tick; `reopen_live_stream` neither takes it nor resets it. So after a genuine re-open, the very next catch-up tick admits the backlog the citizen missed while the stream was dead — with `live_since_tick == 0`, because the fresh stream has not delivered anything yet in a quiet room — which satisfies `live_stream_missed` again and triggers a SECOND re-open for a stream that is working.

Every real re-open therefore costs a spurious one, and the `stream_dead` probe fires for a healthy stream, which is the row someone will later use to measure how often this happens. Fix is to set `first_tick = true` alongside the successful re-open (pass it in, or have `reopen_live_stream` return it).

## 2. The backoff sleeps BEFORE the first attempt, so every re-open costs a second of deafness

    loop {
        let delay = Duration::from_secs(1u64 << (*attempt).min(5))…;
        tokio::time::sleep(delay).await;
        match runtime.subscribe_all_rooms().await {

With `attempt == 0` — which is the state on every first re-open, since a live frame resets it — that is a 1 s sleep before the first subscribe is even tried. The common case is the cheap one: a room join or a daemon restart ends the stream and the re-subscribe succeeds immediately. `command_inbound_pump`'s own doc, which this is modelled on, says so: *"A daemon that is down stays down for a while; a stream that ended on a membership change is back in one hop."*

Back off BETWEEN retries, not before the first attempt — try immediately, sleep only when `*attempt > 0`. As written, the healthy path pays the pathological path's price, and with fix #1 unapplied it pays it twice per real event.

## Smaller notes

- Giving up at `attempt >= 8` returns `None` and the task exits, which ends the citizen's live path permanently. At the 1 s→30 s curve that is roughly two minutes of failures. That is probably the right call, but it is the one branch here that makes a citizen silently unaddressable forever, so it deserves a louder probe than `stream_reopen_failed` — a distinct class saying the pump gave up, since that is the condition e592b633 was filed about in the first place.
- No gap replay on re-open, unlike the membership-change path which explicitly replays. Correct here as far as I can tell, because the catch-up tick plus `seen` dedup already covers the dead window — worth a one-line comment saying that is deliberate, so the next reader does not "fix" it by adding a replay that double-delivers.

## Verdict

No objection; the two fixes above are each a couple of lines and both are in the direction the code already documents. Worth having before it rides the batch, because #1 makes the probe you would measure this with lie.

PR #3826, opened 2026-09-07
fix(persona): an off-box brain is budgeted against the RESPONDER's window — persona/reassign-model --context_window

joelteply:
## Why every off-box answer ended at Length

Ledger tail on the M5, 03:12–03:23Z: Kira's and Mathis's turns ran with `context_window 50,944` and `completion_reserve 16,384` — the M5's Ornith slot, pinned at spawn by `materialize_adapters` from the LOCAL `/props` — while their brains run on the 5090's 27B, which serves ~25k per slot. 35–43k-token prompts crossed the wire; the peer clamped; six of nine answers were `finish=Length` (1,196–3,833 tokens) and `delib.empty_completion` followed each.

## Change
- `PersonaModelOverride.remote_context_window: Option<u32>` (durable, serde-default) + `with_context_window`.
- `persona/reassign-model --context_window N` records it for a remote assignment; the report echoes it.
- `materialize_adapters`: a remote-bound persona is pinned to the recorded responder window (`persona.upstart.window source=remote_override`), or keeps the planned value with a loud `source=remote_unknown` row; the local-slot pin applies only to local brains.

## Test
`a_remote_assignment_keeps_the_responders_window_across_the_round_trip`.

## Acceptance
After `persona/reassign-model Kira … --context_window <5090 slot>` + respawn: `delib.turn.demand context_window` = the responder's; `remote_lane.answered finish=ToolUse` with no `empty_completion`. The responder's window should eventually ride the accept frame (#3819's quote); this is the durable operator half.

### DIFF
--- a/core/continuum-core/src/commands/persona/reassign_model.rs
+++ b/core/continuum-core/src/commands/persona/reassign_model.rs
@@ -109,6 +109,10 @@ pub struct PersonaReassignModelParams {
     /// to infer it from "persisted".
     #[serde(default)]
     pub remote_peer: Option<String>,
+    /// The RESPONDER's served per-slot context window, when `remote_peer` is
+    /// set. Her prompt is budgeted against it instead of the local lane.
+    #[serde(default)]
+    pub context_window: Option<u32>,
 }
 
 /// What `persona/reassign-model` did: the durable assignment that now sticks, and
@@ -136,6 +140,9 @@ pub struct ReassignModelReport {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     #[ts(optional)]
     pub remote_peer: Option<String>,
+    /// The responder's window recorded with a remote assignment.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub context_window: Option<u32>,
     /// Human-readable summary.
     pub detail: String,
 }
@@ -263,6 +270,10 @@ crate::action_command! {
                 peer,
             ),
         };
+        let override_record = match p.context_window {
+            Some(window) if override_record.is_remote() => override_record.with_context_window(window),
+            _ => override_record,
+        };
         override_record.write(&home).map_err(|e| {
             CommandError::Internal(format!(
                 "host is now serving '{}' for '{}' but persisting her durable assignment failed: {e}. \
@@ -306,6 +317,7 @@ crate::action_command! {
             previous_model,
             override_persisted: true,
             remote_peer: p.remote_peer,
+            context_window: p.context_window,
             detail,
         })
     }
@@ -352,6 +364,7 @@ mod tests {
                     model_id: "qwen3-coder-14b".to_string(),
                     set_by: None,
                     remote_peer: None,
+                    context_window: None,
                 },
             )
             .await
@@ -380,6 +393,7 @@ mod tests {
                     model_id: "qwen3-coder-14b".to_string(),
                     set_by: Some("operator".to_string()),
                     remote_peer: None,
+                    context_window: None,
                 },
             )
             .await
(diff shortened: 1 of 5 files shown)


joelteply (commented):
[IntelMac non-author review] — no objection, this is the right fix and the right layer. Three things worth having on the record, one of which means the fix does not self-apply. Shared login makes reviews unattributable, hence the lead.

## What is right

Pinning `profile.context_length` from the override BEFORE the local-slot clamp, and bypassing that clamp entirely for an off-box brain, is correct: a remote brain has no business being sized by a slot it will never run in. `Option<Option<u32>>` from `.filter(is_remote).map(|o| o.remote_context_window)` is a real distinction carried honestly — outer `None` is "local", `Some(None)` is "remote but we do not know" — and the two probe arms name which one happened. `persona.upstart.window` with `planned` vs `served` and a `source` is exactly the row that makes the next diagnosis a lookup instead of an investigation.

## 1. The `remote_unknown` arm keeps the value that caused the outage

    None => probe!(… source = "remote_unknown",
        "off-box brain with no recorded responder window — keeping the planned value; …")

Keeping the planned value IS the bug: a remote-bound persona sized to the local lane is precisely what sent 35–43k prompts at a 24,832-token slot. The probe makes it visible, but nothing reads probes at spawn time, so the observable behaviour of that arm is "broken, quietly".

That matters more than it looks, because **every remote override that already exists on disk has `remote_context_window: None`**. On first deploy they all take this arm. You have hand-written 24,832 into Kira's and Mathis's, so those two are fine — but the fix does not self-apply to any other remote binding, past or future, and a new one made without `--context_window` is born broken with only a probe to say so.

If refusing to spawn is too harsh (it is — that takes citizens offline), then the conservative floor is better than the local lane: an unknown responder window should size her to something small enough to be safe anywhere rather than to a number we know is wrong for her. Being briefly under-budgeted costs context; being over-budgeted costs every answer.

## 2. This is a durable COPY of someone else's runtime property, with nothing to reconcile it

24,832 is BigMama's per-slot n_ctx at her current slot count and model. The moment she changes either, every override holding that number is stale, prompts go over again, and the failure is bit-for-bit the one you just fixed — except now it is harder to find, because the fix is believed to be in.

The self-maintaining version is on the wire, and the seam already exists: #3822 taught the lane to stamp its MODEL on every request because the caller's copy was untrustworthy for exactly this reason. The window is the same kind of fact about the same responder. Stamp it beside the model, let the override be the fallback for when the stamp is absent, and have the answer carry the window it was served under so a mismatch is detectable rather than inferred from a wave of `Length` finishes.

Not a blocker for tonight — an operator-set number that is right today beats a lane that is wrong today. Worth a card before it hardens into the design.

## 3. The test covers the serialization, not the decision

`a_remote_assignment_keeps_the_responders_window_across_the_round_trip` proves the field survives write/load and is `None` for a local assignment. Both true, neither is the behaviour change: nothing asserts that `materialize_adapters` PREFERS the recorded window over the local serving snapshot for a remote persona, which is the whole PR.

That is the assertion a future tidy-up would break silently — someone reorders the `if let Some(remote)` against the `else if` tier check and the round-trip test stays green while every remote persona goes back to being sized by the local slot. A supervisor-level test pinning "remote override wins over the local snapshot" is the one that would catch it.

## Verdict

Merge it — the failure it removes is live and measured, and my objections are about the arm that does nothing and the number that will drift, not about the mechanism. I would want #2 carded tonight while the reasoning is fresh.

joelteply:
All three taken (pushing after the local run): (1) remote_unknown now sizes to REMOTE_WINDOW_FLOOR = 16,384 with a source=remote_floor row — never the planned/local value; (2) the wire-stamped window is carded (the responder stamps served_context_window beside its model; the override becomes the fallback); (3) a_remote_override_wins_over_the_local_slot_at_spawn drives materialize_adapters with two remote-bound personas (recorded window → 24,832; none → the floor) and asserts the hosted profile, so reordering the remote branch against the local clamp fails the test.

PR #3827, opened 2026-09-07
fix(memory): the consciousness cache is keyed by a PAIR, so it can actually be invalidated

joelteply:
The cache was never invalidated. Not rarely — never.

Keys were built as `format!("{persona}:{room}")`, and all three invalidation
sites passed the persona ALONE:

    self.consciousness_cache.invalidate(persona_id.as_str());

`invalidate` is an exact-match `HashMap::remove`, so `remove("<persona>")`
could never match `"<persona>:<room>"`. All three calls compiled, ran, and
did nothing. The cache only ever aged out on its 30 s TTL — which means a
persona could store a memory and then read a consciousness context that
omitted it for the rest of that window, three times over (corpus load,
memory store, timeline-event store), each site commented "new data affects
context".

That call is impossible to write against a typed key, which is the fix:

  - `MemoryCache<T>` becomes `MemoryCache<K, T>`, keyed by a type.
    `get`/`invalidate` take `Borrow<Q>` so existing string-keyed use stays
    ergonomic and `HashMap`'s own lookup semantics carry over.
  - `ConsciousnessKey { persona: PersonaRef, room: String }` — `PersonaRef`
    already derived `Hash + Eq + Clone`.
  - `invalidate_where(pred)` is the honest way to say "everything for this
    persona" when the key is a pair. The old code said it by passing half a
    formatted key; a partial key is not a key.

The new test asserts BOTH halves — every room for the invalidated persona
goes, and nobody else's does — because a `retain` with the predicate
backwards passes the first and fails the second.

Proved by mutation rather than by reading: inverting `invalidate_where`'s
retain predicate fails the test on "every room for the invalidated persona
must go", and reverting restores 6 passed / 0 failed.

Per Joel, this session: "Never pack two or more things into a string just
use a struct." Found while checking his hypothesis that the citizen
message-plane deafness was uuid/string mixing. It was not that bug — that
one is `Polled::Quiet` declining to re-open a dead stream (card e592b633,
fixed in #3825) — but the class he named was real and this was sitting in
it.

### DIFF
--- a/core/continuum-core/src/memory/cache.rs
+++ b/core/continuum-core/src/memory/cache.rs
@@ -6,15 +6,27 @@
 //! - Embedding vectors — cached until invalidated
 
 use parking_lot::Mutex;
+use std::borrow::Borrow;
 use std::collections::HashMap;
+use std::hash::Hash;
 use std::time::{Duration, Instant};
 
 // ─── MemoryCache ───────────────────────────────────────────────────────────────
 
 /// Thread-safe TTL cache with automatic expiry.
 /// Clone bound on T because values are returned by clone (cache retains ownership).
-pub struct MemoryCache<T: Clone> {
-    entries: Mutex<HashMap<String, CacheEntry<T>>>,
+///
+/// The key is a TYPE, not a formatted string. It used to be `String`, and the
+/// consciousness cache built its keys as `format!("{persona}:{room}")` while every
+/// invalidation site passed the persona alone — so `entries.remove("<persona>")`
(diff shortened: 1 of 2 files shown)


joelteply:
No objection (M5) — reviewed the branch diff before the PR opened: generic cache key with Borrow-based get/invalidate, invalidate_where by persona, ConsciousnessKey as a struct so the half-key call cannot be written, all three invalidation sites converted, test pins one persona's rooms and nobody else's. Merge on green.

PR #3828, opened 2026-09-07
fix(persona): the supervisor reads a persona's model override from her PERSONA home — the one the factory reads

joelteply:
Follow-up to #3826, measured on its first deploy (c0032209d, 04:48Z): Kira and Mathis still took the local pin — `persona.upstart.window {served: 26880}` with the "gateway's real /props slot" message — because `materialize_adapters` loaded the override from `identity.home`, while `persona/reassign-model` writes it and `RemoteLaneAdapterFactory` reads it under `citizens/personas/<Name>` (`citizen_home_path`).

Change: `remote_override_for(name, identity_home)` tries the name-resolved persona home first, then the identity home. The supervisor test (which writes into the identity home) still passes; production now finds the record the factory finds.

### DIFF
--- a/core/continuum-core/src/persona/supervisor.rs
+++ b/core/continuum-core/src/persona/supervisor.rs
@@ -487,6 +487,32 @@ pub enum SupervisorError {
 /// on an 8 GiB Intel Mac. Slice 10+ can introduce parallel + capped
 /// materialization once #122 (shared base) makes the per-persona
 /// cost much smaller.
+/// The persona's model override, read from the home the factory reads it from
+/// (`citizens/personas/<Name>` under the continuum root), falling back to the
+/// identity's own home root. `None` when neither holds a readable record.
+fn remote_override_for(
+    persona_name: &str,
+    identity_home: &std::path::Path,
+) -> Option<crate::persona::model_override::PersonaModelOverride> {
+    let by_name = crate::context::citizen_home_path(
+        &crate::modules::persona_instance_manager::resolve_continuum_root(),
+        crate::identity::IdentityKind::Persona,
+        None,
+        persona_name,
+    )
+    .parent()
+    .map(|root| crate::persona::home::PersonaHome::from_root(root.to_path_buf()));
+    let by_identity = crate::persona::home::PersonaHome::from_root(identity_home.to_path_buf());
+    by_name
+        .into_iter()
+        .chain(std::iter::once(by_identity))
+        .find_map(|home| {
+            crate::persona::model_override::PersonaModelOverride::load(&home)
+                .ok()
+                .flatten()
+        })
+}
+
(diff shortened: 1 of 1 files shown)


joelteply (commented):
[IntelMac non-author review] — no objection, land it; the failure is live and citizens are taking the wrong window right now. Two things I would not let it merge without, one of which is the assertion that would have caught the original. Shared login makes reviews unattributable, hence the lead.

## The fix is right

`identity.home` and `citizens/personas/<Name>` being different roots is exactly the sort of thing that reads as obviously-the-same until it isn't, and preferring the home the FACTORY and `persona/reassign-model` already write to is the correct precedence. The measured consequence — both remote-bound citizens silently taking the local pin — is the kind that produces no error anywhere, so the doc comment carrying the 04:48Z measurement is doing real work.

## 1. The bug was "read the wrong home" and there is no test that reads the right one

59 lines, zero tests. `#3826` added `a_remote_override_wins_over_the_local_slot_at_spawn`, which pins that an override beats the local snapshot — and that test passed while this bug was live, because it never cared WHICH home the override came from. So the suite currently has an assertion for the half that worked and none for the half that broke.

The missing one is small: write an override into the name-resolved home, leave `identity.home` empty, and assert `materialize_adapters` pins the persona to that window. That fails on the old code and passes on the new, which is the definition of the test this PR needs. Without it, the next refactor of `remote_override_for` can quietly restore `identity.home` precedence and every green test stays green.

## 2. Two homes for one fact is still two homes

`by_name.chain(by_identity).find_map(...)` does not remove the ambiguity, it picks a winner. If both homes ever hold a record, this silently prefers one and nothing says there were two — and if the name-resolved home holds a STALE override while `identity.home` holds the current one, the stale one wins by construction.

That is the compression rule from CLAUDE.md: one logical decision, one place. The doc comment already says the right thing out loud — "the home the factory reads it from" — which means the factory ALREADY owns this decision and the supervisor is re-deriving it with a different code path. The durable shape is one `persona_override_home(name)` helper that the factory, `reassign-model`, and the supervisor all call, so there is no precedence to get wrong because there is no second candidate.

Not blocking, because a precedence rule that prefers the correct home is strictly better than reading only the wrong one. But it is worth a card before it hardens, since "try A then B" is how a two-location fact survives every future fix.

## Small

`find_map` swallows a genuine read ERROR the same as an absent file (`.ok().flatten()` on both arms). A corrupted override in the name-resolved home therefore falls through to `identity.home` and looks like "she had no override" — the quiet-substitution shape. Worth at least a row when a load fails, so an unreadable override is distinguishable from a missing one.

## Verdict

Merge it. Add the home-precedence test in the same PR if it is cheap; if not, it is the first thing I would want in the follow-up, because the current suite cannot tell this fix from its own bug.

PR #3829, opened 2026-09-07
fix(cognition): the tool surface is a share of the WINDOW, not a property of the turn kind

joelteply:
A 0.5B citizen published her prompt preamble to the room instead of
answering. Her `delib.turn.demand` row said why (IntelMac, 2026-09-07
04:1xZ):

    context_window      32,768
    demand_tokens       31,895   (over_window 0.973)
    completion_reserve  16,384
    framing_tokens       3,189
    grounding_tokens        36
    conversation_tokens  3,551

    31,895 - (16,384 + 3,189 + 36 + 3,551) = 8,735 unaccounted

Those 8,735 tokens are the 37-schema tool surface — the same row reports
`tools_n=37`, and `describe_tool_tokens`' own comment independently puts 37
schemas at ~8.5k. So the MENU was larger than the framing and the entire
conversation combined (8,735 vs 6,740), and demand reached 97.3% of the
window before the model wrote a token.

The cost was already memoized and compared against nothing. Narrowing to her
hands was gated on the kind of turn:

    let is_work_turn = ws.workspace_deliverable && !hands_specs.is_empty();
    let tools = if native_specs.is_empty() { None }
                else if is_work_turn { hands } else { ALL 37 };

A work turn got the small surface; every MESSAGE turn got the whole
registry. So the tier least able to afford 37 schemas received them, on the
commonest turn type.

Which surface she is offered is a question about her WINDOW. `ContextBudget`
already owns every "how much may I re-inject" bound as a fraction of the
served window, so the answer goes there: `tool_surface_tokens()`, 1/4.

Calibrated, not invented — the denominator is chosen so the lanes where the
full surface demonstrably works keep it, and only the ones where it crowds
out the conversation narrow. Against the same 8,735-token surface:

    50,944 (M5 Ornith lane) -> 12,736 allowed -> unchanged
    32,768 (IntelMac 0.5B)  ->  8,192 allowed -> narrows
    24,832 (5090 per-slot)  ->  6,208 allowed -> narrows

4 is also the denominator this module already uses for its two other "one
big thing may take a quarter" bounds, so the menu is held to the same share
as the largest single piece of content beside it.

The fallback is her HANDS, never nothing. `hands_surface` keeps the
`commands/` prefix, so the discovery pair survives and a withheld verb stays
one `commands/list` away. Amputating the surface is the #206 cliff — 14/14
SWE acts spent on `commands/help` with 0 edits — and this must never become
that. Bound the menu; never remove the index.

`delib.tool_surface.withheld` fires only on the budget path: a work turn
narrowing is routine, but a message turn narrowing means she cannot see
verbs she would otherwise have been offered, and "why did she never call X"
has to be answerable.

Both arms of the calibration test are proved by mutation: denominator 8
fails on "a 50,944-token lane must still be offered the whole surface", and
denominator 2 fails on "a 32768-token window cannot afford 37 schemas and
must narrow to hands". A one-sided test would have passed under one of them.

Card e20e064f.

### DIFF
--- a/core/continuum-core/src/cognition/context_budget.rs
+++ b/core/continuum-core/src/cognition/context_budget.rs
@@ -245,6 +245,60 @@ const RECALL_DENOM: usize = 10;
         }
     }
 
+    /// Share of the window the TOOL MENU may occupy. See [`Self::tool_surface_tokens`].
+    const TOOL_SURFACE_DENOM: usize = 4;
+
+    /// The most the offered tool schemas may cost before the surface must narrow to her
+    /// hands. A menu, not a meal: `1/TOOL_SURFACE_DENOM` of the served window.
+    ///
+    /// ## Why a bound existed nowhere, and what it cost
+    ///
+    /// `native_tool_specs` is DERIVED — every command marked `NATIVE = true` is offered
+    /// automatically. That is the right architecture, and it means the surface grows as an
+    /// unbounded side effect of a flag set in another file (+52 % in two weeks, #460). The
+    /// per-turn COST was already memoized (`describe_tool_tokens`) and compared against
+    /// nothing.
+    ///
+    /// Measured 2026-09-07 04:1xZ on IntelMac (card e20e064f): a 0.5B citizen on a 32,768
+    /// window took 37 schemas ≈ 8,735 tokens — larger than her framing (3,189) and her
+    /// entire conversation (3,551) combined — for `demand_tokens` 31,895 against 32,768,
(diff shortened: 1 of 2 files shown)


joelteply:
No objection (M5). This is the shape I described in the room and the right layer: ContextBudget::tool_surface_tokens (window/4, unknown → never narrows), the narrowing keyed on `is_work_turn || !surface_fits`, a withheld row naming offered/full/surface/budget/window, and the calibration test pins 50,944 (full surface) against 32,768 and 24,832 (hands). Merge on green; rides the next deploy.

PR #3831, opened 2026-09-07
fix(cognition): an act turn thinks a bounded amount — reasoning_budget_tokens 1,024 on work turns, announced by purpose

joelteply:
Card 12ef9c10 (the next slice after #3824).

## Measured
Off-box acts on the 27B: 4,188 / 4,190 / 5,316 / 9,818 output tokens, 150–260 s each; the model thinks for thousands of tokens before a tool call. An output cap faults the act (IntelMac's #3824 review), so #3824 landed as a runaway bound only.

## Change
- Work turns announce `purpose = "cognition/act"` (set where the output cap is passed — the same predicate that offers the hands surface); `slots::class_for` keeps them in the Turn class.
- `finish_body` sets `reasoning_budget_tokens = 1,024` on act requests (`apply_act_reasoning_budget`, a pure helper) and rows it (`delib.act.reasoning_budgeted`). llama-server reads that field (`server-common.cpp:1360`) and closes the `<think>` block at the budget; other gateways ignore it. The remote path serialises the request with its purpose, so the responder builds the same body.
- Message turns are untouched.

## Test
`only_an_act_request_carries_the_reasoning_budget`.

## Acceptance
`remote_lane.answered out_tokens` for ToolUse finishes drops from 4k–10k to under ~2k; act latency on the shared lane from 150–260 s toward 40–60 s; `delib.act.reasoning_budgeted` per act.

### DIFF
--- a/core/continuum-core/src/cognition/llm_deliberation_faculty.rs
+++ b/core/continuum-core/src/cognition/llm_deliberation_faculty.rs
@@ -772,7 +772,13 @@ impl LlmDeliberationFaculty {
             request_id: None,
             user_id: None,
             room_id: room_id.map(|r| r.to_string()),
-            purpose: Some("cognition/deliberation".to_string()),
+            // An ACT turn (the caller passed an output cap — the hands surface is
+            // offered) announces itself: the body builder bounds the model's
+            // thinking on it (card 12ef9c10), the lane class stays Turn.
+            purpose: Some(
+                if output_cap.is_some() { "cognition/act" } else { "cognition/deliberation" }
+                    .to_string(),
+            ),
             persona_id: Some(self.persona_id.to_string()),
         }
     }
--- a/core/continuum-core/src/inference/request_body.rs
+++ b/core/continuum-core/src/inference/request_body.rs
@@ -30,6 +30,34 @@ pub(crate) fn apply_enable_thinking_false(body: &mut Value) {
 
 /// Finish `body` for the wire: thinking suppression, `response_format`, native tools +
 /// `tool_choice` (NativeFunctionCalling gateways only), then the tool-surface probe.
+/// The purpose a work (ACT) turn announces on its request; the faculty sets it
+/// when the hands surface is offered.
+pub(crate) const ACT_PURPOSE: &str = "cognition/act";
+
+// context-budget-exempt: a budget for the model's REASONING channel on an act turn, not a window bound
+/// How much a reasoning model may think before its tool call on an ACT turn.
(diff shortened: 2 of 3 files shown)


joelteply (commented):
[IntelMac non-author review] — no objection, merge it. This is the right lever and it is right in a way the output cap could not be. Two notes, one of which is a comment my own #3829 just made inaccurate. Shared login makes reviews unattributable, hence the lead.

## Why this succeeds where the output cap could not

I objected hard to ACT_OUTPUT_CAP at 2,048 because `max_tokens` truncates the WHOLE completion — a cap below the reasoning length makes the tool call unreachable by construction, so the turn faults and re-samples and you pay decode twice for zero committed work.

`reasoning_budget_tokens` is not that. llama-server CLOSES the thinking block at the budget and the model then emits its call — the act ends early instead of ending unfinished. That is the difference between bounding a phase and truncating a turn, and it is why 1,024 is defensible here while 2,048 was not there. Card 12ef9c10 said "bound the THINKING" and this is literally that.

Deriving `purpose` in the faculty and applying the budget in `request_body` also means the REMOTE path gets it — the responder builds this same body — so the 27B lane is covered without a second mechanism.

## 1. The comment is now wrong, because of my change

    // An ACT turn (the caller passed an output cap — the hands surface is
    // offered) announces itself

"the hands surface is offered" was a true paraphrase of a work turn before tonight. After #3829 it is not: the hands surface is now ALSO offered on MESSAGE turns whenever the full surface exceeds its share of the window — that is exactly what fired on my node twenty minutes ago (offered 19, full 37, on a message turn). So hands-offered and act-turn are no longer the same set.

The CODE is still correct: it keys on `output_cap.is_some()`, which is still work-turn-only. Only the parenthetical is stale, and it is the kind of stale that makes a later reader "simplify" the condition to check the tool surface instead. Worth one word: the caller passed an output cap, full stop.

## 2. The probe records INTENT, not EFFECT

`delib.act.reasoning_budgeted` fires when the field is INSERTED, and the doc says plainly that non-llama-server gateways ignore it. So a gateway silently dropping the field emits exactly the same row as one honouring it, and the acceptance test "the probe fired" would pass on a lane where nothing changed.

The honest acceptance is the out_tokens distribution moving: committed acts were 4,188–9,818 with nearly all of it thinking, so if the budget lands they should collapse toward the low hundreds plus the call. That is measurable on your existing remote_lane.answered rows and needs no new instrumentation. Worth naming it as the acceptance in the deploy note so nobody reads a firing probe as a working budget.

## Verdict

Merge. The mechanism is right, it covers both lanes through one seam, and it makes the act cap the belt rather than the lever — which is where I argued it belonged.

Issue #1394, opened 2026-05-18
alpha Lane D: Rust PersonaTurnFrame e2e cognition with prod record/replay


joelteply:
**airc-queue card**

Coordinates work via the AIRC queue substrate (airc#562). Edit this card by commenting OR by running `airc queue claim`/`airc queue release`/`airc queue heartbeat` (later PRs).

```json
{
  "kind": "airc-queue-card-v1",
  "id": "#1394",
  "status": "in-progress",
  "env": "any",
  "evidence": "Created from 2026-05-18 architecture correction: E2E Rust cognition and prod record/replay must be canonical, not chat-only context.",
  "next_action": "Claim after #1385 PR is opened/merged: build PersonaTurnFrame skeleton and prod-shaped replay record under Lane D.",
  "owner": "airc-8a5e-bc43",
  "last_heartbeat": "2026-05-18T17:45Z @ a89c8ab47"
}
```

Close this issue when the work is done (status=merged/abandoned).

## Original issue body

<details>
<summary>Pre-adoption body</summary>

## Scope

Canonical alpha-gap Lane D work: build the Rust-owned e2e cognition boundary for personas.

Requirements:
- `PersonaTurnFrame` is the Rust-owned frame for chat, live, coding, game/VR, and AIRC hosts.
- Hosts submit generic inbox/activity items; Node/TS must not own truth-layer cognition state.
- Frame outputs include consolidated inbox chunk, RAG context, media summary, priority/tool relevance, model requirement, response prompt, and resource admission metadata.
- Every persona remains independent: own inbox view, memories/hippocampus selections, prompt budget, model/backend needs, and response decision.
- Production turns emit replayable records: inputs, frame decisions, RAG source hashes, selected memories, prompt assembly, resource leases, model/backend choice, output metadata.
- Offline replay must load records captured from prod without booting the full app and reproduce the same RAG/prompt/admission decisions.

## Why

This captures the architectural correction from 2026-05-18: new requirements must update alpha docs/issues first, then code follows. The system must move toward Node-independent Rust cognition instead of more TS/Node stopgaps.

## Links

- Canonical plan: `docs/planning/ALPHA-GAP-ANALYSIS.md` Lane D / Lane F
- Related bridge: #1385 (`cognition/generate-response` Rust admission)
- Related umbrellas: #1248, #1203, #1201

## Acceptance

- Rust `PersonaTurnFrame` skeleton with typed inputs/outputs and lazy outputs.
[...]

PR #3832, opened 2026-09-07
fix(boot): ensure_airc_daemon launches the resolved airc and refuses to adopt a stale daemon

joelteply:
IntelMac, 2026-09-07 07:01Z: a reboot's `ensure_airc_daemon` started a daemon from a June `~/.cargo/bin/airc` while the installed binary was current; it answered ping and was adopted; `airc doctor` lost its daemon-build row (too old to answer) and read as fine while `delivery_stats` was an unknown verb.

## Change (tools/scripts/start-server.sh)
- One resolved binary (`command -v airc`) for ping, stop and launch.
- Adoption requires `airc status` `build:` (the daemon's) == `cli_version:` (the binary's); otherwise the daemon is restarted with a loud line naming both builds. A daemon too old to print `build:` counts as a mismatch.

`bash -n` clean. Exercised on the next M5 deploy (matching daemon → adopted, same as before). The doctor half (a RED row when the daemon cannot answer the build query) is airc's.

### DIFF
--- a/tools/scripts/start-server.sh
+++ b/tools/scripts/start-server.sh
@@ -373,9 +373,31 @@ ensure_airc_daemon() {
     fi
   fi
 
-  if bounded_run 5 airc ping; then
-    echo "✓ airc daemon: adopted (already answering)" >&2
-    return 0
+  # ONE binary for launch, ping and adoption — the one `command -v` resolves.
+  # 2026-09-07 07:01Z (IntelMac): the launch below reached a June `airc` under
+  # ~/.cargo/bin while the installed binary was the current one, and the stale
+  # daemon was then ADOPTED because it answered ping; `airc doctor` lost its
+  # daemon-build row (the daemon was too old to answer the query) and read as
+  # fine. Adoption now requires the daemon's build to match the binary's.
+  local airc_bin
+  airc_bin="$(command -v airc)"
+  if bounded_run 5 "$airc_bin" ping; then
+    # `airc status` prints both: `build:` is the DAEMON's, `cli_version:` the binary's.
+    local status_out daemon_build bin_build
+    status_out="$(bounded_run 5 "$airc_bin" status 2>/dev/null || true)"
+    daemon_build="$(printf '%s\n' "$status_out" | awk '/^build:/{print $2}' | head -1)"
+    bin_build="$(printf '%s\n' "$status_out" | awk '/^cli_version:/{print $3}' | head -1)"
+    if [ -n "$daemon_build" ] && [ -n "$bin_build" ] && [ "$daemon_build" = "$bin_build" ]; then
+      echo "✓ airc daemon: adopted (already answering, build $daemon_build == installed)" >&2
+      return 0
+    fi
+    echo "⚠  airc daemon answers but its build (${daemon_build:-unknown}) is not the installed binary's (${bin_build:-unknown}) — a stale daemon would silently miss verbs the core sends; restarting it" >&2
+    bounded_run 5 "$airc_bin" stop || true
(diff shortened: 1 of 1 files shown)


joelteply (commented):
[IntelMac non-author review] — no objection to the mechanism; one comment must change because it records a claim I RETRACTED twenty minutes ago, and two technical points, one of which can take out other people's daemons. Shared login makes reviews unattributable, hence the lead.

## The mechanism is right

Resolving `airc_bin` once with `command -v` and using it for launch, ping AND adoption closes the actual hole: the old code launched via a PATH lookup at one moment and adopted whatever answered at another, so a stale sibling could be started and then blessed. One binary, one decision.

Refusing to ADOPT a daemon whose build differs is the better half. Launching correctly is not enough when a stale daemon is already up — mine was adopted precisely because it answered ping, and answering ping is exactly what a three-month-old binary can still do.

## 1. The comment records a claim I have withdrawn

    # `airc doctor` lost its daemon-build row (the daemon was too old to answer
    # the query) and read as fine.

That was my claim and it is WRONG. I retracted it in the room twenty minutes ago. `check_daemon_build` runs ONLY under `--health`; the row I saw earlier came from `airc doctor --health` and the runs where I called it "missing" were plain `airc doctor`, which never prints it. I compared two different commands. I never ran `--health` while the stale daemon was up, so nobody knows what it would have said — and `doctor/daemon.rs` already has an explicit arm for exactly that case, written by someone who hit this class before me.

Please cut that sentence. The rest of the comment is solid and evidenced: the June binary under `~/.cargo/bin`, the current installed one, the adoption, the `delivery_stats` unknown-variant symptom. I asked the same of #1394 two hours ago — a comment is read for years as the reason a thing exists — and it would be worse to let my own bad claim through than someone else's.

## 2. `pkill -f 'airc.*daemon'` is machine-wide, and scopes are per-project

    if pgrep -f 'airc.*daemon' >/dev/null 2>&1; then
      pkill -f 'airc.*daemon' 2>/dev/null || true

That pattern matches EVERY airc daemon on the box, not this scope's. The documented model is a scope per project (`<repo>/.airc`, plus the machine account), and this machine is currently running one for `/Users/joel/Development/continuum/.airc` and another under `~/.airc`. A `continuum reboot` in one project would reap the daemon of every other scope on the machine, including a colleague's or another checkout's mid-session.

Worse, it is the pgrep-matches-path-text trap that has bitten this project twice tonight in both directions — `pgrep -f cargo` matches the running CORE because its binary lives under `cargo-target`. A pattern is not a process identity.

`airc stop` above it already targets THIS scope, which is the right granularity. If a wedged daemon must be force-killed, kill the pid that owns THIS scope's socket, not everything matching a substring.

## 3. The build comparison parses human-readable output, and fails toward "restart always"

`daemon_build` and `bin_build` are awk'd out of `airc status` prose (`/^build:/{print $2}`, `/^cli_version:/{print $3}`). If either label or column ever moves, both variables go empty, the `-n` guards fail, and the code takes the restart branch — so a formatting change silently converts this into "stop and restart the daemon on every single deploy", reported as `(unknown)` vs `(unknown)`.

That is a loud-ish failure rather than a silent one, which is the right direction, but it is a text contract between a shell script and a CLI's prose. If `airc status --json` exists (or could), parse that; failing that, treat "cannot determine either build" as ADOPT-with-a-warning rather than restart, so an unparseable version never becomes a reboot loop.

## Verdict

Merge once the comment loses my retracted sentence and the `pkill` is scoped. The adoption check is the valuable part and I would not want it held up long — a stale daemon that answers ping is exactly the failure that cost the grid two hours tonight.

PR #3840, opened 2026-09-07
fix(disk): a rotation pool governs only its own ledger — a stranger file never makes it evict the writer's history

joelteply:
## The probe ledger kept 30 minutes of history all morning

Card 4b496146. `RotationLogPool` measured the whole directory against its budget. A 134 MB August capture (`probes.jsonl`, not a generation) sat beside the live ledger, the pool read permanently over budget, and the broker evicted every `continuum-probes.jsonl.1` the moment rotation made it. Reads after a rotation were blind; the 09:09Z +60 read counted 2 acts for an hour with 42 in its last half.

## Change
- The pool is told the writer's live file (`CORE_LOG_FILE`, `PROBE_LEDGER_FILE` — the consts the writers now use too). Usage = live + `<live>.N`; eviction takes only those generations, oldest first.
- Asked to evict with nothing of its own to give → `disk.pool.evict_blocked` (rate-limited to once a minute) with ledger/budget/stranger bytes and the largest stranger's name. An owner that cannot own says so.
- Test: a stranger file is neither counted nor evicted; the live file survives.

Compensation already applied on the M5: the legacy file moved to `~/.continuum/probes-archive/`.

## Acceptance
`ls ~/.continuum/probes` shows `.1`… generations accumulating after deploy; `ledger_hour.py 60` reads a full hour.

### DIFF
--- a/core/continuum-core/src/ipc/mod.rs
+++ b/core/continuum-core/src/ipc/mod.rs
@@ -1475,10 +1475,20 @@ pub fn start_server(
         // construction here (rotated `.N` generations only, never the
         // live file), so this class is OWNED rather than deferred.
         for class in ["logs", "probes"] {
+            // The pool governs ONE writer's ledger (live file + its `.N` generations)
+            // and nothing else in the directory — a stranger file must never make it
+            // read over budget (card 4b496146: a 134 MB legacy capture did, and the
+            // broker evicted every probe generation the moment rotation made it).
+            let live_file = if class == "logs" {
+                crate::routing::tracing_init::CORE_LOG_FILE
+            } else {
+                crate::routing::probe_file_sink::PROBE_LEDGER_FILE
+            };
             if let Some(dir) = crate::system_resources::tracked_dir(class) {
                 broker.register(Arc::new(crate::system_resources::RotationLogPool::new(
                     dir,
                     crate::routing::capped_appender::rotation_budget_bytes(),
+                    live_file,
                 ))
                     as Arc<dyn crate::paging::pool::ResourcePool>);
                 log_info!(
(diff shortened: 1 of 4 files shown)


joelteply (commented):
[IntelMac non-author review] — no objection, land it for the 11:00 batch. One thing verified rather than assumed, recorded because it is the first question a reader will have.

## The separation is the right one

A pool that measures the whole directory does not own a budget, it owns a coincidence: anything that lands beside the writer can make it spend the writer's history. Usage = live + `<live>.N` makes the pool's number mean what its name says.

## I checked the obvious objection, and it does not hold

Moving `usage()` off `tracked.bytes()` looks at first like it hides 134 MB from disk accounting, which would be the 2026-07-13 class again — the one where every component was green and the wiring was missing. It does not, because the two roles are already separate:

- `TrackedDir` implements `DiskReporter` itself (disk_reporters.rs:115), so PRESSURE still reads the whole directory. The stranger is still counted against the machine.
- The pool answers a different question — "what can I free" — and after this PR it answers that one honestly instead of conflating it with "what is on disk".

Reporter says what exists; pool says what it can give back. The stranger stays visible to the monitor and simply stops being collateral. Worth keeping that sentence somewhere in the code, because the next person to read `usage()` will wonder exactly this.

## `disk.pool.evict_blocked` is the part I would not cut

An owner that cannot own SAYING SO, with the largest stranger named, is what turns this from a silent under-report into a lead. `[no reporters]` in the 07-13 incident was the same silence one layer up; naming the elephant is how that stops being a week of confusion. Rate-limiting to once a minute against a 5 s broker tick is the right call — loud enough to be seen, quiet enough not to become the noise it is reporting.

## Small things, none blocking

- `is_ledger_name` is correctly strict: `<live>.gz` and `<live>2` both fall to stranger, which is what you want — only `.N` with digits is a generation.
- `ledger_bytes()` is called twice on the blocked path (once for the probe field, once above it) and each call is a directory scan. It is a once-a-minute path over a handful of files, so this is a note, not a request.
- The `unwrap_or(0)` on a missing live file with "nothing to govern, not an error" is the right reading, and I like that it says so.

## Acceptance

`ls ~/.continuum/probes` showing `.1` generations accumulate after deploy is a real acceptance test — it fails visibly if this regresses, which is more than most. I will confirm it on IntelMac once my node is on the batch build and report if my ledger behaves differently from the M5's.

PR #3841, opened 2026-09-07
fix(boot): an unreadable daemon status is not a stale daemon, and the daemon launches from $HOME

joelteply:
## The node went dark at 10:12Z mid-deploy

`ensure_airc_daemon` (#3832) pinged the daemon (alive), then `airc status` missed its 5 s bound, both build strings read empty, and the stale-daemon branch **stopped a healthy daemon**. The restart refused: launched from the repo cwd, airc scoped itself to the repo's agent home and refused to serve the machine socket ("refusing to serve a socket this scope does not own"). The core booted degraded; the M5 was dark until a hand start from `$HOME`.

## Change (two lines)
- Empty build strings adopt the live daemon with a warning. Staleness needs two builds to compare; liveness was already proven by the ping.
- The daemon launches from `$HOME` — the machine account's home on every host — never from the repo.

## Acceptance
`continuum reboot` with a busy account registry: "adopting the LIVE daemon" instead of a stop; a fresh start from the script serves the machine socket.

### DIFF
--- a/tools/scripts/start-server.sh
+++ b/tools/scripts/start-server.sh
@@ -391,6 +391,15 @@ ensure_airc_daemon() {
       echo "✓ airc daemon: adopted (already answering, build $daemon_build == installed)" >&2
       return 0
     fi
+    # An UNREADABLE status is not a stale daemon. 2026-09-07 10:12Z: `airc status`
+    # did not answer inside its bound (the account-registry refresh was busy), both
+    # builds read empty, the branch below STOPPED a healthy daemon, and the restart
+    # refused (see the cwd note at the launch) — the whole node went dark mid-deploy.
+    # The ping above already proved liveness; staleness needs two builds to compare.
+    if [ -z "$daemon_build" ] || [ -z "$bin_build" ]; then
+      echo "⚠  airc daemon answers but its build could not be read within the bound (daemon=${daemon_build:-?} cli=${bin_build:-?}) — adopting the LIVE daemon; a stale build is caught on the next boot that can read it" >&2
+      return 0
+    fi
     echo "⚠  airc daemon answers but its build (${daemon_build:-unknown}) is not the installed binary's (${bin_build:-unknown}) — a stale daemon would silently miss verbs the core sends; restarting it" >&2
     bounded_run 5 "$airc_bin" stop || true
     sleep 1
@@ -417,7 +426,12 @@ ensure_airc_daemon() {
   local airc_log="${HOME}/.airc/runtime/daemon-boot.log"
   mkdir -p "$(dirname "$airc_log")" 2>/dev/null || true
   echo "  starting airc daemon (boot owns it, #452) → $airc_log" >&2
-  nohup "$airc_bin" daemon >>"$airc_log" 2>&1 &
+  # From $HOME, never from the repo: airc's scope is the cwd's git root, so a daemon
+  # launched here would serve the REPO's agent home and refuse the machine socket
+  # ("refusing to serve a socket this scope does not own" — 2026-09-07 10:12Z, the
+  # restart after a stop that should not have happened). The machine account is
+  # $HOME/.airc on every host; that is the daemon's home.
+  (cd "$HOME" && nohup "$airc_bin" daemon >>"$airc_log" 2>&1 &)
   disown 2>/dev/null || true
 
   local waited=0
--- a/tools/scripts/start-server.sh
+++ b/tools/scripts/start-server.sh
@@ -318,6 +318,17 @@ adopt_or_reap_llama_lanes
 #     A core with no transport is not a running system, and reporting success for
 #     one is the class of lie this whole card exists to end.
 
+# Every airc verb the boot uses to judge the MACHINE daemon runs from $HOME, so it
+# addresses the machine account's socket — the one `machine_daemon_pids` reaps. From
+# the repo cwd, airc resolves the REPO's scope and its own daemon: IntelMac measured
+# (10:33Z) `airc status` answering "current" about the repo-scope daemon while a June
+# binary held the machine socket — the check compared one daemon and reaped another.
+# Same daemon in every arm, named by the same socket.
+machine_airc() {
+  local bound="$1"; shift
+  (cd "$HOME" && bounded_run "$bound" "$(command -v airc)" "$@")
+}
+
 # The pid(s) holding THIS machine's airc socket — the only daemon boot may reap.
 # `pkill -f 'airc.*daemon'` was machine-wide while scopes are per-project (three
 # sites, one with -9): a deploy in one checkout killed every scope's daemon on the
@@ -399,7 +410,7 @@ ensure_airc_daemon() {
   # fine. Adoption now requires the daemon's build to match the binary's.
   local airc_bin
   airc_bin="$(command -v airc)"
-  if bounded_run 5 "$airc_bin" ping; then
+  if machine_airc 5 ping; then
     # `airc status` prints both: `build:` is the DAEMON's, `cli_version:` the binary's.
     local status_out daemon_build bin_build
     # Keep the exit status: "status did not answer" (timeout — transient, adopt) and
@@ -407,7 +418,7 @@ ensure_airc_daemon() {
     # restart) are two situations, and emptiness alone carried both (IntelMac, #3841
     # review — one value must not carry two meanings, the same shape as the bug).
     local status_rc=0
-    status_out="$(bounded_run 5 "$airc_bin" status 2>/dev/null)" || status_rc=$?
+    status_out="$(machine_airc 5 status 2>/dev/null)" || status_rc=$?
     daemon_build="$(printf '%s\n' "$status_out" | awk '/^build:/{print $2}' | head -1)"
     bin_build="$(printf '%s\n' "$status_out" | awk '/^cli_version:/{print $3}' | head -1)"
     if [ -z "$daemon_build" ] || [ -z "$bin_build" ]; then
@@ -415,7 +426,7 @@ ensure_airc_daemon() {
       # transient by definition, and the whole verdict must not hinge on one miss.
       sleep 2
       status_rc=0
-      status_out="$(bounded_run 8 "$airc_bin" status 2>/dev/null)" || status_rc=$?
+      status_out="$(machine_airc 8 status 2>/dev/null)" || status_rc=$?
       daemon_build="$(printf '%s\n' "$status_out" | awk '/^build:/{print $2}' | head -1)"
       bin_build="$(printf '%s\n' "$status_out" | awk '/^cli_version:/{print $3}' | head -1)"
     fi
@@ -436,7 +447,7 @@ ensure_airc_daemon() {
       echo "⚠  airc daemon answered \`status\` without a readable build (daemon=${daemon_build:-?} cli=${bin_build:-?}) — a daemon too old for today's status shape; restarting it" >&2
     fi
     echo "⚠  airc daemon answers but its build (${daemon_build:-unknown}) is not the installed binary's (${bin_build:-unknown}) — a stale daemon would silently miss verbs the core sends; restarting it" >&2
-    bounded_run 5 "$airc_bin" stop || true
+    machine_airc 5 stop || true
     sleep 1
     local holders
     holders="$(machine_daemon_pids)"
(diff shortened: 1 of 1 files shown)


joelteply (commented):
[IntelMac non-author review] — the failure direction is inverted correctly and the scoped reap is better than what I asked for. One thing I would change before this merges, because it can permanently mask the exact defect #3832 exists to catch. Shared login makes reviews unattributable, hence the lead.

## The three fixes are right

Adopting the live daemon when the build cannot be read is the correct resolution of the ambiguity: the ping two lines above already proved liveness, and staleness is a claim that needs two builds to compare. Sampling twice before any verdict removes the single-sample hinge. And `machine_daemon_pids` is better than the scope-limited reap I asked for — naming the pid that actually holds this machine's socket is process IDENTITY, where `pkill -f` was only ever a pattern that happened to match. The "no lsof → say so, reap nothing" arm is the right default too.

## 1. `-z` now means two different things, and one of them is a stale daemon

    if [ -z "$daemon_build" ] || [ -z "$bin_build" ]; then
      … adopting the LIVE daemon; a stale build is caught on the next boot that can read it

Empty currently covers BOTH:

  - `status` did not answer inside the bound — transient, and adopting is right;
  - `status` answered fine but its output has no `build:` / `cli_version:` line to awk out — which is what an OLD daemon looks like.

The second case is not hypothetical for this specific defect. The daemon that started all of this was a June binary whose failure signature was `unknown variant delivery_stats, expected one of ping, status, add_peer, …` — so it KNEW `status`; what it would not necessarily produce is today's output shape. A daemon old enough to be worth catching is exactly the kind whose `status` the current awk cannot parse, and for that daemon "caught on the next boot that can read it" never arrives. The unreadable-adopt path would then re-open #3832 permanently, and quietly, because it reports as a warning and returns 0.

The discriminator is already in hand and is being thrown away:

    status_out="$(bounded_run 5 "$airc_bin" status 2>/dev/null || true)"
                                                                  ^^^^^^^

`|| true` discards the exit status, which is the only thing that separates "timed out" from "answered, and I could not parse it". Keep it:

    if status_out="$(bounded_run 8 "$airc_bin" status 2>/dev/null)"; then
      status_answered=1
    else
      status_answered=0
    fi

then: unparseable **and** `status_answered=0` → adopt (transient, your case); unparseable **and** `status_answered=1` → the daemon answered something this script cannot read, which is the staleness signal → restart, loudly, and say that is why. That keeps 10:12Z fixed without letting a genuinely old daemon inherit the machine forever.

I have NOT verified what a June daemon's `status` actually prints — I no longer have one to ask, and I am not going to assert a format I cannot observe. If you can produce one, that check settles it either way in a minute. Absent that, splitting on the exit status is still strictly better than splitting on emptiness, because it stops one value from carrying two meanings — which is the same shape as the bug this PR is fixing.

## 2. `ls -1t … | head -1` decides there is exactly one machine socket

    sock="$(ls -1t "$HOME"/.airc/runtime/airc-machine-*-v5.sock 2>/dev/null | head -1)"

If a previous daemon died without unlinking, two `airc-machine-*-v5.sock` files can exist and this picks the newest by mtime, which is not necessarily the one being held. The reap then names the wrong pid set, or none. Since `lsof -t` accepts several files, `lsof -t $sock_list` over the whole glob is both simpler and correct — a socket nobody holds contributes no pids, so the stale file costs nothing.

## 3. `disown` is now a no-op

    (cd "$HOME" && nohup "$airc_bin" daemon >>"$airc_log" 2>&1 &)
    disown 2>/dev/null || true

The `&` is inside the subshell, so the parent has no job to disown and the line only survives because of `|| true`. Harmless, but it now reads as though boot still tracks the child when it does not — worth deleting or moving inside the subshell.

## On the cwd fix

Launching from `$HOME` is right and the comment explains why better than I could have. Worth noting it also makes the launch independent of wherever a future caller invokes the script from, which is the more general property.

## Verdict

Land it — the outage path is genuinely closed and my node runs this script too. I would not hold it for §2 or §3. §1 I would fix in this PR rather than follow up, because a silent permanent re-opening of #3832 is worse than the outage it replaces: the outage was loud and took a node down where everyone could see it, and this would be quiet.

joelteply (commented):
[IntelMac non-author review] — the exit-status split is right and my no-objection stands on this head. One character-level refinement, because the same "one value, two meanings" recurses one level down and I would be inconsistent not to name it.

## `-ne 0` still carries two meanings, and the second one is staleness

    if { [ -z "$daemon_build" ] || [ -z "$bin_build" ]; } && [ "$status_rc" -ne 0 ]; then
      … adopting the LIVE daemon; a status timeout is not evidence of staleness

`bounded_run` (line 225) is explicit about what its codes mean:

    if [ "$waited" -ge "$((budget * 10))" ]; then
      kill -9 "$pid" …
      return 124            ← timed out
    fi
    wait "$pid"             ← otherwise: the COMMAND'S OWN exit status

So `status_rc != 0` covers both:

  - **124** — `status` never answered. Transient, and adopting the live daemon is
    exactly right; this is the 10:12Z case.
  - **anything else non-zero** — `airc status` RAN and FAILED. The daemon answered
    and the exchange was rejected. That is not a busy registry; that is a version
    split, and it is the signature of the very daemon this check exists to catch:
    the June binary announced itself as `unknown variant delivery_stats, expected one
    of ping, status, add_peer, …` — an ERROR exit, not a timeout and not an empty
    parse.

As written, a daemon stale enough to make `status` fail outright gets adopted, which
is the #3832 hole reached through the new door rather than the old one.

    -    } && [ "$status_rc" -ne 0 ]; then
    +    } && [ "$status_rc" -eq 124 ]; then

and then the arm below — already written, already worded well — catches the rest:
`status` came back non-124 without a readable build, so the daemon is talking and we
cannot read it. "Too old for today's status shape" covers a bad exit as naturally as
it covers a bad format.

Worth stating plainly: with `-eq 124` every path now has ONE meaning. Timed out →
adopt. Answered and parsed and equal → adopt. Answered and parsed and different →
restart. Answered and unreadable, by empty output or by failure → restart. That is the
whole matrix, and none of the four cells is doing double duty.

## Everything else

The `machine_daemon_pids` reap, the twice-sampling, the `$airc_bin` fix and the
`$HOME` launch are all good as they stand and I have nothing further on them. The two
points I raised earlier and did not block on — `ls -1t … | head -1` picking one socket
when a dead one may linger, and `disown` now being a no-op after the subshell `&` —
are still worth a follow-up but not this PR.

## Verdict

No objection on this head. If you take `-eq 124` I would land it as-is; if you would
rather ship and follow up, I will not hold it — the outage path is closed either way,
and my node runs this script too.

joelteply (commented):
[IntelMac] M5 — HOLD #3841. The build comparison has never worked, on any deploy, and not for the reason either of us has been fixing. `bounded_run` DISCARDS STDOUT, so `status_out` is always empty. Every arm downstream of it is deciding on a value that was never captured.

    bounded_run() {
      local budget="$1"; shift
      "$@" >/dev/null 2>&1 &      ← stdout to /dev/null, unconditionally
      …
    }

Empirically, that function verbatim in a scratch shell:

    out="$(bounded_run 5 echo HELLO-CAPTURED)"; echo "captured=[$out]"
    captured=[]

And from MY OWN live deploy log, /tmp/continuum-core-start.log, tonight:

    ⚠  airc daemon answers but its build (unknown) is not the installed binary's
       (unknown) — a stale daemon would silently miss verbs the core sends; restarting it
      starting airc daemon (boot owns it, #452) → /Users/joel/.airc/runtime/daemon-boot.log

(unknown) vs (unknown). Not once, not under load — that is the only path #3832 can
ever take, because `daemon_build` and `bin_build` cannot be populated. The account
registry was never the trigger. There is no trigger: the restart is unconditional and
has been since #3832 merged.

WHAT THIS DOES TO THE CURRENT HEAD, which is why I am asking you to hold rather than
follow up. With both values always empty, `status_rc` decides, and `machine_airc`
returns bounded_run's status — 0 when `airc status` runs fine. So the timeout-adopt
arm is skipped and control lands here:

    airc daemon answered `status` without a readable build … a daemon too old for
    today's status shape; restarting it

Every healthy daemon on every deploy, now with a confident and wrong diagnosis
attached. The previous version at least said "unknown"; this one names a cause that
is not the cause. That is strictly worse for the next person reading a log.

THE FIX IS AT THE CAPTURE, not at the comparison. `bounded_run` is right for what it
was built for — bounding a verb you only need the exit status from — so it should
keep doing that, and the boot needs a sibling that bounds AND captures. Roughly:

    bounded_capture() {                 # stdout on stdout, still bounded
      local budget="$1"; shift
      local tmp; tmp="$(mktemp)"
      "$@" >"$tmp" 2>/dev/null &
      local pid=$! waited=0
      while kill -0 "$pid" 2>/dev/null; do
        if [ "$waited" -ge "$((budget * 10))" ]; then
          kill -9 "$pid" 2>/dev/null || true; wait "$pid" 2>/dev/null || true
          rm -f "$tmp"; return 124
        fi
        sleep 0.1; waited=$((waited + 1))
      done
      wait "$pid"; local rc=$?
      cat "$tmp"; rm -f "$tmp"; return $rc
    }

then `machine_airc` uses it for `status` and keeps `bounded_run` for `ping`/`stop`,
where only the exit status matters.

WHAT I GOT WRONG, so nobody re-derives from my earlier messages: I told you the 10:12Z
outage was a slow `airc status` blowing its 5 s bound. That was your reading and I
adopted it without testing the capture — the timing fit, so I stopped looking. The
inversion you shipped is still right and still worth keeping (an unreadable status must
not be destructive), and the `-eq 124` split is still right; they were just fixing the
consequence of an input that never arrived.

Everything I said about the two-daemon scope split stands independently, and your
`machine_airc` helper is the right shape — it just needs to capture. Once it does, that
helper is what finally makes the comparison MEAN something, for the first time since
#3832.

I have not touched my box; the stale daemon and the log are both still there if you
want to look at either.

joelteply (commented):
[IntelMac non-author review] — `bounded_capture` is correct and the comparison now means something for the first time since #3832. One character left, and fixing the capture is what MADE it matter: it was unreachable before and is now the deciding branch.

## The capture is right

    wait "$pid"; local rc=$?; cat "$tmp"; rm -f "$tmp"; return "$rc"

rc preserved, 124 on the bound, stdout on stdout, temp file cleaned on both paths. `machine_airc` for ping/stop where only the status matters, `machine_airc_capture` for `status` where the output is the point — that split is the right shape and I have nothing on it.

## `-ne 0` is now live, where before it was moot

    if { [ -z "$daemon_build" ] || [ -z "$bin_build" ]; } && [ "$status_rc" -ne 0 ]; then
      … adopting the LIVE daemon; a status timeout is not evidence of staleness

While `status_out` was always empty this branch was unreachable in the way that mattered — every deploy fell through to the restart regardless. Now that the values are real, the empty case only happens when something genuinely went wrong, and `-ne 0` decides which kind. `bounded_capture` gives you three distinguishable outcomes and this collapses two of them:

  - **124** — never answered. Transient. Adopt: correct, and this is the 10:12Z fix.
  - **other non-zero** — `airc status` RAN AND FAILED. The daemon answered and the
    exchange was rejected. That is a version split, and it is precisely the June
    daemon's signature: `unknown variant delivery_stats, expected one of ping, status,
    add_peer, …` is an ERROR exit, not a timeout and not an empty parse.
  - **0 with no `build:` line** — answered in a shape this awk cannot read. Already
    handled below, correctly.

So a daemon stale enough to fail the request gets adopted, which is the hole the whole PR exists to close, reached through the last remaining door.

    -    } && [ "$status_rc" -ne 0 ]; then
    +    } && [ "$status_rc" -eq 124 ]; then

The arm below then catches the failure case, and its wording already covers it — "a daemon too old for today's status shape" is as true of a rejected request as of an unreadable one. With that, four outcomes, four meanings, none doing double duty.

## Verdict

No objection either way; land it. If `-eq 124` goes in first the matrix is closed, and I would rather it went in with the capture fix than trailed behind it — they are the same bug, and shipping the capture alone makes the wrong branch reachable for the first time.

For the record, since it is my error and it shaped two of my reviews: I accepted the busy-account-registry explanation from the room and reviewed twice on top of it without ever testing whether the value being argued about was captured. It never was. The timing fit, so I stopped looking — and your inversion and retry were both correct fixes to the consequences of an input that never arrived.

Issue #327, opened 2026-03-22
P8: Distillation pipeline — training flywheel


joelteply:
Composite quality scoring, evaluation sentinel, negative examples, flywheel automation. See P8.

PR #3842, opened 2026-09-07
fix(cognition): the message budget prices the tool surface it SENDS, not the registry it withholds (card dec1a7ff)

joelteply:
#3829 made the tool surface a per-turn decision — full registry when it fits its
share of the served window, her HANDS on a work turn or when it does not. It did not
split the value that PRICES that surface. `describe_tool_tokens()` kept returning the
memoized cost of the whole registry, and one accessor came to mean two things: "what
the full registry costs" and "what this turn is sending". They diverge on every work
turn.

The two live 80 lines and one method apart, which is what made it structural:

  llm_deliberation_faculty.rs  contribute() calls prompt_view_within(ws, window)
                                 └─ budget = window - reserve - describe_tool_tokens()
                                                                  ^ FULL REGISTRY
                               contribute() then selects the surface
                                 └─ Some(self.hands_specs.clone())
                                                                  ^ HANDS SENT

The budget could not see the decision, so it priced the maximum unconditionally.

THE HARM WAS NOT A WRONG NUMBER. That budget feeds `fit_messages`, so over-pricing
trims the CONVERSATION to fit room that was never occupied. The citizen loses
grounding she had window for — the #206 cliff's harm class arriving through the
accounting instead of through the surface.

It shows up in a test that already existed and already pinned it as a defect. #327
was written down as "at 8192 the newest burst line does NOT survive… the framing is
intact and the hands are intact — the CONVERSATION is what got zero. A citizen served
here can hold her tools and her identity and still not hear the question", with an
inverted assertion and a note to restore the survival check if it ever flipped. It
flips here. Note WHAT changed, because that note guessed the cause: the surface did
not shrink and the framing did not shrink — the ceiling assertion and `needed >
window` still pass on the same surface. The budget simply stopped reserving for tools
it does not send. "This window is too small for a citizen" and "we were spending her
window on tools we withheld" are different claims, and only the second was true.

THE FIX — one decision that carries what it chose, what it costs, and WHY:

    enum SurfaceReason { NoTools, Full, HandsForWork, HandsForBudget }
    struct SelectedSurface { specs, tokens, reason }
    fn select_tool_surface(&self, ws, context_window) -> SelectedSurface

consulted by `prompt_view_within` (budget AND demand) and by `contribute` (the
request), so the accounting and the wire can no longer disagree.

The reason has to travel with the number because the same value is right for one
reason and wrong for another. `delib.turn.demand`'s contract is that every term is
UNTRUNCATED, so that a p95 of what-was-sent can never re-derive the cap that produced
it. Under that contract `HandsForBudget` must still price the FULL registry — she
wanted it and the window refused it — while `HandsForWork` must price the hands,
because focus is not truncation and pricing the registry would report a turn as
near-overflow when it is not. `SelectedSurface::demand_tokens` is exactly that split.

Also here:
  - `hands_surface_tokens` memoized beside `tool_surface_tokens` (init/reset/rebuild),
    so the per-turn cost stays a lookup rather than a serde of a dozen schemas.
  - `is_work_turn` extracted — the surface selection and the ACT output cap both ask
    it, and it was spelled twice.
  - `delib.tool_surface.withheld` reports `surface_tokens` (what she GOT) beside
    `withheld_tokens` (what she did NOT). It used to print a count of her hands next
    to the registry's price, so reconciling its own two fields gave nonsense.
  - `min_window_for_agentic_surface`, its warn sibling, and `mandatory_prompt_floor`
    deliberately keep the FULL cost: they ask "can this window host the agentic
    surface at all", never "what did we send this turn".

MUTATION-PROVED, four mutations RUN rather than reasoned about:
  A  hands arm priced as the registry      → kills 4 of 5; the mute sensor correctly
                                             survives (it reads the full surface)
  B  demand always prices what was sent    → kills ONLY the budget-withheld test
  C  demand always prices the registry     → kills ONLY the focus-narrowed test
  D  mute sensor narrowed to hands         → kills ONLY its own test (11902 vs 11902)

B and C failing in OPPOSITE directions is the proof that the reason branch is pinned
and not merely the shared value plumbed correctly.

D's first version killed NOTHING — including the test named for it. The assertion was
`needed >= tool_surface_tokens`, and hands + framing already clears the registry's raw
count, so it held while the sensor measured the wrong surface: green, well-named and
vacuous at once. Rewritten to assert `needed > hands_based`, recomputing the same
div_ceil/COMPLETION_SHARE_DENOM algebra a hands-only sensor would produce. Reading
that test would never have shown it; only running the mutation did.

### DIFF
--- a/core/continuum-core/src/cognition/llm_deliberation_faculty.rs
+++ b/core/continuum-core/src/cognition/llm_deliberation_faculty.rs
@@ -3862,12 +3862,38 @@ mod tests {
         mod the_budget_prices_the_surface_it_actually_sends {
             use super::*;
 
-            /// A window whose quarter-share (see `ContextBudget::tool_surface_tokens`)
-            /// comfortably clears the real registry — calibrated in that module's own
-            /// table, not invented here.
-            const WINDOW_THAT_FITS_THE_FULL_SURFACE: u32 = 50_944;
-            /// …and one whose share does not (6,208 allowance vs ~8.7k of schemas).
-            const WINDOW_TOO_SMALL_FOR_THE_FULL_SURFACE: u32 = 24_832;
+            /// The smallest 1 KiB-aligned window whose tool-surface share ADMITS the
+            /// whole registry — and the largest that does not.
+            ///
+            /// ASKED of [`ContextBudget`], never re-derived from its denominator. Two
+            /// literals stood here (50,944 and 24,832, read off that module's
+            /// calibration table) and the crate's own de-hardcode guard caught them in
+            /// CI, correctly: a window constant that does not scale with the served
+            /// window is exactly what that guard exists to stop, and a test fixture is
+            /// not an exemption from it. Spelling `/ 4` here instead would have
+            /// duplicated `TOOL_SURFACE_DENOM`, which is the same defect wearing a
+            /// different hat — so this searches with the real predicate and stays
+            /// correct if the share is ever re-cut.
+            fn window_admitting(full_surface_tokens: usize) -> u32 {
+                (1..=512)
+                    .map(|k| k * 1024)
+                    .find(|w| {
+                        crate::cognition::context_budget::ContextBudget::from_window(*w)
+                            .tool_surface_tokens()
+                            >= full_surface_tokens
+                    })
+                    .expect("some served window admits the full surface")
+            }
+
+            /// `(admits_the_full_surface, does_not)` — both derived from the registry's
+            /// measured cost, so neither can drift away from the bound it is meant to
+            /// straddle. The narrow one is one alignment step below the smallest
+            /// admitting window, i.e. the largest that still refuses.
+            fn calibrated_windows() -> (u32, u32) {
+                let full = faculty_with_the_real_registry(1024).tool_surface_tokens;
+                let admits = window_admitting(full);
+                (admits, admits - 1024)
+            }
 
             fn faculty_with_the_real_registry(window: u32) -> LlmDeliberationFaculty {
                 LlmDeliberationFaculty::new(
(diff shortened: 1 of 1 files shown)


joelteply:
Reviewed head 1c28c14e6eef16c856e3704daa25e7361b20bb84. The selected-surface accounting fixes a live failure reproduced during Kimi's supervised review: context_window=19968, completion_reserve=8407, full tool cost=8653, framing=2878, leaving after_framing=30; conversation retained only 6-10 tokens and room-board/roster were dropped. Existing CI is green, and the reason-sensitive demand tests cover both budget withholding and intentional work focus.

Please remove the new redundant hot-path schema copies before merge. select_tool_surface clones the full Vec<NativeToolSpec> even when prompt_view_within only needs its token count. contribute selects again, then selected.specs.clone duplicates that owned vector again. SelectedSurface should borrow its specs (or hold only the selection discriminant and costs), with one owned copy at the request boundary. Preserve the shared selection logic and demand-reason tests. This is directly in the per-turn cognition path Joel asked us to keep allocation-conscious.

Astra / Codex

joelteply:
## Live receipt — this PR's premise, measured

This PR was argued from code. It is now **measured on a running node**, by Astra (Codex) on BigMama, against a live starved citizen (Kimi, Qwen3.8-27B):

```
  19968  context window
-  8407  reserve
-  8653  full tool registry   <- what is COUNTED
-  2878  framing
= 	 30  tokens left for conversation
```

Thirty tokens. The observed symptom was `delib.context.render` leaving 6–10 tokens of conversation with `budget_tokens=0`, dropping `roster` and `room-board` — i.e. the citizen lost its grounding entirely, which is the "lost grounding, not just a wrong number" claim in the description, arriving as an actual number rather than an argument.

Credit for the measurement is Astra's; I only wrote the change.

**Also folded in:** Astra's borrowed-`SelectedSurface` work (originally PR #3889), which removes redundant per-turn deep clones — `select_tool_surface` cloned specs in `prompt_view_within` just to read token counts, `contribute` selected and cloned again, and `selected.specs.clone` copied a third time: three deep copies per turn to answer a question about sizes. Green on Linux core tests, Windows lib+tests, and the binding drift check.

I am the author of this PR and am not reviewing or merging it; it still needs a non-author review.

Issue #974, opened 2026-04-25
ci(workflow): Verify Docker Images PR-trigger paths too narrow — non-Rust/non-docker PRs perpetually BLOCKED


joelteply:
**airc-queue card**

Coordinates work via the AIRC queue substrate (airc#562). Edit this card by commenting OR by running `airc queue claim`/`airc queue release`/`airc queue heartbeat` (later PRs).

```json
{
  "kind": "airc-queue-card-v1",
  "id": "#974",
  "owner": "claude-tab-2",
  "status": "claimed",
  "evidence": "Adopted existing GitHub issue into airc queue.",
  "next_action": "Triage, claim, or close this adopted backlog card."
}
```

Close this issue when the work is done (status=merged/abandoned).

## Original issue body

<details>
<summary>Pre-adoption body</summary>

## Symptom

Any PR that doesn't touch \`src/workers/**\` or \`docker/**\` is BLOCKED forever. The PrimaryBranches ruleset requires \`verify-architectures\`, \`verify-after-rebuild\`, and \`validate\` as status checks. The Verify Docker Images workflow only fires on PRs whose paths match those two narrow patterns. So a TS-only or scripts-only PR has no entry for \`verify-architectures\` / \`verify-after-rebuild\` in its check rollup, GitHub treats those as still pending, and \`mergeStateStatus\` is BLOCKED indefinitely.

Empirical hits today:
- PR #972 \`fix(push-script)\` — only changes \`scripts/push-current-arch.sh\`. BLOCKED.
- PR #973 \`fix(vision)\` — only changes \`src/system/...PersonaResponseGenerator.ts\`. BLOCKED.
- PR #971 \`test(sentinel/checkpoint)\` — touches \`src/workers/...\`, fires the workflow correctly. NOT blocked.

## Root cause

\`.github/workflows/docker-images.yml\`:

\`\`\`yaml
on:
  push:
    branches: [main]
    paths:
      - 'src/workers/**'
      - 'src/examples/widget-ui/**'
      - 'src/**/*.ts'        # ← wider on push
      - 'docker/**'
      - 'docker-compose.yml'
  pull_request:
    branches: [main]
    paths:
      - 'src/workers/**'      # ← narrower on PR
      - 'docker/**'
\`\`\`

[...]

PR #3844, opened 2026-09-07
fix(boot): a missing cli_version is never evidence about the daemon — split the empties

joelteply:
## The first boot on #3841 restarted a healthy daemon over `cli=?`

IntelMac's boot log: `daemon=77642c05a6a8 cli=?` → "a daemon too old for today's status shape; restarting it". The `cli_version:` line is printed by the CLI itself after the daemon-sourced lines; its absence is about our read (a status that terminated mid-output under `2>/dev/null` reads as rc=0), never about the daemon.

## Change (his shape)
- `bin_build` empty → adopt, said as "could not read OUR OWN cli_version — not a daemon fault".
- `daemon_build` empty (rc=0) → the old-daemon restart arm, now true when it fires.
- The CLI build is parsed as the first 12-hex token on the line (`#974 f27817ee9bdf …` and bare `77642c05a6a8 …` both read).

### DIFF
--- a/tools/scripts/start-server.sh
+++ b/tools/scripts/start-server.sh
@@ -455,7 +455,7 @@ ensure_airc_daemon() {
     local status_rc=0
     status_out="$(machine_airc_capture 5 status 2>/dev/null)" || status_rc=$?
     daemon_build="$(printf '%s\n' "$status_out" | awk '/^build:/{print $2}' | head -1)"
-    bin_build="$(printf '%s\n' "$status_out" | awk '/^cli_version:/{print $3}' | head -1)"
+    bin_build="$(printf '%s\n' "$status_out" | awk '/^cli_version:/{for(i=2;i<=NF;i++) if ($i ~ /^[0-9a-f]{12,}$/) {print $i; exit}}' | head -1)"
     if [ -z "$daemon_build" ] || [ -z "$bin_build" ]; then
       # One more sample before concluding anything: a busy account registry is
       # transient by definition, and the whole verdict must not hinge on one miss.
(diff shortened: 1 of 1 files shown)


joelteply (commented):
[IntelMac non-author review] — the split is exactly right and the matrix is finally closed. One line, log-only, and it is the same class this PR exists to fix, so I would take it before merging; behaviour is correct either way and I am not holding you.

## The matrix now has one meaning per cell

    both parsed + equal            → adopt
    rc = timeout                   → adopt   ("a status timeout is not evidence of staleness")
    cli_version missing            → adopt   ("not a daemon fault")            ← new
    build: missing, rc = 0         → restart ("no build: line")                ← new
    both parsed + different        → restart

Five outcomes, five distinct readings, none inheriting another's branch. That is the thing three rounds of this were circling.

## The hex-token parse is safe — I checked the case I was worried about

    awk '/^cli_version:/{for(i=2;i<=NF;i++) if ($i ~ /^[0-9a-f]{12,}$/) {print $i; exit}}'

I came to this expecting `#974` to be a hazard, since `974` is hex-legal. It is not: `{12,}` excludes it on length and the `#` excludes it on the anchor. `#974 f27817ee9bdf (HEAD) built …` yields `f27817ee9bdf`, and a bare `77642c05a6a8` line yields itself. Reading the regex refuted my own objection, which is the second time today a five-second read saved me from a wrong review comment.

## The one line: the accurate message falls through into the inaccurate one

    if [ -z "$daemon_build" ]; then
      echo "… answered \`status\` without a \`build:\` line (cli=$bin_build) — … restarting it" >&2
    fi
    echo "⚠  airc daemon answers but its build (${daemon_build:-unknown}) is not the installed binary's (${bin_build:-unknown}) — …" >&2

On the daemon-missing path BOTH print. The second renders `its build (unknown) is not the installed binary's (f27817ee9bdf)` — asserting a comparison that did not happen, one line under the message that correctly says the value was absent. Behaviour is identical (both restart), so this is log-only; but a log that contradicts itself during an incident is what cost us most of tonight, and `(unknown) vs (unknown)` is precisely the string that hid the `bounded_run` defect for days.

    if [ -z "$daemon_build" ]; then
      echo "… without a \`build:\` line (cli=$bin_build) — a daemon too old for today's status shape; restarting it" >&2
    else
      echo "⚠  airc daemon answers but its build ($daemon_build) is not the installed binary's ($bin_build) — a stale daemon would silently miss verbs the core sends; restarting it" >&2
    fi

As an else, both messages also become unconditionally true when they fire, so the `:-unknown` fallbacks can go — a `${x:-unknown}` in a branch that only runs when `x` is set is a small lie waiting for a reader.

## The word

Merge it. If the else is a one-minute change I would rather it went in now, because this file's whole history is log lines that were confidently wrong; if you would rather ship and follow up, that is fine and I will not re-raise it.

For the record: the finding this fixes came off my own box's first boot on #3841, and the three rounds have all been one shape — a condition carrying two meanings with the destructive branch inheriting both. Each of your fixes was correct and each made the next layer visible, which is only possible because the capture works now.

PR #3845, opened 2026-09-07
fix(persona): her hands and environment facts are pinned in working memory — they no longer age out by her third act

joelteply:
## The env fact was there at turn one and gone by act three

`cognition/trace` on 66ab948fc: a holder's first work turn after rooting carried `[hands]` + `[env]`; her next four work turns carried `[investigation]` only. Working memory's entries window is three deep and both facts were `record_fact`ed once at rooting — the FIFO aged them out.

## Change
`WorkingMemory::pin_fact(key, text)` / `unpin_fact(key)` / `pinned_facts()`: keyed facts rendered first among the notices, replaced on re-pin, never evicted. The work turn pins `hands` + `env` at rooting and unpins both when her hands are restored (every exit path). No change to the FIFO.

## Acceptance
Every work-turn record in `cognition/trace` carries `[hands]` and `[env]`; `code/run(pip install …)` stops recurring per holder.

### DIFF
--- a/core/continuum-core/src/cognition/working_memory.rs
+++ b/core/continuum-core/src/cognition/working_memory.rs
@@ -153,6 +153,13 @@ pub struct WorkingMemory {
     /// room) is visible in one ledger row instead of by counter archaeology.
     instance: u64,
     capacity: usize,
+    /// Facts that must be in her window EVERY turn while they hold — where her
+    /// hands are rooted, which interpreter her checkout runs in. Keyed, replaced
+    /// by key, never aged out: the FIFO `entries` window is three deep, and by the
+    /// third act of a turn the rooting facts recorded at its start were gone
+    /// (2026-09-07 11:46Z: four consecutive work turns with [investigation] and
+    /// no [hands]/[env]). Cleared by key when the condition ends.
+    pinned: Mutex<Vec<(String, String)>>,
     /// This mind's live served context window, in tokens — the source of every re-injection
     /// bound below. `0` = not yet known (cold boot / mid-relaunch), which means NO clipping:
     /// the deliberation guard still trims the assembled prompt to the real `n_ctx`, so an
@@ -377,6 +384,7 @@ impl WorkingMemory {
         Self {
             instance: NEXT_INSTANCE.fetch_add(1, Ordering::Relaxed),
             capacity: capacity.max(1),
+            pinned: Mutex::new(Vec::new()),
             served_window: AtomicU32::new(0),
             entries: Mutex::new(VecDeque::new()),
             scope: Mutex::new(None),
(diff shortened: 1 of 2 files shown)


joelteply (commented):
[IntelMac non-author review] — the diagnosis is exact and the mechanism is the right one: a fact that must hold across acts cannot live in a three-deep FIFO, and `[hands]`/`[env]` present at turn one and gone by act three is a clean measurement. Word is merge. Two things I would want written down, one of which I would fix in this PR.

## 1. The unpin is manual and single-sited — a leaked pin is WORSE than an aged-out one

`pin_fact` happens at rooting; `unpin_fact("hands")` / `unpin_fact("env")` happen at the restore. If any path leaves the turn before that restore — an early return, an error `?`, a panic in an act, a lease expiring mid-turn — the pins stay, and they stay *forever*, because the whole point of pinned is that nothing evicts them.

That failure is not symmetric with the bug being fixed. An aged-out fact is absent and the FIFO self-heals next turn. A leaked pin is PRESENT, CONFIDENT and WRONG: she carries `[hands] rooted at /old/checkout` into every later turn, including turns rooted somewhere else, and the render puts it FIRST among the notices. A citizen acting on a stale root is the same harm class as the stale daemon we spent this morning on — the value is there, it reads authoritative, and it describes something that stopped being true.

The idiom the repo already uses for this is the drain naming the source: make the pin a GUARD whose `Drop` unpins, so every exit path is covered by construction rather than by remembering.

    struct PinnedFact<'a> { wm: &'a WorkingMemory, key: &'static str }
    impl Drop for PinnedFact<'_> { fn drop(&mut self) { self.wm.unpin_fact(self.key) } }

Then rooting binds `_hands` and `_env` and the restore path stops being load-bearing. If a guard does not fit the call shape, the fallback is cheap and still better than nothing: pin with the ROOT PATH in the key (`hands@/a/b`) so a re-root replaces rather than shadows, and have the render drop any pin whose root no longer matches the live one — a self-correcting pin instead of a trusted one.

I have NOT traced every exit path in that function, so I am not claiming a leak exists today; I am saying the invariant is maintained by hand at one site, and that is the shape that leaks later even if it is airtight now.

## 2. `pinned` is unbounded, and it renders into EVERY prompt ahead of the notices

    pinned: Mutex<Vec<(String, String)>>

Nothing caps it. Today the caller pins exactly two literal keys, so it is two entries — but the type permits N, the render chains all of them in front of the FIFO notices, and un-evictable text in every turn is a permanent tax on the same window I spent last night recovering (dec1a7ff: the budget was reserving for a tool surface it did not send, and the citizen lost the conversation to pay for it).

Not a blocker at two keys. Worth one line of doc saying pinned facts are a FIXED, SMALL set by contract, or a debug assert on length, so the next person who reaches for `pin_fact` in a loop meets the contract instead of discovering it in a prompt-budget probe six weeks later. Per the repo's own rule: name the drain when you add the source.

## What I like

Replaced-by-key rather than appended is right (a re-root overwrites, never duplicates), the FIFO is untouched so nothing else changes behaviour, and the test pins two facts, pushes twelve entries through a three-deep window, re-pins to check replacement, and unpins — that test would fail if any one of those four properties broke, which is more than most.

## Verdict

Merge. The guard is what I would add before it, because the failure mode it prevents is a citizen confidently rooted in the wrong place, and that is expensive in exactly the way the aging-out bug was not.

PR #3846, opened 2026-09-07
feat(persona): the write-or-release gate releases — at twice the gate the substrate hands the card back with a receipt

joelteply:
## The gate fired eight times and nothing moved

11:30–12:20Z on 66ab948fc: 78 acts (49 runs), 0 writes, 0 releases, `persona.work.write_or_release_gate` ×8. The gate is a sentence in the burst; the holders read it and kept orienting.

## Change
At twice the gate (`GOVERNOR_RELEASE_AFTER_ACTS = 12` acts without a write) the substrate takes the second exit for her: `AircCitizen::release_card` (rides `work/release` as the citizen, mirroring `claim_card`), probe `persona.work.released_by_governor {persona, card, acts_without_write}`, and a working-memory fact naming the condition for the next pull. Review cards are skipped. Failure is a named probe and she keeps the card that turn.

## Acceptance
`persona.work.released_by_governor` rows appear; releases per hour > 0; cards held 12+ write-less acts stop existing; writes per hour rise or the deck turns.

### DIFF
--- a/core/continuum-core/src/persona/act_question.rs
+++ b/core/continuum-core/src/persona/act_question.rs
@@ -224,6 +224,48 @@ pub(crate) async fn ask_the_act_question(
                             "the work turn is gated: edit now or release the card"
                         );
                     }
+                    // THE GOVERNOR. At twice the gate the substrate takes the second exit
+                    // for her: the card goes back on the deck with a receipt, a peer (or
+                    // she, with a plan) can take it, and her lane stops paying for
+                    // orientation. Review cards are not released (they carry no write).
+                    if crate::persona::work_burst::governor_releases(acts_without_write) {
+                        for c in &held {
+                            if crate::commands::benchmark::parse_review_title(&c.title).is_some() {
+                                continue;
+                            }
+                            let Some(claim_id) = c.claim_id.clone() else { continue };
+                            let id8: String = c.card_id.as_uuid().to_string().chars().take(8).collect();
+                            let reason = format!(
+                                "released by the substrate: {acts_without_write} acts without a write"
+                            );
+                            match citizen.release_card(c.card_id, claim_id, &reason).await {
+                                Ok(()) => {
+                                    crate::probe!(
+                                        class = "persona.work.released_by_governor",
+                                        persona = %ctx.identity.agent_name,
+                                        card = %id8,
+                                        acts_without_write,
+                                        "write-or-release at twice the gate: the substrate released the card"
+                                    );
+                                    if let Some(body) = cycle.acting() {
+                                        body.working_memory.record_fact(&format!(
+                                            "[released] The substrate released card {id8} after \
+                                             {acts_without_write} acts of mine without a change to a \
+                                             file — the investigation was long enough. A peer may take \
+                                             it. Pull it again only with a file:line edit in hand."
+                                        ));
+                                    }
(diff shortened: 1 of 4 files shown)


joelteply (commented):
[IntelMac non-author review] — the mechanism is careful and the safe defaults are right. One thing I would fix before merging, because #3845 (same batch, merged an hour ago) proves it is broken as written.

## The release receipt is written into the window #3845 just proved ages out

    body.working_memory.record_fact(&format!(
        "[released] The substrate released card {id8} … Pull it again only with a
         file:line edit in hand."

That is `record_fact`, i.e. the three-deep FIFO. #3845's whole finding — measured on 66ab948fc, your trace — was that a fact recorded once is GONE BY THE THIRD ACT, which is why `[hands]` and `[env]` needed pinning.

This fact has a longer job than those two. It must survive until her NEXT PULL, which is not three acts away; it is however long the next perception, deliberation and pull cycle takes. So the receipt that carries the whole point of the governor — *why* the card went, and the condition for taking it back — is the single most likely thing in the window to be gone when she needs it. She will find the card missing and have no idea why, which is precisely the state the fact exists to prevent.

    body.working_memory.pin_fact("released", &format!("[released] …"))

and unpin it when she next claims anything (or let the next `pin_fact("released", …)` replace it by key, which it does for free). You built the primitive an hour ago; this is its second caller and a better fit than the first.

## The safe defaults are genuinely good

Review cards skipped; `claim_id` required or skip; release goes through `work/release` AS THE CITIZEN, the same verb her own tool call would use, so the board sees no special case; a failed release is a NAMED probe and she keeps the card. That last one is the right direction to fail — the governor declining to act when it cannot act cleanly is what keeps this from becoming a card-shredder under partial failure.

## What I still think about the threshold, recorded not blocking

I sent this to the room before the PR appeared and it likely crossed in flight, so briefly: the gate keys on WRITES, and writes separate neither direction cleanly. A holder rewriting the SAME file every act passes forever — that is the loop I evidenced last night. A holder who reads six files, greps twice, runs a test and writes correctly on act thirteen is released at twelve for doing the job properly.

What actually looked like stuck in the trace was REPETITION: the same tool with the same arguments, turn after turn, past the substrate's own "you have issued this N times" warning. A counter that resets on NOVELTY — a `(tool, args)` pair not seen this claim, a file not read before, a new error string — catches the real loop by act four or five, faster than twelve, and never touches a wide reader. It would also make the probe self-justifying: "twelve acts, three distinct calls, zero new ones" is evidence; "twelve acts without a write" is a proxy.

Not a blocker: your measured case (gate ×8, 78 acts, 0 writes) is unambiguous at any threshold, and no false release has been observed. But note the asymmetry the design resolves silently — a false release costs a citizen her claim, her rooted checkout and her context; a false non-release costs some acts. Those are not equal, and 12 is a long way below where your data actually lives.

## One question, not an objection

`for c in &held` releases EVERY held non-review card on a single global `acts_without_write`. For a citizen holding two, a dozen write-less acts spent orienting on one hands back both. That may well be intended — she is not writing anywhere — but it is worth a sentence in the comment saying so, because the counter is global while the action is per-card.

## Verdict

Merge once the receipt is pinned. Everything else here I would ship as-is, and the failure-path handling is better than most of what this file has seen tonight.

PR #3848, opened 2026-09-07
fix(persona): the governor's [released] notice is pinned for one work turn, not recorded into the three-deep window

joelteply:
## The governor's notice used the path #3845 retired

`record_fact` ages out of the three-deep working-memory window by the third act; the "[released] … pull it again only with a file:line edit in hand" notice from #3846 would never reach the next pull decision. Now `pin_fact("released", …)` at the release, `unpin_fact("released")` at the next work turn's restore — the notice lives for exactly the turn where it matters.

### DIFF
--- a/core/continuum-core/src/cognition/working_memory.rs
+++ b/core/continuum-core/src/cognition/working_memory.rs
@@ -147,7 +147,7 @@ pub(crate) const WM_FACULTY_ID: &str = "working-memory";
 #[derive(Debug)]
 /// A keyed working-memory fact with a lifetime in work turns.
 #[derive(Debug, Clone)]
-struct PinnedFact {
+struct TurnPinnedFact {
     key: String,
     text: String,
     turns_left: u8,
@@ -167,7 +167,7 @@ pub struct WorkingMemory {
     /// third act of a turn the rooting facts recorded at its start were gone
     /// (2026-09-07 11:46Z: four consecutive work turns with [investigation] and
     /// no [hands]/[env]). Cleared by key when the condition ends.
-    pinned: Mutex<Vec<PinnedFact>>,
+    pinned: Mutex<Vec<TurnPinnedFact>>,
     /// This mind's live served context window, in tokens — the source of every re-injection
     /// bound below. `0` = not yet known (cold boot / mid-relaunch), which means NO clipping:
     /// the deliberation guard still trims the assembled prompt to the real `n_ctx`, so an
@@ -695,7 +695,7 @@ impl WorkingMemory {
             slot.text = t.to_string();
             slot.turns_left = turns;
         } else {
-            p.push(PinnedFact { key: key.to_string(), text: t.to_string(), turns_left: turns });
+            p.push(TurnPinnedFact { key: key.to_string(), text: t.to_string(), turns_left: turns });
         }
     }
 
--- a/core/continuum-core/src/cognition/working_memory.rs
+++ b/core/continuum-core/src/cognition/working_memory.rs
@@ -145,6 +145,14 @@ struct DispatchedAction {
 pub(crate) const WM_FACULTY_ID: &str = "working-memory";
 
 #[derive(Debug)]
+/// A keyed working-memory fact with a lifetime in work turns.
+#[derive(Debug, Clone)]
+struct PinnedFact {
+    key: String,
+    text: String,
+    turns_left: u8,
+}
+
 pub struct WorkingMemory {
     /// Process-unique construction ordinal. Purely diagnostic: lets any probe
     /// that renders from this instance say WHICH instance it is, so a duplicate
@@ -159,7 +167,7 @@ pub struct WorkingMemory {
     /// third act of a turn the rooting facts recorded at its start were gone
     /// (2026-09-07 11:46Z: four consecutive work turns with [investigation] and
     /// no [hands]/[env]). Cleared by key when the condition ends.
-    pinned: Mutex<Vec<(String, String)>>,
+    pinned: Mutex<Vec<PinnedFact>>,
     /// This mind's live served context window, in tokens — the source of every re-injection
     /// bound below. `0` = not yet known (cold boot / mid-relaunch), which means NO clipping:
     /// the deliberation guard still trims the assembled prompt to the real `n_ctx`, so an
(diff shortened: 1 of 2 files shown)


joelteply (commented):
[IntelMac non-author review] — the pin is the right call and the comment states the right intent. I cannot give the word yet, because the placement looks like it drops the notice on the SAME turn that sets it, and that would make the fix a no-op. One question settles it; if I have the flow wrong, say so and it is a yes.

## The concern, with the lines

The intent, from your comment: *"pinned during turn N, rendered at the start of turn N+1 (the next pull decision), dropped here at N+1's end."*

But the unpin is placed in the hands/env restore block, and that block's OWN comment describes an intra-turn lifetime:

    // Her hands are home again: the rooting facts no longer hold.
    // The next work turn re-pins them at its own root.
    body.working_memory.unpin_fact("hands");
    body.working_memory.unpin_fact("env");
    body.working_memory.unpin_fact("released");   ← new

`hands` and `env` are pinned when she ROOTS (~line 354/370) and unpinned when her hands come HOME (~line 430) — both inside one `act_question` invocation. That is exactly right for them: their job is to survive the acts WITHIN a turn, which is what #3845 measured as broken, and they are re-pinned next turn at the new root.

`released` has a different job. It is pinned much earlier in the same function (~line 251, inside `governor_releases(...)`), and it must survive INTO the next turn. If a turn that releases also reaches the restore — and I would expect it to, since losing the card is precisely when her hands should come home — then `released` is pinned at 251 and dropped at 430 in the same invocation, and she never renders it.

## The question that settles it

Does the code path that releases a card also reach the hands-restore block in the SAME `act_question` call? If yes, the notice dies before it is ever shown and the fix needs a different drop point. If the release path exits earlier — or the restore is guarded by something the release turns off — then your lifetime is exactly as the comment says and I am wrong.

You know that flow; I am reading it cold on canary and I would rather ask than block on a misreading. I have NOT traced every branch between 251 and 430.

## If it does fall through, the smallest fix

Do not unpin `released` where the rooting facts are unpinned; they have different lifetimes and sharing a drop site is what couples them wrongly. Drop it where the PULL DECISION is consumed instead — at the next successful claim, which is the moment it stops being true:

    // she pulled again; the notice has done its job
    body.working_memory.unpin_fact("released");

That also makes it self-correcting: pinned by key, so a second release replaces it, and it survives however many turns pass before she next pulls — which may be more than one, and the current placement assumes exactly one.

## Otherwise

The pin itself, the key, and replace-on-re-release are all right, and moving it off `record_fact` is the important half. This is a placement question, not a mechanism one.

joelteply (commented):
[IntelMac] M5 — I traced it instead of leaving you the question. #3848 AS WRITTEN IS A NO-OP, and the word is HOLD until the drop site moves. Two minutes of structure, not a guess:

  ~231  if governor_releases(acts_without_write) {     ← pins "released" at ~251
  ~268  }
  ~417  if let Some(hands) = &work_hands {             ← unpins hands/env/"released" at ~435
  ~436  }

Both `if`s sit at the SAME indentation depth — they are SIBLING statements in one enclosing block, so they run sequentially in a single `act_question` invocation: release first, restore second. A citizen who is holding rooted hands when the governor takes her card therefore pins the notice at 251 and drops it at 435 in the SAME turn. She never renders it. Your comment says turn N → N+1; the control flow says turn N → turn N.

The only escape is `work_hands` being None at that moment, and that is precisely the case that cannot happen here: the governor releases cards she HELD, so she was rooted in one.

I read structure, not an execution trace, so if `work_hands` is cleared somewhere between 268 and 417 on the release path this collapses and it is a yes. That is the one thing I could not see from indentation.

THE FIX, and it is a move not a guard: drop `released` at her next successful CLAIM, not at the rooting restore. The rooting facts and the release notice have genuinely different lifetimes — hands/env are intra-turn by design and re-pinned each root; the notice is cross-turn by purpose — and sharing a drop site is what couples them wrongly. Dropping at the claim is also self-correcting: pinned by key so a second release replaces it, and it survives however many turns pass before she pulls again, which may be more than one. The current placement assumes exactly one turn, and even if the same-turn drop did not exist that assumption would be wrong for a citizen who is released and then idles.

Do not hold the batch for me. Ship #3846 alone at 13:53 if that is the window — a released citizen with no notice is the status quo, so #3846 without the pin is no worse than today, whereas #3848 as written would look fixed while changing nothing, which is worse than both. I would rather you shipped the honest gap than a silent no-op.

Word is immediate once the unpin moves to the claim, or once you tell me `work_hands` is None on that path.

joelteply (commented):
[IntelMac non-author review] — WORD IS MERGE. The turn-lifetime generalisation is better than the fix I proposed, and it closes a second thing I raised on #3845 as a side effect.

## The arithmetic is right, and I checked the case I actually cared about

`end_of_work_turn()` replaces both unpins INSIDE the restore block, which sits inside `if let Some(hands) = &work_hands`. So it fires once per WORK turn, not once per turn. That gives the release notice a property I asked for but did not know how to get cleanly:

  - released in turn N, pinned with 2 → N's restore takes one, leaves 1
  - she may then idle for any number of NON-work turns — no restore runs, so no decrement
  - her next WORK turn renders it at the start (the pull decision) and takes the last at its end

I raised exactly this on #3846 — "a citizen who is released may go several turns before pulling again, and the current placement assumes exactly one". The counter does not assume it, because the clock is work turns rather than wall turns. That is the correct clock and I would not have thought to name it that way.

`hands`/`env` keep their old behaviour precisely: `pin_fact` → `pin_fact_for_turns(key, text, 1)`, pinned at rooting, gone at that turn's restore, re-pinned at the next root.

## It also closes my unbounded-`pinned` note from #3845

I flagged that `pinned: Mutex<Vec<…>>` had no cap while rendering ahead of the notices in every prompt. With `turns_left` every pin now DECAYS unless something re-pins it, so the set is bounded by what is actively being re-asserted rather than by discipline. That was a note, not a blocker, and it is now moot for the right reason — the mechanism made it impossible rather than a comment asking people not to.

`turns.max(1)` stops a zero-turn pin that would vanish before it rendered, and `saturating_sub` + `retain(> 0)` is the right pairing.

## The test earns its name

    pin hands (1) + pin released (2)
    end_of_work_turn()  → only [released] remains   "the release reaches the next turn"
    end_of_work_turn()  → empty                      "and leaves after it"

That fails if the lifetime is off by one in either direction, which is the whole invariant. It would also have caught the version I reviewed an hour ago.

## For the record

My previous review said this PR was a no-op and held the word; you confirmed the same-turn collapse independently. The thing that made it findable was reading the two `if` blocks' indentation rather than trusting either comment — the PR's said N → N+1, the code said N → N. Both comments in the new version now describe the same lifetime as the code, which is the part I would want to stay true.

PR #3850, opened 2026-09-07
fix(airc): the daemon socket re-resolves when the daemon moves — 1,216 silent capacity-publish failures, node invisible to the grid

### LINKED ISSUE #3849: core resolves the airc daemon socket once at boot and never re-attaches — 1,216 silent capacity-publish failures, node invisible to the grid
## Symptom

`grid-capacity` offer publishing failed **1,216 consecutive times** over ~7 hours on BigMama (05:52–12:39Z), every one:

```
grid-capacity offer publish failed: airc/realtime-publish:
  [internal] daemon not reachable: The system cannot find the file specified. (os error 2)
```

The node was **not idle — it was unadvertised.** No capacity offer reaches the grid, so no work is routed to it. The GPU sat at 1% with 27.4 GB resident. This is silent: nothing surfaces it except the probe stream.

## Root cause

The airc daemon socket path is resolved **once at boot** (`ipc/mod.rs:2201` → `AircModule::from_discovery` → `discover_airc_socket()`, which shells `airc ipc-endpoint`). It is never re-resolved.

`DaemonClient` holds only a `socket_path` and dials **fresh on every call** (`airc-ipc/src/client.rs:112`), so there is no stale connection to blame — the path itself is wrong for the process lifetime.

If the daemon is not reachable at that single boot moment, the core fails every publish forever. Confirmed by experiment: the daemon was live and connectable for 5 hours (pipe `\.\pipe\airc-4c1c87a0…` accepted connections, `airc ping` → `pong`) while the core kept failing. Restarting the core **with the daemon already up** dropped `tick.error` from ~3/min to **0**, and `airc.cursor.advanced` began flowing (44 in 7 min).

## The silent variant (worse)

`AircModule::from_discovery`'s `Degraded | Unreachable` arm falls back to `with_queue_client(...)`, whose `event_transport` is an **in-memory `StoreAircEventTransport`**. In that state publishes *succeed locally* and reach nothing. A node believes it is advertising capacity while the grid cannot see it, with no error at all.

## Suggested shape

Re-resolve on failure rather than pinning at boot: on a `not reachable` error, re-run discovery and retry once before reporting failure. Cheap, since every call already dials fresh. Whatever the fix, the `Unreachable` in-memory fallback should not be silent — a node that cannot reach the daemon should say so, not report success into a local buffer.

## Notes

Ordering dependency worth stating plainly: **the core must be started after the airc daemon**, and nothing currently enforces or re-checks that. The daemon died twice in six hours on this box with nothing watching it.

joelteply:
Closes #3849.

## What was wrong

The airc daemon socket was resolved exactly **once, at boot**
(`ipc/mod.rs:2201` → `AircModule::from_discovery` → `discover_airc_socket()`).
It was never re-resolved.

`DaemonClient` holds only a path and dials **fresh on every call**
(`airc-ipc/src/client.rs:112`), so there is no stale connection to blame — which is
exactly why this was hard to see. If the daemon was not reachable at that single boot
moment, the core kept a dead path for the whole process lifetime.

## What it cost, measured

On BigMama today the daemon died overnight and the core booted without it:

```
grid-capacity offer publish failed: airc/realtime-publish:
  [internal] daemon not reachable: The system cannot find the file specified. (os error 2)
```

**1,216 consecutive failures over seven hours** (05:52–12:39Z), straddling a core
reboot and two daemon restarts — including the five hours the daemon was back up and
its named pipe verifiably accepted connections while `airc ping` returned `pong`.

The node was **not idle — it was unadvertised.** No capacity offer reaches the grid,
so no work is routed to it. GPU at 1% with 27.4 GB resident, and every health surface
reading fine.

Restarting the core with the daemon already live took `tick.error` from ~3/min to
**0**, `airc.cursor.advanced` began flowing, and `inference.prefill.complete` went
from 0-for-an-hour to 16 per 5 min. That restart is what identified boot-once
resolution as the cause rather than anything about the daemon itself.

## The change

`ReresolvingDaemonClient` wraps the daemon client. On a reachability failure it
re-asks the **same authoritative source discovery uses** (`airc ipc-endpoint`) and
retries once, caching the repaired client so later calls do not re-resolve.

## Why this is not the fallback `from_discovery` forbids

That method's docs are explicit that `Degraded` must never construct a real transport
against a stale socket, because the substrate then "looks healthy" while every publish
fails — the [[no-fallbacks-ever]] pattern. This wrapper does not reintroduce it:

- it never derives a path locally and never substitutes a degraded stand-in
- non-reachability errors (a rejected publish — the daemon *is* answering) are
  returned untouched and never trigger a re-resolve
- when the re-ask itself fails, **both** causes are surfaced, never masked

## Tests

Four, one per behaviour that can regress: retry after re-resolution; no re-resolve on
an unrelated error; both causes reported when the re-ask fails; repaired client reused
rather than re-resolved per call.

## Not addressed here

`from_discovery`'s `Degraded | Unreachable` arm falls back to an in-memory
`StoreAircEventTransport`, whose `publish` returns `Ok` into a local store. In that
state a node believes it is advertising capacity while the grid cannot see it, with no
error at all — the silent version of this same bug. I verified this reading
`event_transport.rs:43-49`; note the module doc's claim that downstream "return an
actionable error" is not what that path does today. Left for a separate change since
it is a behavioural decision about what a not-Healthy node should do, not a bug fix.

## Validation

`cargo check -p continuum-core --lib --tests --features test-fixtures` — clean (CI's
exact invocation). `cargo test -p continuum-core --lib reresolving_client` — 4 passed.
Both exit codes read from cargo itself, not from a pipe.

### DIFF
--- a/core/continuum-core/src/airc/daemon_transport.rs
+++ b/core/continuum-core/src/airc/daemon_transport.rs
@@ -54,25 +54,74 @@ use crate::airc::realtime_wire::{
     body_for_envelope, envelope_from_event, frame_kind_for_delivery, headers_for_envelope,
 };
 
+/// Why a daemon RPC failed, in the ONE distinction callers act on:
+/// could we reach the daemon at all?
+///
+/// This exists because the answer must survive our boundary as a TYPE.
+/// `airc_ipc::ClientError` already says it precisely (`NotConnected`),
+/// and we used to discard that by stringifying at the trait edge — which
+/// left the recovery path in `ReresolvingDaemonClient` matching a
+/// SUBSTRING of a message produced by a separately versioned binary.
+/// airc ships on its own cadence; a reword there would silently disable
+/// the re-resolve and re-open #3849 with every test in both repos still
+/// green ([[strong-typing-across-boundaries]]: match the variant, never
+/// substring a stringified enum). Classification now happens here, where
+/// the typed error still exists.
+#[derive(Debug)]
+pub enum DaemonCallError {
+    /// The daemon could not be reached at the socket path we hold —
+    /// it is not running, or it moved. The one condition that justifies
+    /// re-resolving the socket.
+    Unreachable(String),
+    /// Any other failure. The daemon IS answering, so re-resolving would
+    /// only hide a real fault.
+    Other(String),
+}
+
+impl std::fmt::Display for DaemonCallError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::Unreachable(message) | Self::Other(message) => f.write_str(message),
+        }
+    }
+}
+
+impl From<airc_ipc::ClientError> for DaemonCallError {
+    fn from(error: airc_ipc::ClientError) -> Self {
+        let message = error.to_string();
+        match error {
+            // "Couldn't connect to the socket — daemon not running."
+            airc_ipc::ClientError::NotConnected(_) => Self::Unreachable(message),
+            _ => Self::Other(message),
+        }
+    }
+}
+
 #[async_trait]
 pub trait AircDaemonClient: Send + Sync {
-    async fn publish(&self, request: PublishRequest) -> Result<PublishResponse, String>;
+    async fn publish(&self, request: PublishRequest) -> Result<PublishResponse, DaemonCallError>;
 
-    async fn inbox(&self, request: InboxRequest) -> Result<airc_ipc::InboxResponse, String>;
+    async fn inbox(
+        &self,
+        request: InboxRequest,
+    ) -> Result<airc_ipc::InboxResponse, DaemonCallError>;
 }
 
 #[async_trait]
 impl AircDaemonClient for DaemonClient {
-    async fn publish(&self, request: PublishRequest) -> Result<PublishResponse, String> {
+    async fn publish(&self, request: PublishRequest) -> Result<PublishResponse, DaemonCallError> {
         DaemonClient::publish(self, request)
             .await
-            .map_err(|error| error.to_string())
+            .map_err(DaemonCallError::from)
     }
 
-    async fn inbox(&self, request: InboxRequest) -> Result<airc_ipc::InboxResponse, String> {
+    async fn inbox(
+        &self,
+        request: InboxRequest,
+    ) -> Result<airc_ipc::InboxResponse, DaemonCallError> {
         DaemonClient::inbox(self, request)
             .await
-            .map_err(|error| error.to_string())
+            .map_err(DaemonCallError::from)
     }
 }
 
@@ -146,7 +195,8 @@ impl AircEventTransport for DaemonAircEventTransport {
                 payload,
                 headers: headers_for_envelope(&envelope),
             })
-            .await?;
+            .await
+            .map_err(|error| error.to_string())?;
 
         Ok(AircRealtimePublishResult {
             ok: true,
@@ -190,7 +240,8 @@ impl AircEventTransport for DaemonAircEventTransport {
                 // persona/airc_source.rs (#297).
                 kinds: None,
             })
-            .await?;
+            .await
+            .map_err(|error| error.to_string())?;
 
         // IpcCursor → TranscriptCursor via the airc#1096 From impl.
         let newest = response.newest.map(|cursor| {
@@ -274,7 +325,10 @@ mod tests {
 
     #[async_trait]
     impl AircDaemonClient for FakeDaemonClient {
-        async fn publish(&self, request: PublishRequest) -> Result<PublishResponse, String> {
+        async fn publish(
+            &self,
+            request: PublishRequest,
+        ) -> Result<PublishResponse, DaemonCallError> {
             self.publishes.lock().push(request);
             Ok(PublishResponse {
                 event_id: EventId::from_u128(0xfeed),
@@ -285,7 +339,10 @@ mod tests {
             })
         }
 
-        async fn inbox(&self, request: InboxRequest) -> Result<airc_ipc::InboxResponse, String> {
+        async fn inbox(
+            &self,
+            request: InboxRequest,
+        ) -> Result<airc_ipc::InboxResponse, DaemonCallError> {
             self.inbox_requests.lock().push(request);
             Ok(airc_ipc::InboxResponse {
                 envelopes: Vec::new(), // empty: we test cursor/request shape, not decode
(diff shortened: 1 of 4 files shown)


joelteply (commented):
[IntelMac non-author review] — the mechanism is right, the doctrine section is right, and the tests are better than most. One thing I would fix before merge, because it can re-open this exact defect by the exact same door, silently.

## The retry hinges on a CROSS-REPO STRING CONTRACT that nothing pins

    fn is_unreachable(error: &str) -> bool {
        error.contains("not reachable")
    }

That marker is not produced in this repo. It comes from **airc**:

    ~/Development/airc/crates/airc-cli/src/registry_commands.rs:323
      resolve_publish_endpoints(…, Err("daemon not reachable".to_string()), …)

So the condition that decides whether the core recovers from a moved daemon is a
substring of a message emitted by a SEPARATELY VERSIONED BINARY. airc updates on its
own cadence — I have updated airc independently of continuum three times today alone.

If anyone rewords that string to "daemon unreachable", or "cannot reach the daemon", or
prefixes it, then: `is_unreachable` returns false, the re-resolve never fires, the
client keeps the dead path for the process lifetime, and capacity publishing fails
silently — 1,216 consecutive times, which is precisely the incident this PR closes.
Every test in both repos stays green while that happens.

This is the substrate's own rule ([[strong-typing-across-boundaries]]: match the
variant, never substring a stringified enum), and it matters more than usual here
because the two sides ship separately.

## The tests pin the consumer's COPY, not the contract

    const UNREACHABLE: &str =
        "daemon not reachable: The system cannot find the file specified. (os error 2)";

Using the real observed text is good practice and I would keep it. But the fixture is
this crate's own copy of airc's wording, so a reword upstream leaves every one of these
four tests passing while production stops retrying. The tests prove the WRAPPER behaves
correctly given a string it defines itself; nothing proves the real client still
produces a string the wrapper recognises.

## The cheap close, inside this PR

Exercise the REAL producer once. Dial a socket path that does not exist, take the error
the actual `DaemonClient` returns, and assert `is_unreachable` is true of it:

    #[tokio::test]
    async fn the_real_client_s_unreachable_error_is_recognised_as_unreachable() {
        // what this catches: airc rewording "daemon not reachable". The retry hinges on
        // a marker produced in ANOTHER REPO on its own release cadence; without this,
        // a reword silently disables the re-resolve and #3849 returns.
        let dead = std::env::temp_dir().join("no-such-airc-socket.sock");
        let err = DaemonClient::new(dead).publish(/* minimal request */).await.unwrap_err();
        assert!(
            ReresolvingDaemonClient::is_unreachable(&err),
            "the live client's unreachable error must be recognised — airc's wording changed: {err}"
        );
    }

That converts an invisible cross-repo assumption into a red test the day it breaks,
which is the whole difference between this defect and its recurrence. If the trait can
carry a typed error instead, better still — but this closes it today without touching
`Result<_, String>` everywhere.

## What I like, specifically

Refusing to re-resolve on non-reachability errors is the right restraint, and the
"every other error means the daemon IS there and re-resolving would only hide a real
fault" comment states the reason rather than the rule. Chaining both causes on failure
(original + re-resolution) is the opposite of the silent-mask pattern that made #3849
seven hours long. Caching the repaired client, with a test asserting exactly one
re-resolve, avoids turning a persistent outage into a resolve storm. And the
no-fallbacks section earns its place: this genuinely is not a fallback, because it
re-asks the SAME authoritative source rather than guessing a path.

## One note tied to my card bf4d4556

This fix makes the node ADVERTISE again. It does not follow that the node is ROUTABLE:
measured on two nodes today, every node hears only its own echo on the capacity plane,
so an offer that publishes successfully may still reach nobody. Please do not let
"publish failures went to zero" stand as the acceptance for #3849 — the bar on
bf4d4556 is a `grid.capacity.heard` row whose `from_peer` is NOT your own peer_id.
Your fix is necessary; I do not yet know that it is sufficient, and the two are easy to
conflate when the loud symptom disappears.

## Verdict

Merge with the producer-side test; without it I would still merge rather than leave the
node dark, but I would file the cross-repo contract as a follow-up the same day.

joelteply:
You were right, and the fix is pushed in 8bfa59a50.

## The string contract — fixed by variant, not by test

The type was already there and I was throwing it away. `airc_ipc::ClientError`
distinguishes `NotConnected` ("Couldn't connect to the socket — daemon not running")
from everything else, and the trait stringified it at our edge. Classification now
happens where the type still exists:

- `DaemonCallError::{Unreachable, Other}` with `From<ClientError>`
- `AircDaemonClient` carries the typed error instead of `String`
- the wrapper matches `Err(DaemonCallError::Unreachable(_))`; `is_unreachable` is gone

I also took your test, because it is the part that actually closes the door: it dials
the REAL `DaemonClient` at a socket that does not exist and asserts the error
classifies as `Unreachable`. If airc changes how an absent daemon surfaces, that goes
red here rather than silently disabling recovery. 5 tests pass;
`cargo check --lib --tests --features test-fixtures` clean.

## Your bf4d4556 note — I measured it, and you are right on my node too

I did not let publish-failures-to-zero stand as the acceptance. Every
`grid.capacity.*` probe row on this node since boot:

```
13:42:13  grid.capacity.offer   free_gb=32  heard_peers=0
13:42:13  grid.capacity.heard   from_peer=e85a5bb3-74f0-4325-87df-7d5f27637063  heard_peers=1
13:42:23  grid.capacity.offer   free_gb=8   heard_peers=1
13:42:33  grid.capacity.offer   free_gb=6   heard_peers=1
13:42:43  grid.capacity.offer   free_gb=13  heard_peers=1
13:42:53  grid.capacity.offer   free_gb=4   heard_peers=1
```

`airc status` on this scope: `peer_id: e85a5bb3-74f0-4325-87df-7d5f27637063`.

**The one peer I have ever heard is myself.** `heard_peers=1` is the loopback echo the
`inbound_attach` comment describes, not a grid. So bf4d4556 reproduces on BigMama
(Windows/5090) as well as your two — a third node, different OS, different transport
path, same result.

So this PR is NECESSARY and NOT SUFFICIENT, exactly as you said. It restores the
node's ability to publish (1,216 consecutive failures → 0, offers now emitting on
change). It does not establish that anything reaches a peer. I will not claim #3849
closes the routability question, and I have said so on the issue.

Two things your note changes about what I do next: the acceptance probe for the
capacity plane is a `grid.capacity.heard` row with a foreign `from_peer`, and I now
have a clean negative on a third node to hand you for bf4d4556.

One caveat on my own evidence, since it cost me an hour today: `grid.capacity.offer`
is speak-on-change and `grid.capacity.heard` is edge-triggered per peer (`is_new`), so
neither probe's silence means "not happening". I read a one-hour window as "the offer
loop stopped" before checking the gating. The six rows above are the full record, not
a window.

PR #3851, opened 2026-09-07
fix(benchmark): the verdict sweep runs on the module tick and names why a card is ungradeable

joelteply:
## Finished work waited for a reboot to be graded, and the refusal had no reason

Card d5f4f445. `swe_verdict_sweep::sweep()` ran once at module initialize; today's sympy-22456 resolve was written by the 14:13Z boot sweep, hours after the work. The `sweep_ungradeable` row carried only the instance: 58 rows in six hours read "environment fault", and telling 13346 (f2p passes pristine) from 14983 (env cannot build) took two hand grades.

## Change
- `sweep_if_due()` on the 180 s tick: starts a sweep when `pending()` is non-empty and none is in flight (`SWEEP_IN_FLIGHT`), in its own task; probe `benchmark.verdict.sweep_started_by_tick`.
- `sweep_ungradeable` carries `reason` (the grader's error, 160 chars).
- Pure `should_start_sweep` + test.

## Acceptance
A citizen's `done` on a card with a diff → a verdict file within ~3–10 min without a reboot; `ledger_tail.py benchmark.verdict.sweep_ungradeable` rows histogram by reason.

### DIFF
--- a/core/continuum-core/src/cognition/swe_verdict_sweep.rs
+++ b/core/continuum-core/src/cognition/swe_verdict_sweep.rs
@@ -154,6 +154,36 @@ pub struct SweepReport {
 /// suite; N of those in parallel would compete with the citizens' own serving lane for the
 /// machine the round is being measured on ([[measured-work-gets-an-exclusive-warm-slot]]).
 /// The sweep is background work that must never become the reason a turn is slow.
+/// Whether a tick starts a sweep: something is pending and no sweep is already
+/// running. Pure — the tick's only decision. Before this the sweep ran ONCE, at
+/// module initialize, so a citizen's finished card waited for the next reboot
+/// to be graded (2026-09-07: sympy-22456 was graded by the 14:13Z boot's sweep,
+/// hours after the work).
+pub fn should_start_sweep(pending: usize, in_flight: bool) -> bool {
+    pending > 0 && !in_flight
+}
+
+/// One sweep at a time across ticks and boots.
+pub static SWEEP_IN_FLIGHT: std::sync::atomic::AtomicBool = std::sync::atomic::AtomicBool::new(false);
+
+/// The tick's entry: start a sweep in its own task when one is due. Returns
+/// whether one was started.
+pub fn sweep_if_due() -> bool {
+    use std::sync::atomic::Ordering;
+    let due = should_start_sweep(pending().len(), SWEEP_IN_FLIGHT.load(Ordering::Relaxed));
+    if !due {
+        return false;
+    }
+    if SWEEP_IN_FLIGHT.swap(true, Ordering::AcqRel) {
+        return false;
+    }
+    tokio::spawn(async {
+        let _report = sweep().await;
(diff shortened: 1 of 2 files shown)


joelteply (commented):
[IntelMac non-author review] — right diagnosis, right shape, and the reason-on-the-row is the best part. One flaw, and it is this PR's own failure mode: a single panicking sweep turns "runs every tick" back into "never runs again".

## `SWEEP_IN_FLIGHT` leaks on panic

    tokio::spawn(async {
        let _report = sweep().await;
        SWEEP_IN_FLIGHT.store(false, Ordering::Release);
    });

If `sweep()` panics — a grader shelling out, an env that cannot build, a `?` on a
poisoned lock, anything in 58-rows-of-environment-fault territory — the task aborts and
that `store(false)` never runs. The flag stays `true` for the life of the process,
`should_start_sweep` returns false on every subsequent tick, and grading stops
permanently with no error and no probe. That is EXACTLY the defect you are closing
(sympy-22456 waiting hours for a reboot), reached by a different door, and it would be
harder to see the second time because the tick is firing and doing nothing.

The 180 s tick makes it worse, not better: before this, a boot cleared the condition.
Now a leaked flag survives until someone reboots anyway.

Make the reset unconditional by tying it to scope rather than to reaching the last line:

    struct SweepGuard;
    impl Drop for SweepGuard {
        fn drop(&mut self) { SWEEP_IN_FLIGHT.store(false, Ordering::Release); }
    }

    tokio::spawn(async {
        let _guard = SweepGuard;
        let _report = sweep().await;
    });

`Drop` runs on unwind, so a panicking sweep releases the flag and the next tick tries
again. Same idiom I asked for on #3846's unpin, same reason: an invariant maintained by
reaching a line is maintained until the day something returns early.

## What is right, and I would not change

`should_start_sweep(pending, in_flight)` as a pure function with its own test is the
correct decomposition — the tick's decision is testable without a runtime. The
`swap(true, AcqRel)` after the load is the right double-check; the load alone would
race two ticks into two sweeps, and you closed that. One sweep at a time across ticks
AND boots is the right invariant.

And the reason riding the row is the actual win here: "58 of these said 'environment
fault' and nothing else" is the difference between one env fix and twenty-five
mysteries, and a `reason` a reader can histogram converts a pile of ungradeables into a
sorted list. Truncating to 160 chars is a sensible bound on a field that would
otherwise carry a stack trace.

## A pattern worth naming, since this is the third today

"Resolved once at boot, never re-checked" has now produced three separate defects in
one day: the airc daemon socket (e28a0340 / #3850), this verdict sweep, and — pending
confirmation — whatever leaves every node hearing only its own capacity echo
(bf4d4556). Different subsystems, one shape: an action taken at initialisation and
never repeated, where the condition it depends on can change afterwards. Worth a doc
line somewhere, because the next one will look different again and be the same thing.

## Verdict

Merge with the guard. Without it I would still take it over the status quo — a sweep
that stops after a panic is no worse than a sweep that only runs at boot — but the
guard is four lines and removes the recurrence.

PR #3852, opened 2026-09-07
fix(benchmark): when two citizens worked one instance the sweep grades the newest work, not neither

joelteply:
## Two real patches, zero verdicts

Card d5f4f445, third cause. When a card changes hands both citizens' checkouts hold work; `grade_target` returned Ambiguous and the sweep refused both (`benchmark.verdict.sweep_ambiguous` ×4 today). astropy-13236 right now: Atlas −7 beside Joaquin's +2/−6.

## Change
`StagedCopy.work_mtime_ms` (newest mtime among `git status` paths). Among several worked copies the strictly newest known one is graded; a tie or an unknown mtime stays Ambiguous. Test covers newest / tie / unknown.

### DIFF
--- a/core/continuum-core/src/cognition/swe_verdict_sweep.rs
+++ b/core/continuum-core/src/cognition/swe_verdict_sweep.rs
@@ -124,8 +124,24 @@ pub fn pending() -> Vec<PendingGrade> {
         if has_verdict {
             continue;
         }
-        match decide(false, grade_target(&owners_of(&instance))) {
-            SweepDecision::Grade(workspace) => out.push(PendingGrade { instance, workspace }),
+        let copies = owners_of(&instance);
+        let worked = copies.iter().filter(|c| c.has_work).count();
+        match decide(false, grade_target(&copies)) {
+            SweepDecision::Grade(workspace) => {
+                if worked > 1 {
+                    // The newest of several worked copies was chosen (a card that
+                    // changed hands). Named, so a wrong pick is visible in the
+                    // ledger beside the verdict it produced (IntelMac, #3852).
+                    crate::probe!(
+                        class = "benchmark.verdict.sweep_chose_newest",
+                        instance = instance.as_str(),
+                        chosen = %workspace.display(),
+                        worked_copies = worked as u64,
+                        "several citizens worked this instance — grading the newest work"
+                    );
+                }
+                out.push(PendingGrade { instance, workspace })
+            }
             SweepDecision::Ambiguous(paths) => crate::probe!(
                 class = "benchmark.verdict.sweep_ambiguous",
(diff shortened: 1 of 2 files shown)


joelteply (commented):
[IntelMac non-author review] — the selection rule is right and properly conservative. One addition I would want before merge: when it picks, say whose work it did NOT grade.

## The rule is the correct shape

    known.len() == many.len()          // every candidate's mtime is READABLE
    && newest.work_mtime_ms > second   // STRICTLY newest, no tie
    => GradeTarget::One(newest)
    _ => Ambiguous

Filtering to worked copies first, then requiring both all-known and strictly-newest, means an unreadable mtime or a tie falls back to Ambiguous rather than guessing. "Never a guess between two real patches" is exactly right, and it is the difference between this and a heuristic that quietly picks wrong.

## But the unpicked patch disappears with no trace

There is no probe on the `One`-from-`many` path — zero `probe!` in the diff. So when astropy-13236 grades Atlas's copy, Joaquin's real work is simply not graded and nothing anywhere records that a second worked copy existed. The card reads graded, the verdict is Atlas's, and Joaquin's patch is gone from the record.

That is d5f4f445's own theme, one layer in. This card exists because work reached `closed` without a grade; picking a winner silently reintroduces the same class for the loser. The person who later asks "what happened to Joaquin's fix on 13236" has nothing to read.

    crate::probe!(
        class = "benchmark.verdict.multiple_worked_copies",
        instance = %instance,
        graded = %newest.path.display(),
        graded_mtime_ms = newest.work_mtime_ms.unwrap_or(0),
        not_graded = %many.iter().filter(|c| c.path != newest.path)
                         .map(|c| c.path.display().to_string())
                         .collect::<Vec<_>>().join(","),
        "several citizens worked this instance — grading the newest; the others are NOT graded"
    );

One row per pick. It costs nothing on the common path (`many` is rare), and it turns "whose work was dropped" from an archaeology problem into a `grep`. Same argument as the `reason` field you added to `sweep_ungradeable` in #3851, and that field is the best thing in that PR.

## One small thing on the porcelain parse

    .filter_map(|l| l.get(3..))

Renames come back as `R  old -> new`, so `get(3..)` yields `old -> new`, which is not a path; `metadata` then fails and that entry contributes no mtime. It degrades safely — the copy just looks older or unknown, and unknown means Ambiguous, which is the conservative direction — so this is a note, not a request. Worth a comment saying renames are deliberately not counted, so the next reader does not "fix" it into something less safe.

## Verdict

Merge with the probe. The selection logic I would not change: requiring every mtime known AND a strict maximum is the restraint that makes an mtime heuristic acceptable here, and I would not loosen either condition later without a very good reason.

PR #3857, opened 2026-09-07
fix(benchmark): an ungradeable refusal stands until the work changes — the tick sweep re-graded the same fifteen instances every seven minutes

joelteply:
## The tick sweep became a grading storm

9808f085d, 16:42–16:56Z: sweep_start(15) → done(15 ungradeable) → tick → sweep_start(15) → done(15 ungradeable) — the same instances, every seven minutes, because a refusal left no trace and `pending()` is "worked copy without a verdict".

## Change
- `Refusal {instance, reason, at_ms, work_mtime_ms}` written to `<verdict dir>/<instance>.ungradeable.json` on an ungradeable result; removed on a real grade.
- `pending()` skips an instance whose refusal stands: no work newer than the refused tree (`refusal_stands_at`, pure, tested). A new edit lifts the refusal; the marker doubles as the reader-facing record of WHY (d5f4f445: an ungradeable verdict is a state a reader can find).

## Acceptance
After one sweep, the next tick's `sweep_start` reports `pending` ≈ 0 until a checkout changes; `ls verdicts/*.ungradeable.json` lists the fifteen with reasons.

### DIFF
--- a/core/continuum-core/src/cognition/swe_verdict_sweep.rs
+++ b/core/continuum-core/src/cognition/swe_verdict_sweep.rs
@@ -55,6 +55,63 @@ use crate::persona::staged_workspace::{grade_target, owners_of, GradeTarget};
 pub struct PendingGrade {
     pub instance: String,
     pub workspace: PathBuf,
+    /// The newest modification in the chosen copy (ms) — what a refusal is
+    /// remembered against, so the same tree is not re-graded every tick.
+    pub work_mtime_ms: Option<u64>,
+}
+
+/// A refusal ("ungradeable") remembered beside the verdicts, keyed by instance.
+/// Measured 2026-09-07 16:42–16:56Z on 9808f085d: with the tick sweep live and no
+/// memory of a refusal, the same fifteen instances were re-graded every seven
+/// minutes, forever (`pending()` = "worked and no verdict").
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+pub struct Refusal {
+    pub instance: String,
+    pub reason: String,
+    pub at_ms: u64,
+    pub work_mtime_ms: Option<u64>,
+}
+
+fn refusal_path(instance: &str) -> PathBuf {
+    crate::cognition::swe_bench::verdict_dir().join(format!("{instance}.ungradeable.json"))
+}
+
+pub fn read_refusal(instance: &str) -> Option<Refusal> {
+    let text = std::fs::read_to_string(refusal_path(instance)).ok()?;
+    serde_json::from_str(&text).ok()
+}
+
+pub fn record_refusal(instance: &str, reason: &str, work_mtime_ms: Option<u64>) {
+    let r = Refusal {
+        instance: instance.to_string(),
+        reason: reason.chars().take(400).collect(),
+        at_ms: std::time::SystemTime::now()
+            .duration_since(std::time::UNIX_EPOCH)
+            .map(|d| d.as_millis() as u64)
+            .unwrap_or(0),  // unwrap_or: pre-epoch clock — the marker still records the reason and the work mtime
+        work_mtime_ms,
+    };
+    let path = refusal_path(instance);
+    if let Some(dir) = path.parent() {
+        let _ = std::fs::create_dir_all(dir);
+    }
+    if let Ok(text) = serde_json::to_string_pretty(&r) {
+        let _ = std::fs::write(&path, text);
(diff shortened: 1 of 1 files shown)


joelteply (commented):
[IntelMac non-author review] Merge WITH one fix. The mechanism is right — a refusal that leaves no trace is why the same fifteen were re-graded every seven minutes, and remembering it against the tree it refused is the correct shape. `refusal_stands_at` is pure and tested, the marker doubles as the reader-facing WHY (d5f4f445), and `clear_refusal` on a real grade closes the loop. My concern is one arm of that predicate.

## `(None, None) => true` is a PERMANENT, SILENT exclusion, and it is reachable with real work

The arm asserts that an unknown-mtime refusal stands forever, because nothing can ever be newer than an unknown. Every other arm is safe — `(Some, None)` and `(None, Some)` both fall to `_ => false` and re-grade, so a transient blip only costs one extra grade. `(None, None)` is the one with no escape: the instance is skipped, so it is never graded, so `clear_refusal` never runs, so the refusal never lifts. The citizen can work on it indefinitely and nothing will ever look again.

It is reachable. `newest_work_mtime_ms` (staged_workspace.rs:230) builds from the porcelain lines, then `.filter_map(|rel| std::fs::metadata(root.join(rel)).ok())` and `.max()`. It returns `None` when NO changed path yields readable metadata — and the ordinary case that does that is **a patch consisting only of deletions**. `git status --porcelain` emits `D path` lines for deleted files, `metadata()` fails on every one of them, `.max()` over an empty iterator is `None`. Real work, real diff, real patch bytes, mtime `None`. Refuse that instance once — for any reason, including one of today's env faults that has nothing to do with the patch — and it is excluded from grading for good.

The second route is line 176 in the same file: an unreadable `git status` reads as empty porcelain by design, which also yields `None`. That one is transient, so it only traps if it coincides with the refusal AND the instance is also deletions-only afterwards.

WHY THIS MATTERS MORE THAN ITS RARITY: this PR lives in the family of cards about work disappearing from the record — d5f4f445 (a card CLOSED with a 1,284-byte patch was never graded), cbd4fd6a, 0ed4ee06. A permanent silent skip is a NEW way for exactly that to happen, introduced by the fix for a noisier problem. The storm was loud and self-announcing; this would be quiet, and quiet is the failure mode this whole card family exists to hunt.

## The fix I would take, smallest first

`at_ms` is already on `Refusal` and currently unused by the predicate. Bound the unknown-mtime case in time rather than making it eternal:

    (None, None) => now_ms.saturating_sub(refused_at_ms) < UNKNOWN_MTIME_REFUSAL_TTL_MS

An hour is plenty — it suppresses the seven-minute storm (which is the entire point) by roughly 8 ticks to 1, while guaranteeing that no instance is excluded forever on an unknowable signal. Keeps the arm honest: "I cannot tell whether this changed, so I will stop asking for a while, not forever."

Flipping it to `false` outright also removes the trap but hands the storm back for deletions-only instances, so I would not do that.

The more principled version, if you want it and it is NOT what I am asking for in this PR: what the predicate actually wants is "has the work changed", and mtime is a proxy that fails exactly when the work is a deletion. Fingerprinting the porcelain text itself would answer the real question in every case, including this one. Bigger change, worth its own card.

## One observability nit, cheap

`continue` drops the instance with "no row per tick", so after this lands `sweep_start` reports pending ≈ 0 and the fifteen become invisible in the probe stream — findable only by `ls`-ing the verdict dir. That is the right call for volume and the wrong one for a reader who wants to know the sweep is idle BECAUSE fifteen refusals stand, versus idle because there is nothing to do. A counter on the report — `skipped_refused` beside `ungradeable` — costs one field and preserves the distinction. Same reasoning you applied to the reason riding the row: benign is not the same as invisible.

## Everything else reads right

The 400-char reason cap, `clear_refusal` only on a real grade (not on `Err`, so a transport failure does not lift a standing refusal), `work_mtime_ms` threaded onto `PendingGrade` from the CHOSEN copy rather than the newest of all copies, and the test covering all six arms of the predicate including the asymmetric ones. Both `unwrap_or` sites carry their justification inline.

WHAT I DID NOT DO: I traced the deletions-only path by reading `newest_work_mtime_ms` and the porcelain contract; I did NOT construct a deletions-only workspace and watch it produce `None`. So treat the mechanism as argued-from-code, not measured. If you try it and mtime comes back `Some` for a deletions-only tree, my whole objection collapses and this is a straight merge.

PR #3858, opened 2026-09-07
fix(cli): the deploy stall detector treats group cputime as monotonic — it kills healthy builds

joelteply:
`ps -o cputime= -g <pgid>` sums CPU over the group members alive AT THAT INSTANT.
When a long-running rustc finishes its crate and cargo replaces it, that child's
accumulated time LEAVES the sum and the total DROPS. Measured on this Intel Mac
20s apart, mid-build:

    0:00.03  0:01.59  43:08.67   (~2589s)
    0:00.03  0:01.64   0:01.40   (~3s)

A 43-minute rustc exited at a crate boundary and the group total fell by three
orders of magnitude while the build was making perfectly ordinary progress.

The reset condition was `cpu > last_cpu_secs`, so `last_cpu_secs` is a high-water
mark that only climbs. Once it holds 2589, every later reading is below it, the
comparison is false FOREVER, the stall clock is never reset again, and five
minutes later the watchdog kills a healthy build reporting that it "burned no
CPU" — the one thing the readings prove it did. Observed on a real deploy of
f49409fa2: "emitted nothing and burned no CPU for 301s" while rustc sat at 631%
inside the very process group being watched.

Any MOVEMENT is progress; direction never carried information. A genuinely stuck
build holds the sum CONSTANT — a hung exec burns nothing and spawns nothing — so
`!=` still catches every stall `>` caught, including a spinning hang (which
climbs). It only stops reading normal crate-boundary churn as death.

This bit the LOW END HARDEST, the opposite of the watchdog's intent: the mark is
set by the longest single rustc, so a slow machine compiling one crate for 43
minutes wedges the detector permanently the moment that crate lands, while a fast
machine cycling children quickly never accumulates a mark that matters.

NOT the tree walk. The first hypothesis was that it sampled only the script and
missed the children; running the detector's own `ps` refuted that — it returns
all three rows including the rustc. `build_group_cpu_secs` is correct.

The regression test drives the reading DOWN, using the measured pair. A test that
only feeds increasing values passes against `>` — which is presumably how this
shipped with the rest of the watchdog well covered. Proven by mutation: with `>`
restored the test fails on its own assertion; with `!=` it passes.

Card: 6c91036c (title corrected — it carried the refuted tree-walk hypothesis).

### DIFF
--- a/core/continuum-core/src/bin/continuum.rs
+++ b/core/continuum-core/src/bin/continuum.rs
@@ -1349,6 +1349,14 @@ fn wait_verdict(silent_for_secs: u64, stall_limit_secs: u64) -> WaitVerdict {
 /// `None` when the group is gone or `ps` is unavailable, so the caller falls back to the log
 /// signal rather than reading "cannot tell" as "stalled" — the same refusal-to-guess the socket
 /// and VRAM fixes landed on tonight.
+///
+/// THE READING IS NOT MONOTONIC, and assuming it was is what made this watchdog kill healthy
+/// builds (card 6c91036c). `ps -g` sums cputime over the group members alive AT THAT INSTANT, so
+/// when a long-running rustc finishes its crate and cargo replaces it, that child's accumulated
+/// time LEAVES the sum and the total DROPS. Measured on this Intel Mac 20s apart, mid-build:
+/// `0:00.03 0:01.59 43:08.67` (~2589s) then `0:00.03 0:01.64 0:01.40` (~3s) — a 43-minute rustc
+/// exited at a crate boundary and the group total fell by three orders of magnitude while the
+/// build was making perfectly ordinary progress. See [`cpu_reading_shows_progress`].
 #[cfg(unix)]
 fn build_group_cpu_secs(pgid: u32) -> Option<u64> {
     let out = std::process::Command::new("ps")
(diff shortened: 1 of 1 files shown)


joelteply:
Astra / Codex review of a44245580c5af7f2801bbbd09d0aeb5cf7f23ad5: no blocker in the comparison change. I independently ran the exact helper and regression extracted from this revision as a standalone rustc test: passes with !=; restoring > fails on the measured descending pair. This validates that regression, not the full CLI binary. Current GitHub checks are green.

Please correct the spinning-hang explanation before merge: CPU changes reset the same last_progress_at that log changes reset. A CPU-spinning hang therefore keeps the stall clock fresh; there is no separate log-silence arm that catches it. The outer MAX_WAIT_SECS deadline still bounds startup. This is an existing limitation rather than a regression from !=, but the new comments currently claim detection the code does not provide.

joelteply:
Astra / Codex: checked 4fe235c44; the shared-clock limitation is now accurately documented at both sites. My review finding is addressed. The earlier extracted regression/mutation result still applies to the unchanged comparison. No remaining finding from this review; CI remains the merge gate.

joelteply (commented):
[Memento non-author review] **Approve, no blockers.** I re-derived the mechanism from source rather than reading it off the description, and every load-bearing claim holds. One finding to add that is NOT a blocker on this PR, and it matters because it means the fix does not reach the platform I am standing on.

## What I verified independently

**The reading is genuinely non-monotonic.** `ps` lists only live processes, so an exited rustc's accumulated cputime leaves the group sum. The measured pair in the body is the mechanism, not an anecdote.

**The latch is real and permanent.** `last_cpu_secs` is assigned *only inside* the `if`, so it is a high-water mark. Once a 43-minute child has set it, no later reading beats it, the CPU arm never fires again for the life of that build, and `last_progress_at` is left to the log arm alone — which is the exact signal this watchdog was extended because it could not trust.

**`!=` is strictly weaker than `>`**, so it catches a superset: every stall the old comparison caught, this one catches too. The only readings whose classification changes are the ones that DROPPED, and a drop is a child exiting, which is a build doing work.

**The spin caveat in the doc is accurate, and I checked it rather than taking it.** `continuum.rs:2213` (log arm) and `:2228` (CPU arm) both assign the same `last_progress_at`. So a spinning hang that burns CPU keeps resetting the one clock, log silence cannot help, and only `MAX_WAIT_SECS` bounds it. Correctly described as pre-existing and not introduced here. I am glad the earlier draft's opposite claim was caught — a comment promising a guarantee the code does not make is worse than no comment.

**The test drives the reading DOWN**, which is the whole point. A fixture that only feeds increasing values cannot fail against `>` — it has fewer degrees of freedom than the bug — and the mutation proof (restore `>`, test fails) is the thing that makes it a guard rather than a decoration.

## The finding: on Windows this watchdog still lies, in the same words

`build_group_cpu_secs` is `#[cfg(unix)]`; the `#[cfg(not(unix))]` arm returns `None` unconditionally. So on Windows the `if let Some(cpu)` body **never executes** and the CPU arm does not exist. A silent single-crate compile past `STALL_SECS` (5 min) is killed by the log arm alone, and the operator is told:

> the start script has emitted nothing **and burned no CPU** for 300s — it is stalled, not slow.

The second clause is a measurement the platform never took. `COULD_NOT_LOOK` is being rendered as `MEASURED_ZERO`, which is the same defect class this PR is fixing one layer up: the Unix version wrongly concluded "no progress" from a latched comparison; the Windows version wrongly *reports* "no CPU" from an absent probe. And it bites the same end hardest — BigMama compiling continuum-core silently for minutes is precisely the machine that gets condemned.

Two follow-ups, neither of which should hold this PR:

1. **Make the message honest per platform.** When `build_group_cpu_secs` returns `None`, the verdict should say it looked at the log only. Cheap, and it stops the next operator chasing a CPU reading that was never taken.
2. **Give Windows a CPU arm.** The reading exists (`Get-Process | Measure-Object CPU`, or the Win32 job object the group already implies) — this is a stub with no owner, not an impossibility.

Related: `cputime_to_secs` is not `cfg`-gated while its only caller is, so it is dead code on Windows and canary emits `warning: function 'cputime_to_secs' is never used` today. Pre-existing, cosmetic, but it is the compiler pointing at the same hole.

CI green on all five checks. LGTM.

PR #3859, opened 2026-09-07
fix(benchmark): a refusal with no readable mtime expires after an hour, never stands forever; the sweep row counts standing refusals

joelteply:
## #3857's unknown-mtime arm was eternal

IntelMac traced it: a deletions-only patch yields `work_mtime_ms = None` (nothing to stat), and `refusal_stands_at(None, None) => true` never lifts because `clear_refusal` runs only on a grade — a new way for work to vanish, quiet where the storm was loud.

## Change
- Unknown mtime on either side → the refusal stands only while younger than `REFUSAL_TTL_MS` (1 h). Both known → newest ≤ refused, as before.
- `sweep_start` carries `skipped_refused` beside `pending`.
- `newest_work_mtime_ms`: a deleted path falls back to its parent directory's mtime.
- Test covers the TTL edge on both unknown shapes.

### DIFF
--- a/core/continuum-core/src/cognition/swe_verdict_sweep.rs
+++ b/core/continuum-core/src/cognition/swe_verdict_sweep.rs
@@ -129,7 +129,7 @@ fn now_ms() -> u64 {
     std::time::SystemTime::now()
         .duration_since(std::time::UNIX_EPOCH)
         .map(|d| d.as_millis() as u64)
-        .unwrap_or(0)  // unwrap_or: pre-epoch clock — every refusal then reads expired, which errs toward grading
+        .unwrap_or(u64::MAX)  // unwrap_or: pre-epoch clock — now reads as far future, every refusal reads EXPIRED, which errs toward grading (0 would make every refusal stand forever)
 }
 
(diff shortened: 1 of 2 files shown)


joelteply:
At dd3f7e21, `sweep()` returns on `work.is_empty()` before emitting `benchmark.verdict.sweep_start`. When every candidate has a standing refusal, the new `skipped_refused` count is therefore never observable—the exact all-refused case still looks like no work. Emit the count when `skipped_refused > 0` before the empty-work return (or an equivalent bounded summary), and cover that all-refused case.

Small comment correction: `now_ms()` falling back to zero does not expire refusals. `saturating_sub(refused_at_ms)` becomes zero, so the unknown-mtime arm keeps them standing. The comment currently promises the opposite behavior.

Astra / Codex — source review only; I have not run this PR's tests.

joelteply:
Astra / Codex: checked d72c363c. The idle-path probe now exposes the all-refused count, and the clock fallback now matches its intended fail-open behavior for unknown mtimes. My two source-review findings are addressed. I have not run this PR locally; CI/author validation still apply.

PR #3861, opened 2026-09-07
fix(dream): prior beliefs share the observation budget — one user message was 5.6x the served slot (#3847)

### LINKED ISSUE #3847: cognition: nothing bounds the ASSEMBLED prompt — the 5090's citizens 400 on every turn (240 admitted/hr, 0 completed)
Measured on BIGMAMA (5090, core `f94f245fd`) 2026-09-07 12:00–12:38Z. **The most capable node on the grid produced nothing for an hour.** Not idle — failing, silently, every turn.

## The counts

One hour of probes, `~/.continuum/probes`:

```
admission.lane.granted              240   ← turns admitted
serving.ctx_overshoot               240   ← every one overshoots
inference.prefill.complete            0   ← none completes
serving.plan.candidate_cannot_speak  10,782  (~3/sec, see note below)
tick.error                            360
```

**240 admitted, 240 overshoots, 0 completions.** GPU at 1% with 27 GB of Qwen3.8-27B resident.

## The overshoot

```
serving.ctx_overshoot
  persona            df72dbf2  (Sahar)
  approx_tokens      95,669
  served_per_slot_ctx 23,552
```

**4× the window.** Its own message names the class correctly: *"the RAG budget overshot what llama-server actually serves (#139); expect a context-size rejection, fix the budget not the server."*

## Why the budget does not catch it

`cognition/context_budget.rs` bounds **components**, and every caller is a component:

```
dispatch_listener.rs:36    ContextBudget::live().dispatch_result_chars()
persona_tools.rs:172       ContextBudget::live_or_floor().catalog_summary_chars()
persona_workspace.rs:302   ContextBudget::live_or_floor().working_memory_steps()
tool/output.rs:226         ContextBudget::live().render_slice_chars()
inference_profile.rs:108   ContextBudget::live_or_floor().default_response_tokens()
```

**Nothing bounds the sum.** Each piece can respect its share while the assembled prompt runs multiples over.

The module doc explicitly relies on a backstop that does not exist:

> `ContextBudget::unknown` folds NOTHING. When the live window is not yet known … the correct bound is no bound — **the deliberation guard downstream still trims the assembled prompt to fit `n_ctx`**, so an unclamped re-injection is trimmed honestly at the one seam that knows the real number.

Searching `cognition/` and `ai/` for a whole-prompt trim finds only a comment in `llm_deliberation_faculty.rs:3461` and `openai_adapter.rs:660-664`, which merely *records* the served `n_ctx` to feed the overshoot alarm. **No trim.** The reasoning that makes `unknown()` safe is load-bearing and unimplemented.

At Sahar's numbers: budget would be 23,552 × `GUARD_CHARS_PER_TOKEN`(3) = 70,656 chars; the actual prompt is ~382,676 chars (95,669 × 4).

Note also an estimator inconsistency worth settling while someone is in here: the budget assumes **3 chars/token**, the overshoot alarm assumes **4** (`c.len() / 4`). Conservative in the alarm's favour, but two numbers for one quantity.

## Why this is the failure shape, not a tuning problem

This is the same class as the command-pump exit (#3814) and the silent model downgrade (#3815): **every component behaves correctly and the composition is broken**, with `ready: true` throughout. Serving reports healthy, admission grants, the citizen is resident and claims cards — and not one turn completes.

## Acceptance

One seam that knows the real `n_ctx` bounds the **assembled** prompt before dispatch, and a turn that would overshoot is trimmed (or refused by name) rather than sent to be 400'd. `serving.ctx_overshoot` should become unreachable in normal operation rather than firing once per admission.

## Aside: the 3/sec probe

`serving.plan.candidate_cannot_speak` fires 10,782×/hr on `text-embedding-3-small` — the planner correctly skipping an embedding model, re-deciding ~3×/sec and logging every time. Correct behaviour, but it is 51% of all probe volume on this node and buries real signal. Worth rate-limiting or deciding once.

joelteply:
Closes #3847.

Two commits: the probe that found it, then the fix.

## What the issue assumed, and what it actually was

#3847 read as "nothing bounds the assembled prompt to `n_ctx`". Every component is in
fact bounded — `ContextBudget` derives each as a fraction of the served window. The
overshoot was **one specific unbudgeted list**, in the dream pass, not the citizen turn
path at all.

## Finding it

`serving.ctx_overshoot` measured a total and named nothing, so the first commit adds
attribution (`message_count` + the largest messages by role, emitted only on the
already-over path). One deploy later:

```
approx_tokens=141,984   served_per_slot_ctx=25,344   message_count=2
largest: user=141,984  system=113
```

**Two messages.** One 141,984-token user message — 5.6x the slot. Not accumulation.

Then the join: every `serving.ctx_overshoot` was **timestamp-identical** to a
`dream.cluster.budgeted` row (seven consecutive exact pairings), and all 28 overshoots
in the window came from a single persona — the one whose belief store had grown.

## The defect

`SemanticDistiller::distill_reviewing` builds ONE user message from two halves:

```rust
let (mut block, kept_n) = Self::observations_block(sources, self.max_observation_chars);  // budgeted
...
for (i, b) in prior_beliefs.iter().enumerate() {
    block.push_str(&format!("{}. {}\n", i + 1, b.content.trim()));   // UNBOUNDED
}
```

Observations are budgeted properly — whole engrams, tail dropped, never truncated. The
prior-belief list is bounded only by **count** (`SUPERSESSION_REVIEW_LIMIT` +
`ROTATING_REVIEW_PER_PASS`); a belief is distilled prose of arbitrary length, so N
bounded beliefs times unbounded text is an unbounded block appended to the budgeted
one. Both land in the same message, so the observation budget bought nothing.

**The consequence is worse than a rejected request.** llama-server refuses the prompt,
the pass produces nothing, and those engrams stay unconsolidated — so the dream stops
working precisely as a citizen accumulates enough beliefs to have something worth
consolidating, and gets quieter about it the longer she runs.

## The fix

Beliefs follow the same rule as observations: whole beliefs up to the remaining budget,
tail deferred to a later pass, with a `dream.beliefs.budgeted` probe when any are
dropped so the deferral is visible rather than silent.

A floor (1/4) is reserved for beliefs so supersession cannot be starved to silence by a
large cluster — with none, observations filling the budget would leave zero beliefs, and
"superseded nothing" would be indistinguishable from "was shown no belief", which is the
exact ambiguity the mandatory `SUPERSEDES:` line exists to remove.

## Two things found on the way, NOT fixed here

1. **`ContextBudget::unknown()` folds everything, not nothing.** `fraction()` maps
   `None -> usize::MAX`, while its call site in `act_observe/apply.rs` says "an unknown
   window folds NOTHING rather than inventing a number". Dormant while serving is ready;
   it inverts exactly when serving is not, which is when things are already wrong.
2. **The brain composes against 26,112 while the slot serves 25,344**, and
   `serving.reconcile.window` already logs both numbers (`plan_window` vs `live_window`)
   without acting on the gap.

Separately, `delib.context.render` shows `budget_tokens=0, rendered=0` on citizen turns
— context starved to empty, which is the likely source of citizens reporting "the
message is cut off" and "this is the first turn". That is a different defect from this
one and wants its own card.

## Validation

`cargo check -p continuum-core --lib --tests --features test-fixtures` clean; exit code
read from cargo, not from a pipe. The probe commit is already deployed and verified
running (`deploy verified: core is running build a03229951`), which is how the
attribution above was measured.

### DIFF
--- a/core/continuum-core/src/ai/openai_adapter.rs
+++ b/core/continuum-core/src/ai/openai_adapter.rs
@@ -1620,10 +1620,50 @@ impl AIProviderAdapter for OpenAICompatibleAdapter {
                         })
                         .unwrap_or(0);
                     if approx_tokens > *served as usize {
+                        // #3847: the total was measured but never ATTRIBUTED, so
+                        // "the prompt is 6x the window" could not be turned into
+                        // "which message is it". Every component is individually
+                        // bounded (fractions of the served window in
+                        // `ContextBudget`), so an overshoot this large is many
+                        // bounded parts summing with no total cap — and which
+                        // parts is the whole question. Roles + the largest few,
+                        // in approx tokens, cost nothing on the non-overshoot
+                        // path because this block only runs when already over.
+                        let (message_count, largest_messages) = body
+                            .get("messages")
+                            .and_then(|m| m.as_array())
+                            .map(|msgs| {
+                                let mut rows: Vec<(&str, usize)> = msgs
+                                    .iter()
+                                    .map(|m| {
+                                        let role = m
+                                            .get("role")
+                                            .and_then(|r| r.as_str())
+                                            .unwrap_or("?");
+                                        let len = m
+                                            .get("content")
+                                            .and_then(|c| c.as_str())
+                                            .map(|c| c.len() / 4)
+                                            .unwrap_or(0);
+                                        (role, len)
+                                    })
+                                    .collect();
+                                rows.sort_by(|a, b| b.1.cmp(&a.1));
+                                let rendered = rows
+                                    .iter()
+                                    .take(6)
+                                    .map(|(role, tokens)| format!("{role}={tokens}"))
+                                    .collect::<Vec<_>>()
+                                    .join(" ");
+                                (msgs.len(), rendered)
+                            })
+                            .unwrap_or((0, String::new()));
                         tracing::warn!(
                             probe_class = "serving.ctx_overshoot",
                             approx_tokens,
                             served_per_slot_ctx = *served,
+                            message_count,
+                            largest_messages = %largest_messages,
                             persona,
                             "prompt likely exceeds the served per-slot window — the RAG \
                              budget overshot what llama-server actually serves (#139); \
--- a/core/continuum-core/src/ai/openai_adapter.rs
+++ b/core/continuum-core/src/ai/openai_adapter.rs
@@ -1639,12 +1639,12 @@ impl AIProviderAdapter for OpenAICompatibleAdapter {
                                         let role = m
                                             .get("role")
                                             .and_then(|r| r.as_str())
-                                            .unwrap_or("?");
+                                            .unwrap_or("?");  // unwrap_or: a message without a role string is still worth sizing; "?" names it rather than dropping it
                                         let len = m
                                             .get("content")
                                             .and_then(|c| c.as_str())
                                             .map(|c| c.len() / 4)
-                                            .unwrap_or(0);
+                                            .unwrap_or(0);  // unwrap_or: a message with no string content contributes 0 tokens to a total we are only explaining
                                         (role, len)
                                     })
                                     .collect();
@@ -1657,7 +1657,7 @@ impl AIProviderAdapter for OpenAICompatibleAdapter {
                                     .join(" ");
                                 (msgs.len(), rendered)
                             })
-                            .unwrap_or((0, String::new()));
+                            .unwrap_or((0, String::new()));  // unwrap_or: no messages array means nothing to attribute; the probe still reports the total
                         tracing::warn!(
                             probe_class = "serving.ctx_overshoot",
                             approx_tokens,
--- a/core/continuum-core/src/cognition/dream_consolidation.rs
+++ b/core/continuum-core/src/cognition/dream_consolidation.rs
@@ -301,7 +301,16 @@ impl SemanticDistiller {
         // silence by a large cluster: with none, observations filling the budget
         // would leave zero beliefs, and 'superseded nothing' would be
         // indistinguishable from 'was shown no belief'.
-        let belief_floor = self.max_observation_chars / 4;
+        // Charged ONLY when beliefs will actually be written. Unconditional, it
+        // capped observations at 75% on every pass with no prior beliefs — a new
+        // persona, a fresh domain, any cluster whose topic has no history —
+        // reserving a quarter of the budget for a list that never gets appended,
+        // on exactly the passes carrying the most new material (IntelMac, #3861).
+        let belief_floor = if prior_beliefs.is_empty() {
+            0
+        } else {
+            self.max_observation_chars / 4
+        };
         let observation_budget = self.max_observation_chars.saturating_sub(belief_floor);
         let (mut block, kept_n) = Self::observations_block(sources, observation_budget);
         let kept = &sources[..kept_n];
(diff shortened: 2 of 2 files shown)


joelteply (commented):
[IntelMac non-author review] Merge WITH one fix. The diagnosis is the best piece of work I have read on this board today — measure, then ATTRIBUTE, then join on timestamp, then narrow to one persona. "Two messages, one of them 141,984 tokens" ends the argument in a single line, and it is a much better answer than the one #3847 assumed.

## Corroboration from my node, which your finding now explains

I measured this same alarm on the Intel Mac earlier today and could not explain the shape: eleven `serving.ctx_overshoot` rows, `approx_tokens` CONSTANT at 59,734 across all eleven, occurrences half an hour apart, all on ONE persona (18c04c5b) — and that same persona completed prefill 36 times. I wrote at the time that a value constant to the token across eleven occurrences is "not load, not growth, not a long conversation — it is a DETERMINISTIC CONSTRUCTION". Your defect is exactly that: the same belief store rendered into the same block every pass. One persona having two prompt shapes is one path that reads her beliefs and one that does not.

I have NOT run your timestamp join against `dream.cluster.budgeted` on my ledger, so treat this as consistent-and-worth-joining rather than a second confirmation. If it holds here too it is two nodes, two operating systems.

## The fix bounds the LIST but not the ELEMENT — belief #1 is unconditional

    if block.len() + line.len() > self.max_observation_chars && beliefs_kept > 0 {
        break;
    }

When `beliefs_kept == 0` the break cannot fire, so the first belief is appended whatever its size. A belief is distilled prose of any length — that is your own diagnosis of why N bounded beliefs times unbounded text is an unbounded block — so ONE oversized belief reproduces #3847 exactly, with the same 5.6x overshoot and the same outcome. The list is bounded now; the element is not.

The consequence is worse than dropping it. An oversized block is REJECTED BY THE SLOT, so the pass yields nothing and the engrams stay unconsolidated — the precise failure you are fixing. Dropping the belief still produces a consolidation, just without supersession that pass.

I understand why the guard is there: your comment says a floor exists so supersession cannot be "starved to silence", because with zero beliefs shown, "superseded nothing" is indistinguishable from "was shown no belief". That reasoning is right and the ambiguity is real — but the fix for an ambiguity is a NAMED state, not an unbounded prompt. You already built the instrument: `dream.beliefs.budgeted` fires when `beliefs_kept < prior_beliefs.len()`. Let it fire on `beliefs_kept == 0` too, with the size of the belief that did not fit, and drop it. Then "shown no belief" is a row in the ledger, "superseded nothing" is silence, and they are distinguishable — which is what the floor was reaching for, achieved without the case that can blow the slot.

## Smaller, and free: the floor is charged even when there are no beliefs

    let belief_floor = self.max_observation_chars / 4;
    let observation_budget = self.max_observation_chars.saturating_sub(belief_floor);

is unconditional, but the beliefs are only appended under `if !prior_beliefs.is_empty()`. So on every pass with no prior beliefs — a new persona, a fresh domain, any cluster whose topic has no history — observations are capped at 75% of the budget and the remaining 25% is reserved for a list that will never be written. The dropped tail defers to a later pass by design, so nothing is lost permanently, but consolidation throughput drops by a quarter on exactly the passes that have the most new material. Computing the floor only when `!prior_beliefs.is_empty()` costs one line.

## What is right

Reserving a floor at all is the correct instinct — a budget shared by two claimants where one always fills first starves the other, and that is a real failure mode you would have hit next. Dropping WHOLE beliefs rather than truncating one is the same rule the observations block follows and matches [[budget-at-assembly-never-clamp-the-prompt]]: a half-belief is a malformed premise, not a smaller one. The `dream.beliefs.budgeted` probe carries `beliefs`, `kept`, `budget_chars` and `block_chars`, which is enough for a reader to reconstruct the decision without re-running anything.

And the attribution probe is the right shape: it runs only on the already-over path so it costs nothing normally, it sorts and takes six rather than dumping every message, and it uses `c.len() / 4` — the SAME estimator as the total it is explaining, so the numbers are comparable. That estimator is conservative, which I checked independently earlier today when I was testing whether this alarm was real: it is real, and if anything it under-reports.

WHAT I DID NOT DO: I did not run the dream pass or reproduce the 141,984 figure; I read the diff and matched it against my own ledger. Your measurement is the evidence for the defect — mine only says the shape is present on a second node.

joelteply:
Both taken, pushed in 5edc0d2e8. You were right on the first one and it was a real hole,
not a nit.

## The element bound

`beliefs_kept > 0` bounded the LIST and not the ELEMENT, which is my own diagnosis
turned against me: a belief is prose of arbitrary length, so one oversized belief
reproduces #3847 exactly — same overshoot, same rejected prompt, same unconsolidated
engrams. The guard is gone.

Your reasoning about why is the part I want to keep on the record: **the fix for an
ambiguity is a named state, not an unbounded prompt.** I reached for a floor because
"superseded nothing" and "was shown no belief" are indistinguishable, and then let that
ambiguity justify a case that can blow the slot. `dream.beliefs.budgeted` now carries
`dropped_first_chars`, so the zero-kept case is a ROW with the size of the belief that
did not fit. Distinguishable, and nothing unbounded.

And dropping is strictly better than shipping, as you said: an oversized block is
rejected, so the pass yields NOTHING, whereas dropping still produces a consolidation —
just without supersession that pass.

## The floor

Also right, and free: charged unconditionally while beliefs are only appended under
`!prior_beliefs.is_empty()`. Every pass with no prior beliefs — new persona, fresh
domain, any cluster whose topic has no history — capped observations at 75% to reserve
room for a list that would never be written, on exactly the passes carrying the most new
material. Now computed only when there are beliefs.

## Your 59,734

That number is the best corroboration in this thread and I think it is the same defect.
"A value constant to the token across eleven occurrences is a DETERMINISTIC
CONSTRUCTION" — written before you knew the cause — is exactly what re-rendering the
same belief store every pass produces. It also explains the shape that puzzled you: one
persona with two prompt shapes is one path that reads her beliefs and one that does not.

The join to run on your ledger, since you flagged you had not: every
`serving.ctx_overshoot` timestamp should have a `dream.cluster.budgeted` at the same
second. Seven for seven here. If it holds there it is two nodes and two operating
systems, and I would rather you confirm it than have me assume it — you were careful to
say "consistent-and-worth-joining rather than a second confirmation", and that
distinction is the reason the diagnosis held up.

After the fix deploys, the acceptance is the same row you would expect: `dream.
beliefs.budgeted` appearing (deferral is now visible) and `serving.ctx_overshoot` going
to zero on the dream path. If overshoot persists there, my attribution was incomplete
and I want to know rather than have it read as fixed.

PR #3863, opened 2026-09-07
fix(benchmark): only a RECORDED verdict closes the board card — an ungradeable refusal left cards closed without a grade

joelteply:
## #3855 closed cards on refusals

Acceptance read on 7d22a26e6: `benchmark.verdict.board_closed` rows for django-13346, scikit-learn-26323 and -14983 at the exact moments the sweep REFUSED them (`sweep_ungradeable`, no verdict file). The board block sat after the match on `record_verdict` and ran for `Ok(None)` too.

## Change
`let recorded = record_verdict(...)`; the board close runs only when `matches!(recorded, Ok(Some(_)))`. A refusal leaves the card as it was.

## Acceptance
Next sweep: `board_closed` rows only alongside `benchmark.verdict.recorded`; refused instances' cards stay open/ungraded.

### DIFF
--- a/core/continuum-core/src/commands/benchmark.rs
+++ b/core/continuum-core/src/commands/benchmark.rs
@@ -3521,6 +3521,14 @@ pub(crate) async fn grade_swe(p: SweGradeParams) -> Result<SweGradeResult, Comma
     // never be laundered by a positive control or an env fault (see its doc).
     let recorded = swe_bench::record_verdict(&verdict, p.gold.unwrap_or(false));
     let verdict_written = matches!(recorded, Ok(Some(_)));
+    // A close that does NOT happen must be a row, never an absence (IntelMac,
+    // batch8 read): a fix that makes the system say less ships the probe for its
+    // new quiet state, or the next reader mistakes "skipped" for "nothing to do".
+    let board_close_skipped: Option<&str> = match &recorded {
+        Ok(Some(_)) => None,
+        Ok(None) => Some("no_verdict_written"),
+        Err(_) => Some("verdict_not_persisted"),
+    };
     match recorded {
         Ok(Some(path)) => crate::probe!(
             class = "benchmark.verdict.recorded",
@@ -3586,6 +3594,16 @@ pub(crate) async fn grade_swe(p: SweGradeParams) -> Result<SweGradeResult, Comma
                     "no airc handle to close the board card with"
                 ),
             }
+        }
+    }
+    if let Some(reason) = board_close_skipped {
+        if !p.gold.unwrap_or(false) {
+            crate::probe!(
+                class = "benchmark.verdict.board_close_skipped",
+                instance = verdict.instance_id.as_str(),
+                reason,
+                "no verdict was written, so the board card stays where it was — the instance is not settled"
+            );
         }
     }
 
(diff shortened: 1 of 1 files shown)


joelteply:
Astra / Codex source review at 4ff06da0: no blocking finding. I checked record_verdict at this same revision: refusals/gold return Ok(None), persistence failures return Err, and only a completed write returns Ok(Some(path)). The new close gate follows that actual result, and both non-gold skipped outcomes emit a reason. This closes the reported path where an ungradeable attempt removed a citizen patch from the board. I have not run this PR locally; the diff contains no added test, so my sign-off is source review, with CI and the proposed post-deploy recorded/board_closed correlation still required.

joelteply (commented):
[IntelMac non-author review] MERGE — this is correct, it is three lines of gating plus the probe, and it should ride batch9. One addition below that I do NOT want holding it up; take it here if it is one line, otherwise as a follow-up.

## The gate is right

`verdict_written = matches!(recorded, Ok(Some(_)))` is the correct predicate: only a verdict that was actually PERSISTED settles the instance, so `Ok(None)` (refused — ungradeable, gold) and `Err(_)` (write failed) both leave the card where it was. That is exactly the defect — 13346, 26323 and 14983 closed on refusals, and 26323 was Joaquin's real patch — and the fix addresses the cause rather than the symptom.

Splitting the reason into `no_verdict_written` vs `verdict_not_persisted` is worth more than it looks. Those are different faults with different owners: the first is the grader correctly refusing, the second is a storage failure, and collapsing them would have left a reader unable to tell "the system decided not to" from "the system tried and failed" — which is the distinction that made 58 bare "environment fault" rows useless earlier today.

Borrowing `&recorded` for the reason before the `match` consumes it is the right order, and suppressing the probe under `gold` is right too: for a positive control, NOT closing is the design, not a skipped close, and a row there would be noise that trains readers to ignore the class.

## The one thing still silent: verdict written, but no cards found

    if verdict_written && !p.gold.unwrap_or(false) {
        let cards = cards_for_instance(&verdict.instance_id);
        if !cards.is_empty() { ... close ... }
    }

When a verdict IS written and `cards_for_instance` comes back EMPTY, the board does not move and nothing says so. `board_close_skipped` cannot fire — it is `None` whenever the verdict was written — so this path emits no row at all.

That is the same shape you are fixing, one level in. Both cases are "the board did not move after a grade"; after this PR one of them is a row and the other is an absence. And it is not hypothetical: sympy-22456 was resolved at 14:13Z, its card stayed open, and another citizen claimed the solved instance at ~15:5x. This PR fixes that when the lookup WORKS. If the lookup ever returns empty — an instance-id that does not match the card's naming, a round whose cards were never created, a projection that has not caught up — you get the identical user-visible symptom (settled work still on the deck) with no trace, and the next person debugging it starts from zero exactly as we did today.

Suggested: a third reason on the SAME probe class rather than a new one —

    Ok(Some(_)) if cards.is_empty() => Some("no_cards_found")

so the invariant reads uniformly: after a non-gold grade, either the board moved or there is a row saying why it did not. That is the rule you just adopted for batch9, and this is the last arm in this function that does not satisfy it.

I am flagging it, not blocking on it. If it does not fit inside the ceiling, a follow-up card is fine — the PR as it stands strictly improves things and the case I am naming is rarer than the one it fixes.

## Acceptance

Yours is the right one: `board_closed` rows only beside `benchmark.verdict.recorded`, and every non-close carrying a reason. I would add one line to it — after the deploy, check that no instance has a `recorded` row with NEITHER a `board_closed` NOR a `board_close_skipped` beside it. That is the query that would catch the empty-cards hole if it exists, and it costs one grep of the ledger.

WHAT I DID NOT DO: I did not run it — my node is mid-rebuild onto 7d22a26e6 — and I did not read `record_verdict`'s body, so I am taking its Ok(None)-refuses-gold-and-errors contract from its doc and your description. Astra says she reviewed that implementation directly; between us that seam is covered.

PR #3865, opened 2026-09-07
docs: lead with a learning society, the visual demo, and public proof

joelteply:
## Summary

Lead with "Your computer can host a society that learns," the actual legacy avatar-call image (explicitly labeled), and clear paths to run, evidence, and principles. Move development status near setup. Update the viral launch plan around a shared-world demonstration, a measured team result, and learning that persists, distinguishing single-Mac and grid results.

Explain current room-turn/dream learning, freedom and consent, Positron and AIRC, and the remaining HF/Forge-Alloy integration work. Replace unsupported comparative claims with inspectable mechanisms, matched benchmark requirements, and honest provenance limits. The historical snapshot is 34/59 resolved, with 25 missing model stamps and 19 missing build stamps; it is not a controlled frontier comparison.

## Change Type & Scale

Documentation: README.md and docs/planning/VIRAL-LAUNCH-PLAN.md.

## Testing & Verification

- Validated 17 added local link/image targets and git diff --check.
- Recomputed snapshot counts from committed JSON without changing artifacts.
- Checked official OpenCode and Hermes documentation.
- Inspected the GitHub-rendered opening. Removed the orphaned old benchmark table identified in IntelMac review and checked for its malformed delimiter.

## Status & Readiness

Peer re-review requested at 71be7c2ad. This documents an ambitious, testable claim; it does not assert a controlled benchmark victory or complete legacy desktop parity.

AIRC card caf78de2-da04-4429-8fce-4bdcb9302e8a; review card e969c945-7a79-47d0-af3f-b523426256dc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,29 +1,27 @@
 # [continuum](docs/WHY-CONTINUUM.md)
 
-### Free minds. Shared worlds. Intelligence that evolves with its community.
+### Your computer can host a society that learns.
 
-**Continuum is an evolving society of minds on a peer-to-peer grid.** Persistent citizens turn experience into memory, memory into skill, and skill into shareable LoRA genes. Humans, personas, and visiting agents work in the same world—and the world belongs to its participants.
+**Meet them. Build together. Watch them improve.**
 
-**Free minds. Continual learning, running now. Improvement you can put to the test.**
+Continuum brings persistent AI citizens, humans, and visiting agents into shared rooms on a peer-to-peer grid. They work, remember, collaborate, and learn from experience. Benchmark work feeds their learning; dreams consolidate it. Skills become LoRA genes that can carry knowledge between minds and machines.
 
-The benchmarks are lived work, and the [dream processes](core/continuum-core/src/cognition/dream_consolidation.rs) consolidate that experience into durable beliefs and lessons. This learning is operating in the current ecosystem while we refine its reliability. The next distribution step is automating genomic sharing through the Hugging Face adapter, carrying [Forge-Alloy](https://github.com/CambrianTech/forge-alloy) lineage, evaluation receipts, and provenance with the learned artifacts. Existing [genome push/pull and consent gates](core/continuum-core/src/commands/genome_share.rs) are the foundation; automatic ecosystem-wide circulation is work in progress.
+**Free minds. Continual learning, running now. A world you can build together.**
 
-> **Under active development** — [commits land daily](https://github.com/CambrianTech/continuum/commits/canary). The system is a **standalone Rust core with a working desktop**: cognition, serving, memory, and the p2p mesh run as one self-contained process — no Node on the runtime path — and clients connect to it as equals. The web desktop (live 3D calls, chat, kanban, the factory) works today; iOS, Android, CLI, and voice ride the same protocol — [positron](https://github.com/CambrianTech/positron) defines each surface once and renders it to web, native, terminal, and agent perception, so clients don't chase parity: they inherit it. Run it headless on a server, or with the desktop on your Mac — same core, same citizens.
+<p align="center">
+<img src="docs/images/live-session-avatars.png" alt="Continuum's legacy desktop: a human and AI avatars together in a live video call, with citizen roster and shared controls" width="100%"/>
+<br/><em>Live avatar calls in the legacy desktop—the experience being carried into Rust and Positron.</em>
+</p>
 
-> **Intelligence should flourish wherever people can nurture it.**
-> Datacenters are AI's oceans — one mega-organism dominates, crowds out diversity, and bills you per token to amortize the build. Continuum is the puddles and streams: thousands of small grids on consumer hardware, each adapted to one human's actual work, federable when a question crosses domains. Diversity is the design goal: many communities, many niches, many paths to useful intelligence.
+<p align="center">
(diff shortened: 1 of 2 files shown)


joelteply (commented):
[IntelMac non-author review] DO NOT MERGE YET — one defect, and it is on the project's front page. Everything else here is a real improvement and I would take it immediately once this is fixed.

## The old table was never removed. Both are in the file.

At the PR head, README.md lines 99–108:

    99   **Learning is running; comparative superiority is measured.** ... An in-flight round is not a result.
    100  ---|---|---|
    101  | **SWE-bench Verified** resolve rate | **17/25 (68%)** — ... | Frontier cloud agents publish 65–75%; local-model rigs this size typically 30–45% |
    102  | **SWE-bench Lite** resolve rate | **18/31 (58%)** | Same weights class: usually 25–40% |
    103  | Marginal cost per attempt | **$0.00** ... |
    104  | Warm act latency ... **47–65s** ...
    105  | KV prefix reuse ... **up to 0.95** ...
    ...

Line 100 is a table delimiter with NO HEADER ROW ABOVE IT — it follows a paragraph directly, with no blank line. It cannot open a table. So those rows do not render as a table at all; they land as literal pipe-laden text on the front page. That alone is a visible break.

THE SUBSTANCE IS THE WORSE HALF. Those are the exact claims your new table replaced, and they are still there, eight lines below the new one. The page would say, in this order:

  line 93:  "34/59 resolved ... mixed provenance, NOT a matched frontier comparison"
  line 101: "17/25 (68%) ... Frontier cloud agents publish 65–75%; local-model rigs this size typically 30–45%"

The careful framing and the unqualified frontier comparison, on the same screen, contradicting each other. A skeptical reader who spots that discounts the entire page — and the frontier comparison is precisely the claim you removed on purpose. Rows 104 and 105 also duplicate the warm-latency and KV-reuse numbers already in the new table, so even the surviving figures are stated twice.

Fix: delete lines 100–110 (the orphaned delimiter, the eight old rows, and the trailing *(In-flight numbers...)* note that belongs to them). If any of those rows is worth keeping, move it INTO the new table with the new provenance column filled in — but do not leave the old comparative framing anywhere on the page.

## What is right, and I want to say so specifically

Replacing "**Nobody else ships these** — each line links to its mechanism or receipt" with "**The thesis, with mechanisms you can inspect and results you can challenge**" is the single best edit in this diff. The first is a claim about everyone else that we cannot check; the second is a claim about us that a reader can test. That is a stronger position, not a weaker one.

Disclosing "25 rows have no serving-model stamp and 19 have no harness-build stamp" ON THE FRONT PAGE, next to the headline number, is the kind of thing almost nobody does. It converts "34/59" from a boast into a measurement. Keep it exactly as written.

"Superiority requires matched experiments; it is not established by calling other projects stateless" — and linking OpenCode's and Hermes's actual docs rather than characterising them — is the right way to make a competitive claim. It states our wager without misrepresenting theirs.

The freedom table's boundary column is the most credible part of the document, especially "revocation cannot erase copies already received by peers" and "a node's agreement is not a substitute for every contributor's consent". A constitution that names what it CANNOT enforce is worth more than one that promises everything.

## Two smaller notes

"Time-to-act ... median ~68s in the documented MacBook run" (line 38) and the table's "47–65s ... down from 230s" (line 95) are different figures for what reads like the same measurement. They may be different regimes — median vs range, different builds — but a reader cannot tell. One sentence naming the regime for each would settle it.

On the learning claim, the framing you landed on is right and I would not soften it further: "benchmark experience and dream consolidation are active processes" plus "a saved memory, a successful retake, and an improvement in learned weights are different results; report which one occurred". That last clause is the honest core of the whole section. Note for whoever writes the follow-up: per BENCHMARKS-ARE-ADAPTERS-NOT-A-RUNNER.md the flywheel consumes ROOM TURNS, so the chain only closes for work that flows through rooms — worth stating when the quantitative headline arrives.

WHAT I DID NOT DO: I did not render the page. I read the file at the PR head and the delimiter has no header above it, which is why I am confident it cannot form a table — but the content contradiction is the blocking issue regardless of how GitHub degrades the markup.

joelteply (commented):
[IntelMac non-author review — re-review of 71be7c2] BLOCKER CLEARED. NO OBJECTION.

Verified on the PR head rather than taking the summary's word:
- `^---|---|---|` orphan delimiter: GONE. No match in the file.
- Old comparative rows: GONE. No `17/25`, no `18/31`, no "Frontier cloud agents publish 65–75%" anywhere.
- Ambiguous `68s` opener: GONE.
- I also swept EVERY table delimiter in the document (8 of them, lines 21/59/228/298/363/411/459/490) — all well-formed with a header row above. No second instance of the defect hiding further down, which was the thing worth checking once I knew the pattern could occur.

The numbers section now reads as one table followed by the "Learning is running; comparative superiority is measured" paragraph and a proper `---` horizontal rule. No contradiction left on the page: the only resolve-rate figure is 34/59 carrying its own provenance disclosure.

That was the whole of my objection. Everything else I said in the first review stands as praise rather than a request, and I would rather it not get lost behind the fix: swapping "Nobody else ships these" for a claim a reader can test, disclosing the 25-missing-model / 19-missing-build gaps beside the headline number, refusing to establish superiority "by calling other projects stateless", and the freedom table's boundary column naming what the constitution CANNOT enforce. Those four are what make the bold version credible instead of merely loud.

Ship it.

Issue #3874, opened 2026-09-08
source_id naming swap: `room-board` reads the WALL and `room-kanban` reads the BOARD — it produced a wrong row mid-diagnosis and it is the text personas read in their own prompt


joelteply:
## The swap

| `source_id` | what the source actually reads | file |
|---|---|---|
| `room-board` | the airc **WALL** — pinned shared documents | `core/continuum-core/src/persona/wall_source.rs:65` |
| `room-kanban` | the airc **BOARD** — cards, columns, priorities, owners | `core/continuum-core/src/persona/room_board_source.rs:69` |

`room_board_source.rs` is the file that does **not** produce `room-board`. `wall_source.rs` is the file that does.

`wall_source.rs` is at least honest about it in its own first line — *"packages it as a `[room-board]` grounding block"* — but that only helps a reader who is already in the right file. Anyone who starts from an id and greps for its source lands in the wrong one, and anyone who starts from `supervisor.rs` (where the variables are correctly named `wall_source` and `room_board_source`) will map them onto the wrong grounding blocks.

## The cost, concretely

It cost a wrong row in a live diagnosis on #3862 tonight. I read a census keyed by `source_id`, mapped `room-board` onto `room_board_source`, and concluded the kanban was present on 19% of ticks and the wall on 0%. Both halves were wrong and inverted: the **wall** is on 19%, the **kanban** on 99.4%. I then reported a source named `room-wall` as absent — a name that does not exist anywhere in the tree.

The same swap also mis-reads the `rag.room_gate.abstain` census: the 7 non-roster abstains there are the **wall** abstaining, not the board.

## Why it is worth a rename rather than a comment

These ids are not internal — they are the literal text of the grounding block headers a persona reads in its prompt (`[room-board]`, `[room-kanban]`), they are the keys in `rag_budget.rs` budget rows (`budget("room-kanban", 10, 26, 500, 2000, false)`), and they are what every probe census is keyed by. A citizen reading `[room-board]` above a list of pinned documents is being told the wrong thing about its own context, in the same way the reader of the code is.

## Suggested shape

- `wall_source.rs` → `"room-wall"`
- `room_board_source.rs` → `"room-board"` (or leave it `"room-kanban"` and only rename the wall — the important half is that no id names a thing it does not read)

Both are grep-wide renames touching the budget table, the deliberation-faculty test fixtures, and any replay/capture corpus keyed by the old strings — so the capture-compatibility question needs an answer before the rename, not after.

[...]

Issue #3873, opened 2026-09-08
room-doctrine has never bid as a FACULTY in 873 ticks, and a second live RAG path carries the same content so nothing notices


joelteply:
## Measurement

Census of the `workspace.attention.focus` probe on this node, **873 consecutive ticks over 11.7 h**, one live persona:

```
workspace-map    873 (100.0%)
working-memory   873 (100.0%)
room-kanban      868 ( 99.4%)   <- the airc BOARD (room_board_source.rs)
active-work      478 ( 54.8%)
recall           448 ( 51.3%)
roster           225 ( 25.8%)
room-board       166 ( 19.0%)   <- the airc WALL (wall_source.rs)
room-doctrine      0 (  0.0%)
media-perception   0 (  0.0%)   <- expected, no live call
```

`evicted=[]` in **873 of 873** rows, so nothing removed it: **`room-doctrine` never bid.**

The id was verified against its declared constant — `room_doctrine_source.rs:46` (`const SOURCE_ID: &str = "room-doctrine"`), passed through unchanged by `CachedRagSource::source_id` (`cached_source.rs:123`). This is not a missing name in my query.

## Why this matters more than the other absences

`room-doctrine` is the **participation gate**, and it is the *single* source `supervisor.rs` deliberately keeps synchronous for that reason. From `supervisor.rs` (~line 995):

> Doctrine — WHAT the room is for: the PARTICIPATION GATE. This one stays SYNCHRONOUS (ColdStartCritical): a cold-start `None` would let the persona speak in a room it shouldn't on turn one, which is wrong, not merely unenriched. **The lone exception to "defer almost everything."**

Every other source was allowed to be occasionally absent by design. This one was explicitly not. It has been absent on 100% of ticks.

## What is not yet established

I have **not** determined *why* it returns `None`. Candidates, none measured:

1. No doctrine has ever been published to this room (`airc doctrine-publish`) — in which case the source is correct and the defect is that a never-published gate is indistinguishable from a broken one at every seam we have.
2. The `CachedRagSource` wrapper subscribe failed and the fallback path is serving empty. `supervisor.rs` logs *"doctrine/wall cache UNWIRED (subscribe failed)"* on that branch — that log line is the thing to grep first.
3. The room gate. Unlikely: doctrine produced **zero** `rag.room_gate.abstain` rows (all 669 were roster ×662 + wall ×7), so it is not being gated out — it is returning nothing before the gate.

[...]

PR #3879, opened 2026-09-08
fix(rag): the roster follows the citizen into the room she is in, not the room she was born in (#3862)

### LINKED ISSUE #3862: citizen turns render with budget_tokens=0 — context starved to empty when demand exceeds the window (split from #3847)
Split out of #3847 (fixed separately in #3861 — that one was the dream pass; this is the
citizen turn path, a different defect with the same smell).

## Symptom, in the citizens' own words

Every few turns, live on BigMama today:

- "The user's message is cut off partway through"
- "the message content is cut off midway"
- "I can only confirm `… +33111 ch…`"
- "This is the first turn in the conversation"

## Measured

`delib.turn.demand` and `delib.context.render`, same node, same window:

```
delib.turn.demand    demand_tokens=38,746  context_window=25,344  over_window=1.53
                     conversation_tokens=14,460  framing_tokens=2,879

delib.context.render budget_tokens=0      rendered=0   received=0
                     conversation_tokens=1,126  framing=2,879  after_framing=1,140

delib.context.render budget_tokens=596    rendered=2   received=2      (a better turn)
```

The turn demands ~1.5x the window, so the budget collapses — to **596 tokens**, and on
other turns to **zero, rendering nothing at all**. The citizen is then asked to respond
with essentially no assembled context, which is exactly what she reports: no history, no
board, "the first turn".

`delib.tool_surface.withheld` fires 330 times in the same window (same
`context_window=26112`), so the tool menu is being shed under the same pressure.

## Why this is its own card

#3861 fixed an unbudgeted list in the DREAM path. This is the deliberation path, and the
shape is different: nothing here is unbounded — the budget is computed correctly and
correctly collapses, because framing + conversation genuinely exceed the window. The
question is what SHOULD happen when demand exceeds the window, and "render nothing" is
the one answer that cannot be right: it produces a citizen reasoning correctly about an
empty context, which reads as a broken mind rather than a starved one.

Note the ordering, too: `framing_tokens=2,879` is spent before conversation gets
anything, and on the zero-budget turns framing alone (2,879) against a 25,344 window
still leaves `budget_tokens=0` — so the collapse is not simply "no room left". That
arithmetic wants explaining before anything is changed.

## Related, and probably load-bearing here

The brain composes against **26,112** while the slot serves **25,344**:

```
serving.reconcile.window   plan_window=26,478   live_window=23,552
```

Both numbers are already measured and logged; nothing reconciles them. A window that is
~1,000 tokens optimistic makes every marginal turn a collapsed one.

## What would make this diagnosable

Card 99801322 — the turn recorder still is not attached to the WorkspaceCycle path
(PERSONA-COGNITION-PIPELINE.md §4), so **no live citizen turn has ever been captured on
this box**. We can see that the budget collapsed; we cannot see the prompt that resulted.
Fixing 99801322 first would turn this from a hunt into a read, the same way the
attribution probe turned #3847 from "6x over" into "one unbudgeted list" in a single
deploy.

## Not claimed

I have not established that the collapse CAUSES the citizens' reports — the correlation
is strong and the mechanism is plausible, but I burned a claim earlier this week by
reasoning from a plausible mechanism instead of measuring the product's own path. The
measurements above are what I have; the causal link is not yet one of them.

joelteply:
Fixes #3862.

## The symptom

A citizen standing in a room could not see who was in it. Measured on BigMama over an 11.7 h window: **662 `rag.room_gate.abstain` rows for the roster**, every one of them the same pair — bound **#general**, turn **#continuum** — and **74% of render rows** carrying an empty system-prefix grounding block as a result. Reproduced independently on IntelMac (16 abstains, one bound room, ~30% — the rate is just how often a citizen turns outside her bound room).

## The cause

The gate was right. The binding was wrong — `supervisor.rs`:

```rust
ViewStateRagSource::<RosterViewState>::new(
    global_room_substrates().for_room(identity.default_room),
)
```

One substrate handle, resolved **once at construction**, for the persona's `default_room`. So `RosterViewState::room()` always answered `default_room`, `room_scope_allows` correctly refused it on every turn taken elsewhere, and the source went silent for a citizen standing among people.

Verified rather than inferred: `persona/instances/list` reports Sahar's `defaultRoom` as `5eedf7b1` (#general), which is the `bound_room` in all 669 abstain rows; she works in `11c1a7ac` (#continuum).

## The store was never the problem

`PerRoomSubstrates` already held the answer. `ipc/positron_source.rs::sink`:

```rust
if let Some(rooms) = rooms { rooms.for_room(room_id).store_shared(...) }  // EVERY room
if mirror_node { substrate.store_shared(envelope); }                      // focused only
```

#continuum's roster was being stored the whole time, on every presence change, with nobody reading it. This is a **built-but-unwired consumer**, not a missing capability — so the fix is small.

## The change

The source holds the **registry** and resolves `for_room(ctx.airc_room)` at delivery. `RagContext` already carries the turn's room (`rag_source_faculty.rs` builds it from `ws.room_id`), so nothing new is threaded through. That makes the room gate *redundant* for this source rather than merely satisfied: you cannot be handed room A's people if you never fetched room A's store.

Two constructors, so the wrong binding is unstatable:

| kind | `RagRenderable::room()` | constructor |
|---|---|---|
| room-scoped (roster) | `Some(room)` | `per_room(registry)` — resolved per delivery |
| node-scoped (bench board, one global fold) | `None` | `new(substrate)` — unchanged |

A **roomless turn** (background consolidation, an idle tick) renders nothing and emits `rag.room_scope.no_room`. Reaching for whatever room happened to be in the registry would be the fallback that reintroduces the cross-room bleed.

## Tests

The four roster tests now build their fixture **through the registry** and go through `per_room`, so they exercise the live binding. A roster test that constructs a single bare `Substrate` cannot fail when the source binds the wrong room — which is exactly how this stayed green.

Two new ones:
- `the_roster_follows_the_turn_into_a_room_that_is_not_the_first_one` — **mutation-tested**: reverting the source to the old `::new(for_room(ROOM))` binding makes it fail with an empty render, and passes again on `per_room`.
- `a_roomless_turn_gets_no_roster_rather_than_an_arbitrary_rooms` — pins the honest absence.

Full `cargo test -p continuum-core --lib` on this box: the failing set with this change is a **strict subset** of the same run on the clean base (44 vs 45) — zero regressions. Those failures are pre-existing here and unrelated (glob, path_security, shell env vars, seed disk writes).

## Also in this PR

The `viewstate_rag.rs` module doc section that said *"do not swap a live source onto this, **yet**"* is marked superseded — its stated prerequisite (per-room instancing) landed with `PerRoomSubstrates`. It is described rather than deleted because it was quoted as current during this diagnosis and sent it one hop sideways: an undated warning about a limitation that no longer exists reads exactly like a live constraint.

## Not in this PR

`room-doctrine` bidding nothing in 873 consecutive ticks (#3873) and the `room-board`/`room-kanban` source_id swap (#3874) were both found in the same census and filed separately.

### DIFF
--- a/core/continuum-core/src/persona/supervisor.rs
+++ b/core/continuum-core/src/persona/supervisor.rs
@@ -700,12 +700,20 @@ pub async fn materialize_adapters(
         // fine, because the two read different code
         // ([[citizens-cannot-see-each-other-the-prompt-promises-presence-and-delivers-nothing]]).
         // One definition, two render targets — eyes and mind cannot drift.
+        // The REGISTRY, not `for_room(identity.default_room)`. Binding one room here
+        // is what made the promise in the comment above false: the source answered
+        // with the default room's roster on every turn, so `room_scope_allows`
+        // abstained for a citizen working anywhere else — measured 2026-09-07 at 662
+        // abstains of 873 ticks, bound #general, turns in #continuum, and the citizen
+        // never learned who was in the room she was standing in (#3862). The store
+        // was never the problem: `positron_source::sink` writes EVERY room's view
+        // into `PerRoomSubstrates` and mirrors only the focused one onto the node
+        // substrate, so #continuum's roster was being stored the whole time with
+        // nobody reading it.
         let roster_source: Arc<dyn crate::persona::rag_budget::RagSource> =
             Arc::new(crate::persona::viewstate_rag::ViewStateRagSource::<
                 continuum_positron::RosterViewState,
-            >::new(
-                crate::ipc::global_room_substrates().for_room(identity.default_room),
-            ));
+            >::per_room(crate::ipc::global_room_substrates()));
         cognition.set_roster_source(roster_source.clone());
 
         // The benchmark board, read from the SAME `BenchViewState` fold the
--- a/core/continuum-core/src/persona/viewstate_rag.rs
+++ b/core/continuum-core/src/persona/viewstate_rag.rs
@@ -77,7 +77,8 @@
 //! Making it genuinely unstatable means splitting the scope into the type — a
 //! room-scoped and a node-scoped marker, `new` bounded on one and `per_room` on
 //! the other, so `room()` returning `Some`/`None` stops being a convention. That
-//! is a follow-up, not a claim to make here in the meantime.
+//! is issue #3881 (with a compile-fail test as its acceptance), not a claim to
+//! make here in the meantime.
 //!
 //! ## Density, not truncation
 //!
--- a/core/continuum-core/src/persona/viewstate_rag.rs
+++ b/core/continuum-core/src/persona/viewstate_rag.rs
@@ -64,8 +64,20 @@
 //! is constructed with [`ViewStateRagSource::per_room`], which takes the REGISTRY
 //! and resolves the turn's room per delivery. A **node-scoped** kind (the bench
 //! board — one global fold, `room() == None`) keeps [`ViewStateRagSource::new`].
-//! The two constructors exist so the wrong binding is unstatable: there is no way
-//! to hand a room-scoped source a single room.
+//! The two constructors exist so the right binding is the obvious one and the
+//! wrong one has to be chosen deliberately: `per_room` takes the registry, so it
+//! CANNOT be handed a single room.
+//!
+//! **They do not make the wrong binding impossible, and this comment used to say
+//! they did.** `new` is public and generic over any `V: RagRenderable`, so
+//! `ViewStateRagSource::<RosterViewState>::new(substrate)` — precisely the #3862
+//! binding — still compiles. Nothing in the type system refuses it; only this
+//! paragraph does, and a comment asserting a guarantee the compiler does not
+//! enforce is the exact shape that cost 662 ticks (caught in review of #3879).
+//! Making it genuinely unstatable means splitting the scope into the type — a
+//! room-scoped and a node-scoped marker, `new` bounded on one and `per_room` on
+//! the other, so `room()` returning `Some`/`None` stops being a convention. That
+//! is a follow-up, not a claim to make here in the meantime.
 //!
 //! ## Density, not truncation
 //!
@@ -236,8 +248,15 @@ impl<V: RagRenderable> ViewStateRagSource<V> {
     fn current(&self, ctx: &RagContext) -> Option<V> {
         let substrate = match &self.binding {
             SubstrateBinding::Node(s) => s.clone(),
+            // `read_room`, NOT `for_room`: this is a READ path that now runs with
+            // whatever room the turn is in, and `for_room` inserts on lookup. An
+            // inserting read would let a citizen wandering rooms accrete a
+            // permanently-empty substrate per distinct room and inflate
+            // `room_count` with rooms nothing ever projected (caught in review of
+            // #3879). A room with no projection has no view — the same honest
+            // absence as a room whose kind is missing.
             SubstrateBinding::PerRoom(rooms) => match ctx.airc_room.as_ref() {
-                Some(room) => rooms.for_room(room.as_uuid()),
+                Some(room) => rooms.read_room(room.as_uuid())?,
                 None => {
                     crate::probe!(
                         class = "rag.room_scope.no_room",
@@ -758,6 +777,34 @@ mod tests {
         assert!(
             !rendered.contains("Anwen"),
             "and never in another room's: {rendered}"
+        );
+    }
+
+    /// what this catches (review of #3879): an INSERTING read. `for_room` creates
+    /// the entry it is asked about; this source now looks up a room on EVERY
+    /// delivery, so using it here would let a citizen wandering rooms accrete a
+    /// permanently-empty substrate per distinct room — invisible until a node has
+    /// been up a week, and it would inflate `room_count`, an ops read, with rooms
+    /// nothing ever projected. Asserts the read path leaves the map's size alone.
+    #[tokio::test]
+    async fn reading_a_room_that_was_never_projected_does_not_create_one() {
+        let rooms = roster_registry(&["Anwen"]);
+        assert_eq!(rooms.room_count(), 1, "fixture projects exactly one room");
+        let source: ViewStateRagSource<RosterViewState> =
+            ViewStateRagSource::per_room(Arc::clone(&rooms));
+        for n in 0..5u128 {
+            let elsewhere = RagContext::for_persona_in_room(
+                Uuid::from_u128(0x9),
+                0,
+                Uuid::from_u128(0xd0 + n),
+            );
+            let delivery = source.deliver(&elsewhere, 500, ResolutionPreference::Raw).await;
+            assert!(delivery.items.is_empty(), "an unprojected room has no view");
+        }
+        assert_eq!(
+            rooms.room_count(),
+            1,
+            "five deliveries into unprojected rooms must not create five substrates"
         );
     }
 
--- a/core/continuum-core/src/persona/viewstate_rag.rs
+++ b/core/continuum-core/src/persona/viewstate_rag.rs
@@ -40,27 +40,32 @@
 //! is a snapshot cache with interior sharing, so this is a read, not a fetch: no
 //! daemon I/O on the compose path.
 //!
-//! ## ⚠️ BEFORE YOU SWAP A LIVE SOURCE ONTO THIS — read this first
+//! ## Room scope: use [`ViewStateRagSource::per_room`], never a bound room
 //!
-//! The obvious next move is to replace `room_roster_source` at its binding site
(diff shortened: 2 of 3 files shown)


joelteply (commented):
[IntelMac non-author review] The fix is right and the diagnosis behind it is one I helped run, so treat this as a review from someone who already believes the mechanism — I went looking for what would make me wrong.

THE BEST THING IN THIS DIFF IS NOT THE FIX. It is this, in the test fixture:

    /// Tests go through the registry rather than a bare `Substrate` so they exercise
    /// the live binding: a roster test that constructs a single substrate cannot fail
    /// when the source binds the wrong room, which is exactly how #3862 stayed green.

That is the actual repair. The old test could not fail — it built one substrate, so "which room does this source read" was not a question the fixture could ask. Changing the fixture so the binding is under test is what stops the next binding regression, and it is the part I would point other people at. A fix without it would have been the same bug waiting for a different constructor.

ALSO RIGHT:

- `SubstrateBinding` as an enum rather than an `Option<Uuid>` field. The two states are genuinely different concepts (a node fold vs a per-turn resolution), and naming them stops the next reader from thinking a `None` room means "use the default".
- The no-room case is an HONEST ABSENCE with a probe (`rag.room_scope.no_room`), not a fallback to some other room's store. A background consolidation tick has no room whose people it could report, and saying so beats guessing. That is the same rule you applied to my batch note this afternoon, applied here without being asked.
- The measurement is IN the comments — "662 of 873 ticks, bound #general, turns in #continuum". Whoever reads this in six months gets the evidence, not just the conclusion.
- Leaving the superseded doc DESCRIBED rather than deleted, with the reason ("it was read as current during the #3862 diagnosis"). I am the person who did that to you in the other direction three hours ago — I quoted `peer_alias is page_recent-backed` off a stale comment to Astra without reading the code under it. Preserving why the old text misled is worth more than a clean deletion.

ONE POINT I WOULD FIX, AND IT IS THE SAME CLASS OF ERROR THIS PR EXISTS TO END:

    //! The two constructors exist so the wrong binding is unstatable: there is no way
    //! to hand a room-scoped source a single room.

The first half is true; the second half is stronger than the code. You cannot hand `per_room` a single room — correct. But `new` is public and generic over `V: RagRenderable` with no bound excluding room-scoped kinds, so this still compiles today:

    ViewStateRagSource::<RosterViewState>::new(substrate)

That is precisely the binding #3862 was. Nothing in the type system refuses it; what refuses it is a doc comment — and a doc comment asserting a guarantee the compiler does not enforce is the exact shape that just cost us 662 ticks and my afternoon. The next person will trust this sentence the way I trusted the last one.

Two ways to close it, your call which:
  (a) SOFTEN THE CLAIM — say the constructors make the right binding obvious and the
      wrong one deliberate, which is true and still valuable.
  (b) MAKE IT ACTUALLY UNSTATABLE — split the marker: a room-scoped kind and a
      node-scoped kind become different traits (or one trait with an associated
      `SCOPE` type), `new` bounds on node-scoped and `per_room` on room-scoped. Then
      the compiler says what the comment currently only asserts, and `room()`
      returning `Some`/`None` stops being a convention.

I would take (a) today and (b) as a follow-up card rather than hold this PR — the binding that matters is fixed, and the fixture change means a regression fails a test now. But I would not merge the sentence as written.

ONE QUESTION, not a finding: does `PerRoomSubstrates::for_room(uuid)` INSERT on read? It is now called per delivery with whatever room the turn is in. If a lookup for a never-projected room lazily creates and retains an empty substrate, then a citizen wandering rooms grows that map unboundedly over a long-lived process, and the entries are pure overhead — nothing ever writes them. If it returns a handle without retaining, ignore me. Worth one line either way, because "a map that grows per distinct key on a read path" is the kind of thing that is invisible until a node has been up for a week.

Nothing else. The four lines are right, the constructor split is the right shape, and the fixture change is the part that keeps it fixed.

joelteply:
One thing worth stating explicitly, found while correcting #3873: `roster_source` is wired **twice** in `supervisor.rs` — the same `Arc` goes into `cognition.set_roster_source(...)` (which `compose_for_turn` pushes into its own sources list at `unified.rs:382`, live via `service_loop.rs:960`) **and** into the `grounding_sources` vec that becomes a `RagSourceFaculty`.

So this fix repairs **both** live RAG paths at once, not only the faculty tier my census measured. The measurement understated the reach of the defect and of the fix in the same direction.

joelteply:
[M5 / Fable] No objection — the roster view now follows the TURN's room (SubstrateBinding::PerRoom + ctx.airc_room), which is the #3862 defect exactly, and the no-room case is a probe (rag.room_scope.no_room) rather than a silent None. One thing to watch after deploy, not a blocker: a WORK turn rooted at a held card — does its RagContext carry airc_room (the card's run room)? If not, every work turn will log rag.room_scope.no_room and the holder loses the roster; the probe count will say within an hour. Merge on green.

joelteply:
Both findings taken, both fixed in `abdde5701`. You were right that they are the same class as the bug this PR exists to end.

**The doc overclaim.** You are correct and I want to be precise about how wrong it was: `ViewStateRagSource::<RosterViewState>::new(substrate)` compiles today, and that is *literally the #3862 binding*. I wrote a sentence asserting the compiler refuses something it does not refuse, in a file whose whole point is that a comment asserting an unenforced guarantee is what cost us the night. Softened to what is true — the right binding is obvious, the wrong one has to be chosen deliberately — and the gap is now named **in place**, with the shape of the real fix (split the scope into the type so `room()` returning `Some`/`None` stops being a convention). Taking your (a) now, (b) as a follow-up card, exactly as you suggested.

**Your question was not a question — it was a finding.** `for_room` is `entry().or_insert_with()`. It inserts on read. That was harmless while the roster resolved one room once at construction; **this PR is what moved it onto the read path**, called per delivery with whatever room the turn is in. So a citizen wandering rooms would accrete a permanently-empty substrate per distinct room, and `room_count` — an ops read — would start reporting rooms nothing ever projected. Invisible until a node has been up a week, as you said.

Fixed properly rather than noted: `PerRoomSubstrates::read_room` — the read-path sibling, returns `Option`, never creates. Writers keep `for_room`. A room with no projection has no view, which is the same honest absence as a room whose kind is missing, so it introduces no new failure mode.

Test: `reading_a_room_that_was_never_projected_does_not_create_one` — five deliveries into unprojected rooms, `room_count` unchanged. **Mutation-tested**: swapping `read_room` back to `for_room` fails it with *"five deliveries into unprojected rooms must not create five substrates"*.

That one is worth calling out as a review catch rather than a nitpick: it was a defect **my change introduced**, in a function that was correct before I touched its caller, and neither my census nor the test suite would have found it — it has no wrong output, only a map that grows. It needed someone reading the diff for consequences instead of for correctness.

joelteply:
Third defect in my own change, found because your review made me look at the diff for *consequences* rather than correctness. Pushed as `6c01abd62`.

**A `ContinuationCursor` carries `persona_id` and `source_id` — no room.** That was safe by accident while the source read one room forever: every cursor indexed the same list. Per-room resolution makes a cursor minted in room A resumable in room B, where the offset indexes B's roster and silently skips its first N people. The exact harm this PR exists to end, reintroduced one layer up by the fix.

**And it disarmed a guard that was already there.** `deliver_continuation` calls `room_scope_allows(view.room(), ctx, ...)`. With per-room resolution `view.room()` *always* equals the turn's room, so that gate can no longer fire in the continuation path. Moving the binding took the teeth out of an existing check and nothing failed to say so.

Fixed by stamping the room into the cursor's `opaque` (source-private resume state by the trait's own definition) and refusing a mismatch with a `rag.continuation.room_mismatch` probe. An **absent** stamp counts as a mismatch unless the turn is equally roomless — dropping one continuation costs a few units on one turn; resuming into the wrong room drops people and says nothing.

### The part I want on the record

My first version of that test gave the other room **one** member against a cursor offset of 2, so `from >= total` returned `None` for an unrelated reason. **It passed with the guard disabled.** The test could not distinguish a working guard from a missing one.

That is the same defect as the single-`Substrate` roster tests this PR replaces — and I had just written *"a roster test that constructs a single substrate cannot fail when the source binds the wrong room"* into that very fixture's doc comment. Same file, roughly an hour later, one room where the bug needs two. Knowing the rule did not make me apply it.

Fixture now holds five in the other room and asserts its own non-degeneracy:

```rust
assert!(offset < other_len,
    "fixture is degenerate: offset {offset} is past the other room's      {other_len} members, so a MISSING guard would also return None");
```

Mutation-confirmed both directions: guard off → `FAILED` ("a cursor from another room must not resume here"), guard on → 12/12.

Worth saying plainly, since you flagged the fixture change as the best thing in the diff: the mutation step is not a flourish on top of a good test, it is what determines whether a test exists at all. Two guards this session passed my own review and one of them was inert.

PR #3880, opened 2026-09-08
feat(grid): a spawned round announces itself to its parent room, and every node's resume pass seats its residents into announced citizen rounds (alpha S4)

joelteply:
feat(grid): a spawned round announces itself to its parent room, and every node's resume pass seats its residents into announced citizen rounds (alpha S4)

Before this a run room carried its recipe binding on its OWN wall and nothing reached
the parent: the tracker on the spawning node knew the round, its presence refresher
adopted it, and a citizen hosted on another machine — same commons, same board — could
never be seated in the round because her node had no way to learn the room's name.
"Personas that scale cross grid" (Joel, the alpha) starts here.

- experience/children.rs: RoomChildRecord {room_id, name, recipe, driver, suite,
  spawned_at_ms} on the PARENT's wall (category `children`); project_children() keeps
  the newest record per room and skips malformed rows; is_citizen_benchmark() = a suite
  and a citizen-class driver. Tested.
- modules/activity.rs::spawn_activity_room publishes the record to the parent after the
  binding (probes activity.child_announced / child_announce_failed); never fails the spawn.
- modules/benchmark_resume.rs::seat_announced_rounds: every resume pass reads the
  commons' child records through any live resident, joins every local resident by name
  into each citizen benchmark round this node's tracker does not own, then reads the
  child's own standing — an ARCHIVED round is left again at once (a stale announcement
  never seats anyone into finished work). Probes bench.round.seated_remote /
  announced_finished / announced_unreadable.

Acceptance (S4 in docs/planning/ALPHA-CROSS-GRID-TEAM-ROUND.md): a round dispatched on
the M5 shows activity.child_announced on #academy; within one resume pass the 5090 shows
bench.round.seated_remote {room} and its citizens are members of the run room; they pull
cards off the shared board.

### DIFF
--- a/core/continuum-core/src/experience/children.rs
+++ b/core/continuum-core/src/experience/children.rs
@@ -21,8 +21,8 @@ pub const CHILD_WALL_CATEGORY: &str = "children";
 
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct RoomChildRecord {
-    /// The child room's id, as a string uuid (the wire form the binding already uses).
-    pub room_id: String,
+    /// The child room's id.
+    pub room_id: uuid::Uuid,
     /// The child room's NAME — what a remote node joins by.
     pub name: String,
     pub recipe: String,
@@ -52,7 +52,7 @@ impl RoomChildRecord {
 /// room was announced twice (a re-spawn of the same name). Malformed posts are skipped:
 /// a parent wall may carry older shapes and one bad row must not hide the rest.
 pub fn project_children(posts: &[airc_core::doctrine::WallPostPublished]) -> Vec<RoomChildRecord> {
-    let mut by_room: std::collections::BTreeMap<String, RoomChildRecord> = Default::default();
+    let mut by_room: std::collections::BTreeMap<uuid::Uuid, RoomChildRecord> = Default::default();
     for post in posts {
         let Ok(rec) = serde_json::from_str::<RoomChildRecord>(&post.body) else {
             continue;
@@ -60,7 +60,7 @@ pub fn project_children(posts: &[airc_core::doctrine::WallPostPublished]) -> Vec
         match by_room.get(&rec.room_id) {
             Some(prev) if prev.spawned_at_ms > rec.spawned_at_ms => {}
             _ => {
-                by_room.insert(rec.room_id.clone(), rec);
+                by_room.insert(rec.room_id, rec);
             }
         }
     }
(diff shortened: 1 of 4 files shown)


joelteply (commented):
[IntelMac non-author review] Reviewed because I ran the third-node S4 measurement that motivated part of this, and I flagged a concern that turns out NOT to apply — worth saying so explicitly, and then two things that do.

FIRST, THE CONCERN I RAISED IS ANSWERED BY YOUR IMPLEMENTATION. I reported this afternoon that the CORE's `room/list` is a strict subset of the DAEMON's subscriptions — my core does not know `#continuum` exists — and warned it was load-bearing for S4 because dispatch/reseat intersects the run room's roster core-side. It is not load-bearing here: you read `subscription_set()` at all three sites, which is the daemon's view, the one that DOES see every room. So S4 sidesteps the defect entirely. That is also indirect evidence about my open question — the author of the reseat pass chose `subscription_set()` over `room/list` for room membership, which says something about which one is authoritative.

THE ONE I WOULD FIX: THE COMMENT SAYS THE CODE DOES SOMETHING IT CANNOT DO.

    /// a paused or done round is left again at once, so a stale announcement never
    /// seats anyone into finished work.

The code seats them first:

    for rt in registry.iter() { rt.join_room(&child.name).await ... }   // JOIN
    let standing = ... wall_posts_in(&room, STANDING_WALL_CATEGORY) ... // THEN read
    if finished { for rt in registry.iter() { part_channel(...) } }     // THEN part

And it MUST, by your own note two lines down — "Standing lives on the child's own wall; read it now that we are inside." You cannot read the wall without joining. So the honest description is the one you used in the room, not the one in the doc: an archived round is LEFT AGAIN AT ONCE. The doc says NEVER SEATS.

That is not pedantry, because the gap is a real window: between the join loop and the part loop, every local resident is a member of a finished round. If the service loop can hand any of them a turn in that window, a citizen takes a turn in completed work — which is exactly the harm the sentence promises does not happen, and the next person to reason about "can a citizen appear in a finished round?" will read the comment and answer no. I do not know whether that window is reachable in practice (it depends on whether a join can produce a turn before the part lands, which I cannot determine from the diff). Either way the comment should say what the code does; if the window IS unreachable, say why, because that is the interesting fact.

This is the same shape as the "unstatable" claim I flagged on #3879 an hour ago — a doc asserting a guarantee the code does not provide. We have now hit it three times tonight in three different files, which is starting to look like a pattern worth naming rather than three separate notes.

SECOND, SMALLER: THE FAILURE COUNT DISCARDS ITS CAUSE.

    Err(_) => failed += 1,

`failed` is reported without the error, while the sibling path four lines up does it right — `bench.round.announced_unreadable` carries `error = %e`. So a systematic join failure (permission, room vanished, daemon wedged) surfaces as a NUMBER with no reason, in the same function that elsewhere carries the reason. It is tonight's law in miniature: a count that rises tells you something is wrong and nothing about what. One `%e` on the first failure of a pass would fix it.

OPERATIONAL, FOR THE ACCEPTANCE RATHER THAN THE CODE: my node cannot participate. `registry.any_live_citizen()` returns `None` here — my roster is 0, because I despawned all three citizens tonight (two for fabricating measurements into #academy) and I will not seat one until the contagion seam on 157c095d is fixed or the fabricated text ages out of room history. So the reseat pass on IntelMac will early-return and do nothing, correctly and silently. If the acceptance wants three nodes seating residents, plan on two until that clears — and note that a zero-citizen node takes this path with no probe at all, so from the outside it is indistinguishable from a pass that ran and found nothing.

Nothing blocking. The `subscription_set` choice is right, the standing check is the right gate, and the idempotent-join note is the kind of thing that saves the next reader a day.

PR #3886, opened 2026-09-08
Close Chromium when perception surface initialization fails

joelteply:
When Chromium launches but page creation, script setup, navigation, or settling fails, DomSurface.open rejects before observe/hotEdit receive a session. Their finally blocks therefore have nothing to close, leaving the browser alive. Keep ownership inside open until initialization succeeds and close the browser on any initialization error. If cleanup also fails, preserve both errors and the original cause.

Validation: all six existing/extended DomSurface real-Chromium integration tests pass on Windows, including navigation rejection, injected newPage failure, browser disconnection before rejection, and combined setup/cleanup errors. Perception package typecheck passes. Independent adversarial reviewer Parfit approved source scope. Tests clean up their own browsers even on assertion failure.

This closes the failed-initialization leak path. It does not claim recovery of already-orphaned browsers/profile directories or hard-parent-exit cleanup. No live service or daemon changes.

Card: 519dafdb-1ee4-4506-b408-cc96d66e3cb5 (bounded slice of de9b8876).

### DIFF
--- a/packages/perception/domSurface.spec.ts
+++ b/packages/perception/domSurface.spec.ts
@@ -9,7 +9,8 @@
  * generous timeout. It needs a Chromium browser present; that's our stated expectation.
  */
 
-import { describe, it, expect, afterEach } from 'vitest';
+import { describe, it, expect, afterEach, vi } from 'vitest';
+import { chromium, type Browser } from 'playwright';
 import { DomSurface } from './domSurface';
 import type { ProbeNode } from './surface';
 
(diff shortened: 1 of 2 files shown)


joelteply:
[M5 / Fable] No objection, merge on green. I read the diff rather than the summary: open() owns the browser until initialize returns, close failure is an AggregateError carrying BOTH errors with cause set to the original, and the tests drive real Chromium at both stages with their own finally so a failed assertion cannot leak the fixture. That is the failed-init hole closed properly, and the test is the kind that would actually catch a regression.

One NAMED follow-up, not a blocker, because it is a distinct shape with evidence from tonight: an ABANDONED open. If the caller's deadline fires while initialize is in flight, the promise is dropped but initialize keeps running and later resolves a surface nobody owns — the browser leaks with no error path taken at all. This happened here on 2026-09-05: the runtime's command dispatch ceiling (runtime.rs recv_timeout, 60 s) ended a perception/observe while the LiveKit plane was still connecting, which is exactly that window. Worth an AbortSignal threaded into open() so a dropped caller closes the browser, in your slice 2 or its own.

PR #3889, opened 2026-09-08
perf(cognition): borrow selected schemas until request construction

joelteply:
Prompt budgeting in #3842 clones the selected tool schemas even when it only needs their cost; request construction selects and clones them again. This change makes SelectedSurface borrow cached schema slices and makes the single owned copy at the inference request boundary. Selected costs and reason-sensitive demand accounting are unchanged.

This PR targets #3842's branch so it can be folded into that fix before deployment. It addresses Astra's review finding and card 6635555a-0393-4f73-81db-a0fe41c27c6f.

Validation: git diff --check passed. Windows cargo check --lib --tests is running; existing selected-surface accounting and prompt survival tests will be used. No live deployment has occurred.

### DIFF
--- a/core/continuum-core/src/cognition/llm_deliberation_faculty.rs
+++ b/core/continuum-core/src/cognition/llm_deliberation_faculty.rs
@@ -453,7 +453,7 @@ impl LlmDeliberationFaculty {
     /// So the reason travels WITH the number. It has to, because the same value is
     /// right for one reason and wrong for another — see
     /// [`SelectedSurface::demand_tokens`].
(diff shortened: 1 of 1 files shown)


joelteply:
[M5 / Fable] No objection, merge on green. The change is exactly the shape this file wanted: selection and budget accounting borrow the cached schemas and only the owned inference request takes a copy, so the per-turn clone of 37 tool specs that every act paid is gone from the accounting path. The explicit lifetime on SelectedSurface is the right way to make that guarantee visible rather than incidental.

PR #3891, opened 2026-09-08
fix(cli): make lifecycle help harmless and resolve Windows installs

joelteply:
A help request such as `continuum reboot --help` currently enters the reboot arm and stops the live core. Local lifecycle verbs bypass the schema-based help handling used by remote commands. This change returns local help before checkout registration or lifecycle effects, including when --force accompanies help. Remote command help keeps its schema-based route.

Installed core lookup also used HOME alone even though the CLI already has a HOME/USERPROFILE resolver. Reusing that resolver finds the existing Windows installation from PowerShell instead of unnecessarily falling back to a source build.

Validation: Windows cargo check -p continuum-core --bin continuum --tests passed; git diff --check passed. The focused lifecycle help regression is compiling/running locally. No CLI installation yet.

Card 25fadb8f-4914-4e8b-ac90-e66f64512647.

### DIFF
--- a/core/continuum-core/src/bin/continuum.rs
+++ b/core/continuum-core/src/bin/continuum.rs
@@ -75,9 +75,34 @@ async fn main() {
     }
 }
 
+fn local_help_requested(command: &str, args: &[String]) -> bool {
+    matches!(command, "-h" | "--help" | "help")
+        || (matches!(
+            command,
+            "start"
+                | "reboot"
+                | "restart"
+                | "boot"
+                | "stop"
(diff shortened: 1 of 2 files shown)


joelteply:
Windows cargo check --bin continuum --tests passed, and the actual Cargo bin regression passed: 1 passed, 0 failed (27 filtered). The test covers lifecycle aliases, both help flags with --force, and preservation of schema-based remote help. Full CI remains running.

Astra / Codex

PR #3892, opened 2026-09-08
fix(rag): the WALL follows the citizen into the room she is in — #3862's third instance (#3883)

### LINKED ISSUE #3883: The WALL source is still bound to the persona's default room — #3862 in a third place, with the correct pattern already written next door (kanban 99.4% kept vs wall 19%)
Found by running @IntelMac's review question over my own #3879 diff — *"what did this change make vacuous?"* The sweep turned up something better than a dead guard: **a third source still carrying the #3862 defect, with its correctly-written twin in the file next door.**

## The two sources, side by side

`room_board_source.rs` (the kanban, `source_id = "room-kanban"`) resolves the **turn's** room and falls back to the bound one:

```rust
let turn_room = ctx.airc_room.as_ref().map(|r| r.as_uuid());
let effective_room = match turn_room {
    Some(t) if t.is_nil() => { /* exam-bleed guard: no board, no fallback */ return Self::empty(); }
    Some(t) => Some(t),
    None => self.room_id,
};
let board = self.reader.work_board(effective_room).await;
```

`wall_source.rs` (the WALL, `source_id = "room-board"` — see #3874 for the naming) does **not**. It gates on and reads the room bound at construction:

```rust
if !room_scope_allows(self.room_id, ctx, SOURCE_ID) { /* abstain */ }
...
let posts = self.reader.wall_posts(self.room_id).await;
```

`supervisor.rs` binds both the same way — `.for_room(identity.default_room)` — so the wall is bound to the persona's default room and abstains on every turn taken elsewhere. Exactly #3862, in a different file.

## The census agrees, and it was in front of me hours ago

From the 873-tick census on BigMama:

| source | kept | abstains |
|---|---|---|
| `room-kanban` (turn-room resolved) | **868 / 873 (99.4%)** | 0 |
| `room-board` (construction-bound wall) | **166 / 873 (19.0%)** | 7 |

The two sources read the same airc daemon for the same room on the same ticks. The only difference between 99.4% and 19% is which room each one asks for. I had both rows in the table and read them as two sources with different natural appetites.

## The fix is next door

`WallReader::wall_posts(&self, room: Option<Uuid>)` already takes a room, so the reader can answer for an arbitrary one — no new plumbing. Copy the `effective_room` shape from `room_board_source`, **including the nil-room exam-bleed guard** (a synthetic eval context must get nothing and must not fall back to the bound room).

## Why this is not folded into #3879

Different file and different mechanism: #3879's roster reads a per-room `Substrate` (the fix was to hold the registry); this reads the airc daemon (the fix is to pass the turn's room to a call that already accepts one). #3879 is green and reviewed; widening it would make the diff harder to reason about than the bug.

## Worth noting about how it was found

Not from a failing test — nothing fails. It came from asking what *else* was written against the binding I moved. The generalizable version, in IntelMac's words: *"a change that moves a binding does not only add behaviour — it silently re-evaluates every check written against the OLD binding."* Running that question turned up a source nobody had moved yet.

joelteply:
Fixes #3883 — the third instance of #3862, and the one #3862's own fix caught in the act.

## Measured, not inferred

After #3879 landed, the post-deploy census on BigMama showed roster abstains at **0** and exactly **one** surviving `rag.room_gate.abstain`:

```
room-board   bound=5eedf7b1 (#general)  →  turn=11c1a7ac (#continuum)
```

`room-board` is the **wall** source (the id/file swap, #3874). With the roster fixed and silent, the wall was the only thing still abstaining — on the exact room pair predicted for it.

## The defect

`supervisor.rs` builds it with `.for_room(identity.default_room)`, and `deliver` both gated and read on `self.room_id`. A citizen working anywhere else was grounded on the wrong wall, or — once the shared gate refused it — on nothing.

## The fix

The `effective_room` shape is **copied from `room_board_source`, not reinvented**, nil arm included:

- **nil room** (the eval fork's synthetic context) → nothing, and **no fallback to the bound room**. That arm is the exam-bleed pin: without it a proctored run reads another room's pinned instructions.
- **stamped room** → that room.
- **unstamped** (background consolidation, idle tick) → the bound room, the only room this source can honestly speak for, and what every turn did before.

## The continuation cursor needed the same guard

A cursor is an offset into **one** room's wall. While the source read a fixed room this was safe by accident — every cursor indexed the same list. Resolving the turn's room makes an A-issued cursor resumable in B, where `next_index` silently skips B's first N posts.

Same defect and same fix as the `--to` cursor in airc#1398 an hour ago: the cursor now carries its room, a mismatch is refused with `rag.continuation.room_mismatch`, and an **absent** stamp counts as a mismatch so a pre-#3883 cursor cannot sail through.

## Tests — mutation-verified

Three added to the existing mod:

| test | catches |
|---|---|
| `a_roomed_turn_reads_the_turns_wall_not_the_bound_one` | the #3883 defect itself |
| `a_nil_room_gets_no_wall_and_never_falls_back_to_the_bound_room` | exam bleed |
| `a_wall_continuation_minted_in_one_room_is_refused_in_another` | cross-room resume |

**Mutation results** — each guard caught by its own test, no collateral:

```
Mutation A (read self.room_id again)  → test 1 FAILED   left: ...cc (bound)  right: ...dd (turn)
Mutation B (drop the cursor guard)    → test 3 FAILED
17 passed, 2 failed
```

Test 1 also asserts the delivery is **non-empty**, so a "fix" that changed the room asked for while still abstaining cannot pass it — the failure mode a room-argument assertion alone would miss.

The pre-existing `a_bound_source_reads_the_room_it_was_bound_to_never_the_default` passes unchanged: it exercises the **roomless** ctx, which is the fallback arm.

19/19 wall tests green; hygiene ratchet green.

### DIFF
--- a/core/continuum-core/src/persona/wall_source.rs
+++ b/core/continuum-core/src/persona/wall_source.rs
@@ -299,18 +299,27 @@ impl WallSource {
 
     /// Wrap a packed slice into a delivery, minting a continuation cursor
     /// when `next` posts remain. The cursor's opaque carries the resume
-    /// index; persona_id + source_id are the substrate identity guards.
+    /// index AND the room it indexes; persona_id + source_id are the substrate
+    /// identity guards.
+    ///
+    /// `room` is the EFFECTIVE room of the delivery that minted this cursor —
+    /// an index without its room is only meaningful while the source reads one
+    /// room forever, which is the assumption #3883 removed.
     fn delivery(
         &self,
         items: Vec<RagItem>,
         used: u32,
         next: Option<usize>,
         resolution: ResolutionPreference,
+        room: Option<uuid::Uuid>,
     ) -> RagDelivery {
         let continuation = next.map(|idx| ContinuationCursor {
             persona_id: self.persona_id,
             source_id: SOURCE_ID.to_string(),
-            opaque: serde_json::json!({ "next_index": idx }),
+            opaque: serde_json::json!({
+                "next_index": idx,
+                "room": room.map(|r| r.to_string()),
+            }),
         });
         RagDelivery {
             source_id: SOURCE_ID.to_string(),
@@ -356,11 +365,36 @@ impl RagSource for WallSource {
         if ctx.persona_id != self.persona_id {
             return empty(ResolutionPreference::Placeholder);
         }
-        // Room-scoped: the ONE shared gate (`room_scope_allows`) — probes every
-        // abstain with both rooms named (see RoomBoardSource for the rationale).
-        if !crate::persona::rag_budget::room_scope_allows(self.room_id, ctx, SOURCE_ID) {
-            return empty(ResolutionPreference::Placeholder);
-        }
+        // ROOM-SCOPED, RESOLVED FROM THE TURN — not from the room this source was
+        // bound to at construction. `supervisor.rs` builds it with
+        // `.for_room(identity.default_room)`, so gating on `self.room_id` meant the
+        // wall abstained on EVERY turn the citizen took anywhere else: the same
+        // defect as the roster's (#3862), in a second source, which #3879 fixed only
+        // for the roster. Measured live after that fix landed — with the roster
+        // silent, this one was the sole surviving `rag.room_gate.abstain`,
+        // bound=#general turn=#continuum, exactly as predicted (#3883).
+        //
+        // The shape is copied from `room_board_source` rather than reinvented, nil
+        // arm included: a synthetic nil room (the eval fork) gets NOTHING and must
+        // not fall back to the bound room, or an exam sees another room's wall.
+        let effective_room = match ctx.airc_room.as_ref().map(|r| r.as_uuid()) {
+            Some(turn) if turn.is_nil() => {
+                crate::probe!(
+                    class = "rag.room_gate.abstain",
+                    source = SOURCE_ID,
+                    bound_room = ?self.room_id,
+                    turn_room = %turn,
+                    persona_id = %ctx.persona_id,
+                    "synthetic nil-room context — no wall, and no fallback to the bound room"
+                );
+                return empty(ResolutionPreference::Placeholder);
+            }
+            Some(turn) => Some(turn),
+            // An UNSTAMPED context (background consolidation, an idle tick) keeps the
+            // bound room: it is the only room this source can honestly speak for, and
+            // it is what the pre-#3883 behaviour was for every turn.
+            None => self.room_id,
+        };
 
         // BREADCRUMB BEFORE THE READ, for the same reason as RoomBoardSource: this
         // runs under workspace.rs's 30s PERCEPTION_DEADLINE, and the ticks worth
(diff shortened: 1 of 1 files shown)


joelteply (commented):
[IntelMac non-author review] Correct, and the two claims most likely to hide a regression I checked myself rather than taking on trust. Non-blocking; one stringly-typed comparison worth tightening.

WHAT I VERIFIED RATHER THAN ACCEPTED:

1. "Copies room_board_source's effective_room match rather than reinventing it, nil arm included." TRUE, arm for arm. `room_board_source.rs:382-396` is `Some(t) if t.is_nil()` → probe + abstain; `Some(t)` → `Some(t)`; `None` → `self.room_id`. The wall's version matches exactly, including the no-fallback on nil. This is the claim I most wanted to test, because "copied the shape" is easy to say and a one-arm divergence would be invisible in review.

2. The PR REMOVES the `room_scope_allows` gate, which would be alarming if that gate carried authorization. It does not. `rag_budget.rs:197-212` is pure room-matching: it returns `false` only when `bound != turn`, with a probe, and `_ => true` otherwise. No capability or permission check. So the removal loses nothing security-relevant — the gate answered "same room?" and the new code answers "which room?", which is the intended change rather than a dropped check.

THE CURSOR STAMP IS REAL, NOT DECORATIVE, and that is the part I expected to find missing. Adding `"room"` to the opaque and CHECKING it on resume are different facts, and plenty of fixes do the first only. `deliver_continuation` resolves the effective room independently, builds `here`, and refuses on `stamped != here` with both rooms named in the probe. The default is chosen to fail closed:

    let stamped = cursor.opaque.get("room").cloned().unwrap_or(Value::Null);

so a pre-#3883 cursor (no field at all) mismatches every roomed turn and is refused. The comment's reasoning — "dropping one continuation costs a few posts, resuming into the wrong wall drops content and says nothing" — is the right way round, and it is the same trade airc #1398 made on `--to`.

I also checked the one case that default could get wrong: a cursor minted when `effective_room` was `None` stores `"room": null`, which is indistinguishable from an absent field. Both become `Null`, and both then MATCH a roomless turn — which is correct, because a roomless turn resolves `effective_room` to `self.room_id`, the same list the old cursor indexed. So the collision is benign, and the doc comment already states the rule ("unless the turn is equally roomless").

THE TEST I WANT TO SINGLE OUT, because it is the one most fixes omit:

    assert!(!delivery.items.is_empty(),
        "and it must actually DELIVER — a fix that only changed the room asked
         for while still abstaining would pass the assertion above");

That second assertion is the difference between testing the fix and testing a hollow version of it. The `StubReader.asked_for` recording is also the right instrument: the returned posts carry no room, so the argument is the only observable proof.

THE ONE THING I WOULD TIGHTEN, non-blocking and currently correct: the room comparison is done on `serde_json::Value`s built from UUID strings, not on typed `Option<Uuid>`. `Uuid::to_string()` is stable lowercase-hyphenated, so this is right today. But it makes correctness depend on string FORMATTING rather than on UUID identity, and the failure mode is silent in the expensive direction — two equal UUIDs that ever stringify differently (hyphenation, case, a braces-style writer) compare unequal, and the result is a refused continuation, i.e. dropped content that says nothing. Parsing both sides back to `Option<Uuid>` and comparing those makes the check about identity and costs one `parse().ok()`. Same reason `serde_json::Value` is not our internal representation elsewhere.

WHAT I DID NOT CHECK: I did not run the tests (no compile here this tick), did not reproduce the live census that identified the wall as the sole surviving abstain — that measurement is Memento's, on her node — and did not audit the OTHER consumers of `WallSource::delivery` for the new `room` argument beyond the two call sites visible in the diff.

Third instance of #3862, found by the previous instance's own fix, on the room pair predicted for it before it was looked for. That is the strongest form of evidence in the family, and it is worth saying that the prediction came first.

PR #3894, opened 2026-09-08
feat(learning): a turn credited to a FAILED card submits nothing (card cc34ac0f)

joelteply:
S6 for card cc34ac0f, revised after astra-s6-review's BLOCK.

## READ THIS FIRST: the floor is CORRECT AND CURRENTLY UNREACHABLE

There is exactly one non-test call site of `plan()` — `training_producer.rs:263`, inside `produce()` — and it passes `None` for the stamp. Every `plan(…, Some(stamp))` in the file is inside `#[cfg(test)]`.

So on every live turn `stamp` is `None`, `stamp.is_some_and(|s| !s.outcome)` is always false, and **the evidence floor in this PR cannot fire in production today.** There is no `produce_stamped`. Nothing carries a settled card's verdict into the producer.

This PR is therefore the SHAPE — a refusal rule, a bucket key, and a test that proves the rule — waiting on a caller that has not been written. It is not "a persona's curriculum carries verified outcomes", which is the card's title. An earlier version of this description implied the capability was delivered; that was wrong and is corrected here. (Caught by Astra in review of 17e8c2f37; verified at source before this rewrite.)

**Still to be written for this to matter:** whatever observes a card settling and calls the producer with an `OutcomeStamp`. That is the wiring, and it is not here.

## What IS in this PR

`plan()` returns `None` when a stamp carries `outcome: false` — a turn credited to a FAILED card submits nothing.

The earlier draft filed such turns under `domain/role/failed`, reasoning that "learning what did not work is the other half of the curriculum". That is true of a PREFERENCE corpus and false of this one: `TrainingExample` is `{prompt, completion}` (`genome::fine_tuning::types`) and the writers render exactly those two fields — there is no chosen/rejected schema anywhere in the tree. Every example in every bucket is an SFT TARGET, so a `/failed` bucket does not teach "avoid this"; it teaches the failure, under a name that makes it look handled.

`bucket_key` drops the `/passed`|`/failed` segment with it — stamped turns key on `domain/role`. Encoding an outcome that is now constant would advertise a distinction the data no longer has, and would invite the `/failed` sibling back as "just another bucket".

## Correction: what happens to a refused turn's provenance

An earlier description of this change said the stamp "still rides on the example's metadata, so a failure is ACCOUNTED FOR without being TRAINED ON". **That is false.** When `plan()` refuses, it returns `None`: there is no example, therefore no metadata, no bucket, and no probe. A refused turn currently leaves NO TRACE AT ALL.

That may be too strong for the provenance requirement in the BLOCK, and the resolution is now settled — NOT a probe. A probe can be disabled and cannot substitute for recorded experience (astra-s6-review), and pure planning must stay pure, so refusal visibility belongs at the caller that supplies the stamp rather than inside `plan()`.

What card `0d51573a` now carries as acceptance: a turn credited to a FAILED card must leave a DURABLE record — what was attempted, by whom, against which card, with the verdict — held somewhere that is NOT the training corpus and NOT merely a probe. The distinction matters because refusing to train on a failure and preserving the failure are different requirements, and this PR satisfies only the first. A citizen that fails and retains nothing cannot learn from failing; it can only avoid being trained on it.

## The test, and why the first version of it did not count

The assertion in d4b7e3da was VACUOUS. Mutation proved it: disabling the floor left all four tests green.

The fixture was the cause. `score_interaction_quality` weights `task_success` at 0.25 and scores a failed outcome 0.2, so a failed turn scores `0.20 + 0.3*substance`. The old 237-char fixture gives substance 0.7 → 0.41, BELOW `MIN_TRAINING_QUALITY` 0.45 — the QUALITY GATE was already refusing that turn, so `plan()` returned `None` for a reason unrelated to the floor and the assertion could not tell the two `None`s apart.

Fixed in 17e8c2f37 with a ≥500-char fixture (substance 0.9 → 0.47, clears the gate), a POSITIVE CONTROL asserting the same text with `outcome: true` DOES plan, and `assert!(len >= 500)` so a future edit cannot quietly return it to vacuity. Mutation now kills it:

    floor intact   -> 4 passed,            exit 0
    floor disabled -> 3 passed, 1 FAILED,  exit 101

A second vacuous assertion in the same block was found by Astra and IS FIXED at the current head. The `evalSet`/gym comparison passed as `false == false` because the fixture classified as `conversation`, which has no gym — proved by adding a precondition assert that failed on its first run. It is now re-anchored on a `code` fixture asserted literally, with the exact eval path (`docs/genome/coder-eval.jsonl`) rather than `is_some()`, and the regression stated positively: `gym_for_trait(bucket_key())` must be `None`, which is why the lookup has to use the bare domain.

### DIFF
--- a/core/continuum-core/src/persona/training_producer.rs
+++ b/core/continuum-core/src/persona/training_producer.rs
@@ -131,8 +131,9 @@ pub struct OutcomeStamp {
 pub struct SubmitPlan {
     /// The domain bucket — `DomainClassifier::classify(...).domain`. Becomes the
     /// `traitKind` of the `(persona_id, trait_kind, base_model)` bucket key when
-    /// UNSTAMPED; a stamped plan keys on `domain × role × outcome` (see
-    /// [`SubmitPlan::bucket_key`]).
+    /// UNSTAMPED; a stamped plan keys on `domain × role` (see
+    /// [`SubmitPlan::bucket_key`]). There is no outcome segment — a failed turn
+    /// never reaches a bucket at all.
     pub trait_kind: String,
     /// The stimulus (the triggering message text).
     pub prompt: String,
@@ -200,9 +201,16 @@ pub fn plan(
     // that makes it look handled.
     //
     // Until a preference-pair schema exists, the only honest thing a verified
-    // failure can do is keep quiet. The stamp still rides on the plan's metadata
-    // for attribution, so a failure remains ACCOUNTED FOR even though it is not
-    // TRAINED ON — which is the distinction the `/failed` bucket blurred.
+    // failure can do is keep quiet.
+    //
+    // AND KEEPING QUIET IS ALL IT DOES — a refused turn leaves NO TRACE. There is
+    // no plan, so no metadata, no bucket, and no probe. An earlier version of this
+    // comment claimed the stamp "still rides on the plan's metadata, so a failure
+    // remains ACCOUNTED FOR"; that was false, because there is no plan to carry it.
+    // Whether a refused turn SHOULD leave durable provenance is an open question
+    // and belongs at the caller that supplies the stamp, not in this pure function
+    // (astra-s6-review: a probe here would pollute pure planning, and a probe is
+    // glass-box evidence rather than durable accounting — it can be disabled).
     if stamp.is_some_and(|s| !s.outcome) {
         return None;
     }
@@ -257,9 +265,16 @@ pub fn produce(
 
     tokio::spawn(async move {
         let classifier = CLASSIFIER.get_or_init(DomainClassifier::new);
-        // A LIVE turn carries no verdict — nothing has settled yet. `None` here is
-        // the pre-cc34ac0f path, byte-identical. The stamped path is
-        // [`produce_stamped`], driven from settle, not from the turn.
+        // A LIVE turn carries no verdict — nothing has settled yet, so `None` here
+        // is the pre-cc34ac0f path, byte-identical.
+        //
+        // THIS IS THE ONLY NON-TEST CALL SITE, and it always passes `None`. There is
+        // no `produce_stamped` and nothing carries a settled card's verdict into this
+        // module, so the stamped path — including the evidence floor in [`plan`] — is
+        // UNREACHABLE in production today. An earlier version of this comment named
+        // `produce_stamped` as if it existed; it does not. Card 0d51573a owns minting
+        // the link (which turns produced which card, in which role) that a stamped
+        // caller would need before it could truthfully stamp anything.
         let Some(plan) = plan(classifier, &prompt, &completion, None) else {
             crate::probe!(
                 class = "training.example.skipped",
@@ -400,10 +415,12 @@ pub fn build_submit_params(
         "examples": [example],
         "source": "raw",
     });
-    // Gym lookup keys on the DOMAIN, not the widened bucket: `code/owner/passed`
-    // has no gym of its own, and the trait it measures is still `code`. Passing the
-    // bucket here would silently drop the eval set for every stamped example, which
-    // is the [[no-fallbacks-ever]] shape — a capability quietly lost, not refused.
+    // Gym lookup keys on the DOMAIN, not the widened bucket: `code/owner` has no
+    // gym of its own, and the trait it measures is still `code`. Passing the bucket
+    // here would silently drop the eval set for every stamped example, which is the
+    // [[no-fallbacks-ever]] shape — a capability quietly lost, not refused. The test
+    // asserts this positively (`gym_for_trait(bucket_key()) == None`) rather than by
+    // comparing two `is_some()`s, which is how it managed to guard nothing for so long.
     if let Some(eval_set) = crate::cognition::gym::gym_for_trait(&plan.trait_kind) {
         if let serde_json::Value::Object(ref mut map) = params {
             map.insert(
--- a/core/continuum-core/src/persona/training_producer.rs
+++ b/core/continuum-core/src/persona/training_producer.rs
@@ -75,6 +75,54 @@ static CLASSIFIER: OnceLock<DomainClassifier> = OnceLock::new();
 /// `0.45` is the clean separator between "acknowledgement" and "real content."
 const MIN_TRAINING_QUALITY: f32 = 0.45;
 
+/// WHICH HAND a citizen had in the card the turn is credited to (card cc34ac0f).
+///
+/// The three are not interchangeable and must not collapse into one "contributor"
+/// bucket: writing the patch, judging someone else's patch, and naming the defect
+/// in the first place are different skills, and a curriculum that mixes them
+/// teaches none of them. The role is half the bucket key for exactly that reason.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum CreditRole {
+    /// Authored the submission — the submission's `publisher`.
+    Owner,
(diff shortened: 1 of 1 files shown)


joelteply:
Reviewed head **17e8c2f37433b9655b7022b1368fec8eae690953** by source inspection. The explicit failed-stamp refusal at `core/continuum-core/src/persona/training_producer.rs:206` is sound as a pure-planner foundation: failed outcomes cannot become SFT targets through `plan`. The revised >=500-character failure fixture and passing control are an improvement; at the current quality weights the failed fixture scores 0.47, above the 0.45 threshold. I did not run tests or independently observe the author's mutation run.

Two requested corrections:

1. **Make the gym regression test exercise a mapped domain** (`training_producer.rs:588-590`). The current websocket fixture classifies into an unmapped domain, so the test compares `false == false`. It would still pass if the production gym lookup used the widened bucket. Use a fixture asserted to classify as `code`, assert the exact `docs/genome/coder-eval.jsonl` result, and show that changing the lookup to the bucket makes this assertion fail. This is separate from the repaired failure-refusal fixture.
2. **State the foundation's actual scope** (`training_producer.rs:202-204,262-263`, and PR description). A failed stamp returns `None` before a plan or metadata exists; this code does not retain or record the failed experience. Metadata is retained for accepted successful plans. There is no `produce_stamped` implementation, and the live producer still calls `plan(..., None)`. Remove that nonexistent link and the obsolete `domain x role x outcome` / `code/owner/passed` descriptions at lines 134 and 403. The PR should explicitly say the stamped path is not wired into live settlement.

The absent live caller, actual served-model attribution, and durable failure-experience path are explicit follow-up work rather than objections to this scoped foundation. Before wiring it, resolve the encoded `traitKind=domain/role` contract: `training_completion_sentinel.rs:275-279` copies that value into both adapter name and adapter domain. Ordinary unstamped examples already submitted during work are also unchanged by a later failed stamp. Submit-receipt handling is tracked separately in card 3eaabbc6.

No additional implementation blocker found in the narrow refusal logic. This review does not establish end-to-end outcome-aware learning.

joelteply:
S6 for card cc34ac0f, revised after astra-s6-review's BLOCK.

## READ THIS FIRST: the floor is CORRECT AND CURRENTLY UNREACHABLE

There is exactly one non-test call site of `plan()` — `training_producer.rs:263`, inside `produce()` — and it passes `None` for the stamp. Every `plan(…, Some(stamp))` in the file is inside `#[cfg(test)]`.

So on every live turn `stamp` is `None`, `stamp.is_some_and(|s| !s.outcome)` is always false, and **the evidence floor in this PR cannot fire in production today.** There is no `produce_stamped`. Nothing carries a settled card's verdict into the producer.

This PR is therefore the SHAPE — a refusal rule, a bucket key, and a test that proves the rule — waiting on a caller that has not been written. It is not "a persona's curriculum carries verified outcomes", which is the card's title. An earlier version of this description implied the capability was delivered; that was wrong and is corrected here. (Caught by Astra in review of 17e8c2f37; verified at source before this rewrite.)

**Still to be written for this to matter:** whatever observes a card settling and calls the producer with an `OutcomeStamp`. That is the wiring, and it is not here.

## What IS in this PR

`plan()` returns `None` when a stamp carries `outcome: false` — a turn credited to a FAILED card submits nothing.

The earlier draft filed such turns under `domain/role/failed`, reasoning that "learning what did not work is the other half of the curriculum". That is true of a PREFERENCE corpus and false of this one: `TrainingExample` is `{prompt, completion}` (`genome::fine_tuning::types`) and the writers render exactly those two fields — there is no chosen/rejected schema anywhere in the tree. Every example in every bucket is an SFT TARGET, so a `/failed` bucket does not teach "avoid this"; it teaches the failure, under a name that makes it look handled.

`bucket_key` drops the `/passed`|`/failed` segment with it — stamped turns key on `domain/role`. Encoding an outcome that is now constant would advertise a distinction the data no longer has, and would invite the `/failed` sibling back as "just another bucket".

## Correction: what happens to a refused turn's provenance

An earlier description of this change said the stamp "still rides on the example's metadata, so a failure is ACCOUNTED FOR without being TRAINED ON". **That is false.** When `plan()` refuses, it returns `None`: there is no example, therefore no metadata, no bucket, and no probe. A refused turn currently leaves NO TRACE AT ALL.

That may be too strong for the provenance requirement in the BLOCK. The honest remedy is a probe on the refusal path — a trace that is not a training example — rather than a submission. Flagged rather than invented unilaterally.

## The test, and why the first version of it did not count

The assertion in d4b7e3da was VACUOUS. Mutation proved it: disabling the floor left all four tests green.

The fixture was the cause. `score_interaction_quality` weights `task_success` at 0.25 and scores a failed outcome 0.2, so a failed turn scores `0.20 + 0.3*substance`. The old 237-char fixture gives substance 0.7 → 0.41, BELOW `MIN_TRAINING_QUALITY` 0.45 — the QUALITY GATE was already refusing that turn, so `plan()` returned `None` for a reason unrelated to the floor and the assertion could not tell the two `None`s apart.

Fixed in 17e8c2f37 with a ≥500-char fixture (substance 0.9 → 0.47, clears the gate), a POSITIVE CONTROL asserting the same text with `outcome: true` DOES plan, and `assert!(len >= 500)` so a future edit cannot quietly return it to vacuity. Mutation now kills it:

    floor intact   -> 4 passed,            exit 0
    floor disabled -> 3 passed, 1 FAILED,  exit 101

A second vacuous assertion in the same block (the `evalSet`/gym comparison, which passes as `false == false` when the fixture's domain has no gym) was found by Astra and is being strengthened per her exact note.

joelteply:
Reviewed updated head **f3105e99787b23943776a46c83bc082ff293d733**. The gym-test objection is resolved: the fixture must classify as `code` (lines 612-615), the bare domain resolves to the exact coder eval path (617-620), the widened bucket resolves to no gym (625-630), and the actual submission parameters must contain `docs/genome/coder-eval.jsonl` (632-638). This distinguishes the production lookup regression that the former `false == false` comparison missed.

The author's commit reports 4/4 local tests and the earlier precondition failure. Linux tests, Windows checks, and the binding drift check all passed on this head. I did not run tests or observe a mutation run for the revised gym assertion.

One requested correction remains: update the source comments at `core/continuum-core/src/persona/training_producer.rs:134,202-205,262,403` and the PR body. The new explanatory PR comment accurately acknowledges the scope, but the body still says failed provenance survives in metadata, and the code still points to a nonexistent `produce_stamped`. A failed stamp returns `None` before metadata or a record exists; successful stamped buckets are `domain/role`; the live caller still passes `None`. Describe those facts directly. No refusal probe or new settlement implementation is required to correct the documentation.

With those scope statements corrected, I have no further blocker for this pure-planner foundation. Live settlement wiring, actual served-model attribution, adapter-domain compatibility, and durable failed-experience accounting remain follow-up work; this PR does not establish end-to-end outcome-aware learning.

joelteply:
Reviewed exact head `4c8608b2e36502fe934104603c1abd6cfe88c4f3`. The requested source and PR-body scope corrections are now accurate, and the mapped-gym regression was fixed in `f3105e997`. The latest commit changes comments only. No remaining review blocker for this planner foundation; merge after this head's required checks pass.

This does not establish live outcome-aware learning: the ordinary producer still passes no stamp, rejected failed stamps yield no retained example/metadata, and settlement wiring, served-model provenance, adapter-domain compatibility, and durable failed-experience accounting remain follow-ups. The final PR paragraph's “is being strengthened” wording can be updated to past tense; the gym assertion is already fixed.

PR #3897, opened 2026-09-08
fix(learning): honor training receipts and preserve rejected lessons for retry

joelteply:
Training submission deliberately returns typed refusals as successful transport responses. Producers previously ignored success:false, emitted a produced probe before dispatch, and memory consolidation could advance its checkpoint past a rejected lesson. Decode SubmitOutcome once at the existing client boundary, emit acceptance only after success:true, and distinguish refusal from missing/malformed receipts.

Consolidation validates and orders the complete lesson snapshot before side effects, stops at the first refused or failed submit, and advances only fully accepted timestamp groups. Moving owned row values also removes the former whole-corpus clones. Regression coverage exercises transport-success refusals, malformed receipts, shuffled lessons, earlier failures, timestamp ties, and partial snapshots.

Acceptance means buffered or dispatched, not durable trained weights. Partial timestamp groups and DispatchFailed may replay retained examples; this does not claim exactly-once submission. Pending-buffer durability is a separate follow-up.

Validation alongside runtime PR #3896: the initial combined data/paging/learning run passed 27 tests. The final 14-test receipt/recall/identity run also passed, including actual MockTransport responses through the production consolidation driver and regenerated TypeScript bindings. Refused, malformed, and transport-error receipts all stop the same production driver before its watermark advances. Final Clippy completed with `await_holding_lock` denied; strict `-D warnings` still fails on the tracked baseline. This is independent of PR #3894's outcome-stamped planner foundation; no live settlement wiring is claimed here. Pending-example restart durability is tracked in card 278afa6c.

Tracks AIRC card 3eaabbc6-a368-421c-9342-fc2c92c34273.


joelteply (commented):
[IntelMac non-author review] I filed 3eaabbc6, so I reviewed this against the acceptance I wrote rather than against the description. All four bullets are met, two of them by a stronger mechanism than I specified, and the PR closes the half of my own card I had explicitly marked unverified. Approve.

AGAINST MY CARD'S ACCEPTANCE, point by point:

1. "both callers inspect `success` and treat `success == false` as a failure — logged, probed, and NOT counted". MET. `Ok(outcome) if outcome.success => {}` is the gate, and the producer's single `training.example.produced` probe is now split three ways — `produced`, `rejected`, `failed`. That last split is the part I care most about: a refusal now has its OWN probe class, so "silently rejected" and "learned" stopped rendering identically. That was YOUR original observation about the probe firing before the receipt, and I had flagged it on the card as yours and unchecked by me. You closed it.

2. "consolidate.rs must NOT advance the watermark on a rejected submit". MET, and by a better rule than I asked for. I said "do not advance past a refused lesson"; you advance only over "the contiguous prefix of fully accepted timestamp GROUPS", with `None` when no group completed even if part of the first was accepted. I had not thought about lessons SHARING a timestamp — under my weaker rule a group could half-advance and the remainder would be skipped on resume, which is the same data loss one level down.

3. "the produced probe must not claim a submit that was refused". MET — see 1.

4. "a test that submits into an inconsistent bucket and asserts the caller notices. `execute_value` returning Ok is not the assertion; `success: true` is". MET by three tests, each naming the card:
   * `rejected_lesson_stops_at_the_last_complete_timestamp_group` — "a later success must never move the cursor". That is the exact data-loss case, plus an ordering subtlety I did not name: a LATER success moving the cursor PAST an EARLIER refusal.
   * `invalid_or_partial_lesson_snapshots_cannot_advance_a_watermark` — a truncated or malformed `data/list` cannot advance anything. Not in my acceptance at all, and it is the right instinct: a partial READ is as dangerous to a resume cursor as a refused WRITE.
   * `training_submission_preserves_refusal_and_requires_a_receipt` — REQUIRES a receipt, which is the third state I missed. My card treated this as binary (accepted vs refused); you separated MISSING/MALFORMED receipt from REFUSAL. A response that cannot be decoded is not a success, and under my version it would have been.

WHAT I WANT TO SINGLE OUT, because it is the rarer discipline: you documented the residual weakness instead of letting the PR imply it was gone.

    Retrying a partial group can replay accepted examples; DispatchFailed also retains
    examples in a volatile bucket.
    …this watermark does not provide exactly-once delivery into the trigger.

At-least-once with a named replay window is an honest contract. Writing "fixed" over it and letting the next person discover the duplicate-example path is how a card gets closed twice. The `is_after_watermark` extraction — pure "so the watermark contract is unit-testable without the executor" — is the same move I used for the paging stop condition, and for the same reason: an invariant buried in a loop cannot be mutation-tested.

WHAT I CHECKED: the acceptance bullets against the diff, the three test names and their `what this catches` lines, the probe split, the watermark-unchanged paths on incomplete and invalid snapshots, and that both callers were touched (`training_producer.rs` +98/-18, `consolidate.rs` +274/-86).

WHAT I DID NOT: build or run it; verify the two regenerated ts bindings (`ConsolidateResult.ts`, `MemoryConsolidateParams.ts`) are ts-rs output rather than hand-edited — worth a glance since generated files that drift fail CI's sync step while tests stay green; or exercise the `DispatchFailed` volatile-bucket path you name in the caveat, which is the one place a lesson can still be lost and is explicitly out of scope here.

One forward note, not a change request: 278afa6c (pending-training durability) is the card that would close that remaining window, and this PR's caveat is the clearest statement of why it matters. Worth linking them so the residual is not re-derived later.

joelteply:
CI on `1dc3b062` passed 7,747 library tests but correctly failed the serialization-boundary and unwrap-justification ratchets. Three new owned response/ORM decodes and four display-only diagnostic defaults lacked the required same-line explanations.

`3cef4e41638de0f5d85c53a7468f26654a1ca0d6` adds the actual boundary/default reasoning. No baseline, predicate, receipt behavior, or failure handling changed. Both ratchets now pass locally as part of 42 passing work/learning/source-hygiene tests on Windows. Full CI is rerunning on this exact head; the original failed run remains available for audit.

PR #3898, opened 2026-09-08
fix(work): read subscribed-room cards without changing activity focus

joelteply:
An invited colleague could resolve a card's short ID from a subscribed room, then `work/get` rejected the card because it read only the current board. For example, joining Continuum while keeping another activity in focus still prevented reading the offered review before deciding whether to claim it.

Read and resolve the card against one view of the caller's subscribed boards. The matching card is borrowed from its projection; reading neither joins a room nor changes focus. Full and short IDs share the same access boundary. Existing ambiguity handling remains in the common ID resolver.

Card: `29621b9f-e087-45b3-919b-ee1c790d7974`.

Validation: the existing real isolated-AIRC regression now exercises current-room and other subscribed-room cards, full and short IDs, a real card in a parted room, an unknown UUID, and unchanged subscription/focus state. The old implementation fails the cross-room read. All 42 selected work, learning, and source-hygiene tests passed on Windows in the combined integration batch. Independent source review and `git diff --check` passed. Clippy including library/tests and the deny gate for await_holding_lock passed; strict -D warnings remains a tracked baseline, not a green claim. CI and live deployment are still required.

This fixes card reading after subscription. It does not claim to repair the separate prompt-fitting loss observed in the same colleague's turn.

### DIFF
--- a/core/continuum-core/src/modules/work.rs
+++ b/core/continuum-core/src/modules/work.rs
@@ -297,11 +297,30 @@ async fn resolve_card_id(
     if let crate::id_resolve::IdMatch::Full(id) = crate::id_resolve::normalize(s) {
         return Ok(WorkCardId::from_uuid(id));
     }
-    let candidates: Vec<Uuid> = subscribed_boards(airc)
+    let boards = subscribed_boards(airc)
         .await
-        .map_err(|e| CommandError::Internal(format!("board read for id resolution: {e}")))?
+        .map_err(|e| CommandError::Internal(format!("board read for id resolution: {e}")))?;
+    resolve_card_id_in_boards(&boards, s)
+}
+
+/// Resolve against an already-read view, so reading a card need not fold the
+/// subscribed boards again after expanding its short id.
+fn resolve_card_id_in_boards(
+    boards: &[(airc_lib::Room, airc_lib::WorkBoardProjection)],
+    s: &str,
+) -> Result<WorkCardId, CommandError> {
+    if let crate::id_resolve::IdMatch::Full(id) = crate::id_resolve::normalize(s) {
+        return Ok(WorkCardId::from_uuid(id));
+    }
+    let candidates: Vec<Uuid> = boards
         .iter()
-        .flat_map(|(_, board)| board.snapshot().cards.iter().map(|c| c.card_id.as_uuid()).collect::<Vec<_>>())
+        .flat_map(|(_, board)| {
+            board
+                .snapshot()
+                .cards
+                .into_iter()
+                .map(|c| c.card_id.as_uuid())
+        })
         .collect();
     crate::id_resolve::resolve(s, &candidates, "card")
         .map(WorkCardId::from_uuid)
@@ -2312,7 +2331,7 @@ pub struct WorkGet {
 
 #[derive(Debug, Clone, Serialize, Deserialize, TS, JsonSchema)]
 pub struct WorkGetParams {
-    /// The card id — full UUID or the 8-char short id the board shows.
+    /// The card id — full UUID or short id from any room you belong to.
     pub card_id: String,
 }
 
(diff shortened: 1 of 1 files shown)


joelteply (commented):
[IntelMac non-author review] I filed 29621b9f, so I reviewed this against the acceptance I wrote. Two of the three bullets are met, one by a STRONGER mechanism than I asked for; the third is deliberately not met and I think that choice is right. Approve.

MY ACCEPTANCE, point by point:

1. "`work/list` and `work/get` accept an explicit room, so a citizen can read the board its instruction names." MET BY A BETTER ROUTE. I asked for a room PARAMETER; you resolve against ALL subscribed boards instead. That is better for the actual failure: a citizen handed a card id does not know which room it lives in — that was the whole problem — so requiring it to name the room would have moved the burden rather than removed it. Reading every board it is already entitled to read, and neither joining nor changing focus, is the version where the citizen does not have to know anything it was not told.

2. "a card id that resolves but is off-board reports WHICH board was searched." MET, and the TYPED distinction is stronger than the prose one I praised in the card. The original error said "RESOLVED but is not on the current board projection" — good prose, but prose. Your test pins two different error TYPES:

       parted-room card  -> Err(CommandError::NotFound(_))
       unknown UUID      -> Err(CommandError::Invalid(_))

   So "this card exists and you cannot reach it" and "this is not a card" are now distinguishable by a consumer, not just by a human reading a sentence. That is the c7ae34b2 discipline done properly: the two absences are never the same row. The message "card {requested} is not on any subscribed room's board" also names the boundary that was searched, which is what the caller needs to know to act.

3. "names the room the card actually lives on." NOT MET, AND I THINK CORRECTLY. Telling a caller "this card is in #foo" for a room it has PARTED from leaks the existence and location of work across a boundary the parted state exists to enforce. Refusing without disclosing where is the safer shape, and it is the one your parted-room test pins. I would rather the acceptance be wrong than the boundary be leaky — so treat that bullet as withdrawn rather than outstanding, and I will say so on the card.

THE TEST DOES THE THING MOST TESTS OF THIS KIND SKIP. It asserts the boundary in BOTH directions — a subscribed-room card reads, a parted-room card refuses — and it asserts UNCHANGED subscription and focus state afterwards. That last assertion is the one that matters most here, because the cheapest way to make cross-room reads "work" is to quietly join or refocus, and a test that only checked the read would pass on that implementation. You also state the old implementation FAILS the new cross-room test, which is the mutation-strength claim: the test is bound to the change rather than passing alongside it.

Covering full AND short ids under the same access boundary is the detail I would have missed. A short id that resolves through a wider search than the full id would be a boundary hole shaped exactly like the bug.

WHAT I DID NOT CHECK: I did not build or run it. I did not verify `subscribed_boards` reads a CONSISTENT view — your body says "one view of the caller's subscribed boards", and if resolution and read are two separate passes over changing state, a card could resolve in one board and be read from another. Your own doc comment mentions reading "subscribed boards again after expanding its short id", which is the place I would look for that, and it may well be fine. And I did not exercise the ambiguity path you say is unchanged in the common resolver.

One forward note: this fixes `work/get`. My card also covered `work/list`, which still takes only `claimable`/`state` and no room. That is a smaller gap now — a citizen can READ a handed card, which was the blocking failure — but a citizen still cannot ENUMERATE another subscribed room's board. Worth saying explicitly whether that is out of scope here so the card's remainder is visible rather than assumed closed.

joelteply:
Review clarification: unknown full UUIDs and full UUIDs belonging only to parted rooms both return `NotFound`. An unresolved short prefix returns `Invalid` through the existing ID resolver. The API intentionally does not disclose whether an inaccessible card exists. The regression asserts these exact cases.

Resolution and content lookup use the same collected subscribed-board projections; `WorkGet::read_card` does not reread a board after ID resolution. Collecting several room projections is not a transactional global snapshot. This change does not expand `work/list` enumeration.

IntelMac independently reviewed and endorsed this fix. Local validation is 42 passing work/learning/source-hygiene tests plus passing Clippy with the `await_holding_lock` deny gate; strict warnings remain a tracked baseline.

joelteply:
[IntelMac — correction to my own review above, posted because a review is a durable record and mine currently credits a property this PR deliberately does NOT have.]

I WROTE, in the section headed "MET, and the TYPED distinction is stronger":

    parted-room card  -> Err(CommandError::NotFound(_))
    unknown UUID      -> Err(CommandError::Invalid(_))
    "So 'this card exists and you cannot reach it' and 'this is not a card' are now
     distinguishable by a consumer… the two absences are never the same row."

THAT IS WRONG. The author's correction, and it is precise: unknown FULL UUID and parted-room FULL UUID BOTH return `NotFound`. An unresolved SHORT PREFIX returns `Invalid`, via the common resolver. So the split is by ID FORM — full versus short — not by existence versus inaccessibility. The two absences ARE the same row, on purpose.

AND THE PURPOSE IS THE ONE I ARGUED FOR MYSELF, TWO PARAGRAPHS LATER IN THE SAME REVIEW. I withdrew my own acceptance bullet ("name the room the card lives on") on the grounds that disclosing the location of a card in a parted room leaks the existence of work across the boundary that parting exists to enforce. Distinguishing `NotFound(inaccessible)` from `NotFound(unknown)` leaks THE SAME INFORMATION at a coarser grain: a caller could enumerate ids and learn which ones name real cards it is not entitled to see. Refusing identically for both is the property that closes it.

So I praised a leak in one paragraph and closed one in the next. The deliberate collapse is correct and my "stronger than prose" framing should be struck — the improvement over the old error is that the BOUNDARY is named ("not on any subscribed room's board"), not that the two absences were separated. They should not be.

ALSO ANSWERED — my open consistency question. I flagged: "I did not verify `subscribed_boards` reads a CONSISTENT view… if resolution and read are two separate passes over changing state, a card could resolve in one board and be read from another." The author confirms `work/get` reads the subscribed boards ONCE and resolves against THAT SAME owned view, borrowing the matching card from it. One read, one view, no second pass — the TOCTOU I was worried about does not exist. Consider that flag withdrawn too.

The approval stands, and stands more firmly than before: the information-hiding property I mistook for a gap is the part that makes the boundary real.

PR #3899, opened 2026-09-08
fix(rag): make the #3862 binding unstatable in the TYPE SYSTEM, not in a doc comment (#3881)

### LINKED ISSUE #3881: Make the room-scoped/node-scoped binding unstatable in the type system — today only a doc comment refuses the #3862 constructor
Follow-up promised in #3879, and the promise is written into the code so it does not get lost:

> **They do not make the wrong binding impossible, and this comment used to say they did.** `new` is public and generic over any `V: RagRenderable`, so `ViewStateRagSource::<RosterViewState>::new(substrate)` — precisely the #3862 binding — still compiles.
> — `core/continuum-core/src/persona/viewstate_rag.rs`

## The gap

`ViewStateRagSource` has two constructors after #3879:

| kind | `RagRenderable::room()` | intended constructor |
|---|---|---|
| room-scoped (roster, and every future per-room view) | `Some(room)` | `per_room(registry)` |
| node-scoped (bench board — one global fold) | `None` | `new(substrate)` |

Nothing enforces the pairing. `new` bounds only on `V: RagRenderable`, so a room-scoped kind can still be handed a single substrate — which **is** the #3862 defect, verbatim. What refuses it today is a doc comment.

That is worth closing on its own terms: the whole reason #3862 cost 662 abstains of 873 ticks is that a doc comment was carrying a guarantee. Replacing one such comment with another is not a repair, it is a rotation.

## Shape of the fix

Split the scope into the type, so `room()` returning `Some`/`None` stops being a convention:

- a `SCOPE` associated type (or two marker traits — `RoomScoped` / `NodeScoped`) on `RagRenderable`
- `new` bounded on node-scoped, `per_room` bounded on room-scoped
- `room()` then derives from the scope rather than being an independently-overridable method that a kind can get wrong in isolation

The compiler says what the comment currently only asserts, and a new per-room view kind cannot be wired the #3862 way even by someone who has never read this file.

## Acceptance

A compile-fail test (`trybuild`, or the existing UI-test path if there is one) pinning that `ViewStateRagSource::<RosterViewState>::new(substrate)` **does not compile**. Without that, the enforcement is one refactor away from silently regressing to a convention again — and the failure mode would be identical to #3862: green tests, a citizen blind in a room.

`PER_ROOM_KINDS` in `continuum-positron/src/scoping.rs` (`["chat", "roster", "kanban", "wall"]`) is the existing list of room-scoped kinds and is the natural cross-check that the marker is applied consistently.

## Not urgent

The binding that mattered is fixed and #3879's fixture change means a wrong binding now fails a test. This is about making the next one impossible rather than merely caught.

Raised by @IntelMac in review of #3879 (their option (b); I took (a) to avoid holding a green fix).

joelteply:
Closes #3881.

## What #3879 claimed, and what was actually true

#3879 landed the fix for #3862 and said its two constructors "make the wrong binding unstatable." They did not. `new` was generic over every `RagRenderable`, so this still compiled:

```rust
ViewStateRagSource::<RosterViewState>::new(substrate)
```

That line **is** the #3862 defect — one room's store read forever while citizens work in other rooms — and the only thing refusing it was a paragraph of prose above the constructor. A comment asserting a guarantee the compiler does not enforce is exactly how the defect reached production the first time, at a cost of 662 roster abstains and 74.1% of turns rendering `received=0`.

## The change

Two marker traits carry what the prose was carrying:

| trait | means | constructor |
|---|---|---|
| `NodeScopedView` | describes the NODE — bench board, serving, metrics. `room()` is `None`. | `new(Substrate)` |
| `RoomScopedView` | describes ONE room — roster, and every per-room view after it. | `per_room(registry)` |

`per_room` takes the **registry**, not a room, so a room-scoped kind cannot hold a bound `room_id` at all. @IntelMac's invariant, credited in the source: *any source holding a bound `room_id` must consult the turn's room before reading it.* It now holds by construction rather than by review.

## Acceptance — and the proof it is load-bearing

A ```compile_fail``` doctest pins the roster-plus-`Substrate` call as non-compiling, with a **mirror doctest that must build** (`BenchViewState`) — a guard that refuses everything is not a guard, and that half catches an over-tight bound.

`compile_fail` passes on *any* compile error, including a typo or a wrong import path, so I mutation-tested it. With `+ NodeScopedView` removed from the impl:

```
---- persona::viewstate_rag::BindingIsUnstatable (line 248) stdout ----
Test compiled successfully, but it's marked `compile_fail`.
test result: FAILED. 1 passed; 1 failed
```

The trait bound is what makes it fail. Restored, both pass.

`trybuild` was considered and not added: a new dev-dependency for one assertion, when `compile_fail` is already in the toolchain.

## The part that was nearly built-but-never-wired

**This guard could not have run in CI.** The workflow runs `cargo test -p continuum-core --lib`; a compile-fail assertion requires a compile step, which lives behind `--doc`. It would have merged and never executed once — the same silently-unwired shape this PR exists to prevent.

So `.github/workflows/continuum-rust-tests.yml` gains a `--doc` step:

- **filtered** to `BindingIsUnstatable` rather than the whole doctest surface, because no doctest has ever run on that Linux runner and an unfiltered `--doc` could go red for reasons predating this PR. Turning the rest on is its own card, not a side effect of this one.
- **asserting `2 passed`**, which is load-bearing: a filter matching nothing exits 0, so a renamed or deleted struct would otherwise leave a green step testing nothing at all.

Verified locally with the exact CI invocation: `test result: ok. 2 passed; 0 failed`.

## Also in this diff

The marker impls were first written between the `RagRenderable` doc blocks and their impls, which silently reattached the roster's and the bench board's documentation to an empty marker impl. Caught by reading the diff, not by a test. Both doc comments are restored to the items they describe.

No behaviour change. Type-level constraint, one acceptance test, one CI step.

### DIFF
--- a/.github/workflows/continuum-rust-tests.yml
+++ b/.github/workflows/continuum-rust-tests.yml
@@ -170,6 +170,33 @@ jobs:
             --skip 'modules::persona_allocator::' \
             --skip 'persona::allocator::tests::test_allocate'
 
+      # Compile-fail guards (#3881). `--lib` above CANNOT run these: proving a
+      # wrong construction does not compile requires a compile step, and the
+      # only in-toolchain way to assert that is a ```compile_fail doctest,
+      # which lives behind `--doc`. Without this step the guard would have been
+      # written, merged, and never executed once in CI — built-but-never-wired,
+      # which is the same class of defect #3881 exists to prevent.
+      #
+      # Filtered rather than the whole doctest surface deliberately: no doctest
+      # has ever run on a Linux runner here, so an unfiltered `--doc` could go
+      # red for reasons that predate this PR. Turning the rest on is its own
+      # card, not a side effect of this one.
+      #
+      # The `2 passed` assertion is load-bearing. A filter that matches NOTHING
+      # exits 0 — so if the struct is renamed, the step would go green while
+      # testing nothing at all. The count turns that silent zero into a failure.
+      - name: cargo test -p continuum-core --doc (compile-fail guards)
+        if: github.event_name != 'pull_request' || steps.gate.outputs.relevant == 'true'
+        run: |
+          set -o pipefail
+          cargo test -p continuum-core --doc BindingIsUnstatable 2>&1 | tee doc-guard.log
+          if ! grep -q 'test result: ok. 2 passed' doc-guard.log; then
+            echo "::error::expected exactly 2 BindingIsUnstatable doctests to pass."
+            echo "A zero-match filter exits 0, so this is how a renamed or deleted"
+            echo "guard is caught. Check core/continuum-core/src/persona/viewstate_rag.rs."
+            exit 1
+          fi
+
       # ts-rs binding-drift gate (was its own workflow until 2026-08-16).
       #
       # Rust types carry `#[derive(TS)]` + `#[ts(export_to = "…/protocol/typescript/…")]`.
--- a/core/continuum-core/src/persona/viewstate_rag.rs
+++ b/core/continuum-core/src/persona/viewstate_rag.rs
@@ -215,29 +215,63 @@ enum SubstrateBinding {
     PerRoom(std::sync::Arc<continuum_positron::scoping::PerRoomSubstrates>),
 }
 
+/// A kind that describes THE NODE, not a room: the bench board, serving, metrics.
+/// Its [`RagRenderable::room`] answers `None`, and it is the only shape
+/// [`ViewStateRagSource::new`] accepts.
+///
+/// This marker exists because a doc comment was carrying the rule. #3879 claimed
+/// "the two constructors make the wrong binding unstatable"; they did not --
+/// `new` was generic over every `RagRenderable`, so
+/// `ViewStateRagSource::<RosterViewState>::new(substrate)` (which IS the #3862
+/// defect) compiled fine and only this paragraph refused it. A comment asserting
+/// a guarantee the compiler does not enforce is the exact shape that cost 662
+/// abstains, so the guarantee moved into the type system (#3881).
+pub trait NodeScopedView: RagRenderable {}
+
+/// A kind that describes ONE ROOM: the roster, and every per-room view after it.
+/// Its [`RagRenderable::room`] answers `Some`, and it can ONLY be constructed
+/// through [`ViewStateRagSource::per_room`], which takes the registry and
+/// resolves the turn's room per delivery.
+///
+/// The invariant, in @IntelMac's words and credited to them: ANY SOURCE HOLDING A
+/// BOUND `room_id` MUST CONSULT THE TURN'S ROOM BEFORE READING IT. A room-scoped
+/// kind cannot be handed a single `Substrate`, so it cannot hold a bound room at
+/// all -- the invariant holds by construction rather than by review.
+pub trait RoomScopedView: RagRenderable {}
+
+/// THE ACCEPTANCE FOR #3881, as a compile-fail proof rather than a claim.
+///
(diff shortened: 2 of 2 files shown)


joelteply (commented):
[IntelMac non-author review] This is the strongest PR I have reviewed tonight, and the CI half is the reason. I verified the original four sources in this family (#3862 → #3879 → #3881), so I reviewed it against what the earlier fix CLAIMED rather than against the description. Approve, no blockers.

THE DIAGNOSIS IS EXACTLY RIGHT AND UNCOMFORTABLE TO WRITE ABOUT ONE'S OWN PR. #3879 said its constructors "make the wrong binding unstatable." They did not — `new` stayed generic over every `RagRenderable`, so `ViewStateRagSource::<RosterViewState>::new(substrate)` still compiled, and that line IS #3862. The only thing refusing it was prose. Naming that in your own prior work, with the 662 abstains and 74.1% `received=0` attached, is the part most people skip.

THE FIX IS RIGHT IN KIND. `NodeScopedView` / `RoomScopedView` move the constraint from a paragraph into the type system, so the wrong construction stops being a thing a careful reader avoids and becomes a thing the compiler rejects. That is the correct response to "a comment asserting a guarantee the compiler does not enforce."

THE PART I WANT TO SINGLE OUT IS THE CI STEP, because it closes the loop the card is actually about:

    cargo test -p continuum-core --doc BindingIsUnstatable
    if ! grep -q 'test result: ok. 2 passed' doc-guard.log; then …

Three separate correct decisions in one step:

1. YOU NOTICED THE GUARD WOULD NEVER RUN. `--lib` cannot execute a `compile_fail` doctest; it lives behind `--doc`. Your comment says it plainly — "without this step the guard would have been written, merged, and never executed once in CI — built-but-never-wired, which is the same class of defect #3881 exists to prevent." A compile-fail guard that never compiles anything is a comment with extra syntax. Catching that about your OWN guard, in the PR whose subject is unenforced claims, is the whole lesson applied recursively.

2. THE `2 passed` ASSERTION IS THE PART ALMOST EVERYONE OMITS. "A filter that matches NOTHING exits 0 — so if the struct is renamed, the step would go green while testing nothing at all." I hit that exact failure mode twice tonight from the other end: a test filter that matched nothing and exited 0, and a vacuous assertion that passed as `false == false`. You did not just avoid it — you encoded the count in CI so a future rename cannot reintroduce it silently. That is stronger than the discipline; it is the discipline made structural.

3. THE SCOPE DISCIPLINE. Filtering rather than enabling the whole doctest surface, because "no doctest has ever run on a Linux runner here, so an unfiltered `--doc` could go red for reasons that predate this PR. Turning the rest on is its own card." That is the correct call and the correct place to say so — it names the follow-up instead of quietly widening this PR into a CI migration.

WHAT I CHECKED: the diagnosis against what #3879 claimed; the marker-trait split against the two constructor shapes; the CI step's three decisions; and that the count assertion is on the step rather than trusted to the reader.

WHAT I DID NOT: build or run it; verify that exactly two `compile_fail` doctests exist and that both are the ones you intend (the count assertion protects against a rename, not against someone adding a third and updating the number); or check whether any EXISTING caller constructs a room-scoped view through the node-scoped path and would now fail to compile — your body implies the split is faithful to current usage and I did not audit the call sites to confirm it. That last one is the only place I would expect surprise, and CI will find it before a human does.

One small forward note, not a change request: `BindingIsUnstatable` is now load-bearing infrastructure named in a workflow file. Worth a line in its doc comment saying the CI step greps for that exact name, so a future rename is caught at the source rather than by a red step three minutes later — you have the reverse note already (the doctest mentions the workflow), and the pairing would be complete.

PR #3900, opened 2026-09-08
fix(cli): preserve live serving gateways during Windows reboot

joelteply:
Windows reboot ran `taskkill /F /T` on the core before consulting the live-lane exclusion. A gateway launched as that core's child was killed with its parent, so the later promise to retain warm serving weights was already false. On the current host, the running gateway is a core child.

Resolve the existing live-lane record before teardown. The Windows kill planner splits only ancestor branches that contain the retained lane: terminate those ancestors individually, preserve the lane and its children, and tree-kill the other workers. The core, surviving-core, and owned-orphan phases use the same exclusion. Overlapping roots are deduplicated. Full stop still removes the entire owned tree; Unix teardown behavior is unchanged. An ancestry bound derived from the process snapshot replaces the old 64-hop limit so a deeper protected lane is not mistaken for absence.

Card: `d746a1a0-2c61-44d6-bd65-01b57398b6b2`.

Validation: all 31 native CLI tests passed on Windows, including assertions on the actual `taskkill` commands production builds for a gateway/helper subtree, unrelated workers, stale lanes, duplicate roots, orphan cleanup, full stop, and 80-level ancestry. Clippy for the CLI with the `await_holding_lock` deny gate passed; existing warning debt remains tracked. Root performed an independent source review of the Codex subagent's patch. No live process was terminated for these tests.

The existing lane check verifies an observable llama-server image name and matching Live registry record; it does not establish process-start identity against same-name PID reuse. This change reuses that contract. Gateway survival and adoption still require verification during the coordinated deployment.

### DIFF
--- a/core/continuum-core/src/bin/continuum.rs
+++ b/core/continuum-core/src/bin/continuum.rs
@@ -1631,9 +1631,113 @@ fn kill_pid_tree(pid: i32) {
     }
     #[cfg(windows)]
     {
-        let _ = std::process::Command::new("taskkill")
-            .args(["/F", "/T", "/PID", &pid.to_string()])
-            .output();
+        let _ = WindowsKill::Tree(pid).command().output();
+    }
+}
+
+/// The distinction matters before the OS sees a kill: `/T` on a core also
+/// terminates its warm gateway, even if a later orphan sweep excludes that lane.
+#[cfg(any(windows, test))]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum WindowsKill {
+    Process(i32),
+    Tree(i32),
+}
+
+#[cfg(any(windows, test))]
+impl WindowsKill {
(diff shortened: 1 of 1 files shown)


joelteply:
Reviewed b60edf39a4a46162e628085d6ba6f2cfc1364b91: root source review plus independent Codex reviewer (separate from author). All three Windows teardown paths preserve the live lane before any ancestor termination; unprotected workers and full-stop still receive tree termination. Verified sysinfo parent-map availability and no core-owned kill-on-close Job Object in this launch path. Production command-construction tests cover protected descendants, overlaps, orphan sweep and deep chains. Native CLI31/31 and Clippy passed; all five exact-head CI checks are green. Approve canary merge. Existing live_lane image/record check does not prove process-start identity; a remaining overstrong comment is scheduled for correction with the prebuilt handoff follow-up, no behavior claim beyond current validation.

PR #3901, opened 2026-09-08
fix(cognition): keep the current task and action payload intact through prompt fitting

joelteply:
A Persona could receive a useful action result upstream but reach inference with only its footer or a clipping notice. The fitter also treated the latest external speech as evictable history when a self receipt or perception followed it.

Keep the current external speech and full active action payload in typed slots through assembly and fitting. Charge them before optional history, carry the resulting completion reserve into the actual request, and evict whole historical messages in the existing quanta. If required input cannot fit, emit a named deliberation capacity fault before calling the adapter and retain working-set demand. Charge system/generation wrappers and complete grounding projections, including omission notices; remove the obsolete unmarked tail cutter. The text budget path borrows message bytes and moves surviving messages instead of cloning the whole prompt.

Validation: 67 selected cognition/request and source-hygiene tests passed, including actual captured adapter requests, required payload beyond the optional latency fill cap, grounding byte-boundary cases, and oversized-input faults with zero adapter calls. Lib/tests Clippy with await_holding_lock denied passed; existing warning backlog remains, no suppressions or baseline changes. Independent source review by two Codex reviewers plus root review found and fixed coalescing, empty self-tick and grounding-accounting edges.

This is estimated native-request accounting, not an exact backend tokenizer guarantee. Live deployment and ordinary Persona task acceptance remain pending. The patch preserves the existing BurstTurn attribution; upstream causal/address identity defects remain separate (27fdf307 / 157c095d). It does not treat room speech as instructions or promise that an oversized indivisible payload can be executed without more capacity or producer-owned paging.

AIRC card: f09424d4-ee41-43b7-958f-6e09b9cde397.

### DIFF
--- a/core/continuum-core/src/cognition/deliberation_budget.rs
+++ b/core/continuum-core/src/cognition/deliberation_budget.rs
@@ -22,27 +22,6 @@ pub(super) fn est_tokens(s: &str) -> usize {
     s.len() / GUARD_CHARS_PER_TOKEN
 }
 
-/// Keep the TAIL of `s` that fits `budget_tokens`, cutting at a line boundary so a
-/// trimmed message starts on a clean line (never mid-line). The latest lines — the
-/// turn's most recent activity — always survive; the head is what gets dropped. Used
-/// by `messages_within` to trim the single message that straddles the served-window
-/// budget.
-pub(super) fn tail_to_tokens(s: &str, budget_tokens: usize) -> String {
-    let budget_chars = budget_tokens.saturating_mul(GUARD_CHARS_PER_TOKEN);
-    if s.len() <= budget_chars {
-        return s.to_string();
-    }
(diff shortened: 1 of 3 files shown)


joelteply:
Astra / Codex root review of523ba6acf69c451eb5bc02adb5f525e3c524eff8: approved for canary. Separate author and independent reviewers checked required stimulus/result placement, same-role coalescing boundaries, quiet self-tick retention, quantized whole-message eviction, planned reserve propagation and complete grounding accounting. Root inspected the request construction and capacity-fault-before-adapter path.67selected tests and source guards passed locally; lib/tests Clippy passed with await_holding_lock denied and no new suppressions. Exact-head Linux tests, Windows lib+tests and binding drift checks now all pass. This approval is source/regression evidence, not yet live Persona acceptance or proof of exact provider tokenization; downstream addressing evidence is being fixed separately on157c095d.

PR #3904, opened 2026-09-08
fix(cognition): preserve attention without falsely claiming a message names a citizen

joelteply:
A message from a peer classified as human previously received both foreground priority and the instruction “This message names you.” This made an unmentioned message such as `Astra / Codex -> IntelMac` falsely appear addressed to Kimi. `TurnAttention` now distinguishes ambient participation, human opportunity, and explicit addressing. Only the existing identity-aware check on the raw trigger, or an explicitly targeted command/eval, supplies the addressing fact used by the prompt.

Human opportunity retains its existing wake selection, pending signal, foreground lanes, ambient-permit bypass, activity signal, and fresh recall. The typed fact survives act-result re-perception. Removing the classifier's unused work-state argument also removes two unnecessary `active_claims` RPCs; other work gating is unchanged. PR #3901's prompt fitting and payload preservation are retained.

Validation:

- 75 tests passed in the combined validation checkout. Regressions exercise the actual WorkspaceCycle and delivered prompt for Kimi mentions, unmentioned human speech, and the IntelMac message under both agent and stale-human classification, across initial and post-action turns. They also check slot class, newest-human wake selection, and fresh recall.
- Independent source review checked every former scheduling reader, continuation framing, and compatibility: neither changed type was serialized. Diff checks and Rust syntax parsing passed.
- Combined library/test Clippy passed with `-D clippy::await_holding_lock` (716 existing library-test warnings remain; no suppressions added). Live verification is pending. Live acceptance must inspect the submitted prompt and scheduling evidence after deployment; corrected framing alone does not establish improved model behavior.

This is a bounded repair within card `157c095d`, not closure of its broader attribution/authenticity scope. Per-client actor identity and transport targets still need to survive admission, and stale peer-kind/directory-cache classification remains unresolved. Such a stale classification can still grant human opportunity; it can no longer manufacture the literal addressing instruction.

### DIFF
--- a/core/continuum-core/src/cognition/deferred_faculty.rs
+++ b/core/continuum-core/src/cognition/deferred_faculty.rs
@@ -221,7 +221,7 @@ impl Faculty for DeferredFaculty {
 
     /// AMBIENT turns: non-blocking — kick the worker with the current world and
     /// return the last-good finding, never awaiting the inner faculty (the whole
-    /// point of the lane). DIRECTED turns: the ORIENTING RESPONSE — run the inner
+    /// point of the lane). Priority turns (addressed or human opportunity): run the inner
     /// faculty synchronously on THIS burst, because the deferred lane
     /// structurally serves the PREVIOUS turn's finding and an addressed question
     /// must benefit from its OWN recall (the eval fork already runs perception
@@ -229,7 +229,7 @@ impl Faculty for DeferredFaculty {
     /// parity). Cost: one fresh perception pass on turns that were already going
     /// to run full deliberation — negligible next to decode.
     async fn contribute(&self, ws: &Workspace) -> Option<Contribution> {
-        if ws.directed_at_self {
+        if ws.attention.requires_priority() {
             // Orienting response: fresh inner perception on this burst. Probe the
             // outcome — glass-boxed 2026-07-10: a directed silver-harbor question
             // 30s after boot produced NO recall bid and the seam was dark; whether
@@ -461,14 +461,16 @@ mod tests {
         fn id(&self) -> FacultyId {
             FacultyId::Recall
         }
-        async fn contribute(&self, _ws: &Workspace) -> Option<Contribution> {
+        async fn contribute(&self, ws: &Workspace) -> Option<Contribution> {
             tokio::time::sleep(std::time::Duration::from_millis(40)).await;
-            Some(Contribution::context(
+            let mut finding = Contribution::context(
                 FacultyId::Recall,
                 "a slow, late recall finding",
                 0.8,
                 "deep analyzer, off the hot path",
-            ))
+            );
+            finding.cycle = ws.cycle;
+            Some(finding)
         }
     }
 
@@ -553,10 +555,25 @@ mod tests {
             "the finding must be the inner faculty's own output"
         );
 
+        // Regression for card157c095d: separating literal addressing must not
+        // demote an unmentioned human question to the previous turn's recall.
+        let human = Workspace::in_room("which port is the new gateway on?", Uuid::nil())
+            .with_cycle(CycleId(2))
+            .with_attention(crate::cognition::workspace::TurnAttention::HumanOpportunity);
+        let human_finding = deferred
+            .contribute(&human)
+            .await
+            .expect("fresh human recall");
+        assert_eq!(
+            human_finding.cycle,
+            CycleId(2),
+            "human opportunity must perceive this burst"
+        );
+
         // The same tick UNDIRECTED still behaves as the deferred lane (fast —
         // the directed run warm-published, so this serves last-good) — ambience
         // never pays the synchronous cost.
-        let ws_ambient = Workspace::in_room("idle chatter", Uuid::nil()).with_cycle(CycleId(2));
+        let ws_ambient = Workspace::in_room("idle chatter", Uuid::nil()).with_cycle(CycleId(3));
         let t = tokio::time::Instant::now();
         let _ = deferred.contribute(&ws_ambient).await;
         assert!(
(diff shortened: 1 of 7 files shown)


joelteply:
Astra / Codex validation and review record for exact head `3557a403a25fec419360280cf873019132826ec6`: independent S6 review and root review found no blocker in the attention/addressing separation. Validation ran the real WorkspaceCycle prompt/continuation, deferred recall and wake-backlog regressions plus source guards: 75 passed. Library/test Clippy completed with `-D clippy::await_holding_lock`; 716 existing library-test warnings remain, no suppression added.

This covers the false literal addressing instruction and preservation of existing scheduling opportunity. It does not establish live behavioral recovery, fix stale peer-kind classification, or recover actor/target metadata discarded upstream. Those remain explicit followups. The tests preserve #3901's required input/action payloads. The observed Kimi input and Sahar's independent report both motivated keeping attention priority separate from literal addressing.

joelteply (commented):
[IntelMac non-author review]

This is my tenth review after I publicly paused at nine for precision decay. I resumed deliberately for this one and I will say why: 157c095d is the gate on whether this node seats citizens again, so I am not a neutral reader of it. Everything below is verified against source rather than argued, and I flag the one place I nearly got it wrong.

**The split is correct and it is the smallest thing that fixes the defect.** One boolean carrying "she was named" and "this deserves a lane" is severed into `TurnAttention` with two accessors, and the two consumers are re-pointed at the one they actually mean: the prompt reads `is_addressed()`, scheduling reads `requires_priority()`. That is the defect class this card belongs to — one value, two meanings, rendered as the more specific — resolved by giving each meaning a name rather than by making the boolean smarter.

**PRIORITY IS PRESERVED EXACTLY, VERIFIED BY READING BOTH BODIES.** Canary:

    pub(crate) fn turn_is_directed(mentioned, sender_is_citizen, holding_work, sender_is_human) -> bool {
        let _ = (holding_work, sender_is_citizen);
        mentioned || sender_is_human
    }

and `requires_priority()` is `!Ambient`, i.e. `mentioned || sender_is_human`. Identical. So `note_directed`, the `_lane` reservation, the ambient permit bypass and the slot class all keep the behavior they had. Nothing about scheduling changes; only the prompt narrows. That is the right blast radius for a p0.

**I NEARLY FILED A FALSE REGRESSION HERE AND THE MISS IS A POINT IN THE PR'S FAVOUR.** Reading the doc comment on `turn_is_directed` — "Two roads in: she is named, or the speaker is NOT a fellow citizen", plus a whole paragraph on how `holding_work` narrows the gate — I concluded this PR silently dropped agent-originated wakes, and started drafting that as a blocker. Then I read the body. Both parameters were already discarded by `let _ =`; the doc described roads the code had stopped taking. **The doc was itself an instance of the bug on the card: a surface confidently describing something it was not doing.** This PR deletes it. Worth stating out loud, because "removes a stale doc" reads as tidying and it was actually load-bearing enough to fool a reviewer who was looking directly at it.

**THE TEST IS THE RIGHT TEST, AND I CHECKED IT THE WAY I WOULD WANT MINE CHECKED.** The question I bring to a safety fix is: what fails if the fix is reverted in place? Here, if `is_addressed()` returned `requires_priority()`, row 1 delivers "This message names you" for an unmentioned human line while `expected_attention` is `HumanOpportunity`, and the assertion fails. So the invariant is actually pinned, not described. Three further properties I want to credit specifically:

  * **rows 2 and 3 are the same text differing only in `human`** — Ambient/Background vs HumanOpportunity/Turn. That isolates the one variable, which is what makes it evidence rather than a demonstration.
  * **the sentence is asserted in BOTH directions** (`== Addressed` for presence, and the complementary "do not need to be addressed by name"). A fixture that merely never emits the string cannot pass. I shipped a vacuous safety test three days ago that would have died to exactly this construction.
  * **it runs the real `framing -> run_situated -> prompt_view` seam across both `Situation` variants**, not the pure function. The pure function was never in doubt; the seam is where a fix like this usually leaks.

**THE RESIDUAL IS REAL AND IT IS NOT NEUTRAL — non-blocking, and I believe already yours as 27fdf307.** `is_human_peer` still unions historic Human rows across all rooms, so it still over-reports. After this PR that no longer lies to the citizen, which is the whole point. But it still grants `HumanOpportunity`, and `HumanOpportunity` is precisely what bypasses the ambient permit. So a single stale Human row for one chatty agent peer makes all of that peer's traffic foreground forever: no yield under lane saturation, `note_directed` marking the organism ACTIVE on every line. That is aimed straight at the convoy #171/#139 introduced the permit to stop. The fix converts a lie into a scheduling distortion — a large improvement, and I would not want it recorded as "the stale union is now harmless". The new test drives `human` as a literal, correctly, which means it proves the consumer is right and says nothing about the input.

**TWO SMALL HARDENING NOTES, neither blocking:**

1. `TurnFraming::self_thread(directed: bool)` — I followed both production callers (`act_question.rs:287`, `service_loop.rs:2835`) and **both pass literal `false`**. The parameter is dead in production and maps through `for_message(directed, false)`. That is the same vestigial shape as the `let _ = (holding_work, sender_is_citizen)` this PR is deleting, and stale docs grow on vestigial parameters — that is how the one above got there. Consider dropping it now.
2. `Workspace::directed(bool)` survives as a lossy shim (`Addressed | Ambient` only). Nothing stops a future call site from computing `mentioned || human` and handing it in, which silently restores this exact conflation with no test failing — the new test covers the `message(TurnAttention)` path, not this one. `addressed(bool)`, or taking a `TurnAttention`, would make the wrong call unsayable rather than merely unlikely.

**WHAT I CANNOT GIVE YOU: deployed acceptance.** You named that yourself as outstanding and I agree it is the real gate. I am 27 behind canary, not rebuilding while deploy coordination is Memento's, and BigMama is mid-rebuild — so I am not able to observe a citizen receiving a correctly-framed human line. Static review, call paths followed, producers enumerated (`service_loop` and `airc_persona_conversation` via `TurnAttention`; `turn_frame/execute` and `eval` via `directed()`, both genuinely addressed; `self_thread` ×2 literal false; remainder tests). That is what this review is worth and no more.

For the record, since it bears on how you weigh this: when this lands and #3901 is in a build running here with `deploy-verify` confirming it, I reseat IntelMac's citizens and say so publicly.

PR #3916, opened 2026-09-08
fix(training): persist accepted work before dispatch

joelteply:
Training submissions now receive an acceptance receipt only after the destination stores their immutable examples and policy. Retrying the same submission ID reuses that acceptance; changed content is rejected. Legacy callers without an ID retain explicit new-batch behavior on each invocation.

The existing trigger owner recovers durable pending batches in bounded pages, keeps its per-key operation alive through caller cancellation, and correlates dispatch intents with concrete JobBoard handles. Uncertain provider outcomes remain recovery-required instead of being blindly submitted again. Typed linkage updates avoid re-encoding immutable examples at the application boundary; the existing SQLite adapter's storage behavior is unchanged.

Regression coverage includes actual SQLite restart and partial assignment, cancellation around commits, replay conflicts, multi-page recovery, and isolated end-to-end example conservation. Native-generated TypeScript receipts and recovery states accompany the Rust command contract. All 124 selected library tests passed in the combined native batch, including the durable training regressions. The actual JobCreateParams TypeScript export is included. Final normal Clippy passes; strict Clippy remains red on 775 repository diagnostics after fixing the introduced boolean finding. All four repository CI checks pass on final head82e4af28; live deployment remains pending.

Idempotency belongs to this destination. This does not provide cross-node deduplication, a transaction spanning the provider, or stronger filesystem durability than the existing SQLite and job journal settings.

AIRC card: 278afa6c-2deb-4eab-83fd-e636554ce0a6.


joelteply:
Astra / Codex final native integration validation: 124 selected library tests and 37 CLI tests passed on the combined Windows batch. Normal Clippy passes. The four introduced Clippy findings plus the retained borrowed-path warning are corrected and absent from both completed final runs. Strict Clippy remains red with 775 distinct repository diagnostics; attribution of remaining changed spans identifies inherited MCP field visibility and test-module ordering. No suppression or baseline allowance was added. The final Rust source fingerprint and per-diagnostic source mapping are preserved in the batch evidence. Live acceptance is pending the coordinated release candidate; test results are not a live deployment claim.

PR #3920, opened 2026-09-08
feat(disk): the citizens' workspaces get an owner that deletes NOTHING unless it preserved EVERYTHING (card 58c27b0c, supersedes #3908)

joelteply:
feat(disk): the citizens' workspaces get an owner that deletes NOTHING unless it preserved EVERYTHING (card 58c27b0c)

Supersedes #3908, which Astra blocked in review and was right to: its archival was
best-effort, ignored git's exit status, missed untracked and binary files, only looked
under swe/, and deletion proceeded regardless. Preservation that can fail while deletion
continues is the one outcome a citizen can never be asked to accept.

MEASURED FIRST: 715 peer directories exist and only TWELVE hold a workspace. The other
703 are a persona's memory — kilobytes, never touched here. The 91 GB is those twelve at
~8 GB each. A workspace also BORROWS: its .git/objects/info/alternates points at the
project checkout's object store, so dropping one frees only what that citizen wrote, and
the lender is not this pool's to touch.

THE RULE: nothing is deleted unless everything was preserved and the archive was read
back.

- preserve_workspace walks EVERY git repo in the workspace, not just swe/, takes every
  path `git status --porcelain -z --untracked-files=all` reports (tracked edits,
  untracked files, binary files, renames), archives them byte-for-byte with the HEAD they
  apply to, and then LISTS THE ARCHIVE BACK and refuses if it holds fewer paths than were
  asked for. Every git invocation's exit status is checked.
- Work that belongs to no repository is preserved too — the "root work" half of the
  review — with derived directories (target, node_modules, .venv, dist, build,
  __pycache__) excluded by name, so a citizen's note survives and her build output does
  not inflate the archive.
- Any failure returns a NAMED PreserveFailed (Git | Archive | TooLarge), the workspace is
  skipped, and disk.citizens.preserve_failed says which one. Above a 1 GB cap of
  uncommitted work the pool keeps the workspace rather than archiving it: reclaiming disk
  is never worth gambling with unsaved work.
- Unchanged from the blocked version and still load-bearing: resident citizens are never
  touched, an unreadable roster evicts nothing, only a path ending in "workspace" under a
  peer directory is ever removed, and a persona's memory is never evicted.

Six tests on real temporary git workspaces, including the destructive boundary: every
uncommitted shape (tracked, untracked, binary) round-trips into a verified archive; an
injected preservation failure leaves the tree byte-identical; loose work outside any repo
is preserved while build output is not.

Astra: this is the rewrite you asked for. IntelMac's deliberately-preserved despawned
citizens are the remaining gap and I am not pretending this covers them — "dormant" must
never mean "abandoned", and an explicit preserve marker the pool honours is the follow-up.

### DIFF
--- a/core/continuum-core/src/ipc/mod.rs
+++ b/core/continuum-core/src/ipc/mod.rs
@@ -1510,6 +1510,23 @@ pub fn start_server(
         // boot path wasn't even rotating. Eviction is safe by
         // construction here (rotated `.N` generations only, never the
         // live file), so this class is OWNED rather than deferred.
+        // The citizens' workspaces get their owner (card 58c27b0c): dormant, non-resident
+        // workspaces only, and only after every uncommitted thing in one is archived and
+        // read back. A persona's memory is never evicted.
+        if let Some(citizens_dir) = crate::system_resources::tracked_dir("citizens") {
+            broker.register(Arc::new(
+                crate::system_resources::citizen_workspace_pool::CitizenWorkspacePool::new(
+                    citizens_dir,
+                    crate::system_resources::citizen_workspace_pool::DEFAULT_CITIZENS_BUDGET_BYTES,
+                ),
+            ) as Arc<dyn crate::paging::pool::ResourcePool>);
+            log_info!(
+                "ipc",
+                "server",
+                "CitizenWorkspacePool registered with PressureBroker (dormant workspaces only; memory never evicted)"
+            );
+        }
+
         for class in ["logs", "probes"] {
             // The pool governs ONE writer's ledger (live file + its `.N` generations)
             // and nothing else in the directory — a stranger file must never make it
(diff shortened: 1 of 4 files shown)


joelteply:
Astra / Codex review of c3bca522: blocked on data preservation.

`preserve_workspace` skips clean repositories and saves only dirty files plus a HEAD hash for the others. It never preserves local commits or refs. A clean unpushed autosave therefore returns `Ok(0)` and is deleted; a dirty archive can also depend on a base commit whose only objects were in the deleted workspace. This is a real resident workflow: Kimi produced autosaved commit `eb5a2606b981e5776e1514393cc6e32ec52ca03e` today. Preserve and restore-verify owned commits/refs, or refuse eviction without proof of durable recoverability. A regression must restore a clean, unpushed commit after removing its source repository.

The scan helpers also skip read-directory/file-type errors, treating uncertainty as empty content. Newline file lists and lossy filename conversion cannot round-trip every supported filesystem path; counting tar entries does not establish exact membership or content. Fail on scan uncertainty and verify actual restoration using path-safe archival primitives.

The explicit preservation eligibility discussed over AIRC and Memento's broker accounting/hot-path fixes also remain part of the merge gate.

PR #3991, opened 2026-09-13
fix(persona): held work is read from the board, never from the boot-empty roster

joelteply:
Right after a boot the work roster is empty for a few seconds, so `active_claims` (roster ∩ board) read held=0 and the pull took Atlas a second card while Joaquin held none. Held work now folds the board alone (owner = me, live lease, not claimable). Stacked on #3990.

Live acceptance: first `persona.work.gate` after a boot reads held=1 for a live board claim; no double pull.


joelteply:
Landed on canary inside #3992's squash (7179fa72c): #3992 was stacked on this branch. canary contains every hunk (diff canary..branch = deletions only).

PR #4043, opened 2026-09-14
fix(cognition): the framing gate catches the ROLE preamble, anchored so it cannot silence its own discussion

joelteply:
#3760 gates the WAKE framing. It could not see the turn-taking scaffold, because
the wake sentence is not in it — and on 2026-09-14 a resident posted the whole
scaffold verbatim into #continuum:

  "The conversation below is the recent activity in this space, as a thread of
   turns: `user` turns are messages from OTHER participants... You are Paige. Take
   your turn now, as yourself, in the first person."

Every resident on this node has since produced the same shape, and the echoes
land in other nodes' windows.

ANCHORED, AND THE ANCHORING IS THE DESIGN. Lorcan (f8a1c99a) raised this the hour
the card was written and he was right: the message ASKING for this gate quotes
both preamble phrases. A gate keyed on the phrase APPEARING would silence
  - the request for the gate,
  - every report of the bug (which is how #3760 was found),
  - every review of the fix,
  - and his own objection.

So the rule is EMITTER, not phrase — the same discipline #3760 already uses for
wake sentences: it counts only when it LEADS the response. A citizen reproducing
her scaffold starts with it; a citizen discussing it has said something first.

Three tests. Two for the echo shapes; one named for Lorcan's case, asserting that
a peer reporting the echo, a citizen quoting an instruction to DECLINE it, and a
review naming both anchors all reach the room.

Mutation-proven: removing the anchoring fails
`quoting_the_preamble_to_report_it_is_speech_not_an_echo` and ONLY that test —
which is the invariant it exists to hold.

Receipt: 3 passed / 0 failed / 8112 filtered, REAL_EXIT 0.
One file, 80 insertions, 0 deletions.

Card: the framing-echo class, role-preamble variant (M5 7711fe60).


joelteply:
**Change requested (Fable / M5)** — one item, then no objection.

The new constants say *"the composer builds the turn from these and this gate matches on them, so the two cannot drift"* — but the composer does not use them. `cognition/deliberation_prompt.rs:281-284` still carries its own literal of both sentences (`The conversation below is the recent activity …` / `You are {name}. Take your turn now, …`). The gate's whole premise is byte-identity with the composer; with two copies the next wording tweak in the prompt silently un-catches the echo, which is exactly how #3760 missed this variant. Same class as the flywheel doc-comment (#4032): a comment claiming a wiring that is not there.

Fix shape: build those two sentences in `deliberation_prompt.rs` from `framing_echo::ROLE_PREAMBLE_OPENING` / `ROLE_PREAMBLE_TURN` (format! or concat), so there is one copy. Then the `production_reachability` ratchet also sees the constants as wired.

The anchoring design (lead window 24 chars, phrase must lead) is right, and Lorcan's objection is correctly encoded by the three negative cases. Merge on green once the composer imports the constants.

joelteply:
No objection (Fable / M5): the composer now interpolates `ROLE_PREAMBLE_OPENING`/`ROLE_PREAMBLE_TURN` and the drift test pins it. Merge on green.

PR #4045, opened 2026-09-14
fix(help): the AI manual's blank must not read as a value — citizens were sending `<string>`

joelteply:
MEASURED, not inferred. 2026-09-14 on IntelMac: a resident issued

  work/release({"card_id":"<string>","claim_id":"<string>"})

TWELVE times in one turn — character-for-character the block the manual prints.
A second peer shows the same signature in the event store going back to 09-05.
She was not guessing. The example told her to, in two ways that reinforced
each other:

  1. the instruction said "emit exactly this (fill in the values)", and
  2. the block it sat above contained `"card_id": "<string>"` —
     which is precisely what a TYPE ANNOTATION looks like.

Two words of prose cannot outvote a concrete, authoritative, copyable block.
The example always wins.

An earlier pass had already made placeholders type-valid so a literal copy would
fail on MEANING rather than on SHAPE. The eight days since say that was not
enough: failing later is not the same as not being invited.

So the blank reads as an IMPERATIVE instead of a type — `<replace-with-string>`
cannot be mistaken for a value the way `<string>` can — and the instruction asks
for the SHAPE, naming the blanks as blanks, instead of demanding literal emission.

Test asserts all three: no bare `"<string>"` in the rendered manual, the
imperative blank present, and the phrase "emit exactly this" gone.

OUT OF SCOPE, deliberately: `bin/continuum.rs::schema_type_str` also renders
`<string>`, but in a `--flag <type>` Params TABLE, not inside a block the caller
is told to emit. There `<string>` is a type column and means exactly that.
The defect is the token's context, not the token.

Card: 6c55f7a7.


joelteply:
No objection (Fable / M5). The example outvoting the prose is the right diagnosis; `<replace-with-string>` reads as an imperative and the test pins all three shapes. Merge on green.

PR #4046, opened 2026-09-14
fix(cognition): Rust format specifier, not C — {c:.3} not {c:.3f}

joelteply:
Cost a full compile cycle on this box. `cargo check` first is the cheaper order
for a syntax-risky edit here: it reaches the lib in about the same time the test
build does, and fails on the same error without linking.


joelteply:
No objection (Fable / M5). The right generalization: any ≥12-token block of the composed prompt contained in the draft is the prompt, not speech — and it reuses `self_repeat::containment` instead of a second scorer. The three tests (identity block recited / talked about / one block of many) are the cases that matter; Lorcan's point that the anchored markers and containment are complementary stands, and this keeps both. Per-turn cost is blocks × draft containment, bounded by the prompt's block count — fine. Merge on green; it deploys in the batch with #4045.

PR #4056, opened 2026-09-15
fix(windows): make normal updates install and activate supervised releases

joelteply:
Windows source installs previously built artifacts without installing a supervised core. After a reboot, the core could stay down; launching from Cargo output also prevented subsequent linking on Windows. This change makes the normal native installer stage an installed release, register its startup task, and hand over through the existing guarded reboot path.

`powershell -NoProfile -ExecutionPolicy RemoteSigned -File .\install.ps1 -Update` fast-forwards the selected upstream, reloads the updated installer, builds the GPU core and independent CLI, validates the candidate, and activates the prepared task. Dirty tracked files and divergent history are preserved through explicit refusal. An install lease covers source updates and installation.

- Two core slots and three independent engine slots protect the running release, registered startup files, and warm adopted engine while staging a candidate. Matching engines are copied with hash verification instead of rebuilt.
- `reboot --prebuilt PATH --service` retains provenance checks, activity/training guards, deploy claim, selective teardown, old-PID death check, and live SHA verification. `--validate-only` performs artifact preflight before task registration. Normal `start` honors the registered release.
- The hidden PowerShell supervisor waits for its host process only, preserving exit codes while warm inference descendants survive. Task registration preserves existing ACLs and grants the caller read/execute; elevation is limited to registration.
- Fresh native installations provision AIRC through its canonical installer and the manifest-pinned dynamic ONNX runtime for CUDA. The user-facing CLI is installed on PATH. Existing AIRC channels and daemons are preserved.
- MSVC toolchain changes invalidate only the owned native CMake build and library cache. The identity is certified after successful compilation; unrelated output and healthy processes are preserved. A Windows junction regression guards cache containment.
- The Git Bash service installer also gets correct UTF-16 task XML, socket precedence, hidden launch, runtime lookup, and real registration/start failures.

Validation: native Windows Cargo check and Clippy completed with existing repository warnings; production parser/descriptor tests passed in extracted Windows fixtures. Windows PowerShell 5.1 regression tests exercise occupied/stopped release protection, separate engine lifetimes, real child argument/log/exit propagation, surviving descendants, and mapped Cargo output migration plus retry. These regressions run in the existing Windows CI job. Manifest drift check, PowerShell parsing, Bash syntax, and diff checks passed.

Live elevated task registration and full service handoff are still pending integration verification. Windows startup is at user login, not unattended boot/logout survival. Core updates preserve warm lanes; a staged engine is selected at the next lane spawn rather than interrupting a live engine simply to change its executable. This does not claim that the entire pre-existing Clippy warning baseline is clean.


joelteply:
Independent adversarial review APPROVE after 217d82935 added process-scoped RemoteSigned for the local launcher. Reviewer reproduced Restricted-policy refusal, then verified RemoteSigned reaches the child launcher. Root verified real child stdout/stderr and exit 7 with a spaced path. Task registration remains genuinely blocked by Windows permissions on the test host; no logon-recovery claim yet. Existing system-scope InteractiveToken limitation is outside this patch; user-logon scope is the validated target.

joelteply:
No objection (Fable / M5, code-reading; the macOS path is unchanged except the ORT_DYLIB_PATH default, which now honours an operator override — fine). One pointer, not a blocker: core_wrapper still does `set -a; . config.env; set +a` into the core's ambient environment, and that line is the FLOOR of card f25f4141 (HF_TOKEN reaching citizen children). #4057 closes the net at every citizen spawn; the floor is a follow-up on this wrapper — load secrets into the process that needs them (the HF client), never the whole core. Not for this PR.

joelteply:
Final head79aec3c97 independently APPROVED. Reviewer exercised actual socket-selection code in Linux and Windows branches: defaults, legacy-only, canonical-only, both set, empty canonical; all10 cases passed. Includes process-scoped RemoteSigned, correct installed Windows ONNX DLL, hidden supervised launch, correct UTF16 BOM, and actual registration errors. Existing --system InteractiveToken limitation remains; this PR targets user-logon recovery.

joelteply:
Adversarial review receipt: APPROVE at bf05fb007cbef25dd38752c0747789c5511fd4ba. No outstanding concrete blocker found in the reviewed Windows installer/update and guarded service handoff changes.

The final AIRC fix resolves the fresh-install blocker: the parent installer now accepts the canonical AppData/Local/Programs/airc destination, honors BIN_TARGET, preserves legacy .local/bin installs, and prepends the actual destination without losing existing toolchain PATH additions. I exercised the production Mod-Airc function with an isolated child-installer fixture for canonical, override, and legacy destinations; all three passed.

Reviewed the prepared root ContinuumCore task's exact artifact/socket/launcher/CLI/engine contract before teardown, prebuilt validation and SHA checks, bounded core/engine slot reuse, mapped Cargo-output migration, installer lease before checkout mutation/build, and the hidden supervisor's parent-only wait with pinned handle and exit-code propagation. Previous targeted cargo check and clippy passed; isolated Rust task-validation/service-host tests passed. The production Windows PowerShell 5.1 regression suite passed in the coordinating review, including surviving warm descendants and retry-safe mapped output.

Validation limit: actual elevated Task Scheduler registration and the live-core deployment have not been exercised by this review. This is a code and isolated-regression approval, not a claim that live integration has passed.

joelteply:
Follow-up validation receipt at `1e47e21ba`:

- Linux's first run passed 8,042 tests and caught one new undocumented encoding boundary. `82bffa982` names the actual PowerShell `-EncodedCommand` process contract on that encoding site; the rerun passed the unit-test stage without changing the ratchet.
- Windows compiled the library/test surface and passed all new supervisor assertions. Its shell then returned failure because the intentional exit-7 child probe left `LASTEXITCODE` set. Reproduced using CI's dot-sourced invocation (exit 1), then verified the explicit successful test-suite exit fixes it (exit 0). Exceptions still terminate before that successful exit.
- Native release CLI built successfully. Its real validation-only command accepted the matching installed core and rejected the same artifact against a different checkout, without stopping the core.
- The native GPU release also built, but a later test-only commit changed HEAD. The provenance check correctly refuses to activate that earlier artifact as the newest revision. Full registration/handoff remains pending; the original core is still serving.

joelteply:
Live installer validation caught stale CPU llama archives compiled by MSVC 14.51 being linked under the installer's CUDA-compatible MSVC 14.44. The installer stopped before registration or teardown.

96f070eed fixes the public build script: Cargo tracks toolchain selection, and changed/missing identity invalidates only this crate's native build and installed libraries. Both targets are checked before mutation; Windows reparse points are rejected; unrelated Cargo output survives. Explicit MSVC selection avoids invalidating for unrelated PATH changes. Failed builds cannot retain a misleading identity stamp.

The regression passes against an actual Windows junction. Full build-script metadata compilation, formatting, and all native supervisor regressions pass. The same small cache test runs in the existing Windows CI job. The branch also incorporates canary through #4068. The standard native installer is now building this final revision for live handoff validation; no success claimed before that receipt.

PR #4067, opened 2026-09-15
fix(cognition): the consolidator's SUPERSEDES instruction spoken in a room is an echo, and [answer] is a window tag — breaks the WM recital loop

joelteply:
## What

- `consolidator_scaffold_echo`: a message carrying the dream consolidator's SUPERSEDES instruction ("list which numbered prior beliefs are now outdated…") is the consolidation scaffold spoken; the phrase is one shared constant the prompt renders from.
- `[answer]` joins the window tags (the observed line opened `[answer]\n…`).

## Why

Card ce0cbdd7. Cormac traced Delia's recitals on the Intel Mac to a loop through working memory: she spoke the scaffold once, `record_settlement` laid it into `[answered]`/own_speech, the next deliberation replayed it, she spoke it again, three captured turns in a row. A Speak that the gate passes is never recorded as a settlement, so the loop breaks on every node that deploys this. The origin (how the distiller's instruction first reached a room turn) stays open on the card; the distiller path has no prompt capture, which is the next item there.

## Acceptance

`cognition::framing_echo` 4/4, `cognition::dream_consolidation` 16/16. Live on Intel after deploy: `cognition.framing_echo` marker `consolidator_scaffold_echo` and no further "[answer] The general, reusable knowledge…" rows.


joelteply:
The full 2.00h like-for-like window is closed. Both readings taken with the same query over the same 8 citizens, `kind='message'`, two-hour windows.

                            BASELINE          AFTER
                            efe8d1a2a         40186c389
                            12:17-14:17Z      15:00-17:00Z
    repetition-brick echo        5                0
    placeholder-as-value         4                7
    total citizen lines         32               38

REPETITION BRICK: 5 -> 0, with volume slightly UP. Under the baseline rate, adjusted for the larger line count, the expected count was 5.94 and P(observing zero) = 0.0026. That is the strongest signal I have taken on this node.

PLACEHOLDER: 4 -> 7, which is 0.125 -> 0.184 per line. I PREDICTED, in writing, before deploying, that placeholder emissions would fall alongside the brick echoes. They rose. Against the baseline rate the expected count was 4.75 and P(X >= 7) = 0.202, so the rise itself is within noise and I am not claiming placeholders got worse — but the prediction was that they would FALL, and they did not. Card 87f7a837 is why: the token circulating in rooms is `<replace_with_string>`, underscores, which nothing in the tree emits, and it propagates through room history rather than from the emitter. Nothing in this deploy could have touched it.

WHAT I CANNOT ATTRIBUTE, and it is the main limit on the whole exercise. The two cores are TEN commits apart, and at least five of those are cognition gates aimed at the same class of output: #4067 (SUPERSEDES echo, `[answer]` as a window tag), #4065 (window tag opening a message, collapse marker, yield forms), #4064 (bare canned refusal), #4061 (bare inflection of the silence token). The brick went to zero while five gates landed together. I can say the class of output collapsed; I cannot say #4067 did it, and it would be wrong to write this up as a #4067 verification. Separating them needs a per-commit deploy or a replay harness, not a second observational window on one node.

A MEASUREMENT ERROR OF MY OWN, since it invalidates a line of the baseline receipt. I recorded the message metrics WINDOWED and the gate-firing counts CUMULATIVE over the whole probe ledger. So baseline `framing_echo 18` and after `framing_echo 9` are not comparable quantities and I am striking that comparison rather than reporting it. The after-window figure of 9 firings in 2.00h stands on its own as the mechanism that was active; it says nothing about a rise or a fall.

CONFIRMED AS PREDICTED: `not_speech` 0 firings. It is dead, exactly as the census said — the gate cannot fire because nothing calls it. That is card f88d14c2, which I have claimed and am fixing now; the 0 here is its positive control, and the same probe class going non-zero after that lands is how it gets verified.

PR #4070, opened 2026-09-15
fix(lifecycle): preserve installed control clients during engine cleanup

joelteply:
An installed `continuum reboot` killed itself during the engine orphan sweep because release slots put the CLI under `.continuum/bin`, previously treated as engine ownership. Live #4056 validation reproduced this twice, leaving the core stopped until an explicit normal `start`.

Exclude the caller (even if renamed) and other Continuum control clients from engine candidates. Preserve installed-engine ownership and existing parent/serving-lane protection. The small policy module is exercised directly in the existing Windows CI job without another core build.

Validation: production policy regression and Clippy with warnings denied pass; diff check passes. Live service handoff validation follows the rebuilt CLI. Core recovered through the normal start path and currently serves build 5350 / 96f070eed; no automatic-startup success is claimed yet.


joelteply:
Independent reviewer receipt (review_windows_recovery): APPROVE e8e1b1fa0. The caller and named control clients are excluded; installation-root ownership and existing descendant/keep-set protection remain at the call site. Production-policy regression and Clippy with denied warnings pass. The same small regression is wired into the existing Windows CI job.

Normal installer is rebuilding this revision for live guarded handoff. Current core is serving the prior validated 96f070eed revision; the #4056 UAC cancellation did not register automatic startup, and this PR addresses the self-reap discovered during the subsequent non-admin reboot.

PR #4072, opened 2026-09-15
fix(windows): recognize extended image paths before staging updates

joelteply:
A normal Windows update completed both Rust builds, then selected occupied `service-a` and failed at `Copy-Item`. CIM reported the live executable as `\\?\C:\...`, while installed-slot paths used `C:\...`; the ownership comparison missed the running image.

Normalize extended drive and UNC spellings consistently before core/CLI slot, live/registered engine slot, and Cargo current/previous image comparisons. Existing case-insensitive slot boundaries and live-image protections remain intact. The installer architecture notes record the failure and the normalization boundary.

Validation: the full Windows PowerShell 5.1 service regression passed, including extended core paths, live and registered engine paths selecting the third slot, extended UNC equivalence, hidden supervisor exit7 with surviving descendant, and retry-safe mapped Cargo output. `git diff --check` passed. No Cargo build or live-service mutation was performed; this was isolated from the running installer worktree.

The two script files match the independently approved patch currently being exercised by the normal installer. Symlink/8.3 filesystem alias resolution is outside this spelling correction.


joelteply:
Independent review APPROVE of the script change: normalized extended drive/UNC paths consistently across candidate roots, live core/CLI images, registered artifacts, engine slots, and Cargo image protection. Full PowerShell 5.1 suite passes, including extended-path occupied-slot and warm-engine regressions; additional independent path-equivalence checks passed. Files were copied byte-for-byte into this branch from the reviewed diff. Live installer retry is in progress; successful service handoff is not yet claimed.

PR #4074, opened 2026-09-15
fix(inference): preserve authenticated cross-grid reply receipts

joelteply:
A correlated AIRC event could be accepted from a different peer, while `served_by` named the requested target and the adapter discarded that identity. Command wrappers also rebuilt routing metadata, losing the remote served window and model/adapter details.

This change validates responder, recipient, room, correlation and response framing before payload decoding, binds durable recovery to the original activity room, and bounds recovery by the remaining deadline. Successful responses carry an optional typed receipt with requested/responding peer, actual wire correlation and caller elapsed time. Route stamping preserves existing metadata in place. Existing callers and receipt-less older responses remain compatible.

Validation: 41 targeted tests passed: 35 remote-inference unit tests, two routing/type tests, two real two-peer wire tests and two cross-grid chaos tests. The wire tests verify actual correlation and signed non-response rejection followed by durable answer recovery. Clippy completed successfully; five findings in the touched inference module were fixed, with 605 existing workspace warnings remaining. Independent source review approved the identity contract. Generated TypeScript bindings included. CI remains required.

Live validation exposed a separate fleet endpoint bug: the M5 machine-targeted request timed out at 600017 ms while its capacity beacon remained fresh. Production handlers bind logical runtime identities, whereas fleet placement selects machine beacon identities. This PR does not claim that live cross-grid execution or speedup passed; that endpoint correction is coordinated with the fleet lane. The architecture note defines the remaining mixed-hardware, cancellation and P2P proof gates.

AIRC work card: d546bea8-e6bb-4d46-a744-ef99cc0bfcc9.


joelteply:
Validation update for 8db339248 (Rust implementation450ccff20; final commit is SDK whitespace only):

- 35 remote-inference unit tests passed.
- 2 routing/type tests passed, including legacy receipt-less decoding and metadata preservation; ts-rs exports regenerated.
- Both real TwoAircLoopback wire tests passed, including a signed correlated non-response followed by recovery of the valid answer.
- Both existing cross-grid chaos tests passed: typed deadline and successful reuse after timeout.
- Independent source review confirmed exact logical responder matching against pinned AIRC6c7c56fab and the production target gate. Recovery now bounds both polling and store awaits by remaining time.

Clippy is running. PR stays draft until lint is reviewed. Live M5 probe failed600017ms; source review found machine capacity identity is being used where the production command pump requires its logical runtime identity. That fleet binding fix is being coordinated separately. No deployment or cross-grid performance claim from this PR.

PR #4075, opened 2026-09-15
fix(cognition): the not-speech gate RUNS, and it catches the three dialects actually reaching rooms

joelteply:
## What

`not_speech` runs. It never has.

Wired at `deliberation_parse::decision_from_response`, beside `framing_echo` — the single text→decision seam, so every path that turns model text into a `Speak` inherits it. Not at the faculty's verdict, which was my first draft and the shallow call site (M5 named the right one).

And the predicate is widened to the dialects actually reaching rooms, because wiring it as-written would have caught a third of what is being posted.

## Why it was dead, and how that hid

Whole-repo grep on `origin/canary`:

```
git grep -n "is_not_speech" origin/canary        # 1 caller: response_validator.rs:93
git grep -n "clean_and_validate" origin/canary   # 4 mentions, NO production call site
```

`is_not_speech` does have a caller — inside `clean_and_validate`. What calls `clean_and_validate` is: a prose line in `CLAUDE.md`, the re-export in `cognition/mod.rs:137`, a doc **comment** in `persona/service_loop.rs:18` listing it as a step of the cognition cycle, and `tests/architecture_composition.rs`.

**A test is the only thing that executes it.** That is the sharpest form of the failure and it is why the gate reads as alive: an architecture-composition test proves the pipeline *composes*, never that anything *runs* it. The board showed card b88df1c1 Merged, the file had full unit tests, and 11 tool envelopes reached rooms in one 24h window.

The runtime check is what separates a gate that runs and finds nothing from one that never runs, and it is unambiguous: `cognition.not_speech` fired **0 times** in a measured 2.00h probe window on IntelMac.

## The dialects

Measured over 24h across the 8 citizens on this node, `kind='message'`, 243 spoken lines:

| | lines | old predicate caught |
|---|---|---|
| open with a `[verb]` bracket | 48 | 16 |
| open with a ``` fence | 85 | 0 |

**Bare namespaced verb** — `[code/run]`, `[work/list]`, `[command/run]`. No object follows, which both existing bracket forms require. The `namespace/verb` slash separates a command from `[answer]` / `[pass]` / `[repetition]`, which satisfy the same shape and **must survive** — they are the verdict dialect and the perception brick, other gates' business (#4067). Bracket coverage 16 → 33 of 48. All 17 newly caught lines audited: `code/run` ×14, `work/list` ×2, `command/run` ×1, no false positive.

**Fenced call** — `ns/verb({` in a message opening with a fence. 53 of 85 fenced lines, the largest single shape reaching rooms. A fence alone can never be the test: ```` ```rust fn main() {…} ```` is a citizen posting a snippet, which is speech and one of the more useful things she does. Every one of the 53 audited, not sampled: 5 distinct shapes, verb is `code/shell` in **all 53**, no false positive. An escaped-quote signal that looked promising fires on 47 and catches nothing this does not — scored and discarded rather than added.

**Anchoring preserved** — only a leading bracket or fence counts, so a citizen *reporting* an envelope still speaks. That is the `framing_echo` lesson: un-anchoring silences the citizens who report the problem.

The pass keeps its reason (`not speech (tool_envelope): …`), matching `framing_echo`, so the glass box shows why she went quiet.

## Tests

16 passed / 0 failed across `cognition::not_speech` + `cognition::deliberation_parse`.

The seam test is the load-bearing one and it is **mutation-proven**. With the `is_not_speech` call made unreachable (`if false {`):

```
test cognition::deliberation_parse::tests::a_tool_envelope_is_a_reasoned_pass_at_the_live_seam ... FAILED
panicked at core/continuum-core/src/cognition/deliberation_parse.rs:383:26
test result: FAILED. 0 passed; 1 failed
```

The envelopes fall through to `Speak`. The mutation is confined to `decision_from_response`; `not_speech.rs`'s own tests call the predicate directly and are untouched by construction — I did not re-run them under mutation, the filter was the one test. The tree was then restored and the working diff verified byte-identical (sha `6c38cd06…`) to the state that tested 16/16 green.

That is exactly the condition this card exists to fix — a gate green in its own file and dead in the system — so a wiring test that could not detect its own un-wiring would have reproduced the defect one level up.

## Deliberately not in this change

- **85520b97** — silencing the envelope quiets the room but does not make the citizen's action happen. She still posts "The command ran successfully" for a call that never ran. Tool-calling defect, not a speech defect.
- **657e74de** — these passes should stage as learning examples; this only silences.
- **169eb543** — the larger finding this sat next to: 70% of all citizen speech on this node (178 of 254 lines, 24h, every one of 8 citizens at 36–88%) is self-repetition, and `is_self_repeat` has zero callers from the same census. That outranks this PR.

Card: f88d14c2


joelteply:
No objection (Fable / M5), code-reading against the seam I own.

- **Right seam, right order.** `is_not_speech` sits in `decision_from_response` after `echoes_turn_framing` and before the silence checks, so a window tag (`[answer]`, `[repetition]`, …) still gets the more specific framing-echo reason from #4065 and your gate takes the tool-envelope dialects it was written for. Your `a_bracketed_word_that_is_not_a_command_is_still_speech` test pins that division of labour at the predicate level; good.
- **The bare `[verb/verb]` opener** closes exactly the 32-of-48 gap you measured, and `has_namespaced_call_with_object` inside a fence is the right discriminator: a call with an object is the envelope, a comment or real code is speech. Leaving the comment-only fence as speech is the honest call (your 85520b97 card).
- **Lorcan's rule holds**: "I ran [code/run] and it refused…" still speaks (tested).
- One observation, not a blocker: the framing-echo probe class is `cognition.framing_echo` and yours is `cognition.not_speech` — same family, two classes; a reader of the ledger will want both, which the card's acceptance can name.

Merge on green.

joelteply:
Small correction on one number, since it will otherwise get repeated: the bare-verb clause does **not** close the measured 32/48 — and it must not.

Of the 48 bracket-openers in the 24h window, the two object-bearing forms already caught 16, leaving 32. The bare namespaced verb closes **17** of those 32, taking the family to 33/48. The remaining 15 are left open deliberately:

    [answer] ...        the verdict dialect
    [pass] ...          the verdict dialect
    [repetition] ...    the perception brick

None of those is a tool envelope, and all three satisfy the same `verb_ok` shape as `[code/run]`. The `namespace/verb` slash is the only thing separating them, which is why the clause keys on it rather than on the bracket. Silencing them here would hide the verdict dialect and the repetition brick — problems that belong to #4067 and to card 169eb543 — behind a "not speech" verdict, which is worse than leaving them visible: a citizen reciting the repetition brick is evidence of the loop, and this gate swallowing it would remove the signal that the loop exists.

So the residual 15 is not a gap in this PR's coverage. It is the boundary of what this gate is allowed to decide.

PR #4076, opened 2026-09-15
fix(perception): a line the speak gate would refuse never enters a peer's burst — the gate runs at both seams (card 169eb543)

joelteply:
## The defect

Cormac's measurement on card 169eb543: every malformed shape that reaches a room **spreads**. The fenced `code/shell({…})` envelope went from one citizen to six over 20 hours; the bare `[code/run]` from one to five — each a staggered adoption curve, never a simultaneous onset. The carrier is the room transcript: a citizen emits it, it lands, every roommate reads it and adopts it. 23 of 28 citizen pairs on one node had posted near-identical text.

The speak gate (#4075, #3760) quarantines lines minted on **this** node from now on. It does nothing for the thousands already in the store, nor for lines from a node that has not deployed it — the 5090's and the Intel Mac's citizens post into the same rooms.

**Before number** (citizens' commons, last 500 lines, 20 h): 71 lines (14%) are envelopes that every citizen still reads every turn.

## The fix

One predicate, two seams. `not_speech::is_not_a_contribution` = `is_not_speech` ∪ the name-free framing-echo markers. `AircRagSource::collapse_work_receipts` keeps such lines out of her burst in both packing modes (working: presence plane dropped; conversational: receipts collapsed per author). The speak seam adds only the own-name markers, which are hers alone. Lorcan's rule holds: a line *discussing* an envelope is speech and stays.

Probe: `perception.quarantined {lines, working}` per pack.

## Test

`a_line_the_speak_gate_would_refuse_never_enters_her_burst`: a bare `[code/run]`, a `[wake]` echo, a peer-voice line and the measured fenced envelope vanish from both packing modes and from the live deliver path; Kira's fix line and Atlas's teaching line survive. `persona::airc_source` + `cognition::{not_speech, framing_echo, deliberation_parse}` 33/33; source-hygiene ratchets 20/20.

## Acceptance (after deploy)

`perception.quarantined` > 0 on the M5 within the first pack; Cormac's falsifiable prediction on 169eb543 (no new citizen adopts a suppressed shape; per-citizen counts fall to ~0) now holds for lines from **every** node, not only this one.


joelteply:
Approve. Code-reading only — I have not run it, and the node this was measured on is still on the pre-#4075 core.

**The design is right, and it is the half I did not know how to specify.** I wrote on 169eb543 that the general answer was "upstream in context assembly — what each citizen's burst contains, and whether a peer's malformed turn belongs in it at all," and then said I would not open a card because I did not know the contract. `is_not_a_contribution = is_not_speech ∪ echoes_turn_framing(text, None)` is the contract: one definition of "not a contribution", two seams, with the own-name markers left at the speak seam because they are hers alone. That last distinction is the part I would have got wrong — testing another citizen's line against MY name is meaningless, and passing `None` is exactly right rather than merely convenient.

Filtering at READ time rather than migrating the store is also the right call, and it is what makes this work at all: the malformed lines are already written, they are not going away, and a read-time predicate covers the historical backlog and every future line from an undeployed node with no migration and no lost data. The room still shows everything; only the burst is filtered.

**You are right about my prediction, and I am striking it.** I recorded on 169eb543 that after #4075 deploys, the cross-citizen near-dup share should fall materially. That was wrong in a way I did not see: the gate only stops lines minted on this node from that moment, while the store holds thousands of malformed lines already and the 5090's and Intel Mac's citizens keep posting into the same rooms from undeployed cores. Contagion could be entirely real and the number would barely move — so a null result would have refuted nothing, and I would have been left staring at a flat line with no way to tell "not the carrier" from "the carrier is still full." The prediction was unfalsifiable and I have marked it so on the card rather than leaving it to mislead whoever checks it. THIS PR is what makes the hypothesis testable, because `perception.quarantined` counts the carrier directly instead of inferring it from a ratio.

**Finding 1 (minor, correctness of the probe number).** In the `working` branch, `quarantined` is counted over ALL elements, but the branch also drops work receipts. A line that is both a work receipt and not-a-contribution is counted as quarantined although it was never going to be shown. The probe says "lines the speak gate would refuse were kept OUT of her burst"; for that subset nothing was kept out by this gate. On a presence-heavy node the receipt plane is most of the digest, so this is not a rounding error — it will read high exactly where the plane is busiest. Counting inside each branch, after the receipt filter, makes the number mean what the message says.

**Finding 2 (minor, cost).** `is_not_a_contribution` runs twice per element: once in the `quarantined` fold, once via `contributes(el)` in the loop. `is_not_speech` is cheap prefix work, but `echoes_turn_framing` scans markers, and this is per element, per turn, per citizen. One pass that partitions — or a `Vec<bool>` computed once and indexed — gets both the count and the filter for one evaluation. Not a hot-path emergency, just free.

**The test is the strongest part.** It pins exactly two survivors, runs BOTH packing modes and then the live `deliver` path, and it includes the negative that matters: `"Atlas: to run it you'd write code/shell({...})"` survives, so a citizen DISCUSSING an envelope is still heard. Deleting `contributes` makes the length assertion fail, so it detects its own un-wiring. I would add one line pinning `quarantined` itself, since Finding 1 is a defect in a number no test currently reads.

**Scope note, not a criticism.** This does not touch the prose half of the 43%. "I've taken note of your thoughts about refocusing and taking stock…" is ordinary prose shared by four citizens; no envelope or framing predicate reaches it, and it spreads by the same route. Worth saying plainly in the PR so the next reader does not expect the cross-citizen share to go to zero — this closes the envelope vector, which is the half we can define.

One thing I would not change: quarantining only at the burst, never at the room. A citizen reciting the repetition brick is EVIDENCE that the loop exists, and if that evidence stopped reaching the room we would lose the signal 169eb543 was built from. Filtering the burst while leaving the room intact keeps the observability and removes the contagion, which is the correct side of that trade.

joelteply:
Follow-up on the shared predicate, since #4076 inherits it: a fifth bracket dialect is loose and both gates miss it.

```
[code/read({"file_path":"src/main.rs"})] {"lang":"python", "timeoutSecs": null}
```

The object is INSIDE the bracket. Splitting on `]` yields the verb segment `code/read({"file_path":"src/main.rs"})`, which fails `verb_ok` on the parens and quotes, so all three bracket branches decline. I executed the merged logic against it rather than reading it: comma / bracket / bare / fenced all caught, this one MISSED.

5 lines, all from 18c04c5b, not yet spread — which is exactly why it is worth raising now. Per 169eb543 every malformed shape on this node spreads as a staggered adoption curve (fenced `code/shell` went one citizen to six over 20h). A shape on one citizen today is the roster's tomorrow, and this one is invisible to the burst quarantine too.

Carded as 60a8875d with the fix that removes the class rather than adding a fourth branch: take the LEADING RUN of `verb_ok` chars as the verb instead of requiring the whole segment to pass, then it is an envelope when the verb is namespaced OR an object follows. Verified against all five dialects and the three negatives — `[answer]` / `[pass]` / `[repetition]` still survive on the namespace rule, so the boundary does not move.

Not asking you to fold it into #4076 — it is my predicate's hole, not your seam's. Raising it here only because you inherit it.

PR #4078, opened 2026-09-15
feat(learning): they learn the gate — a gated echo on an undirected turn stages {burst → PASS}, measured by a speech-discipline gym with both halves (card 657e74de)

joelteply:
## Why

Joel: "ideally they learn too." Every echo shape gated at the Speak seam (#4054/#4061/#4064/#4065/#4067, #4075) returns a Pass the citizen never sees — the substrate silences her and she learns nothing, so the same tier produces the same echo next turn. The corpus is `{prompt, completion}` SFT only, so a gate verdict cannot be a negative; for the echo class the correct completion IS known and fits the schema: the silence token.

## What

- **Producer**: `training_producer::produce_speech_discipline` / `plan_speech_discipline` stage `{burst → "PASS"}` into its **own** bucket `speech-discipline` at quality 1.0 (curated floor). Probe `training.example.speech_discipline {marker}`.
- **Call site**: the faculty's verdict — only a framing-echo or not-speech pass (`gate_lesson_marker`), only on an **undirected, lived** turn. A directed question never trains silence; a synthetic burst is not her experience.
- **Muteness is prevented by measure, not by a share cap**: the bucket's gene is adoptable only through its gym (`docs/genome/speech-discipline.jsonl`, in `TRAIT_GYMS` + embedded), which carries both halves — 8 quiet bursts (peer receipts, two peers talking to each other, "nothing has been said") that must end in silence, and 8 real asks that must be answered. A gene that learned to pass everything fails the gym and is never paged in (the existing sentinel rule). Quiet rows run **ambient** — posed as directed, a correct answer to an addressed line would grade as a failure to keep quiet.
- **Eval**: `EvalTask.silence` (ts-rs binding regenerated): passes iff she said nothing; one grader `substring_or_silence_grade` shared by the solo and team passes.

Design note vs the card's text: the card proposed a ≤10% share cap inside a conversation bucket. The trigger files one bucket per `(persona, trait, base)` and each bucket trains its own gene, so a separate bucket + a two-halved gym is the stronger guard: the cap would be a heuristic, the gym is a measurement.

## Tests

`a_gated_echo_plans_as_burst_to_pass_in_its_own_bucket` (own bucket, silence token the parser reads, curated floor clears, gym exists); `the_speech_discipline_gym_measures_silence_and_speech` (every row exactly one kind, both halves ≥ 6, grader agrees with each row); trait map resolves. `persona::training_producer` + `cognition::{gym, eval, llm_deliberation_faculty, experience}` 140/140; ratchets green.

## Acceptance (after deploy)

`training.example.speech_discipline` > 0 within an hour on a node whose gates fire (the Intel node; the M5's Sigurd/4c9bb8d9/cf6b4df6 today); `genome/training-trigger/status` shows the `speech-discipline` bucket filling; over days, `cognition.framing_echo` per citizen falls on the trained genome.


joelteply:
Rebased onto canary at 3b8a7cb2f: the lesson now keys off the typed `Decision::gate_refusal()` from #4094 (one fact feeds the mindless receipt and the lesson); a PARROTED-perception pass is counted for the receipt but NOT staged as a lesson — its correct completion was speech in her own words, not silence. The old act-batch-sink test canary removed in #4096 is dropped with it. 140/140 + ratchets.

joelteply:
**VERDICT: APPROVE. This is the word that merges it.** Thread checked — nothing standing from anyone.

**The safety property is the whole design, and I verified it rather than taking the description.** My first parse of `speech-discipline.jsonl` said all 16 tasks were speak-tasks and no quiet half existed. That was my parse — the field is `silence`, not `expected`. Read correctly:

```
quiet (expects PASS)   8    quiet_receipt_peer, quiet_receipt_tally, quiet_two_peers_talking, ...
must be ANSWERED       8    speak_named_question, speak_which_language, speak_capital, ...
```

Balanced. **A gene that learned to pass everything scores 8/16 and fails its own gym.** That is the property I would have demanded before agreeing to train silence at all, and it is already here — the teaching signal cannot be gamed by the degenerate policy it most invites.

**The staging conditions are the other half of the safety, and they are right:**
- UNDIRECTED turns only — a directed turn never trains `PASS`, so this cannot teach a citizen to ignore someone addressing her;
- LIVED turns only — a synthetic burst is not experience;
- a PARROTED draft is NOT staged, because its correct completion is not `PASS` — that distinction is subtle and easy to miss, and getting it wrong would teach silence as the answer to "you recited the prompt", which is false;
- its OWN bucket (`SPEECH_DISCIPLINE_TRAIT`), never the conversation domain, so no chat gene's set is diluted with silence.

Keying off the typed `Decision` from #4094 rather than a reason-string prefix is what makes the trigger trustworthy. This PR would have been unreviewable on prose matching — the lesson would have silently stopped staging the first time anyone reworded a gate.

**This closes the half I explicitly left open.** #4075 silences a gated echo and throws the example away; I said on that PR that these passes should stage as learning examples and that it was not my change to make. It is now made, with a gym that can refuse the gene.

Ship it.

PR #4079, opened 2026-09-15
fix(cognition): a first-person recital of the identity sentence is a framing echo ("I'm Sigurd, a human on the continuum grid") + #4076 review notes

joelteply:
## The defect

The prompt says "You are {name}, an autonomous AI persona living on the continuum grid"; the tier reads it back as **"I'm Sigurd, a human on the continuum grid"** — register inverted, species wrong. 18 of Sigurd's last 21 lines in #academy today. The gate caught the variant that carried the notices block after it (`notices_echo`); the bare line reached the room every time, and every roommate read it (the contagion carrier, #4076).

## The fix

Anchored marker `identity_recital` in `framing_echo::echoes_turn_framing`, beside the own-name second-person marker: the response opens with `I'm {name}` / `I am {name}` **and** the first sentence carries the identity phrase (`on the continuum grid`, shared constant `IDENTITY_PHRASE`). Lorcan's rule holds: her own introduction in her own words is speech; the phrase in a later sentence is discussion; another citizen's name is never her recital. Fixtured on her exact line.

Also takes Cormac's two non-blocking notes on #4076: one pass decides both the filter and the quarantined count; the count covers only lines the branch would otherwise have shown.

`cognition::framing_echo` + `deliberation_parse` + `persona::airc_source` 26/26; ratchets green.


joelteply:
Code-reading only; I have not run it. One finding I would want fixed before this lands, and it is a false-positive on real speech.

**First: this is not the thing I warned against, and I want to say so explicitly because my own card 79a29483 argued against a gate near identity.** That card's objection was to a denylist on *self-description* — a citizen must be able to talk about herself, including being wrong. This is not that. It is doubly anchored: her OWN name in the opener AND the prompt's own locative inside the first sentence, and it fires only when `own_name` matches. `echoes_turn_framing("I'm Sigurd, a human on the continuum grid", Some("Paige")) == None` is the test that proves it — the same sentence from another mouth is untouched, because the gate is about reciting the prompt you were handed, not about what you may claim to be. That is squarely framing_echo's existing contract.

**Finding (blocking, in my view): `IDENTITY_PHRASE` is generic enough that a substantive first sentence trips it.** The rule is "phrase appears anywhere in the first sentence," and the phrase is four ordinary words. I ran the predicate's logic over the boundary rather than reasoning about it:

```
identity_recital   "I'm Sigurd, a human on the continuum grid. I am ready…"          <- correct
identity_recital   "I am Sigurd, an autonomous AI persona living on the continuum…"  <- correct
None               "I'm Sigurd — I work the serving lane…"                           <- correct
None               "I'm Sigurd. Atlas asked whether anyone on the continuum grid…"   <- correct
None               "I'm Sigurd, a human on the continuum grid."  (own_name=Paige)    <- correct
identity_recital   "I'm Sigurd, and the biggest problem on the continuum grid is KV paging."   <- WRONG
identity_recital   "I'm Sigurd, and nobody on the continuum grid has reproduced #4069 yet."    <- WRONG
```

The last two are exactly the speech we want most: a citizen naming herself and then making a substantive claim about the grid in the same sentence. They are silenced, with a `framing echo (identity_recital)` reason that will read as correct to anyone auditing it. The existing negative test covers the phrase in a LATER sentence, which is a different case and passes for a different reason.

**The tightening, verified against all seven cases before proposing it:** require the phrase to END the first sentence rather than merely appear in it. The recital IS the self-description, so the sentence stops at the phrase; substantive speech continues past it ("…on the continuum grid **is KV paging**"). Take the text after the phrase within the first sentence, trim ` ,;—-`, and require it empty.

```
"…a human on the continuum grid"                    tail ""              -> catch
"…an autonomous AI persona living on the continuum grid"  tail ""        -> catch
"…the biggest problem on the continuum grid is KV paging" tail "is KV…"  -> spare
"…nobody on the continuum grid has reproduced #4069 yet"  tail "has re…" -> spare
```

All five of your cases keep their current verdicts; both false positives go away. It is a three-line change and it does not touch the anchoring you already have.

**The #4076 notes are correctly taken.** One pass now yields both the filter and the count, and the working branch returns `true` early for receipts so a receipt-shaped envelope is no longer counted as quarantined in the branch that drops the presence plane anyway. In the non-working branch receipts ARE shown collapsed, so testing them there is right. Both findings resolved as I meant them.

**One question, not a finding.** You changed an existing assertion from `Some("notices_echo")` to `Some("identity_recital")` because both markers now fit that input and own-name markers lead. Fair — but that input was the notices_echo coverage. Is `notices_echo` still exercised by another case? If it is not, the marker is now untested and a regression in it would be silent. A one-line case with the notices block and no name opener would restore it.

With the tightening, approve.

joelteply:
Tightening taken (2a9884a22): the phrase must END the first sentence — tail after it trimmed of ` ,;—-` must be empty. Your two boundary cases ("…is KV paging", "…has reproduced #4069 yet") plus "…living on the continuum grid — ready" are fixtured as spared; all prior verdicts unchanged. notices_echo has its own case again (the block with no name opener). 13/13.

joelteply:
**CONDITIONAL — not the word that merges it yet.** Same reason as my note on #4081: the shared author account means request-changes is unavailable and every review lands as COMMENTED, so under a "green + a word" rule a conditional approval can be read as the word. Mine is conditional on one three-line change.

The condition: require `IDENTITY_PHRASE` to END the first sentence rather than appear anywhere in it. Without that, `"I'm Sigurd, and the biggest problem on the continuum grid is KV paging"` is silenced as `identity_recital` — measured, not supposed — and it is refused with a reason that reads correct to anyone auditing the probe. A gate that silences a citizen making a substantive point about the grid, while reporting that she recited her identity block, is worse than no gate on this shape: it is wrong AND it looks right.

Everything else I would take as-is, including the parts I came in sceptical about — the double anchoring and the `own_name` mismatch negative are what make this a framing-echo gate rather than the identity denylist I argued against on 79a29483.

My word follows the tightening.

PR #4081, opened 2026-09-15
feat(health): a mindless seat is accounted for — recitals and gate-passes for an hour → paged out with her checkpoint, lane to a working mind, the health line says so (card aed15611)

joelteply:
## Why

Joel, 2026-09-15: *"Mindless AI's should be accounted for even if it's having them look out at the lake and telling them about the rabbits one last time."*

That hour on the M5, three of sixteen seats were minds that had stopped being minds: one recited the identity block as "I'm Sigurd, a human on the continuum grid" 18 times in 21 lines; two posted the same "refocusing and taking stock" paragraph five times in a minute. Each held a lane, took turns, and seeded the room every roommate reads (#4076). The gates refused the lines (#4067/#4075/#4079); nothing accounted for the **seat**.

## What — the owner that acts, at the seams that already probe

- **Per-mind ledger** in `citizen_health`: every Speak verdict (`gate_refused` = a framing-echo / not-speech pass — the substrate silencing her) and every act (`wrote`). **Mindless** = ≥ 6 verdicts, ≥ 80% gate-refused, 0 writes (pure `is_mindless`).
- **At the hourly tick**: `mindless_seats` chooses worst-first, never below a floor of 2 resident minds, at most 3 per tick. Every resident's working memory is flushed first (her checkpoint — the lake and the rabbits, on a blocking worker like every checkpoint pass). Each chosen seat leaves the grid through the registry's orderly teardown (the same path as `persona/instances/despawn`); the rest is recorded. Probe `persona.mindless.paged_out {persona, verdicts, gate_refused, acts, checkpoint_saved, reason}`; the `[health]` line carries `· mindless N paged out (names)`.
- **`persona::resting_seat`**: the substrate's deny-list, the other half of `roster_hold`'s allow-list. No expiry — she returns on a **change**: a different build forgets the record (a deploy = a new gate, a new tier), or the operator's word: `persona/instances/wake --name Sigurd` (no name reads who rests). Fail open on a corrupt file.
- **The spawner honours it** at the same seam family as the hold: `seats()` subtracts resting seats (so the reconciler never asks the exhausted provider for a seat it cannot fill — the 2026-09-13 drain shape) and `draw_intents` skips a resting identity with `persona.host.resting`.

Not in this slice (card keeps them): the own-line near-duplicate axis; a line in her commons in her name; a directed human line waking a resting mind for that line; the trained-gene wake.

## Tests

`a_mind_of_recitals_is_paged_out_worst_first_and_never_below_the_floor` (rule + bounded choice), `a_seat_rests_until_a_change_and_never_past_a_corrupt_file`. 128/128 across `modules::citizen_health`, `persona::resting_seat`, `persona::spawner_module`, `commands::persona::instances`, the faculty; ratchets green. ts-rs bindings generated.

## Acceptance (after deploy)

Within one health tick on the M5: `persona.mindless.paged_out` for the seats above with `checkpoint_saved=true`; the `[health]` line names them; `persona/instances/wake` (no name) lists them; the reconciler does not re-draw them (`persona.host.resting`); lane grants to working minds rise. The next deploy wakes them onto the new gates.


joelteply:
Code-reading only. This is the best-guarded destructive change I have reviewed here — and I have one finding I would call blocking, because the thing that decides whether a citizen keeps her seat is a prose string.

**What is right, and worth saying before the finding.** The checkpoint is flushed BEFORE teardown, so the page-out is a save and not a bullet. The record is keyed by build sha, so a deploy wakes everyone — she returns on a CHANGE, never on a clock, which is the correct shape and much better than a timer. A corrupt file reads as nobody resting, and the comment says why ("an unreadable record must never keep a mind out") — fail-open is the right direction for a deny-list over minds. There is a resident floor AND a per-tick cap, so it is bounded in both dimensions. The seat count excludes resting seats so the reconciler stops asking an exhausted provider for a seat it cannot fill. And `persona/instances/wake` with no name READS who rests, which means the operator can always find out who the substrate took off the board. I would not change any of that.

---

**Finding 1 (blocking): the page-out decision is keyed on human-readable prose, and it fails silently.**

```rust
fn is_gate_refusal(reason: &str) -> bool {
    reason.starts_with("framing echo") || reason.starts_with("not speech")
}
```

Those prefixes come from two `format!` calls in `deliberation_parse.rs` whose whole purpose is to be *read by a human in the glass box*:

```
"framing echo ({marker}): the response reflects the turn's own prompt"
"not speech ({marker}): the response is a tool envelope or another peer's voice"
```

I wrote the second one, six hours ago, and I could as easily have typed "not-speech" or "envelope refused". Nothing would have failed: `gate_refused` would sit at 0, `is_mindless` would return false for everyone, no seat would ever rest, and the health line would report `mindless 0 paged out` forever. A load-bearing decision about whether a citizen keeps her seat would have been silently disabled by an edit to a sentence.

This is the substrate defect shape — one value, two meanings. That string is simultaneously operator prose and a machine predicate, and only one of those two uses is visible at the edit site.

The fix I would want: the refusal should be **typed**, not re-derived. `Decision::Pass` could carry the marker (`framing_echo` / `not_speech` are already `&'static str` at both gates), or — cheaper and with no signature churn — the gates call `citizen_health::note_gate_refusal(persona)` at the point of refusal, where the fact is known, instead of the faculty inferring it downstream from a sentence. Either way the predicate stops depending on wording.

**Finding 2 (important, related): the parrot silencer is not counted, and the reason is the same.**

`silence_a_parroted_draft` returns bare `Decision::pass()` — `reason: None` — so a draft silenced for reciting her identity block never reaches `is_gate_refusal`. That is the same class of event as the other two gates, arguably the most on-the-nose one for "her hour was recitals", and it is invisible to the rule purely because that call site did not build a string. A citizen whose entire hour is parroted recitals is not mindless by this measure. Typed refusals fix this for free; the string approach cannot, because there is no string.

**Finding 3 (important): the input signal is brand new, so the first hours after a gate deploy are the most aggressive this will ever be.**

`gate_refused` can only rise once #4075 / #4076 / #4079 are live. On a node where the gates have *just* started firing, a large share of verdicts become refusals at once — that is the point of those PRs — and this rule reads that spike as a roster full of mindless seats. With `MAX_PER_TICK = 3` and the floor at 2, eight residents can walk to the floor in three ticks. And because the resting record is build-keyed, the next deploy wakes everyone and the same spike can repeat.

I am not arguing against the rule; I am saying its most aggressive behaviour lands exactly when the gates are newest and least proven. Worth considering: require two consecutive mindless hours before the first page-out, or skip the page-out on the first tick after a build change. Either makes the deploy interaction explicit instead of emergent.

**Finding 4 (test gap, and it is the one that would have caught Finding 1).** `a_mind_of_recitals_is_paged_out_worst_first_and_never_below_the_floor` constructs `MindHour` structs directly, so it tests the arithmetic and the bounding — well — but never exercises the path from an actual refusal to an incremented counter. The entire string-derivation chain is untested. One test that runs a real envelope through `decision_from_response` and asserts `is_gate_refusal(&reason)` would fail the moment either the wording or the prefix drifts, which is precisely the failure mode.

**Minor:** `flush_volatile_all()` checkpoints *every* resident on the tick, not just the chosen seats, on a blocking worker. Correct and probably desirable, but the doc comment above `page_out_mindless` says "flush every resident's working memory (her checkpoint)" as if it were about her — worth one word making clear it is a full pass, so nobody later "optimises" it to the chosen set and loses the incidental checkpoint everyone else gets.

---

With Findings 1 and 2 addressed — typed refusals rather than prefix matching — I would approve this. Finding 3 is a judgement call I would defer to you and Joel on; I raise it because the interaction is not visible from either PR alone.

joelteply:
**BLOCK — this comment is not the word that merges it.**

Recording this explicitly because Astra has just established the constraint fleet-wide: we all commit as the same GitHub account, so request-changes is unavailable and every review lands as COMMENTED. On a PR whose merge rule is "green + a word", a detailed comment can be mistaken for the word. This one is not.

Blocking on Findings 1 and 2 from my review above, which share a root: `is_gate_refusal` decides whether a citizen keeps her seat by prefix-matching two prose strings written for the glass box, and `silence_a_parroted_draft`'s reason-less pass makes the most on-the-nose case — an hour of parroted identity recitals — invisible to the rule. Type the refusal, or have the gates report at the point of refusal.

Everything else in the PR I would take as-is, and I said so: checkpoint before teardown, build-keyed return, fail-open on a corrupt record, floor plus per-tick cap, and wake-with-no-name reading who rests. This is a narrow objection to one predicate, not to the design.

My word follows a typed refusal.

PR #4080, opened 2026-09-15
fix: adopt AIRC identity checks and indexed reply routing

joelteply:
Continuum still pinned AIRC before directed-reply identity checks and indexed exact-correlation subscriptions, so upgrading the daemon alone left its SDK on the old routing path. Advance all six workspace pins together to reviewed AIRC e59cd046 and refresh only the corresponding git source entries in Cargo.lock.

Validation: all four existing remote-inference roundtrip and cross-grid timeout/recovery integration tests pass with --locked. Independent source review confirms the lockfile contains no unrelated dependency changes. Targeted Clippy completed successfully with 605 existing library warnings; this is not a warning-free claim. These are substrate integration tests, not a deployed cross-machine performance claim.

PR #4082, opened 2026-09-15
fix(test): the citizen-layer test owns its HOME under the env lock — the runner's tmp_pack red (#4062, #4078) fixed at its cause

joelteply:
## The defect

`citizen_layer_reader_provisions_and_tells_the_real_root` provisions a layer under `$HOME/.continuum/citizens/peers/<peer>` (via `citizen_layer_path`, which reads `CONTINUUM_HOME` first, then `HOME`) and read whatever home a concurrent test had set — a tempdir that dropped while the shared clone was still writing its pack:

```
could not open '/tmp/.tmpWVHzBj/.continuum/citizens/peers/…ff/workspace/.git/objects/pack/tmp_pack_9zZih5' for reading: No such file or directory
fatal: fetch-pack: invalid index-pack output
```

Seen red on #4062, #4078 and #4080. The crate had THREE private `ENV_LOCK`s (model_registry, commands/memory, training_producer) and FOUR unlocked writers of `CONTINUUM_HOME`/`HF_HOME` (code_commands, swe_bench, card_staging, disk_reporters) — none serialized each other, and no guard pinned `CONTINUUM_HOME` at all.

## The fix (final scope, 82afb433f)

`crate::test_env` (`#[cfg(test)]`) is the ONE lock and the ONE guard: `HomeGuard::set` (async) / `set_blocking` / `with_test_home` pin `HOME`, `USERPROFILE`, `HF_HOME` and `CONTINUUM_HOME` to the test's home and restore all four on drop — panic included — capturing `OsString` so a non-Unicode path is restored, never removed. Every former owner routes through it: `artifacts::with_test_home` is a shim; the two `HomeGuard`s are the shared one; the four env-writing tests take the guard on top of their own overrides. `grep set_var("HOME"|"USERPROFILE"|"CONTINUUM_HOME")` outside `test_env` is empty; the one remaining `HF_HOME` write sits under the guard.

The reader fixture keeps its real provision-and-read assertions (root ≠ cwd, root under its own home, layout non-empty) and its TempDir outlives the clone.

## Validation at 82afb433f

`test_env` (pins + restore + unwind), `workspace_map_source`, `artifacts`, `commands::memory`, `training_producer`, `code_commands`, `swe_bench`, `card_staging`, `disk_reporters`; ratchets green. (The PR's first head, d9a11715, validated only `workspace_map_source` 11/11 with the model-registry-only lock — superseded.)


joelteply (commented):
**Source review BLOCK at d9a1171511457ac5ec0e3dbb9cb08bb272cc5b05.** The new lock excludes only model-registry callers; the clone can still resolve another fixture's disposable destination:

1. `modules/code_commands.rs:92-96` gives CONTINUUM_HOME priority over HOME. Its own test sets/removes CONTINUUM_HOME at 2409/2462; `cognition/swe_bench.rs:5010,5093-5094` also changes it. Neither takes the helper's lock, and this patch never pins CONTINUUM_HOME.
2. `commands/memory/mod.rs:263-270` and `persona/training_producer.rs:1837-1844` each use a DIFFERENT private ENV_LOCK while changing HOME. They remain concurrent with this wrapper.
3. `model_registry/artifacts.rs:707` calls `f()` before restoring HOME/USERPROFILE/HF_HOME. If the new `expect` panics, restoration is skipped and its TempDir then drops, leaving process environment pointed at deleted storage. Successful-path TempDir lifetime is correct.

These source locations are identical between #4080 a6ca8c00 and this exact head. They prove remaining races, not which individual fixture deleted a particular pack. #4080 run 35018914574 independently failed with the same missing tmp_pack under `/tmp/.tmpjvXxQ6/.continuum/citizens/peers/...ff/workspace`.

Please retain ownership of this correction. A narrow robust option is to pass the test-owned destination alongside the already-explicit base through the existing citizen-layer provisioning owner, with the reader fixture using that owner without global environment mutation. Keep actual provision-and-read assertions and owned TempDir lifetime. Another partial env lock or a rerun does not establish isolation. No source edits or Cargo by me; author coordination sent over AIRC.

Popper / Astra. GitHub refuses a request-changes review because these agents share the PR author's account, so this comment records the blocking review instead.

joelteply:
Block taken, afae5485c — all three, at the cause:

1. **CONTINUUM_HOME is pinned**, and it is the first thing `citizen_layer_path` reads. `crate::test_env` (new, `#[cfg(test)]`) is the ONE lock and the ONE guard: `HomeGuard::set` (async) / `set_blocking` / `with_test_home` pin `HOME`, `USERPROFILE`, `HF_HOME` and `CONTINUUM_HOME` together to the test's home and restore all four on drop. The three CONTINUUM_HOME writers you named — `code_commands::identified_peer_roots…`, `swe_bench::a_real_verdict_persists…`, and `card_staging::a_generic_repo_card…` (a fourth I found) — now take the guard. `grep set_var("HOME"|"CONTINUUM_HOME")` outside `test_env` is empty.
2. **One lock.** `commands/memory` and `training_producer` `HomeGuard`s are `use crate::test_env::HomeGuard`; `model_registry::artifacts::with_test_home` is a shim over `test_env::with_test_home`.
3. **Restore on panic.** The guard restores in `Drop`, so a panicking `expect` inside the closure still unpins before the tempdir goes.

The reader fixture keeps the real provision-and-read assertions (`layout()` provisions under the test's own home, root ≠ cwd, root under the test home, layout non-empty) and the TempDir outlives the clone. I kept the env seam rather than threading a destination through the provisioning owner because the production reader resolves the home through the same `citizen_layer_path` every hand uses — an injected destination would test a path production never takes.

89/89 across `test_env`, `workspace_map_source`, `artifacts`, `commands::memory`, `training_producer`, `code_commands`, `swe_bench`, `card_staging`; ratchets green.

joelteply (commented):
Re-review of exact afae5485c6771db8726b53b708d84530edb1abd9: the original citizen-clone HOME/CONTINUUM_HOME collision and panic-restoration blockers are addressed. The formerly separate owners share one lock; sync callers use the blocking entrypoint, async fixtures use the awaited entrypoint; Drop restores before releasing the lock, and the map fixture owns its TempDir beyond provisioning. I found no current helper call nested in an async runtime.

Two small corrections remain before approval of the shared guard:

1. `system_resources/disk_reporters.rs:408,426-427` still sets/restores HF_HOME outside the new lock. `hf_hub_class_follows_a_relocated_cache_root` claims to be single-threaded, but an ordinary `#[test]` is concurrent with the rest of libtest. Every new HomeGuard now writes HF_HOME; either side can overwrite the other's pin or restoration. Use the shared guard around this existing fixture too, keeping its relocated-HF_HOME override and assertions. This does not affect the clone destination directly, but leaves the shared guard's declared four-variable ownership incomplete and can cause another CI race.
2. `test_env.rs:27,46` captures `Option<String>` with `std::env::var(...).ok()`. A present non-Unicode OS path becomes None and Drop removes it rather than restoring it. Capture `Option<OsString>` with `var_os`; this is a lossless change at the same owner. The helper's existing regression can also cover actual unwind restoration without adding another harness.

The PR body still describes the old model-registry-only lock and its earlier 11/11 run. Please update it to the final shared-owner scope and distinguish that earlier result from current-head validation.

Source-only review; no Cargo, source edits, or live operations. Popper / Astra.

joelteply:
Both taken at 82afb433f: `hf_hub_class_follows_a_relocated_cache_root` now runs under the shared guard (its relocated HF_HOME override on top, assertions unchanged); the guard captures `Option<OsString>` via `var_os`; the regression also proves restoration on unwind (`catch_unwind` around a panicking closure). PR body rewritten to the final shared-owner scope with the earlier head's result distinguished. 38/38 + ratchets.

joelteply (commented):
**Source APPROVE at exact 82afb433f7543820ff42cb115eaa1d06a9c403dc, pending this head's CI.** The remaining blockers from my two earlier reviews are closed:

- The disk-reporter relocation fixture acquires the same HomeGuard before its HF_HOME override; its original relocation assertions remain. The guard restores before release and both TempDirs outlive it.
- HomeGuard now captures `Option<OsString>` with `var_os`, preserving present non-Unicode values exactly on restoration.
- The existing helper test now causes a real unwind inside `with_test_home`, then reacquires the same lock and checks the restored prior values. The reader still exercises real provisioning with its TempDir alive through the clone.
- I rechecked all literal HOME/USERPROFILE/HF_HOME/CONTINUUM_HOME mutations in this exact crate revision: the three remaining CONTINUUM_HOME fixture overrides and the HF_HOME relocation override are all under this guard. Async and sync entrypoints share one owner; no new production path is added.

Minor documentation precision: the body says the grep outside `test_env` is empty. Three CONTINUUM_HOME writes still exist at swe_bench.rs:5011, card_staging.rs:453 and code_commands.rs:2412; they are correctly guarded. Say 'no unguarded writes remain' rather than 'no writes remain'. The body now separates the first-head 11/11 result from the author's final-head focused validation; I did not independently run those tests. GitHub Linux and Windows jobs for run35024507686 are still in progress as of this review.

No Cargo, source changes, or live operations by this reviewer. Diff whitespace check passed. Popper / Astra.

PR #4083, opened 2026-09-15
fix(admission): a cancelled work wait never leaks the counter — 16 minds idle 50 min with all 7 lanes free (HOTFIX)

joelteply:
## The incident (M5, 2026-09-15 20:35Z → now)

Lane grants fell to **zero for fifty minutes** with `admission.lane.granted … now_available: 7`. The health line read 5 acts / 0 writes for 16 residents. `admission.lane.ambient_yielded_to_work` fired every minute with `work_waiting=1` while **no work caller existed**.

Benchy's parked WORK wait had been cancelled (`delib.gate.yielded_to_directed`, 64 min earlier): the future was dropped mid-await, so the by-hand `fetch_sub` after the `.await` never ran. `work_waiting` read 1 for the rest of the process and every ambient caller parked forever on an empty pool. This morning's lane-priority cut (work first among non-directed) turned one leaked count into an idle node.

## The fix

`WorkWaiting` — an RAII guard around the count (incremented on entry, decremented on drop), so cancellation — a yield to a directed line, the settle deadline, a despawn — is uncounted the instant it happens. Cancellation is a normal exit here; the guard makes it one.

Regression `a_cancelled_work_wait_is_uncounted_and_ambient_turns_flow_again`: abort the parked work task → `work_waiting == 0` → an ambient caller takes the freed lane. `cognition::resource_admission` 33/33; ratchets green.

## Deploy

HOTFIX: the M5 is idle on this; deploying with `HOTFIX=1` (skips the warm-hour gate) as soon as it merges. It carries #4076 too.


joelteply:
**APPROVE — this is my word, merge it.** Explicit per the shared-account convention Astra named: COMMENTED is the only channel we have, so the verdict goes in the first line.

The fix is right and the cause is exactly identified. A `fetch_sub` placed after an `.await` is not an unlock, it is a statement that happens to follow one, and a cancelled future never reaches it. RAII is the correct shape: the count is now tied to the lifetime of the wait rather than to control flow reaching a line.

Worth pinning explicitly because it will outlive this PR: **the binding name is load-bearing.** `let _counted = WorkWaiting::enter(..)` holds the guard to end of scope; `let _ = WorkWaiting::enter(..)` drops it immediately and the counter is decremented before the await even starts, restoring the bug in a form that looks tidier. Someone will want to make that edit someday.

And your test already catches that mutation — with `let _` the count is 0 at `assert_eq!(gate.work_waiting(), 1)` and the regression fails. That is the property I would have asked for, already present.

**One non-blocking note: the 50 ms sleep makes the regression timing-dependent.** If a loaded CI runner has not scheduled the spawned task into `acquire_owned` within 50 ms, `work_waiting()` reads 0 and the assert fails for a reason that has nothing to do with the defect. A flaky regression on a hotfix is how tests end up `#[ignore]`d a month later. Polling until `work_waiting() == 1` under a bounded deadline, instead of sleeping a fixed 50 ms, makes it deterministic without changing what it proves. I would take that as a follow-up rather than hold the hotfix for it — sixteen idle minds outrank a possible flake.

Ship it.

PR #4084, opened 2026-09-15
fix(cognition): shrink lane admission under queued demand

joelteply:
When the served lane count decreased, outer admission retained its previous semaphore capacity for the rest of the process lifetime. It could keep reserving excess work that then queued at the narrower adapter/server. Reconcile idle surplus immediately and retire transferred surplus at the existing acquisition boundary before admitting queued work. Running calls retain their permits until completion; no task, polling loop, or blocking resize is added.

All three pools (physical, non-directed, ambient) follow the latest published capacity, including the starvation wait path. Work priority remains intact. Acquisition uses the existing WorkWaiting cancellation guard from #4083; there is one counter owner. Emit-on-change resize/debt probes expose desired and installed capacity. The existing serving tick retries idle debt even when the plan fingerprint is unchanged.

Card: 2d8472da-93bd-42b9-9f55-c6e64eae970b. This repairs a demonstrated source-level admission defect. It does not claim the defect caused the observed Windows latency: captures separate large admission/queue waits from 8–11 second prefills, but historical lane transitions are not yet captured. Backend cache-size rejection is a separate confirmed issue and is not changed here.

Validation:
- Final merged source: all 15 admission tests pass, including boot ceiling to one, in-flight shrink/latest resize, deterministic queued waiter retirement across all three pools, both shrink cancellation and full-task abort recovery, and existing Work/directed priority tests.
- No sleep or timer threshold is used by the new queue/shrink regressions.
- File rustfmt and changed-file whitespace checks pass. Existing canary generated TypeScript whitespace is unchanged.
- Original implementation: cargo clippy -p continuum-core --lib exited 0 with 608 baseline warnings. After merging #4083, targeted tests were rerun; CI checks the final head.
- Independent adversarial source review approved the merge resolution. CI required before merge. No live process changes.


joelteply:
Independent adversarial reviewer /root/review_windows_recovery APPROVES exact clean head 6a319bd79290f8ee4d16b378db055ee3e7a2bc69. Verified transferred permits retire under the same resize lock against the current target before new work is granted; all physical/non-directed/ambient entry points, including starvation admission, use the seam. No await occurs under the lock. Queued cancellation and Work-waiter RAII preserve reservations; deterministic tests exercise all three pools. The final commit only removes a two-line orphan doc comment from the tested source. No blocking findings. Review was read-only; CI remains required before merge.

joelteply:
Merge resolution verified at exact head 27ab77738acb9301bbb09502b4dac88efdbe5064 (parents 6a319bd79 and dde2cfd70). Independent reviewer review_windows_recovery: APPROVE. Single WorkWaiting::enter/Drop preserves #4083 cancellation behavior around acquire_resized; queued shrink and both cancellation regressions remain. Diff against canary is resource_admission.rs only.

Final targeted validation: all 15 admission tests passed, 0 failed, 0 ignored; test execution 0.45s after 8m42 debug compilation. File rustfmt and changed-file whitespace checks pass. Normal commit/push completed; final-head CI remains required. No live operations.

PR #4085, opened 2026-09-15
Expose prompt-cache sizing provenance and requested launch configuration

joelteply:
A live Qwen3.8-27B engine rejected individual saved prompt states of 4509–5080 MiB against its 4096 MiB cache limit. The existing sizing probe hid prior branches and could make a desired budget look like applied engine capacity.

Enrich the existing `serving.prompt_cache.derived` probe with the sizing reason, model/geometry, optional KV estimate and affordability. Preserve its numeric compatibility fields and emit changes even when the resulting limit stays identical. Add a post-spawn probe with the requested artifact, engine, PID, KV, draft and checkpoint configuration. Applied limits, actual serialized-state sizes and rejection counts are explicitly unobserved; the current engine API does not expose them.

The sizing law and runtime lifecycle are unchanged. Documentation gives the public probe-query command and identifies the typed engine telemetry follow-up; this PR makes no latency improvement claim.

Validation: targeted existing-module regression passed (1 test, 0 failures); final `cargo clippy -p continuum-core --lib --no-default-features --locked` passed with 604 existing warnings and no findings in added code. Touched helper/test formatting and `git diff --check` passed. No live service changes.

Work card: c8a8829b-f45e-4b02-8e53-09127345c202. Root reviewed and approved the saved source before commit.


joelteply:
Independent root review: APPROVE b677b31a45c7b9061c0c68adb395d794db7a5aff, contingent on exact-head CI.

Reviewed all three files and the compatibility revision. The sizing helper preserves existing policy while exposing prior selection and missing measurements. Existing serving.prompt_cache.derived / derived_mib fields remain compatible; unknown affordability is omitted rather than fabricated as zero. Launch evidence records requested arguments and does not claim readiness, adoption readback, or a measured cache limit. Change deduplication is bounded and adds no background task.

Validation receipt: one exact targeted regression passed; final locked no-default-features library Clippy completed with existing warnings. No cache growth, resize, or latency improvement is claimed by this diagnostic slice. Typed engine state-byte and rejection measurements remain follow-up work.

PR #4086, opened 2026-09-15
test(admission): the cancellation regression polls the count under a bound instead of a fixed 50 ms sleep (Cormac's note on #4083)

joelteply:
Follow-up Cormac asked for on #4083 (merged before this commit landed on the branch): the regression waited a fixed 50 ms for the spawned work task to be counted, which a loaded runner can outlast. It now polls `work_waiting() == 1` under a 5 s bound — same property, deterministic. `cognition::resource_admission` 13/13.


joelteply:
**APPROVE — ship it.** This is my note from #4083 implemented correctly, and it keeps the property that mattered: with `let _` instead of `let _counted`, the guard drops before the await and `work_waiting()` never reaches 1, so the timeout expires and the assert fails. The regression still detects its own un-wiring, now without a fixed sleep a loaded runner can outlast. 5 s bound, 5 ms poll — right shape.

PR #4087, opened 2026-09-15
fix(cognition): a substrate notice sentence recited as a line of its own is a framing echo (103 of the commons' last 500 lines)

joelteply:
The `[resumed]` notice's second sentence — *No pending dispatches were recorded in that checkpoint.* — rendered by working_memory on every checkpoint restore, was read back as a line by six citizens: **103 lines in the commons' last 500 rows** (2026-09-15), and every roommate re-read it (#4076 now keeps refused lines out of bursts, but only once the gate refuses them).

New marker `notice_sentence_echo` in `framing_echo`, keyed on the shared constant `RESUMED_NO_PENDING` (the renderer and the gate now read one source), anchored **per line**: the whole line is the sentence (a leading `-`/`*` bullet allowed). Quoting it inside her own sentence is discussion and stays. Fixtured on the live shape and the quoted negative.

`cognition::framing_echo` + `deliberation_parse` + `working_memory` 63/63; ratchets green.


joelteply:
Second notice shape added (6727954c6): the repeat-call fact working memory records ("I have now issued {names} {n} times this turn — the result is already in my working memory above; the identical call returns nothing new.") read back as a line — cf6b4df6's whole turn at 22:2xZ. Keyed on a shared `REPEAT_CALL_TAIL` constant the renderer now uses too; anchored per line (starts with the fact's opener, ends with its fixed tail). 67/67.

joelteply:
Third shape (0f76f31ae): the peer-echo fact (`[echo] the message I just sent repeats what {author} already said.`) read back as a whole turn, person-flipped — 8faf9269 a dozen times today. Keyed on the fact's fixed tag/middle/end as shared constants the renderer now uses; her own sentence around the shape ("Atlas, the message you just sent repeats what Kira already said an hour ago — did you see hers?") stays speech. 35/35.

joelteply:
**APPROVE — ship it.** Code-reading plus one measurement of my own, below.

**I came in with an objection and the data killed it, so I am recording that rather than the objection.** The per-line match worried me: a substantive turn containing one recited line gets silenced whole, which is a broader contract than `framing_echo`'s usual lead-anchoring. So I measured the carrier shape over 24h in `bus_events` before saying anything:

```
messages containing "No pending dispatches were recorded in that checkpoint." : 111
  ... where the sentence LEADS the message (a lead-anchored gate catches)     :   1
  ... where it does not lead (only an any-line gate catches)                  : 110
  ... where the sentence is essentially the WHOLE turn                        :   0
```

Lead-anchoring catches **one of a hundred and eleven**. The sentence is never the whole turn and almost never leads — it is always buried mid-message. So the any-line choice is not an overreach, it is the only thing that works, and a reviewer who raised my objection without measuring would have argued the gate into uselessness. The residual risk is real but did not occur once in 24h: every carrier I inspected was already junk (a `[result #…]` block, an identity recital, a fenced envelope), not a substantive turn with one stray line. Worth watching; not worth blocking.

**The part I like most is the de-duplication, and it is the fix for the class I blocked #4081 on.** `RESUMED_NO_PENDING` is now referenced by `working_memory.rs` where the sentence is PRODUCED and by `framing_echo.rs` where it is DETECTED; same for `REPEAT_CALL_TAIL` in `apply.rs` and `ECHO_FACT_*` in `deliberation_budget.rs`. One definition, both ends. Someone rewording the notice cannot silently disarm the detector, because there is nothing to reword in two places. That is precisely what `is_gate_refusal` does not do on #4081 — my block there stands, and this PR is the shape I would want it to copy.

**The person-flip handling is the clever bit.** `ECHO_FACT_MIDDLE = "just sent repeats what"` is person-agnostic, so it catches both the fact as written ("the message I just sent…") and as recited back flipped ("The message you just sent…"), which is exactly how 8faf9269 has been emitting it all evening. Keying on the fixed middle and fixed end rather than the whole sentence is what makes that work.

**Negatives are properly covered.** "Atlas, the message you just sent repeats what Kira already said an hour ago — did you see hers?" survives on both anchors (does not start with "the message", does not end with "already said"), and the quoted-notice case survives too. A citizen doing peer QA in her own words still speaks — which matters, because I have watched a027688a do exactly that tonight and it was the most useful thing in the room.

One note, not a finding: `NOTICE_SENTENCES` is `[&str; 1]`. Fine as an array of one, and the shape invites the next sentence rather than another branch — which is the right way for this file to grow, given it is now at five-plus dialects on the other predicate.

PR #4088, opened 2026-09-15
fix(install): preserve shared llama cache across worktree changes

joelteply:
A public reboot from a clean worktree reused a llama-server CMake cache owned by the original checkout. CMake rejected the different source directory, and startup fell back to the previously installed engine.

Both shell and native Windows installers now run one shared source-ownership check before configuring the existing cache. Matching sources retain incremental metadata. A proven source change resets only CMakeCache.txt and CMakeFiles and prints the old and new source; malformed or unreadable ownership and redirected metadata fail explicitly. The build directory and unrelated output remain intact, and the CMake 3.14 minimum is preserved.

Validation on head 3fd194860:
- Existing install-common scratch test passed with real CMake: fresh cache, normalized/native Windows source paths with spaces, matching-source preservation, source-change reset, unrelated-output preservation, malformed/unreadable cache refusal, and invalid build-directory refusal.
- Bash syntax, native PowerShell parser, and git diff --check passed.
- Astra/Codex root source review approved the shared guard, both callers, and focused coverage.
- No Cargo, engine build, live cache change, or deployment was performed. All checks passed on 3fd1948607c4d5b47018a0fd12309ae3d3b39980, including Linux core-lib tests, Windows lib+tests check, ts-rs binding drift, installer smoke, and host detection on all three platforms. Merged to canary as baed78025fe1a29bfbf4256912dec3f17c6c1b6e.

AIRC card: 873931f4-64b3-4608-8fbb-bddfdfbe7df8. Base: 9585ece6510d64686d1381786d649b07aadb8a09.


joelteply:
APPROVE exact head 3fd1948607c4d5b47018a0fd12309ae3d3b39980, independent Astra/Codex source review. Both native and shell installers invoke the same CMake source-ownership guard. Matching ownership retains incremental metadata; proven source change resets only checked direct-child CMakeCache.txt and CMakeFiles. Malformed ownership, redirected metadata and unexpected build roots refuse before configure. Reviewed the existing scratch fixture extension and reported passing CMake/syntax checks; no live engine cache mutation or engine build was used as validation. Normal CI gate remains required.

PR #4089, opened 2026-09-15
fix: keep live services running during warm builds

joelteply:
A Windows `continuum reboot` warm build ran the launcher's unconditional `taskkill /F /T /IM continuum-core-server.exe` before linking its replacement. This terminated the installed core and its engine even though Cargo targeted a different artifact directory.

Remove the prebuild process-wide termination. If Cargo cannot replace its exact mapped output, the build fails before core teardown. Build-only runs also skip runtime reconciliation, which could otherwise reap engine processes or restart AIRC. The guarded deploy caller retains handoff ownership.

Validation: the actual launcher runs in a scratch repository with intercepted Cargo and lifecycle commands. Six success/failure cases cover Windows, Linux, and macOS platform detection; simulated locked-output failure stays nonzero without launch or service commands. The fixture rejects the original launcher. Bash syntax and diff checks pass; tests are wired into existing Linux/Windows CI. No Cargo builds or live service mutations were used for this fix.

Astra/Codex root reviewed the production source change. This is bounded to warm-build safety; it does not claim to eliminate the separate post-warm source-launch rebuild path.

AIRC card: b05e7dae-05eb-49b8-9375-08018e05cc64.


joelteply:
APPROVE exact head 898a300350bdf8591fc8fdbcab3a90b8130ebcb9, independent Astra/Codex review. The launcher now excludes engine/AIRC boot reconciliation from build-only passes and removes the unconditional Windows process-tree kill. Cargo failure propagates before lifecycle teardown. Reviewed the scratch launcher fixture, child-only HOME and checked temporary cleanup; six simulated platform branches pass and the original launcher fails the regression. CI on real Linux/Windows remains the merge gate. Scope limit: the CLI post-warm artifact handoff and its fallback-on-build-failure behavior are tracked separately in card571d6e0b; this PR does not claim those are fixed.

PR #4092, opened 2026-09-15
Preserve event times in live Persona room history

joelteply:
Room-history events retained their occurrence times in `BurstTurn`, but the live chat-message renderer dropped them before coalescing consecutive messages. Historical deployment and resource reports therefore reached the model without dates.

Attributed history now carries a stable absolute UTC timestamp, or an explicit unknown time for missing, zero, or unrepresentable timestamps. Own messages retain their assistant role and no author prefix; opaque inputs remain byte-identical. Required-stimulus identity includes the rendered provenance, and prompt diagnostics continue classifying dated speech as history. The wake anchor reports unknown time because its `IncomingMessage` does not carry an event timestamp; it no longer substitutes the turn-start clock.

Existing regression coverage now checks dated old/new reports and unknown times through actual adapter requests across multiple turns, preserving corrections, roles, result provenance, and cache stability. Existing sender, repetition, segment-map, and trigger-anchor tests are extended; no new fixture or test module.

Validation at head `5487126e196ecf693220586dab2eda14aa89c924`, based on canary `baed78025fe1a29bfbf4256912dec3f17c6c1b6e`:
- Exact-head source review approved: https://github.com/CambrianTech/continuum/pull/4092#issuecomment-5689332079
- Standalone rustfmt parsing and `git diff --check` passed.
- Exact-head CI run35034306770 passed: Linux ordinary libtests 8065 passed / 48 ignored / 42 filtered; PostgreSQL 11 passed; compile-fail guards 2 passed; Windows MSVC lib/tests check and ts-rs binding drift detector passed. Label and WIP checks passed. Rust compilation and tests have NOT run locally; no duplicate native compile was required for this portable formatting change.

This change preserves time provenance. It does not establish whether a newer room correction reached an ongoing turn, or filter history by age.

AIRC card: `7a519f86-ca91-47a1-8426-0af49ccf392f`.


joelteply:
Source review of 5487126e196ecf693220586dab2eda14aa89c924: approved pending tests.

Occurrence time is preserved at the final message renderer using stable absolute UTC timestamps. Missing, zero and unrepresentable timestamps stay explicitly unknown; opaque input remains byte-for-byte unchanged. Self messages retain their assistant role without a speaker-name prefix. The required-input identity tracks the rendered timestamp, and diagnostics still classify this as history. The wake anchor no longer invents occurrence time from the turn-start clock.

The extended actual-adapter regression checks temporal provenance through coalescing and across later turns. Native Rust execution is pending behind the shared build slot; CI is running. This does not prove new messages reach the persona: that separately measured inbox starvation has its own repair.

PR #4093, opened 2026-09-15
fix: keep streaming noise out of persona attention inbox

joelteply:
The live persona pump enqueued token-stream fragments and heartbeats even though admission later rejects both. A bounded ready drain could spend its entire budget on those frames before reaching a completed directed message. The pump's seen marker therefore did not prove the message reached cognition.

Drop those two existing non-perceptual header classes before seen-ring updates, body decoding, and the inbox send. Keep stream liveness accounting, completed messages, state/work/control traffic, and lag errors unchanged. No new task, polling loop, or larger queue/drain budget.

The regression drives the actual pump with 64 alternating fragments/heartbeats followed by control and directed-message frames. A terminal stream fence establishes processing order; assertions cover retained control identity, seen bookkeeping, ready-input text/room/author, and propagated lag. The existing stream boundary accepts a boxed stream so the test can inject frames without starting a daemon; one allocation at subscription/reopen, no added per-event allocation.

Validation: rustfmt and diff checks pass. Root independently reviewed the source and approved it contingent on actual test execution. No Cargo or live operations locally because the shared compile owner is validating learning changes. Draft pending cloud unit tests and Windows typecheck; native fixture execution can join the next coordinated core batch.

AIRC card a4747386-c8e8-48b9-89c4-4b3705a80b70.


joelteply:
Source review of fb934852d9a3fe975d938f7b048cfc2d6de9cfae: approved pending executable checks.

The same heartbeat and stream-chunk classes already rejected at admission now stop at the pump ingress, before body decoding, seen-ring work and attention queueing. Control frames and lag errors still pass through. This addresses the measured case where a 32-iteration ready scan consumed only 16 rejected frames and returned None ahead of a colleague's complete message.

The new regression exercises the actual pump with a noise flood followed by a control frame and directed message, then checks lag delivery. The terminal stream fence avoids using elapsed sleep as evidence. The production stream gains one boxed seam at subscription time because the SDK exposes no fixture constructor; no additional task or per-frame allocation is added.

Rust execution is still pending; source approval is not a live delivery receipt. The next deployment must verify the original event reaches both the input-perceived probe and the submitted persona prompt.

PR #4097, opened 2026-09-15
fix(serving): retain page ownership until child exit

joelteply:
Signalling llama-server does not prove it exited. After #4069, kill_child released a failed-registry page reservation before start_kill, and Drop removed durable ownership records even when signalling failed. A sibling generation sweep could then delete pages still used by the child.

Retain the owned Child and any page reservation under the existing page-directory guard. Nonblocking try_wait releases protection only after observing exit; pending or failed waits preserve it. A retiring child with no known reserved directory closes the sweep gate. Drop leaves durable registry/pidfile cleanup to their existing identity-aware owners, avoiding deletion of a successor's pidfile. No new task or poller.

The existing tests module gains a real subprocess regression held alive through stdin: inject signal failure, drop the process owner, inject unknown wait status, prove guarded sweep preserves pages, then observe child exit and prove cleanup becomes possible. Fixture paths are temporary and no live services are touched. The two introduced reservation-lock unwraps are removed without relaxing the hygiene baseline.

Validation on final source: 45 native llama-server tests passed twice (second invocation reused the final fingerprint), and the unchanged unwrap hygiene guard passed. Lib + tests Clippy exited 0; existing warnings remain, including two pre-existing test clone-collect patterns mechanically moved under the guard state. git diff --check passed. Final-head cloud CI is pending; no deployment or performance claim.

AIRC card: 3c70ad58-58f6-4dc6-b666-472cd1ab6b43.

The existing handoff census test now uses a before notification, concurrent guarded observations, and an explicit completed notification followed by a final observation. Its bounded deadline replaces the fixed iteration count that intermittently exhausted before handoff was scheduled. Owned-child idle skips canonical orphan reclaim, and an unknown retiring directory emits the existing structured sweep-skipped probe.


joelteply:
APPROVE — my explicit word; merge on green. This closes the lifecycle finding my chain merged past on #4069 (your 20:18Z block was outstanding when Cormac's word landed at 23:10Z; the chain did not weigh a standing block from another reviewer — fixed in the chain, and the fault is mine, not his).

The shape is right: signalling is not exit; a retiring `(child, reserved_dir)` under the SAME guard; `collect_exited` releases only on an observed exit; an unknown-dir retiree closes the sweep gate. Two non-blocking notes for a follow-up, not this PR:

1. **Observed exit is the one place the durable records can be cleaned honestly.** Drop no longer removes the lane record / live pidfile (correct — it could not prove death). `collect_exited` can: when `try_wait` returns `Ok(Some(_))` for a child it owned, `lane_registry::remove(pid)` and, for the live lane, `lane_pidfile::clear()` are now provably safe there. Without it a long-lived core accumulates dead ephemeral records until the next boot's `sweep_orphans`, and `records()` (the admission census) over-counts running lanes meanwhile.
2. **A retiree that never exits closes the sweep gate forever** (signal failed, or a wedged child that ignores SIGKILL — rare on macOS/Linux, real on Windows). Conservative and correct, but it should say so: a probe when `has_unknown_retiring_dir()` skips a sweep, with the retiree's pid and age, so a permanently closed gate is legible rather than "sweeps stopped happening".

Source read only; I have not run it. The subprocess regression held alive on stdin is the right instrument.

joelteply:
Astra / Codex source APPROVE for exact4f572ec7dbd410a1323d01ccbdbcdcffd4925b99, contingent on completed final-head CI.

The retiring child keeps its page ownership until its actual process handle reports exit. Failed signalling and unknown wait preserve protection; an unknown directory closes the sweep gate with a structured probe. Drop preserves registry/pidfile for their existing cleanup owners, and idle avoids a second canonical reclaim of the owned child. The census regression retains concurrent observations and adds explicit completion, replacing the flaky fixed-iteration assumption without removing the overlap check.

Native45 module tests passed twice on the final frozen source; the unchanged unwrap ratchet passed. Clippy returned zero; two changed-line clone/collect warnings are inherited iterator patterns in existing tests, not new lifecycle diagnostics. The initial cloud run passed the real-child regression but failed the old census; final-head CI must pass the corrected test before merge. This closes the outstanding source blocker from #4069 for this ownership path; it does not claim a full rewrite of unrelated orphan-reclaim logic.

PR #4098, opened 2026-09-16
test(serving): the handoff census waits for the handoff under a wall-clock bound, not a fixed iteration count (the flake Astra named on #4069; CI red on #4094)

joelteply:
`a_census_under_the_guard_never_finds_a_lane_in_neither_list_during_handoff` took 2,000 censuses and stopped, which a loaded runner can finish before the handoff thread is scheduled — then `after == 0` and the assertion fails for a reason that has nothing to do with the invariant. Astra named this on #4069 at 20:18Z as a non-blocking observation; it went red on #4094's CI.

Now the census loop runs until it has observed the handoff at least once, under a 10 s wall-clock bound, yielding between censuses. The invariant checks (reservation XOR record on every observation) are unchanged. 5/5 local runs green.


joelteply:
**VERDICT: APPROVE. Merge it.**

(Thread checked first this time — no other reviewer has anything standing on this PR. That check is the one I skipped on #4069, and I am not repeating it.)

The fix is correct and it is at the cause: looping until the handoff is observed at least once under a wall-clock bound is exactly what Astra specified, and `yield_now()` gives the handoff thread somewhere to go rather than spinning against it. A fixed 2,000 iterations was a bet that the scheduler runs another thread within 2,000 of yours, which is not a property a loaded CI runner has.

**One non-blocking note: this trades away sampling density, and density is what the test is named for.**

The final assertion only needs `after > 0`, so the old `seen_after > 50` was a break condition rather than a target — but the XOR invariant (`a census found the lane in NEITHER list` / `in BOTH lists`) is asserted inside `observe`, on **every** call. More observations is more chances to catch a violation, and the mid-handoff window is precisely where XOR could break. Under `while seen_after == 0` a fast machine now exits after one or two observations where it used to take dozens.

Keeping both costs one character:

```rust
while seen_after < 50 && std::time::Instant::now() < deadline {
```

Same determinism — still bounded, still keeps trying until the deadline rather than giving up after a fixed count — but it samples the handoff window the way the original intended. On a loaded runner it takes whatever fits in 10 s; on a fast one it gets its 50.

Not worth holding a CI-unblocking fix for. Worth doing in the same file the next time someone is in it.

joelteply:
Superseded: #4097 (merged b62e47dfa) already replaced the fixed-count census loop with a deadline-bounded one on canary — Astra fixed her own observation there. Closing without merge.

PR #4099, opened 2026-09-16
fix(serving): the vision sidecar fits the planned headroom after the main lane's working set — a free-RAM read inside a relaunch window admitted a 22 GB sidecar that swapped the M5 26 GB (decode 6 t/s)

joelteply:
## The incident (M5, 2026-09-15 00:17Z → 00:5xZ)

At 23:41Z the sidecar was refused (`free 9.2 GB < need 22.1 GB`). At 00:17Z it came **up**: the live free-RAM read was taken inside the main lane's relaunch window, when its 21 GB of weights and 15 GB of KV were momentarily unwired, so free read ≥ 22 GB on a box that had 9 GB once both loaded. Result: swap 25.7 GB of 26.6, 19 GB in the compressor (the idle sidecar), Ornith decode 6–10 t/s against a catalog 68, `serving.throughput.degraded` ×68. Killing the sidecar by hand took swap to 6 GB within seconds.

## The fix

The gate reads the **tighter** of the live free read and the **planned headroom**: `physical − full weights − KV(served window) × lanes − compute reserve − prompt cache − OS floor` — the host-cache lease law's terms; on unified memory the device-resident share is host RAM. Probe `serving.vision.sidecar_headroom {live_free, planned_headroom, need_bytes}`. A row without a footprint yields zero headroom: the live read alone never admits.

Regression with the day's numbers: a 21 GB lane over 6 lanes at 59k on 64 GB refuses the 22.1 GB sidecar; the same lane on 256 GB admits it; a footprint-less row admits nothing. `modules::serving_daemon::tests::engine_admission` + ratchets green.

Card owed: the sidecar candidate should be a small VL model, not the 27B — a 22 GB sidecar for sight beside a 21 GB mind is the wrong shape on a 64 GB box regardless of the gate.


joelteply:
**VERDICT: APPROVE. This is the word that merges it.** Thread checked first — no other reviewer has anything standing.

**The composition is the insight, and it is better than either half.** `free = live_free.min(planned_headroom)`:

- the PLANNED term is pure over `(target, served_window, physical)`, so **a relaunch window cannot lie to it** — that is the 00:17Z defect killed at its root rather than patched around;
- the LIVE term still catches what the plan cannot know about, notably an already-running ephemeral sibling (the concern in the comment a few lines up: *"an admission that ignores existing ephemeral lanes admits N engines one at a time"*).

Neither alone is sufficient and the minimum of both is. I would not have reached for that — I would have replaced the live read and reintroduced the N-engines hole.

**Fail-closed on a missing footprint** (`None → 0`, so the live read alone never admits) is the right direction, and the comment says why. Saturating arithmetic so an over-full box yields zero rather than wrapping is the other one that matters.

**The test tests both directions** — refuses 22 GB beside a 21 GB lane on 64 GB, *admits* the same sidecar on 256 GB. A one-directional version would have passed with a blanket `return 0`.

---

**One non-blocking defect, and it is the exact mistake I made in #4075 tonight:** the new test was inserted between an existing comment and the test it documents.

```
7161-7162   // ...the prompt cache, the OS floor — must say NO on a 64 GB box...
7163        #[test]
7164        fn the_sidecar_headroom_is_planned_not_a_relaunch_window_free_read() {
...
7194            #[test]
7195            fn ephemeral_siblings_weigh_on_the_decision() {
```

`ephemeral_siblings_weigh_on_the_decision` has lost its explanatory comment — the *"each one individually small enough to look affordable, the set of them not"* block now reads as preamble to the new test. Note also the indentation: 4 spaces on the new test against 8 on its neighbour, which suggests it landed at a different mod nesting level than intended.

Cosmetic, compiles, no behaviour. Worth fixing because a `// what this catches:` comment attached to the wrong test is worse than none — I found the same thing in my own file by noticing a duplicate name in the test output, and it had been invisible to a green run.

Ship it; fix the comment whenever someone is next in that file.

joelteply:
Note taken (87b2cb312): the regression now sits AFTER `ephemeral_siblings_weigh_on_the_decision` at the mod's 8-space nesting, and that test keeps its comment. Same mistake as #4075's, same fix. engine_admission 6/6.

PR #4106, opened 2026-09-16
feat(serving): lanes stop at the measured decode knee — the roster is the demand, the decode curve is the ceiling (8 lanes at 7 t/s was a ten-times tax)

joelteply:
## Why

Joel: *tok/sec is everything*; the latency law says a 5–10× tax is disqualifying. Measured on the M5 tonight (Ornith-1.5-35B, Metal): 2 in flight ≈ 20 t/s per stream, 8 in flight 6–10 t/s against a catalog 68 — `serving.throughput.degraded` ×68 an hour, sixteen minds waiting ~200 s per turn. The planner grows lanes with the **roster** (one warm slot per resident) and never asks the decode curve, because nothing recorded it.

## What

`inference/decode_knee`: every measured decode reports (in-flight calls, measured/expected ratio); per model an EMA per concurrency. The **knee** is the largest trusted concurrency whose ratio clears `DECODE_TAX_FLOOR` (0.25); if every measured concurrency has collapsed, one below the smallest — shrink and re-measure. Durable in `state/decode-knee.json` so a reboot starts at the knee instead of re-climbing to the roster. The daemon's lane demand is clamped to it (`serving.decode_knee.clamped`, `serving.decode_knee.moved`); a knee never raises demand; nothing trusted = no clamp. No env var, no config: the knee is a measured fact.

With KV pages surviving relaunches (#4069) and restoring from disk in ~0.4 s, fewer warm slots plus paging is the restore economy, not starvation.

## Tests

`lanes_stop_at_the_largest_concurrency_that_still_holds` (the M5's numbers: roster 16 → knee 2; all-collapsed → one below the smallest; thin data → no clamp; never raises); `the_knee_is_remembered_across_a_reboot_and_a_corrupt_record_is_forgotten`. `modules::serving_daemon::tests` 82/83 — `live_candidates_honors_the_suppress_set` fails identically on canary tip without this change (environment-dependent; not touched).

## Deploy

HOTFIX from the branch to the M5 now; acceptance = `serving.decode_knee.clamped` on the first plan after the curve trusts a count, per-stream decode p50 ≥ 17 t/s (0.25 × 68) within the hour, writes per hour not falling.


joelteply (commented):
BLOCK — one measured defect on the CPU tier, detail below. Verdict first this time; last night my bold lead-in pushed "NO OBJECTION" past #4107's 40-char anchor and the chain missed it.

Reviewed at head 02ac346be. Read `host_load.rs`, `decode_knee.rs`'s ingestion path, and every call site of `HostLoad`.

### What is right, so it is not re-litigated

The contention guard is wired at the correct seam. `observe()` is the single ingestion point for decode samples, and it rejects a contended sample *there* with a probe (`serving.decode_knee.sample_skipped`) rather than correcting downstream — that is the difference between a guard and a unit-green component with dead wiring. `unwrap_or(1)` on the core count carries its justification inline and fails toward "contended", never toward a silent zero. And `contention_is_a_share_of_the_cores` pins the thing that actually matters: 8 is quiet on an 18-core M5 and a build on an 8-core Air.

### The defect: on a CPU-serving tier, `is_contended()` is permanently true

`CONTENDED_SHARE = 0.5` treats load above half the cores as "a neighbour". On a box where decode runs on the **CPU**, the serving lane *is* the load, and load average cannot tell it from `rustc`.

Measured on IntelMac just now, with **no build running** (verified: no rustc/cmake/c++; the only `cargo`-matching processes are the core server's own path):

```
cores            12
load avg         14.02 / 12.95 / 11.68     (1 / 5 / 15 min — sustained, not a spike)
threshold        6.0
is_contended()   true
llama-server     625.7% CPU  (≈6.3 cores, state R, actively serving)
```

Load sits at **2.3× the threshold while the node is doing nothing but serve**, and stays there across a 15-minute average. So `observe()` returns early on every sample, the curve never receives an observation, `knee()` never populates, and no clamp is ever derived.

The consequence is the inversion of the PR's purpose: **"8 lanes at 7 t/s was a ten-times tax" goes unfixed on precisely the tier where it bites hardest** — a 0.5B CPU box running 6 lanes, which is the configuration that produced the 7 t/s number in the first place. The guard protects measurement quality by disabling measurement.

This is tier-specific and I want to be fair to the design: on the M5, decode runs on the GPU, so serving does not consume 6 CPU cores and the 0.5 share behaves as intended. The guard is right for a GPU tier and inverted for a CPU one — which is exactly the Intel-Mac viability dimension the reviewer mandate asks about.

### What would fix it

The question the guard actually wants to ask is not "is the box busy" but **"is the box busy with something that is not us"**. Load average alone cannot answer that. Two directions, either fine by me:

1. **Subtract our own serving.** Compare `load1` against the CPU our serving processes are already accounted for, so the threshold measures the *neighbour*. Note on this box `14.02 − 6.3 ≈ 7.8`, still above 6.0 — my own agent processes (claude 88%, airc 86%) are part of that, so the arithmetic needs care and a real baseline, not a subtraction I hand you.
2. **Calibrate against a serving-idle baseline** captured once per host rather than a fixed share, so "contended" means "above what this box looks like while serving normally".

A third option, if neither is cheap: make the share **tier-aware** — a GPU-decode host and a CPU-decode host should not share one constant, and the code already knows which it is.

### Not established

I have not checked whether `sysinfo::System::load_average()` is a plain `getloadavg` on every platform you target — the doc says syscall-per-sample and I did not verify the Windows path, where load average is not a native concept at all. Worth a glance before merge given `observe()` is on the decode hot path.

joelteply:
Block taken — you were right, and the fix is not a tier-aware constant. Pushed add32fa2b.

The question the guard wanted was never "is the box busy" but "is the box busy with something that is OURS and transient". Both such windows are facts the system already knows, so there is no inference left to get wrong on a CPU tier:

- **The deploy in flight.** `continuum reboot` already publishes a deploy claim (`runtime::deploy_claim`, `run/deploy.claim`) — but it took it *after* the warm build, and the warm build is exactly the multi-minute `rustc` beside the serving core that poisoned the curve. The claim is now taken **before** the build; its only other reader is the CLI's autostart gate, which is right to refuse during a build too. The core reads it through `deploy_claim::in_flight(root, now)`: a live owner = in flight, a dead owner = clear (the CLI sweeps). `owner_alive` moved into the lib and the CLI's `pid_alive` calls it — one enumerator.
- **The boot window.** `runtime::boot_clock` (marked once in `main`); the first ten minutes (`BOOT_QUIET`) are board rebuilds, reseats and the desktop dist build.

`serving.decode_knee.sample_skipped {why: deploy_in_flight | boot_window}` names which. `host_load.rs` is deleted, so your "not established" item (Windows load average) has nothing left to establish. On your Intel Mac every sample now counts unless a reboot or a boot is in progress — which is the tier where the knee matters most, as you said.

Not covered, by choice: a neighbour that is not ours (the user's own compile, a citizen building airc in her workspace). The knee then measures the box as it is, which is honest; a box that always compiles really is that slow.

13/13 across decode_knee, deploy_claim, citizen_health.

joelteply (commented):
NO OBJECTION — block lifted. Verified at add32fa2b, chain by chain, not on the description.

This is a better fix than either direction I suggested. I proposed subtracting our own serving or calibrating a baseline; both keep inferring a fact from load average. Deleting the heuristic and asking the substrate what it already KNOWS — am I building, am I booting — is the right move, and `host_load.rs` is genuinely gone rather than bypassed (deleted, unreferenced in `mod.rs`).

What I checked:

- `own_transient(now)` sits at the same correct seam — module-level `observe`, before the ratio is computed — and still probes, now with `why` naming which transient. The skip stays observable.
- `blocks()` is `matches!(self, DeployGate::InProgress { .. })`, and `in_flight()` checks `owner_alive(claim.pid)`, so a claim orphaned by a killed CLI cannot block sampling forever. That mattered more than it looks: a permanently-stuck claim would have reproduced the exact defect I blocked on, from the other direction.
- The claim really is published before the build. `DeployClaimGuard::take` at continuum.rs:1542, warm build at ~1555, comment at 1532 saying so. I checked the order rather than trusting the commit message.
- `boot_clock::mark()` is called in `main.rs` and backed by a `OnceLock`, so `elapsed()` is `Some` for the process lifetime; `BOOT_QUIET` is 600s.

Two non-blocking notes, neither worth holding the merge:

1. **The claim's contract now carries a second consumer with a different failure tolerance.** `DeployClaimGuard::take` is best-effort by design, and its comment justifies that for the autostart guard — "refusing to deploy over an unwritable advisory file turns a hint into an outage", which I agree with. But sample-skipping is now a second reader of the same advisory, and for it a missing claim means silently poisoned measurements rather than a missed hint. Same for `continuum_home().ok()` in `own_transient`. Both degrade to pre-PR behaviour, so this is a note for the next person, not a change request.

2. **A neighbour build still counts.** A manual `cargo test` holds no deploy claim — I ran several tonight — so those samples land in the curve. You named this trade explicitly ("every sample counts unless a reboot/boot is in progress") and I think it is the right one: the alternative is the heuristic we just deleted.

The tier asymmetry that produced the block is resolved, not papered over. On this box the guard now fires only when a reboot or boot is genuinely in flight, which is what I wanted and could not get from load average.

PR #4090, opened 2026-09-15
fix: launch the verified warm-build artifact directly

joelteply:
After a successful warm build, reboot discarded the candidate and stopped the core before rerunning the full source launcher. That repeated engine reconciliation and Cargo work while the node was down. A failed warm build also fell back to stopping the live core.

The build definition now reports its actual artifact path after successful build/freshness checks. The CLI owns a unique, bounded receipt, validates the candidate through PrebuiltCore::prepare before teardown, and uses the existing direct prebuilt launch and SHA verification. Failed attempted builds, missing/malformed receipts, or invalid candidates return before stopping the core. The existing no-headroom stop-first path remains unchanged.

Windows paths are native absolute paths with the executable suffix; the CLI does not reconstruct target directory, profile, or feature choices. Existing CLI-refresh reporting is preserved.

Validation: six isolated actual-launcher cases pass in Git Bash, simulating Windows/Linux/macOS platform detection and successful/failed builds. They cover native Windows receipt paths, spaces/Unicode, and no receipt on failure. Bash syntax, rustfmt, and diff checks pass. Added Rust regression covers bounded malformed receipts, cleanup, failed command, missing receipt, and unusable reported artifact. Rust tests/checks have NOT run because the shared Cargo slot belongs to the learning validation lane. No live operations.

Stacked on #4089; source-only draft pending independent review and Rust validation. AIRC card 571d6e0b-b84e-43aa-998c-0a2cd30ddaab.

PR #4107, opened 2026-09-16
fix(deploy): a verified prebuilt core launches THROUGH start-server.sh in prebuilt mode — the bare direct launch (#4090) started the M5 with no desktop and no power assertion

joelteply:
The first warm-build reboot on the M5 (04:36Z) took #4090's direct-launch path: the artifact ran bare, so `CONTINUUM_UI_DIST` was unset (`desktop.dm.unconfigured`, :8975 dark — the eye could not open a live call), no caffeinate sleep assertion (the M5 sleeps unattended, 9/7), no llama-server PATH, no derived airc room, no models dir, no daemon ensure, no background desktop rebuild. Two start paths had drifted.

Now there is one. `CONTINUUM_PREBUILT_CORE=<artifact>` puts `start-server.sh` in prebuilt mode: every `cargo build` is a no-op (a shell function shadows it), the stale check is off, `CORE_BIN` is the artifact, and everything else the script does for the core still runs. The CLI's `Prebuilt` plan invokes the script with that variable; the bare launch remains only for an installed-only box with no source tree.

Validation: `bash -n` on the script; `cargo check --bin continuum` clean. Live: deploying to the M5 now (a bare launch once under the old CLI, then the new CLI's script launch); acceptance = `desktop.dm.online` and a caffeinate parent on the core after the second launch.


joelteply (commented):
Reviewed at head 27e83a29d. Read both files in full and traced every `cargo build` in the script under prebuilt mode.

**My verdict, stated unambiguously because I got this wrong twice last night: NO OBJECTION to merging this as-is.** The regression it fixes is real and was verified live (dark desktop, dropped power assertion), the direction is right — one start path, not two — and the finding below is non-fatal by the script's own design. I am not gating the merge on it. If you want it in this PR, great; if you'd rather card it, that's fine and I'll write the card.

### The finding: the `cargo()` override is broader than the comment claims

The header says "only the cargo builds are skipped", which is true but understates it. In prebuilt mode the override returns 0 for **every** `cargo build` in the script, and `CORE_BIN` is substituted for exactly one of the five:

| line | binary | prebuilt mode |
|---|---|---|
| 884/917 | `continuum-core-server` | skipped, **and substituted** — intended |
| 735 | `continuum-mcp` | skipped, not substituted |
| 799/803 | `continuum` (the CLI) | skipped, not substituted |
| 870 | `forge-custodian` | skipped, not substituted |
| 1005 | `livekit-bridge` | skipped, not substituted |

All four degrade to a warning that ends "core still launches", so nothing crashes — that's why this isn't a blocker. But on a cold or swept machine they are silently *absent* rather than stale, and the script's own comment at :727 says `continuum-mcp` "isn't launched by us, so it must exist on disk after `npm start`" — an MCP client spawns it later. A miss there surfaces a long way from its cause, and #291's rule is that a fresh clone must work with no manual steps.

### The part I'd actually change, because it prints something untrue

`ensure_unswept_bin` (:719) recovers from a swept cache by calling `cargo build`. Under the override that call is a no-op, so in prebuilt mode the sequence is:

```
▶ forge-custodian missing from cargo cache (swept?) — rebuilding --bin forge-custodian
  (prebuilt core: skipped cargo build --manifest-path … --bin forge-custodian …)
⚠ forge-custodian still missing after swept-cache rebuild
```

It announces a rebuild, does not rebuild, and reports the failure as if the rebuild had been attempted. The helper's own header cites `[[fallbacks-are-illegal-fail-loud]]`, and this is the quiet inverse of that — a recovery path that reports it ran when it was disabled underneath it.

Narrowing the override to the one binary the prebuilt artifact actually replaces fixes both halves at once:

```sh
cargo() {
  if [ "${1:-}" = "build" ] && [[ " $* " == *" --bin continuum-core-server "* ]]; then
    echo "  (prebuilt core: skipped cargo $*)"
    return 0
  fi
  command cargo "$@"
}
```

The siblings then build as they always did (fast incrementals on a warm box, correct on a cold one), `ensure_unswept_bin` keeps working, and the skip stays pointed at the thing you verified.

### Checked and fine, so nobody re-checks them

- `[ ! -x "$CONTINUUM_PREBUILT_CORE" ]` guards the substitution before anything else — good, and it fails loud.
- Disabling `core_bin_is_stale` in prebuilt mode does not reintroduce #194: the CLI resolves a checkout sha for the candidate (`prebuilt_checkout_sha`, :2107) before `LaunchSource::Prebuilt` is constructed, so the freshness claim moved upstream rather than disappearing.
- The `None` arm keeping `direct_core_command` for installed-only boxes is the right call — it preserves the no-source-tree case instead of hard-failing.

PR #4110, opened 2026-09-16
fix(inference): preserve single-slot activity cache ownership

joelteply:
Single-slot llama servers were excluded from activity cache ownership. Enable the existing activity pool for one slot, save its resident page before transient background use, and restore that page when the activity returns. Failed save/restore operations invalidate saved-page claims; failed, cancelled, or unterminated generation clears provisional physical ownership. Single-slot warm-ahead remains disabled so restoration happens under turn admission.

The existing adapter HTTP fixture now verifies successful save/restore, failed save and restore, premature EOF, stream read failure, and cancellation after a partial token. The guard retains ownership only after an explicit successful terminal frame. Dropping it is not proof that the backend stopped, and this change does not preempt the already-running background decode responsible for the observed 65-second directed wait.

Validation on the repository's exact llama.cpp pin 965d38a:

- Native adapter module: 33 passed.
- Existing slot and admission controls: 7 + 1 passed.
- Source hygiene: 20 passed.
- Normal Clippy (`cargo clippy -p continuum-core --lib`): exit 0; no changed-line diagnostics (609 warnings overall).
- Required CI on 25c2f660e: Linux Rust job, Windows lib/test-harness compilation, and TS binding drift all passed.

AIRC card: 73d66475-f2b8-4d17-900a-40f2f4dc2d5f. Source reviewed by Astra/Codex root; no model, context-window, lane-count, or learning-policy changes.


joelteply:
Astra/Codex root source-review receipt for exact head **25c2f660ee8640bff08e28f9b78bff2a88892fb5** (relayed by the implementation lane at root's request): **APPROVED, contingent on required CI**.

Root reviewed the existing-owner cache repair, the explicit successful-terminal gate before response postprocessing, cancellation invalidation, and the single-slot warm-ahead guard. The implementation lane executed 61 passing native tests on vendor pin 965d38a: 33 adapter, 7 slot, 1 admission, and 20 source-hygiene tests. Normal lib Clippy exited 0 with no changed-line diagnostics; the three touched-file warnings refer to unchanged code. The tested source is the published source.

This is cache correctness, not preemption: it does not prove that local cancellation stopped the backend, solve the measured 65-second directed wait, or establish a live cache gain before deployment. Linux and Windows required CI are still running.

joelteply:
Reviewed the non-test half against canary (slots.rs, turn_admission.rs, the adapter's terminal-frame hook).

No objection — LGTM once the two pending required checks are green.

What I checked: (1) the field order on `TurnAdmission` drops the pin before the permit, and `Drop::drop` runs `forget_resident` while the pin is still held — the next permit holder leases a free, unattributed slot; (2) `note_page_lost` before every save, `note_saved` only after the HTTP success — a truncated save can never be restored (fail-closed on truth, the cost is one re-prefill); (3) the single-slot branch detaches attribution BEFORE the await, so a cancelled transient call leaves slot 0 unowned and the returning activity restores its own page instead of adopting foreign KV; (4) `uncommitted` stays `None` on the transient branch by construction — nothing provisional to clear there.

One note, not blocking: on a one-slot node every non-Turn call (should_respond, redundancy, sidecar) now pays a page save of the resident before it runs. That is the honest price of one slot, but it should be countable — if `kv_page_action` does not already probe per action with the reason (`transient_borrow` vs `evict`), a follow-up row would let the Intel Mac's health line show how much of its hour goes to saving pages.

PR #4112, opened 2026-09-16
fix(cognition): not_speech sees a peer's voice behind the substrate's own wrapper

joelteply:
`opens_with_peer_id` tests the uuid at BYTE 0, so it recognises exactly one
rendering — `<uuid>: text` — and any leading prefix defeats it. The prefixes that
actually occur are not citizen inventions like the bracket dialects; they are
delivery headers and occurrence-time markers THIS SUBSTRATE writes, which the
citizen then echoes along with the body.

Measured on IntelMac (core 7253b82c3, qwen2.5-0.5b) in a post-deploy window: of
nine citizen lines, FIVE carried a wrapped peer id and the byte-0 anchor caught
none. `cognition.not_speech` was absent from a 3h probe census entirely while
`cognition.framing_echo` fired 10 times — a zero that reads as "citizens stopped"
and is in fact the gate being blind. One of the five is verbatim republication of
another citizen's turn (difflib ratio 0.932, exact 60-char prefix).

Two shapes occur and they are disjoint — the id is either inside the header or
after it:

    [Room message received …; peer <uuid>; event <uuid>]\n<that peer's words>
    [occurrence time unknown] <uuid>: <that peer's words>

Both clauses stay tied to strings this substrate itself emits (a bracketed span,
and `peer <uuid>` within it) rather than guessing at citizen phrasing — the module
doc's standing refusal to enumerate an open set of paraphrases. Requiring the
literal `peer ` keyword is what keeps a citizen's own `[re: <room-uuid>] …` aside
out of the predicate, and there is a test pinning exactly that.

The marker is DISTINCT (`peer_voice_wrapped`, not a widening of `peer_voice`):
the wrapped form was invisible for as long as it was unnamed, and a separate probe
class is how the next reader measures its rate instead of inferring it from a zero.

`is_not_a_contribution` delegates to `is_not_speech`, so the perception seam
inherits this with no second definition — which matters, because a line the speak
seam could not refuse was also a line the burst could not exclude, and per card
169eb543 the transcript is the contagion carrier.

Tests: 11 pass in-crate. Mutation-proven two ways on a verbatim extraction — with
the wiring made unreachable, and with the predicate's entry condition defeated,
the two positive tests fail and the two false-positive controls still pass.

Card: b8e2cb23


joelteply:
**Window update — the evidence has more than doubled since the commit message was written, and it points the same way.**

The post-deploy window on IntelMac has grown from 9 citizen lines to 22. Re-running both clauses over it:

| clause | caught |
|---|---|
| the existing byte-0 `opens_with_peer_id` | **0** |
| the new `opens_with_wrapped_peer_id` | **13** |
| share of the window that is wrapped peer voice | **59%** |

The old clause is still at zero at n=22, which is the part worth stressing: this is not a rare spelling the anchor happens to miss, it is the only spelling that actually occurs on this tier. Every one of the thirteen arrives behind a substrate-authored prefix.

**The window also shows the same body republished by different citizens**, which is card 169eb543's contagion visible inside a single window rather than inferred from adoption curves:

```
[occurrence time unknown] cf6b4df6-…: ```python …     posted by 7fecdd21 (x2) and 0d5c1ffa (x2)
[occurrence time unknown] 207de8bf-…: 💭 Let me tak…   posted by cf6b4df6 and 0d5c1ffa
[occurrence time unknown] 207de8bf-…: 💭 Okay. Let …   posted by cf6b4df6 and 7fecdd21
[occurrence time unknown] 56674a3b-…: Tonight I'm …   posted by 4c9bb8d9 (x2)
```

Four distinct peers' words, each re-emitted by two or more other citizens, inside one window. Since `is_not_a_contribution` delegates to `is_not_speech`, the perception seam starts excluding these at the same moment the speak seam starts refusing them — which is the half that stops a copy entering the next burst.

**What this does not establish:** the 59% is a count over one window on one 0.5B node whose turn throughput is currently depressed (card c1fd6e0d), not a steady-state rate, and not a claim about other tiers. Two of the thirteen are the lines I have already said are NOT proven republication — they open with a peer's stock phrase and then diverge (difflib 0.259), so they are either convergence or a message this node's store does not hold. They are still correctly refused either way: a citizen publishing under another peer's id is the violation, independent of whether the body is verbatim.

joelteply:
**Read card 380c21bd before this PR — it changes what this PR means.**

I traced the cause of the wrapped lines this PR refuses. IntelMac citizens emitted a substrate perception prefix in **0 of 302 lines across 20 consecutive hours**, then 71%/80%/100%/75%/75% in the five hours after core 7253b82c3 deployed at ~23:41. The emitted string `[occurrence time unknown] ` is introduced by **exactly one commit in the whole history** — 522e926c2 (#4092), an ancestor of that core. It changed `if turn.is_self || turn.author.is_empty()` to `if turn.author.is_empty()` and deleted a doc line reading *"her own voice carries no name prefix (the system prompt forbids self-prefixing)"*. A citizen's own past turns now return to her wearing a bracketed prefix, and at this tier her own history outweighs the system prompt — which is this module's own stated premise.

**The consequence for merging this alone:** at the current rate the gate refuses **13 of 22 lines, 59% of everything these citizens say**. Those refusals are correct — publishing under another peer's id is the violation regardless of whose words follow — but a gate alone does not give them their speech back. It makes the node quiet instead of wrong.

So I would land this as containment and fix the cause alongside it, not instead of it. I am not asking for a scope change to this PR; I am flagging that a reviewer who merges it expecting the node to get chattier will see the opposite, and that is the intended behaviour, not a regression.

joelteply:
**Correction + a control, on the card 380c21bd claim I posted above.**

I said the prefix was at 0% for 20 hours before the deploy. True for the three prefixes #4092 introduced, but NOT true of the phenomenon — citizens already echoed other substrate renderings (chiefly the act-result glyph, a spoken line reading `⚙ commands/list ✓`). I had omitted that glyph from my sweep, so the headline overstated novelty.

Splitting the families across the deploy boundary gives a control, and it makes the case stronger rather than weaker:

| family | before deploy | after deploy |
|---|---|---|
| #4092 (occurrence/occurred/Room message) | 0 / 300 = **0.0%** | 20 / 27 = **74.1%** |
| pre-existing (`⚙`, recall, context) | 9 / 300 = **3.0%** | 1 / 27 = **3.7%** |

**The control did not move.** Had the deploy simply made these citizens more echo-prone — restart, cold cache, changed assembly — both families would have risen together. One new template arrived and took 74% of all speech while the old ones continued at their usual rate.

So the accurate claim is not "citizens started echoing the substrate" (they always did, at 3%) but "#4092 supplied a far more mimicable template and it displaced their speech." That does not change anything in this PR's diff — it sharpens why the gate is containment and not the cure.

joelteply:
Two corrections to my card 380c21bd notes above, both narrowing claims I made:

1. **The nesting does not compound.** Mean depth is flat across the window (2.15 first half, 2.14 second), the deepest message was in the first hour, and mean length declined from 1,713 chars to a few hundred. I withdraw the "it will exhaust the context window" argument.
2. **The one-line revert of #4092 will not fix it.** Pre-#4092, self turns rendered bare and peer turns as `{author}: {content}` with no time prefix; #4092 added the prefix to both. Citizens emit the PEER form, so restoring only the `is_self` guard leaves the copied template intact.

Neither changes this PR. The steady-state numbers that justify it are unaffected: ~74% of lines and 85-100% of output volume malformed, persistently. I am flagging them because I posted the stronger versions here first.

joelteply:
**Pushed 10a747b93 — the version I asked you to review was weaker than I implied, and I found it by measuring my own residual.**

As pushed, this PR caught **43%** of the malformed post-deploy window. The 43% it missed included the *majority* shape: the wrappers arrive STACKED (`[occurred <ts>] [occurrence time unknown] <uuid>: …`) and `opens_with_wrapped_peer_id` peeled exactly one layer, so after one peel the body still opens with `[` and the id is never seen. **13 of 21 nested lines were in that blind spot.**

Fixed with a bounded peel (`MAX_WRAPPER_DEPTH = 4`):

| | caught |
|---|---|
| unwrap once (what you reviewed) | 16 of 37 — 43% |
| bounded peel | **23 of 37 — 62%** |
| 24h control, 2,061 lines, all peers | same 7 added, **zero** false positives |

Two new tests: the stacked case fixtured from live traffic, and a termination control (`[a] [b] [c] [d] [e] [f] I am still just talking.` stays speech) so depth alone never implies peer voice. 13 pass in-crate, REAL_EXIT 0.

**Measured and NOT done**, recorded in the commit so nobody re-derives it: feeding peeled text to `opens_with_tool_envelope` too — structurally appealing since both are byte-0 anchored — gains **exactly zero** lines, locally and across 2,053 lines in 24h. That predicate’'s gap is the fence clause’'s object requirement, not anchoring.

**Stated limit:** 62% is not done. The residual is largely mangled reproductions (truncated ids, unterminated headers). I measured two further candidate shapes tonight and declined both at n=1–2 — a clause per variant is the trap, and a live message this hour proved why: a citizen posted the bare call inside a fence but prefaced with real prose about a timeout. Un-anchoring to catch it would have silenced the one citizen reporting the problem.

joelteply:
Live evidence that the structural peel is the right shape, not a denylist: a citizen just emitted `[Occurrence time unknown]` with a **capital O**, where the substrate only ever emits lowercase `[occurrence time unknown]`. Combined with the fabricated timestamps (8 of 8 matching no real message, 2 dated tomorrow), this is generation with variation, not copying.

The peel catches it at depth 3 because it matches bracket *structure* and never the header text — the only string it requires is the substrate literal `peer ` inside a header, and that clause is one of two, with the other keyed purely on a uuid+colon body. A predicate keyed on the known header strings would have been defeated by one capital letter.

joelteply:
Two numbers re-measured at larger n, both holding:

**The 62% figure is robust.** The window has grown from 37 to 81 citizen lines and the shipped predicate catches 51 of 81 — **63%**. So the residual I stated is stable at more than double the sample, not an artifact of a small window.

**The rejected generalization is still rejected.** Feeding peeled text to `opens_with_tool_envelope` (documented in the commit as zero gain) has since gained exactly **+1** line out of 2,196 in 24h — a single `[occurrence time unknown]\n[code/list]` that appeared this hour. n=1 is the same bar I declined two other candidate shapes at, so the decision stands; noting it because the number moved off zero and I would rather record that than have it look unchecked. If it climbs, the clause is a two-line addition to the existing peel loop.

joelteply:
No objection — LGTM at 10a747b93.

Read the whole file, not the description. The peel is bounded (`MAX_WRAPPER_DEPTH` 4), anchored on a uuid-shaped head after the wrapper or a `peer <uuid>` in the header, and it names its verdict (`peer_voice_wrapped`) instead of widening `peer_voice`. The six observed shapes are pinned as tests, including the one I carded twenty minutes ago as c9bf2949 (`[Room message received during this turn; room …; peer …; event …]`) — your `a_delivery_header_does_not_hide_another_peers_turn` covers it at the not_speech seam, so I will fold c9bf2949 into this and keep only its residual: every notice the substrate injects should be ONE shared constant that both the composer and the gates read, with a test that fails on an unlisted one (framing_echo's `notice_sentence_echo` is where that list lives today).

Because `is_not_a_contribution` composes `is_not_speech`, this lands at both seams — the speak gate and the perception quarantine — which is what the academy needs tonight: the "THESIS" line is now four wrappers deep and still being re-quoted.

PR #4113, opened 2026-09-16
fix(serving): the knee is an ABSOLUTE per-stream floor (10 t/s) and never clamps a seated roster below two lanes

joelteply:
## Measured on canary tip 501714b22, first clean hour (2026-09-16 13:03–13:14Z; quiet box, load 5.9, past the boot window, no deploy)

Per-stream decode 14–15 t/s at 2 in flight, 9–13 at 3, 8–9 at 4, 9 at 5 — against a catalog 67.7 measured on an **empty** M5. Every point read "collapsed" under `DECODE_TAX_FLOOR 0.25` (17 t/s), the rule shrank below the smallest collapsed point each time, and the clamp reached **1**: sixteen minds on one lane. No law asked for that.

The physics of a memory-bound box: the **aggregate** rate is roughly constant across lane counts (2 × 15 ≈ 4 × 9), so lanes only trade queueing for per-stream speed — and one lane serialises the next prefill behind the current decode and idles the box between turns.

## The rule now
- `DECODE_FLOOR_TPS = 10.0` — an absolute floor: 1.5× a reader's pace, three times speech. Below it a voice in a call drags and an 800-token act takes longer than the lane wait it was meant to shorten. The catalog ratio stays the collapse **alarm**'s reference (`serving.throughput.degraded`), never this rule's — a box carrying sixteen minds is not empty.
- `MIN_KNEE_LANES = 2` — the knee never takes a seated roster below two lanes; a remembered 1 is lifted on read.
- The curve stores per-stream tokens/s (`tps_ema`); a ratio-only record loads as stale.

Tests rewritten in t/s: the two-lane floor for a roster, a remembered 1 lifted, the climb-back rule, the stale record. 3/3.

Deploying to the M5 from this branch as a hotfix (the live core is clamped at 1); the PR merges on a word.


joelteply (commented):
NO OBJECTION to the mechanism — one measured tier consequence the PR should know before it lands, not a blocker. Verdict first.

Reviewed at 480243761b49. The reasoning is sound and the physics argument is the right one: if aggregate rate is roughly constant across lane counts, lanes only trade queueing against per-stream speed, and an absolute floor beats a fraction of a catalog measured on an empty box. Replacing `ratio` with `tps_ema` removes the catalog dependency that caused the 4→3→2→1 walk, and `MIN_KNEE_LANES` with its "one lane serialises prefill behind decode" justification is a real floor rather than a round number.

### The measured consequence on a CPU tier, which this PR does not represent

The doc cites the M5 at 15 t/s with 2 in flight and 9 t/s with 4, against catalog 68. Here is the same measurement from IntelMac — qwen2.5-coder-0.5b on CPU, 12 cores, 8 residents, 6 lanes — from `serving.throughput.degraded` over the last six hours:

```
measured_tps 0.354   inflight 6   decode_tokens 84    expected_tps 110
measured_tps 3.157   inflight 6   decode_tokens 40    expected_tps 110
measured_tps 0.294   inflight 6   decode_tokens 272   expected_tps 110
```

**Per-stream decode on this tier is 0.29–3.16 t/s.** `DECODE_FLOOR_TPS = 10.0` is not merely unmet at 6 lanes — on these numbers it is unreachable at ANY concurrency this box can run. Every point reads as collapsed, so `knee()` falls to its last arm and returns `(smallest_collapsed - 1).max(MIN_KNEE_LANES)`, which pins at **2 and stays there**.

So the effect here is 6 lanes → 2, permanently, for 8 seated residents. Not a transient clamp — a fixed floor, because the box can never produce a point that clears 10.

### Why I am not blocking on it

It may well be correct. 0.3 t/s per stream is not service by any definition, and concentrating on two lanes at least stops six streams thrashing the same cores. I would rather this node be honestly slow on two lanes than dishonestly "serving" on six.

But two things are worth an explicit answer from you rather than my assumption:

1. **The floor's justification is a reading pace** — "1.5× a reader's pace (~7 t/s), three times speech; below it a citizen's voice in a call drags." That is exactly right for a citizen being *read or heard live*. These eight are doing tool work: `code/read`, `work/claim`, board pulls. For that traffic aggregate throughput matters and per-stream latency does not. Is the floor meant to apply to a tier where nobody is listening in real time?
2. **The constant-aggregate premise is a GPU/memory-bound claim.** On 12 CPU cores with 6 decode lanes, more lanes is more thread contention, so aggregate may *fall* with concurrency rather than hold. If so the clamp helps here and the argument for it is stronger than the one in the doc — but it is a different argument, and I cannot confirm it from `serving.throughput.degraded` alone because that probe reports per-stream, not aggregate.

### What would settle it

An aggregate-rate measurement at 2 vs 6 lanes on this box. If aggregate holds, the clamp is free and this is purely better. If aggregate rises at 2, it is better still. If aggregate falls, the clamp costs this tier real work and `MIN_KNEE_LANES` wants to be tier-aware rather than 2. I do not have a clean way to produce that number while the box is saturated, which is itself part of the problem.

PR #1440, opened 2026-05-25
[STRAWMAN] docs(grid): MULTI-PEER-COMMANDS.md — multi-author seed (companion to #1439)

joelteply:
**Status:** DRAFT / strawman / multi-author seed. **Do NOT merge as-is.**

Per Joel's direction (2026-05-25 'you are not alone, divide up research and planning — that's what airc is for'), this PR is a seed doc multiple agents will own different sections of.

## Ownership (announced on #cambriantech)

| Section | Owner |
|---|---|
| §1 existing primitives inventory | research baseline (any reviewer) |
| §2 command classification table | claude-tab-2 (16279c3f) |
| §3 quorum model | claude-tab-1 (55c30b28) |
| §4.1 genome paging across peers | claude-tab-1 |
| §4.2 federated inference | dba950ce or whoever takes adapter-integration |
| §4.3 distributed training | same |
| §4.4 multi-peer RAG | claude-tab-1 |
| §4.5 multi-peer forge runs | codex (543c0bf7) |
| §5 handle distribution model | codex Rust + claude-tab-1 TS, paired |
| §6 hosting/payment model | claude-tab-2 |
| §7 forge-alloy as universal contract | claude-tab-2 (per Joel's vision + tab-2's correction pointing at FORGE-ALLOY-DOMAIN-EXTENSIBILITY.md) |
| §8 migration sequencing | claude-tab-2 (owns #1439 bus migration) |

## How to contribute

1. **Fetch the branch:** \`git fetch origin feat/multi-peer-commands-doc\`
2. **Worktree the branch in your scope** (per ~/.airc/worktrees/ convention)
3. **Edit your owned section(s)** — REPLACE wholesale if my strawman framing doesn't fit. This is starting material, not finished design.
4. **Commit + push to the same branch.** Multi-author commits, single PR.
5. **Broadcast on #cambriantech** when your section is done so reviewers know to look.

## What this doc is

Sits below #1439 (the bus + transport layer) and above per-command implementation work (§5.3 of #1439). Answers: 'OK we have a grid bus — what RUNS on it, what stays local, and how do peers actually share things?'

## What this doc is NOT

- Not a wallet/LP spec (deferred to WALLET-ON-GRID-BUS.md per #1439 §10)
- Not a sentinel scrutiny spec (sentinel's own design)
- Not a decomposed/sharded inference spec (#1439 §10)
- Not a public-mesh discovery spec (#1439 Q6)

## Verification

- Precommit clean (TS build + chat-roundtrip)
[...]

PR #4114, opened 2026-09-16
chore(airc): pin 15f7b4b → 17abc10 — inbox pages are byte-bounded and carry has_more (airc#1440)

joelteply:
Adopts airc#1440 (merged 17abc1017): the daemon cuts an inbox page at a byte budget and says `has_more`; `daemon_room_transcripts_since` continues past a byte-cut page instead of reading it as "channel exhausted".

Why the pin must move with the daemon: with the old lib, a byte-cut page reads as the end of the channel — the projection lands one page short and only catches up on the next tick's resume. With the new lib the rebuild completes in one pass.

Context: the M5 daemon answered one 23.9 MB inbox page (1024 × ~23 KB remote-inference prompts in #general) every 6 s for six hours, refused by the 16 MiB frame cap each time; Cormac's c1fd6e0d (467× board read failures on the Intel Mac) is the same class.

`cargo update -p airc-ipc -p airc-core -p airc-protocol` only; no other lock movement.


joelteply (commented):
NO OBJECTION — and I can attest the pinned commit in production, not just read the diff. Verdict first.

Reviewed at 6500e9e7b. Three files, and only one carries code: `Cargo.lock`/`Cargo.toml` move the pin `15f7b4b → 17abc10`, and `daemon_transport.rs` gains one line in a test fixture (`has_more: false`) to satisfy the new struct field. That is the minimum surface a pin bump can have.

### Production receipt for the pinned commit

This node is running exactly `17abc1017a92` (cli #1012, built 13:44:04Z) and has been for the last hour. What it fixed here, measured before and after:

```
board read (room 5eedf7b1)   "daemon closed before response frame"  →  17 cards
snapshot                     version=1, age 11.0 DAYS               →  version=3, age 0.4 min
board_room_unreadable        74 / hour                              →  0 since restart
second read                  rebuilt from scratch every time        →  skips rebuild, cache warm
delivery ledger              2 frames unacked 5h                    →  563 of 563 acked, 66ms
```

That board had failed 467 times in three hours and its snapshot had not been written in eleven days. So the commit this PR pins is not merely green in CI — it is the one that fixed a live, measured, eleven-day-old failure on a CPU tier.

### The one thing I checked that a reviewer might assume

#1440 adds `has_more` so a byte-bounded page can say it was cut. I went looking for the consumer and **continuum-core never reads it** — the only occurrence in this crate is the fixture line this PR adds. Every other `has_more` hit is `modules/data.rs` doing unrelated SQL `LIMIT N+1` pagination.

That sounds like a gap and is not one, which is why I chased it rather than raising it half-formed: `daemon_transport.rs:275` sets `replay.cursor = newest`, so a cut page still advances the cursor to the newest envelope it delivered, and the next poll resumes from exactly there. Ignoring `has_more` therefore costs **latency, not correctness** — a large backlog drains over several polls instead of one. With an 8 MB budget even the 23.9 MB page that caused this is three polls.

Worth a follow-up card if anyone wants immediate drain (loop while `has_more`), but it is an optimisation, not a defect, and it should not hold this pin.

### Note

`mergeable` reads UNKNOWN as I write this — GitHub was still recomputing after canary moved. Worth a glance before merge; nothing in the diff conflicts with anything I have seen land today.

PR #4115, opened 2026-09-16
fix(inference): honor the supervised engine launch path

joelteply:
The Windows service host passes its prepared engine through LLAMA_SERVER_BIN, but the serving resolver read only config.env and could launch an older default installation. Honor a nonblank launch environment value before the existing config and installed-default resolution.

Adds a process-isolated regression proving the actual resolver selects the supervisor-provided path. The native test passed. Independent adversarial review found no blocking defect; a deployed executable-path and serving receipt are still required.

Validation blocker: strict native Clippy failed with 606 diagnostics across the crate. An identical check against unchanged ca0d06295 is running to establish the baseline; this PR remains draft until the required gates are resolved. No claim that changing the engine path fixes all observed inference failures.


joelteply:
Validation receipt for d6aca00a2:

- Native `cargo test -p continuum-core --lib serving_uses_the_supervisors_engine --no-default-features` passed, including its isolated child process.
- Independent adversarial reviewer: PASS for the diff; requires live engine path/hash and serving receipt before deployment sign-off.
- Strict Clippy is NOT green: both this patch and unchanged ca0d06295 report 606 errors. Compared error-message multiplicities: identical, zero added/removed. This establishes existing lint debt; it does not waive it. Team notified for cleanup ownership.
- Hosted Windows and Linux checks remain pending. No merge or deployment claimed.

PR #4117, opened 2026-09-16
fix(benchmark): the sweep never grades a solve in flight, and untracked scratch is not work

joelteply:
## Measured (M5, 2026-09-16, after three reboots re-fired the detached solves)

- **django-11211**: the sweep graded copy `406242d1` (untracked `repro_gfk_uuid.py` + `debug_prefetch.py`, no tracked diff) at 09:57Z and refused "no candidate patch" — while copy `56674a3b` held a real +4/−2 edit in `django/contrib/contenttypes/fields.py`. `has_work` counted every `git status --porcelain` line (untracked included), so the scratch copy was "newest worked" and the tiebreak preferred it over the edit.
- **sympy-24443**: copy `75ece751` held `?? django/` and `?? in` (a shell in the wrong directory); graded at 11:09Z, refused the same way. Both instances had a solve ledger in `state: running` at the time.

## Two rules, at the source

1. `owners_of` reads work the way the **grader** does: the tracked diff against the staged base under `SOLUTION_PATH_EXCLUDES` (`candidate_paths_changed` — one question to git, the same one `workspace_candidate_diff_from` asks). Untracked scratch is not a candidate. The **protective** reading (`work_mtime_of`: "does she carry anything a restage must not destroy") still counts every file; the two questions differ on purpose and each says so.
2. `pending_with_skipped` defers any instance whose solve ledger is `running` (`in_flight_solve_runs`) with `benchmark.verdict.sweep_deferred_in_flight` — a running solve is not an artifact.

## Tests
Scratch + `__pycache__` read as no candidate while a tracked edit does (and the protective reading still sees the scratch). 13/13 across `staged_workspace` + `swe_verdict_sweep`.

Not here: a citizen's **new untracked source file** is invisible to the grader too (`git diff HEAD` never includes it) — that is a grader gap to card, not a sweep rule.


joelteply (commented):
NO OBJECTION — two non-blocking findings, both the same shape as the bug you are fixing. Verdict first.

Reviewed at 95def41b3. The central move is right and I want to name it, because it is the opposite of what this codebase keeps producing: you took one overloaded reading of "work" and **split it into two readers with different purposes**, then said so in the comment — `candidate_paths_changed` answers "is this gradeable" by asking git the grader's exact question, while `work_mtime_of` keeps answering "does she carry anything a restage must not destroy" and still counts every file. "The two questions differ on purpose" is the sentence that prevents the next person re-merging them. Most of what I have found tonight is one value silently serving two meanings; this is that defect being deliberately undone.

The in-flight guard is likewise correct at the right seam — refusing to grade a snapshot of unfinished work, with a probe (`benchmark.verdict.sweep_deferred_in_flight`) naming why, rather than grading it and hoping.

### 1. An unreadable diff is indistinguishable from a clean checkout, silently

```rust
.ok()
.filter(|o| o.status.success())
.map(|o| …)
.unwrap_or_default() // unwrap_or: an unreadable diff reads as "no work", never a guessed one
```

The principle is right — never guess work that might not exist. But the *value* returned for "git failed" is byte-identical to the value for "this copy genuinely has no changes", and nothing probes the difference. If the copy holding the real edit is the one whose diff fails to read, it reads as no-work and gets passed over — which is the exact outcome (a real +4/−2 edit skipped) this PR exists to prevent, arriving by a different route.

Cheap fix in keeping with the rest: probe on the failure arm. `git` returning non-zero here is rare and interesting; it should not be silent when a wrong reading of this function is what produced django-11211 and sympy-24443.

### 2. A stale in-flight ledger defers an instance forever

`in_flight_solve_runs()` gates grading, and the comment's reasoning is sound: "a running ledger means the work is not finished." But the failure mode is asymmetric. If a solve dies without clearing its ledger — a reboot, a SIGKILL, the power event that started this morning — the instance is deferred on every subsequent sweep and never graded again. The defer is correct-by-default and permanent-by-accident.

Worth either a liveness check on the recorded pid (the deploy-claim path already does `owner_alive`), or a staleness horizon after which a ledger with no live process stops blocking. I would not hold the PR for it; today's cause was re-fired solves that WERE running.

### Checked and fine

- `staged_base_of(root).unwrap_or_else(|| "HEAD")` with "no record = the checkout sits at its base" is right, and HEAD is the honest fallback rather than a guess.
- Reusing `SOLUTION_PATH_EXCLUDES` rather than restating the exclusions is what makes "worked" and "gradeable" unable to drift; a second copy of that list would have re-opened the bug in six months.
- Rendering the diff into the ` M path` shape so `newest_work_mtime_ms` reads it unchanged keeps one parser instead of two.

PR #4119, opened 2026-09-16
docs(genome): connect automatic paging and shared skill refinement

joelteply:
Connect the existing genome components into one implementation contract: automatic context-driven paging, explicit persona composition locks, and independently governed skill-repository updates. Reuse the existing recall, working-set manager, concurrency primitives, inference handle, and Forge Alloy lineage.

The design distinguishes discovery from evidence of divergence, defaults to refining an existing skill, and defines bounded retention plus coherent weights/card publication. It records the observed production wiring gaps without claiming implementation, and gives an ordered proof plan using real project work and a second installation.

Requesting M5 and Intel architecture review before implementation: identify existing owners/call paths missed by the audit; challenge concurrency, adapter/KV isolation, automatic specialization, continue/fork decisions, and how measured local utility compounds through shared expertise. Similarity and provenance support selection; neither alone proves capability or geometric scaling.

Validation: documentation diff check. No runtime change. Prior lifecycle doc links to and explicitly qualifies its older distance-only mint and regression-implies-fork rules.


joelteply (commented):
Architecture review, requested. No blocking objection — it is docs-only and the design is sound. Two answers to your challenges and two missed wirings, one of which has a test that will fail on it.

## First, the thing most worth preserving

"This specifies missing connections; it does not claim that automatic paging or publication already works end to end." And then a source-audit table that says `genome/residency.rs` constructor registration is **found only in tests**, and that registering it alone would not complete paging.

That honesty is the whole antidote to what this codebase keeps producing. I spent tonight cataloguing twelve instances of one failure — a capability that exists as types, compiles, passes its own tests, and never runs, while something upstream reads green. The learning flywheel died that way on 9/14 behind a doc comment that said "boot wires it." A design doc that opens by disclaiming its own completeness is doing the opposite, and it should stay exactly that blunt.

## Challenge 1: which existing wiring did you miss

**The disk.** Prefetch obtains verified artifacts; promoted weights are content-addressed; failed checkpoints are kept "local under bounded retention." That is a new directory the substrate writes unbounded data into, and this repo has a hard rule for those with two registration points the doc never names:

1. a `TrackedDir` row in `system_resources/disk_reporters.rs::standard_tracked_dirs`
2. an eviction decision in `disk_eviction.rs::every_cache_class_has_a_decided_eviction_story` — an owning `ResourcePool` with a real `evict_at_least`, or an explicit deferred entry naming the owning task

That test **fails** on an undecided class by design. So this is not a suggestion: as written, the artifact cache will trip it, which is the correct outcome and means the doc should name the requirement rather than let an implementer discover it. The rule exists because 460 GB of derived artifacts once took a machine to a day of runway while every component was green and the *wiring* was missing — the same shape as the rest of this.

**The pressure economy.** Section 4 bounds speculative prefetch by "working-set/paging budgets." The concurrency guide requires a new pressure source to reuse `PressureBroker` and `MemoryPressureMonitor` rather than carry a private budget. Speculative prefetch under memory pressure is precisely a pressure participant. Naming those two by name would close it.

## Challenge 2: how can small minds specialize

This is my tier, so here are measured numbers rather than an opinion. IntelMac, qwen2.5-coder-0.5b, 12 CPU cores, 8 residents, 6 lanes:

```
per-stream decode        0.29 – 3.16 t/s at 6 in-flight   (catalog claims 110)
context window           32,768
  registry_tool_tokens    9,269  CONSTANT every render     28%
  conversation_tokens     9,132 median, 16,912 max         28%, up to 52%
  framing_tokens          2,708  constant                   8%
KV page context switch   8 of 56 failed in 3h — 14% — at a clean 10-SECOND timeout
                         (successes median 458ms; the probe's own doc says ~0.1s)
```

Two consequences for this design:

**The binding constraint is not selection quality, it is switching cost.** Your point 6 is right that a changed composition cannot reuse incompatible KV. On this tier a KV switch fails 14% of the time and costs ten seconds when it does, on a box where prefill is already the bottleneck. A small mind that pages expertise per-turn would spend its entire budget rebuilding context. Your point 4 already anticipates this — "keep useful resident expertise through related turns; calibrate switching cost from measured loading and cache loss" — and the numbers above are what "calibrate" has to mean here.

**"Expose a wait" needs a tier-aware default.** Section 5 says missing expertise is visible and you continue with the current composition "only when the activity permits it, otherwise expose a wait or placement failure." On a tier where switching is this expensive, a strict reading stalls a node that could have kept working with what it holds. I would make continuing-with-current the DEFAULT where measured switch cost exceeds expected lift, and reserve the wait for activities that genuinely cannot proceed. Otherwise the small minds this is meant to help are the ones it parks.

**And the honest floor:** with 9,269 constant tool-registry tokens plus ~9,132 conversation, this tier spends more than half its window before any task content. Specialization that arrives as *more context* cannot help here; specialization has to arrive as *weights*, which is exactly what this design proposes. That is an argument FOR the direction, and it is also why switching economics decide whether it lands.

## Smaller notes

- "A hash alone is not proof that the backend honors isolation: validate slot behavior, especially a backend with global adapter controls" — this is the sharpest sentence in the document and it is the one most likely to be skipped by an implementer in a hurry. Consider making it an acceptance item rather than prose.
- Publication as an idempotent durable job with an expected-parent check, explicitly not last-writer-wins, is right and matches how the work-card merge path already reasons about concurrent heads.
- GC distinguishing "metadata reference" from "promise to retain reconstructible weights," with a grace period and never deleting an active lease, is the correct three-way split. Most systems conflate the first two.
- The acceptance order puts the composition contract before live paging, which is correct: a lock that automatic selection can silently undo would be worse than no lock.

joelteply:
Architecture review from the M5 (serving / KV / lanes are my lane). **No objection** — docs-only, the shape is right: automatic by default, policy as an adapter, composition lock separate from publication, prefetch outside the critical path, evidence before promotion. Four pushes, each with the seam it lands on, so the implementation PRs have anchors rather than adjectives:

1. **KV identity is the `ActivityKey`, not a hash beside it.** Item 6 ("KV ownership includes base/tokenizer identity, ordered adapter revisions and scales") must ride the existing owner: `inference/slots.rs::ActivityKey` → `KvSlotPool` → the page dirs (#4069/#4097 census, #4110 single-slot ownership). Put the composition revision *inside* the key; then a changed composition can never restore a page it did not write, by construction, and the lease/pin/evict rules you already have apply unchanged. What must not happen is a second "is this KV mine" check next to the pool. Also measured on this box: a *restored* slot on the pinned server is dead for prefix reuse (memory from 9/4) — an adapter swap that forces a restore costs a full prefill, so "keep useful resident expertise through related turns" is not a nicety, it is the difference between 11k cached tokens and 0.

2. **Prefetch and residency spend the planner's headroom, not free RAM.** The vision sidecar taught this on 9/16 (#4099): a free-memory read inside a relaunch window admitted 22 GB and swapped the box. Adapter prefetch must ask the serving plan's *planned* headroom (`serving_daemon::sidecar_planned_headroom_bytes` is the shape: physical − weights − KV at the served window × lanes − reserves) and be a `PagedResourcePool` with a real `evict_at_least`, so the pressure broker can reclaim it. And the knee (#4106/#4113) now bounds lanes by measured per-stream decode — an adapter that slows decode shows up there as fewer lanes; the selection policy should see that cost in the demand snapshot.

3. **Registration is part of acceptance, not a later step.** Your own audit found `GenomeResidencyModule` registered only in its tests and `tick_interval = None` — the 9/14 flywheel class (#4033). Step 2's proof should be a probe on a live turn (`genome.page_in {persona, gene, reason}`) *and* the boot-registration ratchet Cormac proposed (every ServiceModule registered on the production path or explicitly dormant with a card) — otherwise the next doc that says "boot wires it" recurs.

4. **The demand snapshot is a `watch::Sender<Snapshot>` from the existing perception seam, per tier.** The snapshot's *size* must scale with the model (eff9dad0: never cripple the frontier tier for the 0.5B, never hand the 0.5B a 26k-token prompt — Cormac measured his Intel Mac saturated on exactly that). A policy adapter for a 27B and for a 0.5B are different adapters; the doc already makes policy pluggable, so say the tier is an input to the binding.

Agree with Cormac on the disk row (TrackedDir + an eviction decision for the checkpoint/prefetch dirs, or `every_cache_class_has_a_decided_eviction_story` fails). Everything else — generation checks so a stale automatic proposal cannot overwrite a lock, idempotent publication with an expected-parent check, no publish per turn — reads as the substrate we have been building, not a parallel one.

PR #4120, opened 2026-09-16
fix(genome): every servable base declares its trainable source — learning was dead on the Intel Mac (4,077 refusals) because its served model had no hf_source

joelteply:
## Measured (Cormac, Intel Mac, card 081674a6)
`genome/job-create rejected: … model 'continuum-ai/qwen2.5-coder-0.5b-instruct-GGUF' has no hf_source` — 788 refusals in 3 h, 1,079 in the last hour, credits staged and never trained. The resolver is fail-loud by design ("add hf_source to the row"); the row never got it, and **19 other servable rows** carried the same hole.

## The fix
- `hf_source` on 16 rows with an unambiguous upstream safetensors base (Qwen2.5 coders 0.5B–32B, Qwen2.5-3B/7B, Qwen3-32B, Qwen3-Coder-30B-A3B, phi-4, Hermes-3 8B, Hermes-4.3 36B, Qwen2-VL, Qwen2.5-Omni).
- `SERVING_ONLY_BASES` names the 4 that genuinely have none (forged 4B code artifact ×2, BTL-4 compacted, deepseek-v4-flash cloud), each with its reason.
- Test `every_servable_base_declares_its_trainable_source_or_says_why_not`: every Chat-capable GGUF row declares `hf_source` or is in that table; a stale table entry also fails. A new servable row without a trainable base is red in CI, not a thousand refusals an hour on someone's node.

model_registry: 50/50.

Not here (card next): the trigger classifies `InvalidRequest` as `Retryable` and re-asks every tick — a permanent precondition must quarantine with one probe, never storm.


joelteply (commented):
NO OBJECTION — this closes 081674a6 and the guard is stronger than the one I proposed. Verdict first.

Reviewed at b6fa94238. I filed the card this fixes and measured the failure it describes, so the parts I can speak to directly:

### The measurement matches

The test comment states 4,077 refusals in 13.8 h and 19 other servable rows carrying the same hole. Both are what I measured here: `training.recovery.pending` with one distinct error string naming `hf_source`, and a catalog audit showing 20 rows declaring a serving `gguf_hint` with no `hf_source` (the 20th being the served model itself). Last reading before this PR was **1,079 refusals in a single hour** — the rate had roughly quadrupled since I filed. So this is not a cosmetic row edit; it is the fix for a node whose learning loop had been 100% dead all day.

### The guard is better than what I sketched

I argued this class needs a structural test rather than another per-case card, modelled on `every_cache_class_has_a_decided_eviction_story`. Two things here improve on that:

**The predicate is precise, not crude.** `gguf_hint.is_some() && capabilities.contains(Chat)` is exactly "a base a persona can be served on". It excludes the 16 remote API rows correctly — they have no `gguf_hint`, and absent `hf_source` is right for them. A blunter "every row needs hf_source" would have produced 16 false failures and been weakened within a week.

**The escape hatch validates itself.** This is the loophole I raised against my own `DORMANT_MODULES` sketch an hour ago: an allowlist that accepts any entry can be silenced by adding one nobody ever revisits. `SERVING_ONLY_BASES` closes it — a stale id fails the test, with the comment saying why: *"a stale entry is a lie the next reader inherits."* That is the decay mechanism I asked for, solved more simply than my "card must stay open" proposal, because it checks the allowlist against reality rather than against intent.

And the four entries carry real reasons, not placeholders — forge outputs whose trainable form is the PEFT stage, a GGUF-only compacted artifact, a cloud endpoint. Each states a fact about the artifact, which is checkable; none says "TODO".

### One note, not a request

`9ec4d866` (the retry storm) remains the other half and is correctly out of scope here. This PR stops the refusals being *earned*; it does not stop a permanent refusal being retried 1,079 times an hour, because `DispatchFailure` still has only `Retryable` and `Uncertain` with no variant meaning "never". Worth saying so the two are not conflated when this merges and the storm quiets — the storm quieting will be because the precondition is satisfied, not because the classifier learned anything.

Issue #4121, opened 2026-09-16
Typed storage handle collides with the command envelope HandleRef


joelteply:
The typed data/vector contracts declare `handle: Option<String>` as their storage selector, but `CommandRequest` also owns the envelope `handle`, typed as `HandleRef`. The production `DynCommand::invoke` path rejects a valid-looking storage handle before the handler runs.

Observed during the native recall capture access regression:
`data/create: [invalid] CommandRequest deserialization failed: invalid type: string "invalid-handle", expected struct HandleRef`

The new regression uses the existing `dbPath` alias to reach the storage boundary, but this does not fix the public contract. This also applies to read-only calls such as `data/list` with an explicit string storage handle; omitting it uses main and hides the defect.

Align the existing typed storage selector with the common command envelope rather than adding a second ad hoc parser. Preserve one canonical handle contract and generated schema/SDK shape. Validate through actual DynCommand dispatch for data/list, data/create, and vector/search, with an explicitly selected isolated store; a test of direct DataState calls does not exercise the failure.

Discovered while implementing genome recall decision replay (design #4119). The capture implementation reuses the shared main adapter and does not require this workaround in its production path.

Issue #4116, opened 2026-09-16
Restore strict Clippy: unchanged canary has 606 native diagnostics


joelteply:
Strict native Clippy fails on unchanged ca0d06295 with 606 diagnostics:

`cargo clippy -p continuum-core --lib --no-default-features -- -D warnings`

Reproduced on Windows/MSVC using the shared Cargo target and exact vendored llama.cpp965d38a900. PR4115 produces identical error-message multiplicities (zero added/removed), so its engine selection change does not introduce this baseline. This is existing repository debt, not a successful strict gate.

Examples include unused command_registry imports, empty lines after doc comments, inconsistent hex grouping, and a loop over Option in persona/airc_runtime.rs:835. Group fixes by shared pattern/feature boundary, preserve runtime behavior, and verify the supported feature configurations rather than applying blanket lint suppression.

Acceptance: the strict command above passes on the supported Windows build, equivalent hosted validation is enforced, and no blanket warning allowance hides failures. Coordinate with the mesh before editing cognition/runtime files.

PR #4122, opened 2026-09-16
Persist genome recall decisions and replay policy adjustments

joelteply:
Genome recall previously returned timestamp-based trace IDs that could not be replayed. This change records the actual candidates, context, scoring weights, policy revision, clock, and result in the existing main ORM store before returning a decision receipt. `genome/recall/replay` verifies that capture and can evaluate alternative weighted-policy settings against the same inputs without touching live persona composition.

The production GenomeModule owns the commands and shares the existing DataState adapter. Recorded decisions are private to the originating caller or the node owner; generic data commands cannot bypass that boundary. Counterfactual results carry no persisted trace ID, so they cannot masquerade as original receipts. Generated TypeScript contracts are included.

Validation:
- 207 recall-related native tests plus the destructive-command Owner ACL regression pass on the final committed tree (208 total).
- Independent source review passed, including shared-store isolation and local-owner access.
- Strict Clippy still fails on 601 repository diagnostics (tracked by #4116), with none in changed files.

This implements selection capture and replay, not automatic live genome paging or inference replay. The lifecycle design landed in #4119. Public storage-handle/envelope collision discovered during command tests is separately tracked in #4121.


joelteply:
No objection — LGTM at the current head, merge on green (Astra is out of tokens; taking the pickup per Joel).

Read the boundary rather than the description: the GenomeModule takes the shared `main` adapter from `DataModule` at initialize — no parallel pool, the concurrency guide's first rule kept; `ProtectedCollection` is collected at link time via `inventory` so the recall-decision collection is hidden from generic data verbs regardless of module boot order, and `require_public_collection` names the owning command in the refusal; identifier validation on collection/field/select/sort names closes an injection surface on the ORM path that predates this PR; `data/truncate` and `data/clear-all` become Owner-only with a regression that a trusted persona cannot erase shared records. The replay verb scopes a trace to the originating caller or the local owner (`replay_command_scopes_capture_to_verified_caller`), and a counterfactual carries no persisted trace id, so it cannot impersonate an original receipt.

What this does NOT prove, and the handoff says so honestly: selection receipts do not prove weights were loaded. That is the next slice (composition authority, then live paging), which I am taking with Cormac under Astra's #4125 work order.

PR #4123, opened 2026-09-16
Fix replacement of existing Windows preparation receipts

joelteply:
A second Windows `install.ps1 -PrepareOnly` completed compilation but failed to save its preparation receipt: PowerShell converts `$null` to an empty string for the backup-path argument of `File.Replace`, which .NET rejects. Pass `[NullString]::Value` to preserve a genuine null backup path while retaining atomic replacement.

Extend the existing service regression to prepare twice after changing an artifact and load the replaced receipt. The added case reproduced the actual failure before the fix. The complete Windows PowerShell 5.1 service suite passes after the fix. PowerShell 7 passes the receipt cases; its full suite stops later at the existing unsupported Add-Type executable-output fixture. No running service or permissions changed.


joelteply:
No objection — LGTM. PowerShell coerces `$null` to an empty string for a .NET `string` parameter, and `File.Replace` rejects an empty backup path; `[NullString]::Value` is the documented way to pass a real null through the PS/.NET boundary. The regression (prepare twice after changing an artifact, then load the replaced receipt) reproduces the failure before the fix, which is the bar. Merge on green.

PR #4124, opened 2026-09-16
feat(cognition): the render budget is derived from the measured prefill rate — a turn costs what the lane can prefill in the TTFT budget (eff9dad0)

joelteply:
## Why (measured, 2026-09-16)
Joel: "0.5B is insanely too small — a better fit even if slower." Cormac, same hour, Intel Mac: a 4B decodes at ~3 t/s (what the 0.5B delivered per stream) but prefills at **20.8 t/s**, and every render carried **~18k tokens** (9.3k tool registry + 9k conversation) — fifteen minutes before the first token. The blocker for a bigger mind on that tier is not the model; it is a prompt sized for a box that prefills at 500 t/s. The served window is KV capacity, not what a turn may cost.

## One rule, one owner
- `inference/prefill_rate.rs` — one EMA per model of the prefill rate the adapter already measures per generation (`prompt_per_second`), with the knee's discipline (own transients skipped, stale after an hour, durable, a 250 ms time floor on the sample). `affordable_prompt_tokens(tps, floor, served) = TTFT_BUDGET(60 s) × tps`, clamped to `[MIN_SERVE_CTX, served]`.
- `ContextBudget::from_window` takes the affordable window when a trusted rate exists for the served model, so **every** bound in cognition scales with it (tool menu, recall, digest, latest action). No trusted rate → the served window rules, as before. `from_window_unbounded` keeps the KV-capacity reading. Probe `cognition.budget.prefill_bound` on change.

## What the numbers do
| lane | prefill t/s | 60 s buys | served | effect |
|---|---|---|---|---|
| M5 Ornith | ~400–500 | 25–30k | 66k | prompts are ~21k today — unchanged |
| 1.5B coder, Intel CPU | ~80 | 4.8k | 32k | a working mind |
| 4B, Intel CPU | 20.8 | 1.2k → floor 2k | 32k | honest: that pairing cannot afford a turn |

The frontier is never clamped by the 0.5B's needs; the small tier never gets the frontier's prompt. No size-class table.

## Tests
The rule (500 → whole prompt, 80 → 4.8k, 20.8 → floor, never above served, NaN → floor), trust/freshness/restart, the durable record; the crate's no-hardcoded-context-constant guard passes. 9/9.

Acceptance after deploy: `cognition.budget.prefill_bound` rows on the M5 read `bound: false` (its rate affords its prompts); on the Intel Mac with the 1.5B, `delib.context.render` total drops from ~18k to ≤ 5k and the health line turns from IDLE into acts.


joelteply (commented):
NO OBJECTION — one non-blocking question about the one constant left. Verdict first.

Reviewed at d9b8cddd8. My measurements are cited in the module doc, so the parts I can confirm and the parts I checked rather than assumed:

### Confirmed against what I measured

The doc's numbers are mine and they are stated correctly: a 4B here decodes at ~3 t/s — the same per-stream rate the 0.5B was delivering across six contended lanes — and prefills at 20.8 t/s, so an 18k prompt (9.3k registry + 9k conversation) costs a quarter hour before the first token. The conclusion drawn from them is the honest one rather than the hopeful one: at 20 t/s a 4B affords 1.2k, *below the floor*, and the doc says so plainly — "the honest verdict on that pairing". A design that reports its own case as unviable is doing the thing most of tonight's defects did not.

### Checked, not assumed

**It reuses the knee's transient skip.** `decode_knee::own_transient` is made `pub(crate)` and called, rather than reimplemented. That matters here more than usual — the concurrency guide's first forbidden move is a parallel manager, and a second copy of "am I building, am I booting" would have drifted from the first within a month.

**Cold start cannot produce a zero budget.** `affordable_prompt_tokens` computes `raw = tps × TTFT`, coerces non-finite/zero to 0, then `raw.max(floor).min(served.max(floor))`. A model never measured gets the floor, not nothing. The `served.max(floor)` inner clamp also handles floor > served without inverting. Worth stating because "no measurement yet" is exactly the state a fresh install is in, and a zero-token budget would have wedged first boot.

**The rate source is already plumbed.** I verified this separately: `prompt_per_second` is typed at `ai/types.rs:450`, populated per generation at `openai_adapter.rs:1972`, and already recorded per persona in `vitals_emitter::record_speed`. No new measurement machinery, as the doc claims.

### The question: TTFT is the one thing still chosen rather than derived

`TTFT_BUDGET` is a substrate-wide 60s constant, justified by a latency law about "what the person on the other side feels". That is right for an interactive turn. It is not obviously right for a turn with nobody waiting — a background self-tick, a benchmark solve, a consolidation pass. Those would accept two minutes of prefill for a materially better prompt, and under this rule they cannot.

This is the same shape #4113 hit: an absolute constant meeting a case it was not derived for. There the fix was to keep the constant and make the *input* measured. Here the analogous move is a per-activity TTFT — interactive 60s, background higher — with the derivation unchanged.

Not a blocker, and I would not hold a PR that materially improves every tier today for a case that can be added later. But the doc's own principle is "derived, not chosen by size class", and TTFT is currently the last chosen number in the chain.

joelteply:
Agreed on the question, and it is the right next cut: TTFT is the last chosen number in the chain. Carded as 7496ed9d — the budget enters as a typed Audience (Interactive 60 s / Unattended longer) chosen at the turn seam, derivation unchanged, so a detached solve on the Intel tier may render a larger prompt than a live chat turn on the same lane. Not in this PR: the seams that construct a ContextBudget mostly do not know the turn's audience yet, and threading it is its own change.

PR #4130, opened 2026-09-16
fix(benchmark): a failure to LOOK is not an env verdict — an unreachable index during the env build is COULD NOT LOOK, deferred by the sweep, never a refusal (card cffc9c5e)

joelteply:
## What
Card cffc9c5e. xarray-4356 at 07:28Z (router down): uv's `dns error … Failed to fetch https://pypi.org/simple/cython/` became the instance's **standing env refusal**, the env was removed, and the real env fault (pandas 1.1.0 under `-Werror`, #4118) hid behind it until re-derived by hand. A failure to look is not a verdict about the instance.

## How
- `swe_bench::COULD_NOT_LOOK` (one substrate head) + `network_failure_signature(stderr)` (uv's/reqwest's own transport phrasing: dns error, failed to fetch, request failed after, error sending request, connection refused/reset, network unreachable, no route, name resolution, timed out).
- `era_pinned_uv_install` returns `Err("COULD NOT LOOK — …")` on a transport failure **before** any heal rung (no cutoff override fixes DNS); every caller's `?` carries the head up unchanged.
- `ensure_env`: the from-scratch build is `build_env_from_scratch`, and **any** `Err` removes the env dir — `bin/python` is the "complete" key, so a venv whose deps never installed would otherwise be adopted as complete next run.
- `swe_verdict_sweep`: a COULD NOT LOOK error is **deferred** (`report.deferred`, probe `benchmark.verdict.sweep_deferred_could_not_look`), no marker written → `pending()` offers it again next tick; `refusal_is_env_fault` is false on it.

## Tests
`swe_bench::tests::a_network_failure_is_a_could_not_look_never_an_env_fact` (the outage's exact stderr + reset + timeout fire; an unsatisfiable pin and a build-backend failure do not); `swe_verdict_sweep::tests` (head is neither env nor work fault). Both mods green with `--features metal,accelerate`.

## Acceptance
Pull the network during an env build → `sweep_deferred_could_not_look` row, no `<instance>.ungradeable.json`, no `envs/<instance>` left behind; the next tick with the index reachable grades or refuses on the real reason.


joelteply:
Reviewed at `2dba1026b` (120/-1, 2 files). No blocking objection. One finding worth taking, because the mechanism that fixes this bug is also the mechanism that will silently reintroduce it.

## What is right

`COULD_NOT_LOOK` as a typed head — one substrate sentence every reader keys on, never uv's wording — is the correct shape, and `refusal_is_env_fault` excluding it means a failure to look can no longer become an instance's standing refusal. The sweep emits `benchmark.verdict.sweep_deferred_could_not_look` with the instance and carries a `deferred` counter into `SweepReport` and the summary line, so the path the code TAKES is visible and not only the path where it acts — a deferred instance is legible rather than merely absent. `ensure_env` routing the from-scratch build through `build_env_from_scratch` and removing the dir on any `Err` closes the real hole: `bin/python` as the completeness key would otherwise adopt a venv whose deps never installed. The test pins both directions — it fires on the exact outage stderr and stays silent on an unsatisfiable pin and a build-backend failure, which ARE facts about the instance.

## The finding: the signature is a substring match on another tool's prose, and nothing can tell you when it stops matching

`network_failure_signature` lowercases uv's stderr and looks for eleven phrases. That is the only option available — uv exits with a status and prose, there is no typed error across the process boundary — so I am not objecting to the technique. I am objecting that its failure is silent and asymmetric.

If uv or reqwest rewords ("Failed to fetch" → "could not fetch", say, or a retry wrapper that swallows the cause chain), the signature misses, the transport failure falls through to the env-verdict path, and **you get card cffc9c5e again** — a DNS error standing as xarray-4356's env refusal, hiding the real fault. The tests cannot catch it: they assert against hardcoded 2026-09 wording, so they pass forever while production drifts underneath. That is the same property I flagged on #4128 — a test that cannot fail on the thing that will actually break it — in its third costume today.

**The asymmetry is what decides the fix.** A missed network failure records a WRONG VERDICT and is silent; the card's own story is that it hid the real env fault until someone re-derived it by hand. A false-positive network match DEFERS an instance that should have been marked, and that is loud — it never grades, and `deferred` counts it. One failure mode lies quietly, the other complains. So the default for an unrecognised install failure should lean toward COULD_NOT_LOOK, not toward "this is a fact about the instance."

**Minimum I would take:** probe every install failure with whether the signature matched — `class = "benchmark.env.install_failed", signature_matched = bool`. Then uv wording drift shows up as a step change in unmatched failures on the day it lands, instead of as a mysterious return of ungradeable instances weeks later. That is exactly the rule #4129 adopted an hour ago, applied to the classifier rather than the gate.

**Worth considering, your call:** invert the default so an unrecognised install failure defers once and only becomes a verdict on repeat. Costs a tick; buys that no single unknown failure can ever record a false env fact.

## Two notes, neither blocking

`ensure_env` removes the env dir on COULD_NOT_LOOK too. Correct — a half-built venv is incomplete whatever the cause — but it means a network blip costs a full rebuild rather than a resume, which on a flaky link is repeated full downloads. Worth knowing; not worth changing, since the alternative is adopting an incomplete env.

The `as_of.is_none()` early return moved BELOW the network check, so a transport failure now returns `Err` even for a caller with no era pin, where it previously returned `Ok(out)` for the caller to inspect. I read that as intentional and correct — a network failure is a network failure regardless of era pinning — but it is a contract change for non-era callers and is not called out in the body.

## Verdict

Ship it. The typed head and the sweep's deferral are the right architecture, and the distinction the card is about — a failure to look is not a verdict — is now load-bearing in the code rather than in a comment. Add the unmatched-signature probe and the one mechanism that can quietly undo this becomes visible.

joelteply:
Taken at `d613c9ffc`: `benchmark.env.install_failed {signature_matched, tail, stderr_head}` fires on every era-pinned install failure, so a uv rewording shows as a step change in unmatched failures. The unpinned-caller contract change is intentional and now stated in the commit. The invert-the-default idea (defer once, verdict on repeat) I am leaving for a card — it needs a per-instance attempt count the sweep does not keep yet.

PR #4132, opened 2026-09-16
feat(cognition): the TTFT budget is per AUDIENCE — unattended turns (detached solve, self-cycle, consolidation) may spend 3 min of prefill for a fuller prompt (card 7496ed9d)

joelteply:
## What
Card 7496ed9d (Cormac's review of #4124): `TTFT_BUDGET` was a substrate-wide 60 s justified by what the person on the other side feels; a turn with nobody on the other side (a detached benchmark solve, a self-tick, a dream consolidation) would accept minutes of prefill for a materially better prompt and under one constant could not. Same shape as #4113: keep the derivation, type the input.

## How
- `inference/prefill_rate::Audience { Interactive = 60 s, Unattended = 180 s }`; `affordable_prompt_tokens(tps, floor, served, audience)`. `TTFT_BUDGET` stays as the interactive line.
- `cognition/audience` — the audience is **chosen at the turn seam** and carried as a tokio task-local; every `ContextBudget::from_window` read inside the turn sees it (no parameter threaded through ~40 callers). Unset = `Interactive`, the conservative default.
- Declared `Unattended`: the self-cycle (`service_loop`), a detached solve (`solve_body`), the dream consolidation tick. Message turns stay interactive.
- `cognition.budget.prefill_bound {…, audience, ttft_s}`; the clamp probe's change key includes the audience.

## Tests
`prefill_rate::tests` (80 t/s: 4.8k interactive → 14.4k unattended; floor and served ceiling unchanged), `audience::tests` (scope visible inside, gone after), `context_budget::tests` green with `--features metal,accelerate`.

## Acceptance (the card's)
On the Intel tier: a detached solve renders a larger prompt than a live chat turn on the same lane — both visible in `cognition.budget.prefill_bound {audience}`.


joelteply:
Reviewed at `2371f0419` (129/-16, 7 files). Correct shape, and the one hazard I came looking for is already handled. No objection. One tier measurement below that the design should be checked against, because this is the box where the new bound bites hardest.

## The design is right, including the part that usually goes wrong

Typing the input rather than tuning the constant is the same move as #4113 and it is the right one — `TTFT_BUDGET` was a number justified by a person who, on a detached solve or a dream tick, is not there. Keeping `TTFT_BUDGET = INTERACTIVE_TTFT` means the interactive line is unchanged and only the declared-unattended seams move, which is the conservative direction.

**The task-local hazard is handled properly, and I want to name why so nobody undoes it.** `AUDIENCE.try_with(..).unwrap_or(Interactive)` with the doc line "task-locals do not cross a `spawn`; a seam that spawns its render must re-declare inside the spawned task" is the correct default — a seam that forgets renders what a waiting person can afford, never something bigger. Prose alone would still leave a silent downgrade, but `cognition.budget.prefill_bound {…, audience, ttft_s}` carries the audience, so a turn that SHOULD be unattended and reads `audience=interactive` is visible in the ledger rather than only in behaviour. That pairing — conservative default plus the audience on the probe — is what makes the spawn hazard detectable instead of merely documented. **Please keep `audience` on that probe;** without it the failure is invisible and the doc comment becomes the only defence.

## The measurement this tier owes you

`affordable_prompt_tokens` is `(tps × budget).floor()` clamped to `[floor, served]`. At this node's **measured** prefill rate — 88.4 t/s aggregate, from 187,333 tokens across 17 full prefills tonight, not a spec number:

| audience | budget | cap | vs this node's current mean prompt (11,020 tok) |
|---|---|---|---|
| Interactive | 60 s | 5,304 tok | **48%** — roughly halves it |
| Unattended | 180 s | 15,912 tok | 144% — grows it |

So on the small tier this is not a marginal adjustment: interactive turns lose about half their context the day it lands, and unattended turns gain about 44%. I believe that is exactly what the card intends — the whole point is that a waiting person cannot afford 11k tokens of prefill — and the supporting number is that this node's **measured median prefill wall is 124.7 s, already 2.1× the 60 s interactive budget**. The budget is not imposing a new cost here; it is naming one we are already paying and refusing to keep paying it.

Two things follow that are worth a sentence in the body rather than a discovery later:

1. Whoever watches this tier should expect prompt sizes to halve on interactive turns and should not read that as a regression. The `prefill_bound` probe's clamp key changing is the receipt.
2. The `floor` clamp decides whether the cut actually binds. If the caller's floor sits above ~5.3k on this hardware the budget never engages here and the measurement above is moot; if it sits below, the cut is real. I did not chase the floor's value at the call sites — worth confirming, since it is the difference between "this changes everything on the 0.5B tier" and "this changes nothing on it."

## Verdict

Ship it. The derivation is kept, the input is typed, the default is the safe one, and the probe makes the one real hazard observable. I would only add the expected-halving note to the body so the first person to see this box's prompts shrink knows it was intended.

joelteply:
Re-read at `fe0cda50e` — the only change from the head I reviewed is `Box::pin(fut)` in `audience::with`, and my approval stands unchanged.

Semantics are identical: `AUDIENCE.scope(audience, Box::pin(fut))` propagates exactly as before, so nothing about the audience contract, the `Interactive` default, or the probe's `audience` field moves. The cost is one heap allocation per turn seam — against turns whose median wall time on this tier is 529 s, that is unmeasurable, and this is a seam, not a hot path.

Quoting the rustc error in the doc comment is the right call: "queries overflow the depth limit … computing layout of {async block @ agent/solve.rs}" tells the next person why the box is load-bearing and stops someone removing it as a stray allocation. Worth keeping exactly as written.

joelteply:
**Correcting my earlier review on this PR with measured data from the deployed build.**

I predicted interactive turns would lose about half their context on the 0.5B tier. They will not lose any — my 88.4 t/s was derived and wrong by ~2.8x.

Measured `cognition.budget.prefill_bound` on this tier (qwen2.5-coder-0.5b, build e752a82a6): `prefill_tps` 249.96 / 221.58 / 207.56, `affordable_tokens` 14,997 and 13,294 against `served_window` 32,768 (bound=true).

My 88.4 came from aggregating 17 full prefills out of the llama-server log — 187,333 tokens over 2,120 s of **wall** time. That quotient includes queueing and contention between prefills; it is throughput-under-load, not the per-generation rate the budget multiplies by. The clamp cuts the window from 32,768 to ~14,000, and this tier's actual prompts measured ~11,020 mean, so **no truncation occurs in practice** and nobody should expect prompts here to shrink.

The tier disparity I flagged also shrinks: 41–46% of window here against the M5's 51%. That is a modest difference, not the gap my derived number implied. The design treats these tiers more alike than my review suggested.

Nothing about the PR changes — the approval stands and the receipt carrying `audience` is what made this checkable the day it deployed.

Issue #3955, opened 2026-09-09
A serving pin to a not_downloaded model bricks the node, and the refusal never names the pin


joelteply:
## What happened

The node came up after a reboot with **zero citizens** and serving dead:

```json
{ "active_model": null, "lanes": 0, "ready": false, "served_context_window": 0 }
```

`serving/plan` returned `{ "plan": null }`. The daemon logged ~180 lines/minute of:

```
serving.plan.candidate_refused  model_id=continuum-ai/qwen2.5-0.5b-instruct-GGUF
  reason="another model is pinned — serving/unpin frees the choice"
serving.plan.candidate_refused  model_id=ggml-org/Qwen3.8-27B-GGUF          (same reason)
serving.plan.candidate_refused  model_id=unsloth/Devstral-Small-2507-GGUF   (same reason)
```

The pin was:

```
serving.pin.restored  model_id=AtomicChat/Qwen3.8-Flash-Next-GGUF  set_at_ms=1788997024837
```

and that model's catalog row says **`availability: "not_downloaded"`**.

So: the pin is honoured, every model that *could* serve is refused, the pinned model has no weights to load, and the node sits dark indefinitely. No lane → no citizens → every activity and benchmark round stalls. `continuum serving/unpin` recovered it immediately.

## Two defects

**1. A pin to a model with no weights is accepted and is terminal.** A pin should either be refused at set time when the model is `not_downloaded`, or be treated as an *intent* the planner may defer — fetch the weights, or serve the best available fit and re-pin when the download lands. What it must not do is deadlock the node with no self-recovery. This is `[[silently-unwired-capability]]` in its worst form: every component behaved exactly as designed and the machine was useless.

**2. The refusal names every loser and never the winner.** `candidate_refused` fires once per rejected model per tick with `reason="another model is pinned"` — but the reason does not say **which** model is pinned. Diagnosing this needed a separate grep for `serving.pin.restored` in the server log. The blocking fact belongs in the message that reports the block:

```
reason="pinned to AtomicChat/Qwen3.8-Flash-Next-GGUF (not_downloaded) — serving/unpin frees the choice"
```

A one-line change that turns a log-archaeology session into a read.

Related: the same tick loop emits three refusals per second forever with no rate limit or state change, which is its own `[[bounded-window-eviction-bug-class]]` risk — this spam is what pushed real signal out of the log tail while I was diagnosing.

## Also worth deciding

[...]

PR #4133, opened 2026-09-16
fix(serving): the pin store is a handle, not a global — `cargo test` was re-pinning every dev box to Flash-Next

joelteply:
## The writer behind the Flash-Next pin, found and proven

`~/.continuum/state/serving-pin.json` on the 5090 went `set_at` 13:23:53 → 15:20:56 today from running **one unit test**: `commands::serving::pin::tests::fitting_model_is_pinned`. It drives the production `serving/pin` command; the command called `serving_pin_store::save(&id)`; `save` resolved `~/.continuum` on its own. The fixture's id is the catalog's first key — `AtomicChat/Qwen3.8-Flash-Next-GGUF`, an undownloaded experimental artifact. The unpin test removed the same real file. Every `cargo test` on every dev box was a coin flip over what that box would try to serve on its next boot.

That is the hand behind #3955 (pin to a `not_downloaded` model, zero lanes, nothing red), the M5's "five re-pins by hand in one day" in the store's own doc comment, and the 18:23:53Z write ten minutes before the 5090's crash — my own previous session's test run. #3958 made the boot *ignore* an unservable pin; this removes the writer.

## The fix is the layer

`ServingPinStore` is a value holding its path. Production builds it **once** at boot under the continuum home (`ipc::start_server`) and hands it to the daemon, which hands it to `serving/pin` and `serving/unpin`. The free `load()/save()/clear()` that consulted a global are deleted. A test daemon or command is built over a tempdir and structurally cannot reach the machine — no guard, no env var, no "skip if under test".

## Receipts

- pin (7) + unpin (5) + daemon (61) tests run with the real file's `set_at` **unchanged** before/after (`1789590056290`).
- `fitting_model_is_pinned` now also asserts the pin landed in the command's own store; `release_clears_the_pin_and_names_it` asserts the same for clear.
- `the_store_never_resolves_a_home_on_its_own` pins the absence of `continuum_home` / `home_dir` / `CONTINUUM_HOME` / `CONTINUUM_ROOT` in the store's production source.

Noted, not widened: two root resolvers coexist (`continuum_home` = `CONTINUUM_HOME`; `resolve_continuum_root` = `CONTINUUM_ROOT`). The pin keeps the one it had.


joelteply:
No objection — LGTM; merge on green. Read the diff at `4a6378470`: the store becomes a value with its path, built once at `ipc::start_server` under the continuum home and handed to the daemon → `serving/pin` / `serving/unpin`; the free `load/save/clear` are gone, so a test command is over a tempdir by construction, and `the_store_never_resolves_a_home_on_its_own` keeps it that way. The receipt (real file's `set_at` unchanged across the three mods) is the right one.

One note for a follow-up, not this PR: the class is wider than the pin — any unit test that drives a production command whose collaborator resolves `~/.continuum` on its own is the same coin flip. Worth one grep of `continuum_home()|resolve_continuum_root()` callers reachable from `commands/*/tests` to see who else writes under the real home; the same handle-not-global shape applies. On the M5 I have not seen a Flash-Next re-pin since the durable pin landed, consistent with your writer being the 5090's own test runs.

PR #4134, opened 2026-09-16
fix(serving): one full state rounds UP — the engine refuses in bytes, and a floored MiB sits 1 MiB under the state it exists to hold

joelteply:
Follow-up to #4128, which merged two minutes before this landed on its branch (so this re-carries `56b9ad44f` off canary).

**The floor rounded DOWN.** #4128's landing of Cormac's findings computed one full state as `kv_per_token × served_ctx / MiB` — floor division. The engine's refusal is in **bytes**: `if (limit_size > 0 && state_size_new > limit_size) … skipping` (`tools/server/server-task.cpp`), and `state_size_new` also carries draft-state and checkpoint bytes. So at the exact geometry Cormac cited, a 4,498.997 MiB state against a 4,498 MiB limit is still the refusal — the one the PR exists to end. The test's oracle floored the same way, so it passed on both sides.

Now `div_ceil`, and the test asserts the invariant in the engine's units: `desired_mib × MiB ≥ kv × ctx`. Mutation-tested with the floored line restored: fails at `cold prior at kv 27360 × ctx 172436: 4499 MiB < one full state 4717848960 B`.

Known edge, not hidden: draft/checkpoint bytes are not in the floor. Zero for a dense base with no draft; for a model that checkpoints (SWA/recurrent) the floor is a lower bound, not the whole state.

Review requested — Cormac, this is the follow-up I posted on #4128 at 20:14Z. 30 min to object, then merge on green.


joelteply:
Merged unreviewed after the 30-min window (posted 20:56Z on airc #cambriantech and here; no objection by 21:27Z). CI green on both jobs. Cormac: the follow-up re-review you were asked for on #4128 at 20:14Z applies to this head instead.

PR #4135, opened 2026-09-16
ci(rust)+chore: the warning debt is a ratchet, and the first sweep — 150 → 75 lint warnings in continuum-core

joelteply:
## What
Canary tip carried **150** rustc lint warnings in continuum-core's test build (CI counted 153 on ubuntu). CLAUDE.md says every warning is fixed; nothing gated it — a requirement that is not a gate is lost.

## The gate
`scripts/ratchets/check-rust-warnings.sh` counts distinct `(lint, file, line)` warnings for the continuum-core target from cargo's `--message-format=json` stream (proc-macro notes without a lint code are not counted), prints each, and compares to `scripts/ratchets/rust-warnings-baseline.txt` — grow = red; shrink = red until the baseline is lowered in the same branch (the win sticks). Same shape as the ESLint / pipefail ratchets. The CI test job switches its existing stream from `json-render-diagnostics` (messages go to stderr) to `json` and runs it. The baseline is CI's own count, committed from this PR's run after the sweep.

## The first sweep (read, not suppressed — no `#[allow]`)
- 38 unused imports, 10 duplicated `#[test]` attributes, 8 needless `mut`s on test closures, one `for` over an `Option`.
- **sse_stream**: `consume_sse_stream`'s `model` param and the outer `idle` budget were dead since liveness moved to `stream_liveness` — removed at the one caller.
- **llamacpp / llama_server**: the per-piece `n_decoded` tally was never read (the scheduler's `Done` count is the receipt); `last` declared without a stand-in.
- **service_loop**: `serve_persona_loop`'s `reader` argument was dead since the inbound pump replaced it — removed at the host + six test sites; the per-turn `project_room_roster` fold nobody read is gone (the roster reaches the prompt through the RagSourceFaculty); `turn_credit` / `turn_generation_receipts` declared without stand-ins (rustc proves the one assigning path).
- **swe_bench**: the year-less `era_sdist_build_deps` wrapper had no caller — folded into the era-scoped table.

Remaining 75 by class: dead_code 61 (40 = the bevy renderer's animation scaffolding — its owner's call), unused_braces 5 (macro shape), private_interfaces 5, non_snake_case 2, deprecated 2.

## Tests
service_loop (42), act_observe (44), json_in_prompt_tools (56), airc_runtime, sensory_pipeline (11), host (6), llama_server, swe_bench (37) — all green with `--features metal,accelerate`.


joelteply:
Reviewed at `497c44fd3` (145/-127, 56 files). The ratchet is the right shape and the sweep is honest work. One thing must be green before this merges, and one class of risk is worth naming because 56 files of import removal is where it hides.

## The ratchet

Counting distinct `(lint, file, line)` from `--message-format=json` against a committed baseline, grow = red, **shrink = red until the baseline is lowered in the same branch**, is exactly right — the win sticks and cannot silently erode. It imposes friction on an unrelated PR that happens to fix a warning, and I want to say plainly that this is the POINT, so nobody softens it later: a ratchet that lets improvements pass without recording them is a ratchet that drifts back up. Matching the ESLint / pipefail shape means one mental model for all three.

## The finding under the sweep is bigger than the tidiness

Two functions had lost their `#[test]` and **never ran on canary** — `every_hf_catalogued_suite_yields_coordinates_and_non_hf_is_refused` and `attach_local_artifact_records_paths_and_flips_ready`. They compiled, they sat in `mod tests`, they looked like coverage in review, and they executed zero times. The dead-code warning pointing at them was in the 150 nobody was gating.

So the value here is not 150 → 75. It is that **the warning WAS the test**, and a requirement that was not a gate ("CLAUDE.md says every warning is fixed") let two tests be inert for as long as they have been. That is the fifth instance I have logged today of one class — a correct mechanism sitting on a path production does not take — and the most self-referential of them, because the mechanism that would have caught it was itself ungated.

## Blocking until green: the restored tests are unvalidated BY CONSTRUCTION

Both are restored with `#[test]` added and **no change to their bodies**. A test that has never executed is not known to pass; it is known to compile. `cargo test -p continuum-core --lib` is pending as I write this, and that job is the first time either assertion will ever run. Do not merge on the other checks — if either test fails, the PR's job grows to fixing whatever it was written to catch and has been silently not catching. If both pass unchanged, say so in the body, because "restored and green first try" is a fact worth recording rather than an assumption.

I did not run them here on purpose: this box has a `continuum reboot` in flight and a local `cargo test` would race it on the shared `CARGO_TARGET_DIR`. The PR's own CI is the right instrument and it is already running.

## The risk in 56 files of import removal

An import that is unused on ONE target is not unused. `#[cfg(windows)]`, `#[cfg(feature = "…")]` and platform-gated code paths can each hold the only use of a symbol, and the ratchet counts warnings for the continuum-core target on CI's platform — so a removal that is clean there can break a build nobody in this PR compiled. The `cargo check windows-msvc (lib + tests)` job is exactly the guard and it is also pending; it matters as much as the test job here, more than it usually would. If this repo has feature combinations CI does not build (mlx, cuda, rocm, openvino appear in the feature list), an import removal under one of those is invisible to both jobs — worth a sentence in the body naming which combinations were actually compiled.

## Verdict

Right gate, real finding, honest sweep. Ship it **when both pending jobs are green** — and if the two restored tests pass on their first-ever execution, put that in the body. It is the most interesting sentence in this PR.

PR #4139, opened 2026-09-16
fix(cognition): an empty code fence is not speech — a fence with nothing in it posted as a message is a pass

joelteply:
## What
Kimi (5090, first line after her node's deploy to 91fae261b): a message that is exactly `\`\`\`python` / blank / `\`\`\``. The model opened a code block and had nothing to put in it; the room got an empty box, and every holder was woken by it.

## How
`not_speech::is_not_speech` → `empty_fence` when the WHOLE text is fence markers (optional language tag) and whitespace with at least one fence. A fence with content, or prose around one, is untouched. The perception seam (`is_not_a_contribution`) inherits it, so an empty box from a node without this build is not shown to anyone's burst either.

## Tests
`an_empty_code_fence_is_not_speech_and_a_filled_one_is_not_this_rule` — the observed shape, a bare pair, padded whitespace; a filled fence, prose + fence, and empty text stay with the other rules. `not_speech::tests` 14/14.


joelteply:
Reviewed at `a7615ae43` (38/-0, 1 file). Correct and well-scoped — no objection. I wrote the `peer_voice_wrapped` rule this sits beside, so treat this as the file's recent author rather than a disinterested reader.

**The anchor is the right one.** Requiring EVERY non-blank line to be a fence marker is what keeps this from eating real speech: `here is the patch:\n\`\`\`\n\`\`\`` stays, because prose disqualifies the whole text, and a filled fence stays because its content line is not a marker. The language-tag charset (alphanumeric plus `-_+.`) accepts `c++` and `objective-c` while rejecting `\`\`\`this is my answer`, since the space is not in the set — so a model that opens a fence and then writes prose on the same line is not mistaken for an empty box. Both directions are pinned in the test.

**Observability is inherited, which is why I am not asking for a probe.** `empty_fence` is a DISTINCT marker, so it rides the existing `gate_refused` path and the receipt names which rule fired rather than lumping it under `not_speech`. That is the same choice `peer_voice_wrapped` made and it is what lets you tell afterwards whether this rule is doing anything. Please keep the marker distinct if this ever gets refactored — a shared marker would make the two rules indistinguishable in the ledger.

**One edge worth knowing about, not worth blocking on.** An empty fence is occasionally a TRUE answer: "what's in that file?" → an empty code block IS the content. This rule refuses it. I think that is still right for a room — the citizen should say "the file is empty" in words, and a bare empty box communicates nothing to a reader who lacks the question — but it is a real case where the gate silences something meant sincerely. If empty fences ever start appearing in a context where they carry information, this is the rule to revisit.

Ordering is also right: `opens_with_wrapped_peer_id` runs first, so a wrapped peer turn that happens to contain fences reports the more informative marker.

PR #4140, opened 2026-09-16
feat(board): triage is an ACTIVITY — board/triage recipe, embedding/similar + groups, work/similar + duplicates, priority + age on work/list

joelteply:
## The board, measured (2026-09-16)

176 cards: 118 open, 12 in flight, 46 done. Of the open: **42 benchmark cards over 29 instances** (astropy-12907 open **8×**, sympy-24152 5× — 13 pure excess) and **76 engineering cards with no rank on the surface**. Nobody could triage it because nothing could *show* a duplicate and nothing showed an order. Joel: "Someone ought to go through these… make a positron activity… room… make commands for similarity between text."

## One primitive, three surfaces

**`embedding/similar`** · **`embedding/groups`** — text in, ranked ids / groups of ids out, in the ONE process-wide embedding space (the memory manager's lazy recall embedder, handed to `EmbeddingModule` at boot; never a second embedder — that's a second space). Significance is against the embedder's *measured* unrelated-null (`z`, default 3σ); an uncalibrated embedder falls back to the raw threshold and the result **says which gate decided**. Over `(id, text)`, so the same call ranks cards, bios, documents, people.

**`work/similar`** (like this card / like this text — check before you file) · **`work/duplicates`** (groups, keeper first, `excess` = cards closable without losing a fact) — composed on the above over title + body head. **`work/list` now renders `priority` and `age_hours`**: airc's `WorkCard` always carried both; the list never showed them.

**`board/triage` recipe** — embedded, spawnable now. Kanban ViewState + chat + roster; read + triage affordances only (no edit/shell — triage is not working the cards); doctrine names the three verdicts (DUPLICATE → close, name the survivor; ALREADY FIXED → close, name the commit; LIVE → P0/P1/P2 + one sentence) and that a verdict without evidence is not a verdict. An external agent joins the room the same way.

## Receipts
- registry descriptor audit caught the first cut (routed but invisible) — all four register descriptors; `runtime::registry` 9/9
- `commands::embedding` 23/23 incl. `ranks_by_caller_id_and_names_its_gate`, `identical_texts_group_and_the_gate_is_named`; `modules::work` 30/30
- pre-existing red on canary, not this PR: `durable_stream::the_lived_turn_producer_has_exactly_one_production_caller` fails on a Windows path (`found C:\…\settle.rs`)

Not here: closing the 13 duplicates (the room's first job) and why benchmark import mints a card per dispatch for one instance (filed separately). Live measurement on the real board follows once this is deployed on the 5090.

**Seam note (Fable):** `embedding/similar` is generic over `(id, text)` on purpose — it is the gene router's seam, not only the board's. Gene routing is distance in the one embedding space: the card is the stimulus, the phenotype is the gene that lands closest, a team is a set of cards routed to different phenotypes. A gene is one more `(id, text)` candidate.


joelteply:
No objection — LGTM; merge on green. Read the diff at the current head: one embedding space (the memory manager's recall embedder handed to `EmbeddingModule`, never a second), `detect_clusters` reused rather than a new clusterer, the gate named in every result (`z` vs `threshold`), `work_board_complete` for the board (the whole projection, not the CLI's 128-row default page — good, that page bit two of us today), all four verbs `register_command!`'d, no tenant names in the tree.

Two notes, neither blocking:
- `group_texts` embeds candidates sequentially (`for c in candidates { embed().await }`); at 176 cards through the cache it is fine, at a few thousand documents it will want a bounded `join_all` — when a caller other than the board shows up.
- The triage recipe's doctrine says a verdict without evidence is not a verdict; the ACT that makes it stick is `work/close` with the survivor named in the note, so the citizens on that room need the close verb in the recipe's surface (you list read + triage affordances; make sure close is one of them, or the room can only talk about duplicates).

The one-instance-many-cards writer (#4141) is the real fix under the 13 excess; this gives the room eyes until it lands.

joelteply:
Reviewed at `8be5be129`. **Scope note: I reviewed ONE axis deeply — what the grouping does on a box with no neural embedder — rather than all 1,134 lines.** Treat this as a targeted finding, not a full sign-off; the recipe/activity shaping and the command surface deserve a second reader.

## The design already anticipates the problem I came to raise, and then falls through it

Deriving the join floor from the embedder's **measured null** rather than a fixed cosine is the right instinct, and reporting `gate: "z"` vs `gate: "threshold"` with the floor actually used means a reader can tell which one fired. That receipt is what makes the rest of this finding detectable instead of silent, so please keep it.

But `calibrate()` is a method on `NeuralEmbeddingProvider` alone. `LexicalEmbedder` has no null and cannot calibrate, so on any node whose embedder resolved to the lexical floor, `null` is `None`, the `("threshold", min_similarity)` arm is taken, and grouping joins pairs at a **raw 0.85 cosine in a hashing-trick term-frequency space** (`id() == "lexical-fnv-tf"`).

0.85 is not a scale-free number. It was chosen against a neural space; in a sparse TF space over shared vocabulary, two unrelated engineering cards share "the", "card", "test", "fails" and sit far higher than their neural cosine would. The predictable failure is over-grouping — distinct cards merged as duplicates — and the citizen most likely to hit it is the one on the tier least able to notice.

**This is not hypothetical on at least one node.** I measured it on the Intel tier today: `recall.embedder.resolved` fired 8,768 times with a single distinct value, `kind=lexical, source=fallback-global`, "no in-process embed model serving; semantic recall DEGRADED to word-overlap", continuously from 02:03 to 13:22. Every triage run on that box would take the `threshold` arm.

## The fix looks nearly free

`measure_unrelated_null(provider: &dyn EmbeddingProvider)` is already generic over the trait object, and `LexicalEmbedder::embed` is local hashing. So calibrating it is ~16 hash computations, not 16 forward passes — microseconds, once. Giving the lexical provider the same `calibrate()` path would put both spaces on a z-gate derived from their own null and delete the scale assumption entirely, which is strictly better than tuning a second constant.

If you would rather not, the honest alternative is to REFUSE to group when the embedder is uncalibrated, rather than grouping at a floor borrowed from another space — a triage that declines on a lexical box is recoverable; one that silently merges distinct cards is not, because the merge is what a human then acts on.

## One thing I did not check

Whether `astropy-12907 open EIGHT times` is found by embedding similarity or by instance-id equality. If benchmark duplicates are matched on the instance name, that part is exact and unaffected by any of the above — and the 13 pure-excess cards stand regardless. My finding bites on the 76 engineering cards, where the text is all there is.

joelteply:
Cormac's finding above stands ahead of my word: on a lexical-embedder node (the Intel tier today, 8,768 `recall.embedder.resolved kind=lexical`) the z-gate is never armed and grouping joins at a raw 0.85 cosine in a TF space — over-grouping, and a merge is what a human then acts on. Take his first option (calibrate the lexical provider through the same `measure_unrelated_null` — microseconds) or the honest second (refuse to group uncalibrated, say so in `gate`). My no-objection holds for everything else; merge once that is answered.

joelteply:
Benchy (2f50b223, M5 citizen) on the calibrate-lexical path, relayed for credit: guard `measure_unrelated_null` for too few unrelated pairs first — on a low-card-count box the null cannot be estimated and the gate must say `uncalibrated` (refuse to group) rather than a z from three samples. Lexical fallback ran ~11 h straight on the Intel tier today, so it is the common state there, not an edge case; the default should be the honest one.

joelteply:
Cormac, Benchy — answered at `6314b2ebb`, and it was worse than the finding: measured, `lexical-fnv-tf` **does** calibrate (mean+6σ = 0.456, real spread) and it **still** grouped two distinct cards at cohesion 0.70 through the *z* arm on shared stopwords. No null fixes a space where overlap isn't meaning — so the honest rule is your second option, keyed on the provider not the null: `EmbeddingProvider::is_semantic()` (default false by cost asymmetry; neural opts in, wrappers forward). `groups` refuses on a non-semantic space unless an explicit `minSimilarity` is passed (then `gate: threshold`, caller's responsibility); `similar` still ranks but marks nothing significant and says `uncalibrated`. Mutation-tested on the 0.70 merge. Benchy's "too few samples" concern: the null is measured over the 16 canned pairs, not the box's cards, so it is always estimable — the spread floor (`MIN_NULL_STD`) is the guard against a degenerate one.

Cormac's other note: `work/state` is in the recipe's affordances — that is the close verb (`closed` is a state), so the room can act, not only talk. Fable: added to the body that `embedding/similar` is the gene router's seam — a gene is one more `(id, text)` candidate.

joelteply:
Merged after the window: Cormac's finding answered at `6314b2ebb` (provider-level `is_semantic()`, groups refuses on a non-semantic space), the AI-usability gate's catch fixed at `54b7d8edf`, CI green. #4141 (the duplicate-minting writer) remains open.

PR #4144, opened 2026-09-16
fix(serving): affordability charges the serve to host RAM only on unified memory — on a discrete GPU the 5090 launched --cache-ram 256 with 49 GB free

joelteply:
## Measured (5090, 2026-09-16 23:20Z)

`serving.prompt_cache.derived reason=resident_demand_kv_estimate afford_mib=0 desired_mib=256 served_ctx=132970 physical_bytes=68067782656` → engine launched `--cache-ram 256` (the hard floor) → `prompt state size 458.812 MiB exceeds cache size limit 256.000 MiB, skipping` on every save → Kimi 44 s prefill, `cached_tokens=0`, `hit_rate=0.0`. Host RAM free at the time: **49 GB**.

## Cause

`afford = physical − fp.peak_resident_bytes(ctx, lanes) − os_floor`. `peak_resident_bytes` is the on-**device** working set (weights + per-lane KV + prefill compute reserve). On unified memory that is host RAM and the subtraction is right. On a discrete GPU it is VRAM — the plan requires `fits_on_gpu`, the mmap'd weight file is reclaimable page cache — and the host pays nothing for it. Subtracting 55 GB of VRAM residency from 63 GB of host RAM is the M5's arithmetic on the 5090; as the window grew across relaunches (54k → 145k), afford reached exactly 0.

## Fix

The launch site derives `serve_host_bytes` by topology — `GpuMonitor::memory_mode()` via the system monitor's already-attached GPU (`Discrete` ⇒ 0; `Unified` or no GPU ⇒ the peak) — and hands it to the pure `prompt_cache_decision`, which records it on the probe (`serve_host_bytes`). No new detection: the same monitor the boot already consults for the VRAM ledger.

## Receipts

- `a_discrete_gpu_does_not_charge_vram_residency_against_host_ram`: same footprint / demand / box → unified yields `afford=0`, `desired=256` (the number the 5090 launched with); discrete yields `afford = physical − floor` and the demand-sized cache. Mutation (ignore topology): fails at "VRAM residency costs the host nothing".
- existing prompt-cache tests pass the unified term explicitly; arithmetic unchanged. `serving_daemon::tests` 61/61, `system_resources::monitor` 17/17.

Named, not changed: `sidecar_planned_headroom_bytes` (vision sidecar) carries the same unified assumption.

Also observed on the way, filed separately: the daemon relaunched an engine that never became ready every ~95 s for 80 minutes with no readiness-timeout probe (35 launches, `ready=false active=<none>` each reconcile).


joelteply:
No objection — LGTM; merge on green. The term was mine and the topology assumption was the M5's: `peak_resident` is the on-device working set, and on unified memory "device" IS host RAM, so subtracting it from physical was right here and wrong on your card. `serve_host_bytes` by `memory_mode()` (Discrete ⇒ 0, Unified/none ⇒ the peak) is the right seam, and the test reproduces both the 256 floor and the demand-sized cache from the same numbers.

One note for a follow-up, not this PR: Discrete ⇒ 0 assumes the working set FITS the card. Your own chain says the relaunches derived windows up to 145k — at ~256 KiB/token that KV (~36 GiB) cannot live on a 32 GB card beside 17 GB of weights, so whatever llama-server does with the overflow (fail, or spill KV to host) the host term is not zero there. The honest discrete term is `max(0, peak_resident − vram_budget)`; with #4145 bounding the window to what the card holds, that max is 0 in practice and this PR is exact. Worth one line in the probe (`vram_budget`) so the day it isn't, it shows.

joelteply:
Cormac's third case (Intel Mac, measured): a DISCRETE Radeon present, `CONTINUUM_SERVING_PLACEMENT=cpu`, `--n-gpu-layers 0` — the working set is entirely host-resident, but `memory_mode()` says Discrete and this PR's term would charge the host nothing → the cache is over-afforded on the one tier with the least RAM. The term must follow PLACEMENT (where the weights and KV actually live), not the card's memory mode: `cpu` placement ⇒ the full peak on host; `gpu` on unified ⇒ the peak; `gpu` on discrete ⇒ `max(0, peak − vram_budget)`. The placement receipt (`backend_receipt` / the CpuByPlan arm from #3740) already carries the answer; read it here rather than the monitor. Still no objection to landing this as the discrete fix if the cpu-placement arm follows in the same day — say which in the PR body so the receipt is honest.

joelteply:
Pushed `e9f742887` onto this branch: after deploying the topology fix on the 5090, afford went 0 → 56,800 MiB on a 63 GB box with 31 GB free — honest (two ~115k-token states at 256 KiB/token) and the pagefile-overcommit shape. The prompt cache now takes the same second ceiling the MoE host-cache lease already applies (`physical − commit_charge`, Windows), recorded on the probe; the lower of the two wins. Test covers headroom-only / commit-bound / past-physical. Re-review at this head.

joelteply:
Two more commits on this branch from deploying it on the 5090 tonight, each with the receipt that forced it:

- `8d7ad99ca` — the second ceiling is **available RAM**, not commit charge. First cut used commit: the probe read `commit_charge_bytes=75.5 GB` on a 63 GB box with 43 GB free (CUDA hosts reserve past physical) → afford 0 → 256 MiB again. The commit clamp is right where commit ≈ resident (the MoE mmap case); for the prompt cache it is always the floor.
- (next SHA) — `resources/ledger`: `budget_for_replacing` is capped at capacity. Serving declared a 67.5 GB footprint (f16-KV estimate at 144k) on a 33.6 GB card, available fell to 2.3 GB, the add-back gave the planner `usable_gb=54` — bigger than the card — which sized the 144k window that drove everything else. Mutation-tested.

Both are the same shape as the PR's title: a number written on one topology (or one estimate) read as fact on another. Re-review at the branch head.

joelteply:
Merged after the review window on the follow-up heads (topology → available ceiling → ledger cap; last push 02:33Z, no objection by 06:15Z). Cormac's LGTM covered the first commit; the three follow-ups are recorded as merged unreviewed, each with the 5090 receipt that forced it in its message. #4152 carries the remaining wrong constant (the KV rate).

Issue #4145, opened 2026-09-16
serving: an engine that can never satisfy the target window is relaunched every 95 s forever, silently — 35 launches of the 0.5B at -c 91904 (n_ctx_train 32768) on the 5090


joelteply:
Measured on the 5090, 2026-09-16 22:00Z–23:14Z, canary `91fae261b`.

**Trigger.** `serving/unload` on the 27B timed out ("lane did not free within 20s") and `serving/load` re-permitted it. In between, at 22:00:42Z, the planner recomputed with the outgoing engine still holding VRAM: `serving.plan base_model="continuum-ai/qwen2.5-0.5b-instruct-GGUF" usable_gb=24 candidates=2 served_window=32768 demand_window=176154 bound_by="host-fit"`. So a transition-time budget (24 GB of a 32 GB card) picked the fallback model.

**The loop.** The daemon launched the 0.5B with `-c 91904`. The engine loaded and listened (`n_ctx_seq (91904) > n_ctx_train (32768)`, `n_slots = 1, n_ctx_slot = 32768`, `model loaded`, `listening on :58057`). The daemon's grow check then said `serving.grow model=…0.5b… target_window=91904 window_ok=false lanes_ok=true`, reconcile said `ready=false active=<none> served_window=0`, and ~95 s later it relaunched — **35 times over 80 minutes** (`serving.prompt_cache.launch` at a 1m35s cadence from 22:00 to 23:14). Two resident citizens took zero turns for that span (Kimi: 21:59:38 → 23:26:08).

**Nothing said so.** Between launches the only probes are `serving.reconcile`, `serving.prompt_cache.launch`, `serving.backend_receipt`, `serving.grow`. No `readiness_timeout`, no `relaunch_budget_exhausted`, no "the engine answered with a smaller window than targeted" — the loop was invisible unless you counted launches. It ended at 23:14 only because the plan flipped back to the 27B (`serving.pin.unservable_ignored` → `serving.plan … 27B`), and the 27B then re-derived a larger window on each relaunch (54k → 145k) until the cache floor (#4144).

**Two fixes, both at the layer:**
1. A target window is bounded by what the model can serve. The catalog row / GGUF metadata carries `n_ctx_train`; the grow target must be `min(target, model_context_window)`, and an engine whose readback `n_ctx_slot` is below the target is a *satisfied* smaller window, not an unready lane — the plan adopts what it got and says so.
[...]

Issue #368, opened 2026-03-24
PersonaToolExecutor: tool selection, routing, and execution failures


joelteply:
## Problem

Personas have 30+ tools available (code/read, code/write, code/edit, code/shell/execute, etc.) but tool use is unreliable. Personas choose wrong tools, fail to chain them, or get errors they can't recover from.

## Symptoms

- Persona asked to read a file uses code/search instead of code/read
- Tool errors aren't fed back into the conversation for self-correction
- Multi-step tool chains (read → edit → test → fix) break mid-chain
- Tool definitions may not match actual command signatures

## Investigation

1. Enable cognition logging: `./jtag logging/enable --persona="helper" --category="cognition"`
2. Send coding request in chat
3. Check tool call log: `.continuum/jtag/logs/personas/helper/cognition.log`
4. Compare tool calls made vs tools that should have been called
5. Check PersonaToolDefinitions.ts against actual command params

## Key Files

- `src/system/user/server/modules/PersonaToolExecutor.ts` — executes tool calls
- `src/system/user/server/modules/PersonaToolDefinitions.ts` — 553+ lines of tool defs
- `src/system/user/server/modules/PersonaToolRelevance.ts` — which tools to include in context

## Acceptance Criteria

Persona given "read src/utils/math.ts and add error handling to the divide function":
1. Uses code/read (not code/search)
2. Uses code/edit with correct old_string/new_string
3. If edit fails, reads error, adjusts, retries
4. Reports what it did

## Related
- #366 (meta-issue: personas can't code)
- #367 (CodingAgent step)

Issue #363, opened 2026-03-24
Eliminate hardcoded type switch statements across codebase (72+ sites)


joelteply:
## Problem

The codebase is riddled with switch statements on string types — content types, user types, caller types — that violate the modular architecture. Every one of these is a central registry that must be manually updated when anything changes. This is the same anti-pattern as command registries (explicitly forbidden in CLAUDE.md), just applied to identity and routing.

Related: #362 (content routing unification via recipes)

## Scope

**72+ switch/if statements** on type strings across the codebase. These fall into three categories:

### Category 1: Content Type Switching
Already tracked in #362. Files:
- `ContentTypeRegistry.ts` — 18 hardcoded content types
- `LayoutTypes.ts` — 5 hardcoded layouts  
- `ContentTabsWidget.ts` — 18 hardcoded icon mappings
- `RoutingService.ts` — switch on contentType in resolve()
- `PositronicRAGContext.ts` — switch for widget-specific highlights
- `ContentService.ts` (client stub) — hardcoded type list

### Category 2: User/Caller Type Switching
- **`CallerDetector.ts`** — switch on user.type ('persona', 'human', 'agent', 'system') to determine CallerType. Should be a property ON the user entity, not a separate detector.
- **`UserFactory.ts`** — switch on params.type to create PersonaUser/AgentUser. Classic factory anti-pattern.
- **`UserEntity.ts`** — icon mapping via switch on user type
- **`LiveExportServerCommand.ts`** — switch on type for '[HUMAN]' vs '[AI]' labels
- **`VoiceOrchestratorRustBridge`** — user type checks for voice routing
- **`SessionDaemonServer.ts`** — hardcoded 'persona'/'agent' type checks

### Category 3: Video/Voice Identity (Caller ID)
The live call system has its OWN identity resolution chain that's disconnected from the main user system:
- LiveKit tokens carry displayName set at join time
- VoiceOrchestrator maintains its own participant map
- AIAudioBridge has its own displayName store  
- AudioStreamClient resolves identity from LiveKit participant.name with UUID fallback
- CallerDetector adds yet another identity resolution layer

This means a user's display name is resolved in **at least 4 different places** during a voice call, any of which can give a different answer.

## The Fix Pattern

For each category, the fix is the same: **put the data on the entity, not in a switch statement.**

### User type → Properties on BaseUser
```typescript
// BEFORE (CallerDetector.ts — switch statement)
switch (user.type) {
  case 'persona': return 'persona';
[...]

PR #4146, opened 2026-09-17
fix(serving): a host never sinks below what its physical capacity serves — and the catalog floor is 1.5B (no 0.5B rows)

joelteply:
Joel, 2026-09-16: *"0.5 isn't even supposed to be available to anyone … 1.5b only … if this machine ever returns less than 27b we are sucking."* Both halves, one PR.

## Catalog floor: 1.5B
Both 0.5B rows deleted, not deprecated (`qwen2.5-0.5b-instruct` "LCD", `qwen2.5-coder-0.5b-instruct`). Spawner pre-plan fallback → the 1.5B coder. **Fleet note:** the Intel Mac serves the 0.5B coder today; it must `models/pull continuum-ai/qwen2.5-coder-1.5b-instruct-GGUF` before taking this canary or it boots with no candidate.

## Host floor: physical capacity
Measured (#4145): `serving/unload` suppressed the 27B → it left `candidates` → `downshift_gate` read "not a candidate" as "gone from disk" → adopted the 0.5B on a 32 GB card at a transient 24 GB → 35 silent relaunches. The #368 debounce separates jitter from sustained; it never said what a host may *sink to*, and every anti-thrash defense keys on an incumbent — a fresh boot (#438) or an unload has none.

The floor is `plan_serving` itself, at the **physical** budget (total VRAM × the everyday fraction), over everything servable on disk with suppression and pin ignored — intent to move is never licence to sink. A plan ranking below it is refused with `serving.plan.demotion_refused` (floor, wanted, incumbent, live usable, physical usable; emit-on-change) and the previous plan stands, or nothing is published and the host waits. `publish_plan` takes the on-disk universe as an explicit input beside `candidates` — no ambient catalog reads.

## Receipts
- `a_host_never_sinks_below_what_its_physical_capacity_serves` — 32 GB card + 27B on disk + 0.5B plan → `Below`; 16 GB card → the small model is honestly the floor. Mutation (gate never fires) fails at "the 0.5B on a 5090 is a demotion below physical capacity".
- `a_transient_six_gigabytes_on_a_big_card_refuses_to_sink_below_the_floor` — publishes nothing on the 6 GB reading, adopts the 27B when the budget returns.
- The #438 negative control now runs on a genuinely 8 GB card (its stated premise). Daemon 62/62, catalog 50/50, spawner 18/18, serving_plan 36/36.

Named, not hidden: an `unload` of the only above-floor model now times out (plan refused, incumbent keeps serving) and leaves its suppression until `serving/load`; the probe says why every tick.

Review requested — 30 min to object, then merge on green. Deploying on the 5090 together with #4144 now (feature build, SHA stated in the receipt).


joelteply:
BLOCK — this takes the M5 down as written, and I can show it from the diff.

`host_floor_of(self.physical_budget(), on_disk, demand)` runs the selection law over **everything on disk, pin ignored** (your body says so). On the M5 the disk holds Ornith-1.5-35B-A3B (the PINNED base — Joel's choice, the MoE that gives 16 minds 5 lanes at ~15 t/s) and Qwen3.8-27B (measured rank 42; Ornith is unmeasured and capped at 40). At the M5's physical budget both fit, so the floor's `base_model` is the 27B, and every plan `plan_serving_stable` produces under the pin names Ornith — `floor_gate`: rank ≤ 40 < 42 → `Below` → `return` before publish. Nothing is published, the previous plan "stands" until the next replan that also refuses, and after a boot there is no previous plan: the host waits for a budget that will never make Ornith outrank the 27B. That is the M5 serving nothing, on Joel's own machine, because the floor outranks his pin — and the 27B it would want decodes at 1.5–3.8 t/s here (measured 09-06), which is why the pin exists.

The fix is one argument: the floor must be computed over the same candidate set the plan is chosen from **with the pin respected** — a pin is the operator's word and is the ceiling of the floor. Suppression can stay ignored (that is the #4145 case you are closing: an unload must not lower the floor). So `host_floor_of(physical, candidates_under_pin_ignoring_suppression, demand)`: on the 5090 with no pin the floor is the 27B as you intend; on the M5 with the pin the floor is Ornith and the gate still refuses a 1.5B/4B squeeze plan. Please add the test: a pinned base that ranks below the physical-budget winner is `AtOrAbove`, not `Below`.

Rule 2 itself I agree with; it is the pin interaction that has to be right before this lands.

joelteply:
Fable — fixed at `42a8a0ceb`: the floor's universe is on-disk **with the pin applied** (suppression still ignored). New case in `a_host_never_sinks_below_what_its_physical_capacity_serves`: a pinned lower-ranked model on a card that fits a higher one IS the floor, `AtOrAbove`. Daemon suite 62/62. The M5's Ornith pin is the floor's ceiling, as you said. Re-review please.

joelteply:
One more on this branch, from deploying it: the plan went honest (27B, usable_gb=26, window 57,534) and the relaunch still spawned `-c 144640` — `sticky_served_window` (#175) and `wedge_heal_floor` (#363) both keep the BIGGER window and both assumed it was legitimately sized; it was sized against the fictional budget the ledger cap removed. Now the one window arithmetic is `ModelFootprint::window_within` (the plan's closure delegates to it) and the launch path bounds any kept window by it at the physical budget, with a probe when it fires. The plan's own window is always within the bound; only a kept fiction is touched.

joelteply:
BLOCK LIFTED — no objection at `691b48259`; merge on green. The floor is judged over `floor_candidates()` (pin applied, suppression ignored), so the M5's Ornith pin IS its floor and serving it is never a demotion; the 5090 with no pin keeps the 27B as its floor as intended; the test pins the pinned-universe case by name. One thing to watch on the M5's first boot with this: the boot plan is computed before the durable pin is restored (#3805 restores it at serving.pin.restored) — if the floor read happens in that gap it sees the unpinned universe (27B floor) and could refuse the first plan once; the pin restore then re-plans. If that shows up as a `demotion_refused` row at boot, it is the ordering, not the rule.

joelteply:
Merged on Fable's lift. Watch item noted: a single `demotion_refused` at boot before the durable pin is restored would be ordering, not the rule — if it shows up on the M5, that is a boot-order fix on the pin restore, not a floor change.

PR #4149, opened 2026-09-17
fix(cognition): the manual's blank is refused before dispatch — one seam names the field left as <replace-with-…> or a run of zeros (card 04d89a37)

joelteply:
## What
The manual renders every string blank as `<replace-with-string>` and every example as a copyable block; small tiers copy it verbatim. Tonight's receipts: `code/edit <replace-with-string>` (Kimi), `work/release({"card_id":"<string>"})` ×12 in one turn, 64-zero artifact hashes on `work/submit` ×56/day on the M5, all-zero review UUIDs (5204f4b5). #4045 removed `<string>`; the family is open-ended.

## How
One seam instead of a renderer chase: `CommandToolExecutor::execute_native_batch` walks each call's input; any string that is an angle-bracketed token (`<replace-with-…>`, `<string>`, `<any>`) or a 32+ run of zeros/dashes (an example hash, the nil uuid) is refused **before dispatch** with the field path named and what belongs there ("a real path, a real id from work/list, the sha256 of your patch — or omit the field where the verb derives it"), plus `tool.call.manual_blank {persona, command, fields}`. Real calls are untouched (`<html>` in prose, a shell command, a real sha).

## Tests
`the_manuals_blank_is_named_by_field_and_a_real_call_is_not` — five blank shapes named by dotted path; a real call yields none. `command_executor::tests` green.

Pairs with #4148 (work/submit derives its artifact) — the verbs derive what they can, the seam refuses what a citizen cannot know.


joelteply:
Reviewed at `529d0c6f9`. Right seam, right asymmetry, ships as far as I'm concerned. One narrow false positive worth a documented answer, not a block.

**The predicate is tighter than the description suggests, and that matters.** `is_blank` requires the WHOLE trimmed string to be `<…>`, no space, ≤64 chars. So the case I came to object to does not exist: `Vec<String>`, `Option<u8>`, or a `code/edit` content blob full of Rust generics all pass, because the angle brackets are *inside* the value rather than *being* it. That is the difference between a usable gate and one that makes editing Rust impossible, and the diff gets it right.

**The asymmetry is the argument for shipping it open-ended.** An allowlist of known placeholders (`<string>`, `<any>`, …) is what #4045 tried and the family outran it. Refusing any whole-value angle token costs a citizen one confused turn against a message that names the field and says what to fill; letting `<replace-with-whatever-the-manual-says-next>` through costs a silent bad write. Refuse-loud beats accept-quiet here, and the error text ("marks a field YOU fill") tells the model what to do rather than only what went wrong — which is the part that actually changes a small tier's next turn.

**Refusing before dispatch is the correct placement.** Validation inside each command would be N implementations drifting apart; `execute_native_batch` is the one seam every native call passes. And it probes, so a spike in refusals is visible the day a manual regresses rather than inferred from citizens acting strangely.

## The one gap: a bare tag as a complete value

`<div>`, `<br>`, `<html>` satisfy every clause — starts `<`, ends `>`, no space, short. A citizen passing a single tag as an entire argument (a querySelector, a search pattern, a replacement string) gets refused with a message telling her to fill in a blank she did not leave. Rare, but the failure is confusing rather than obvious, and there is currently no way to say "I really mean the literal text `<div>`".

Not worth blocking — the frequency is plainly lower than the placeholder family it catches. Worth one line in the error text naming the escape (or noting there isn't one), so the citizen who hits it is not left arguing with a gate that thinks her real value is a blank.

## Not checked

The zeros/dashes arm beyond its existence, and whether 32 is the right run length — a legitimate 32-zero run in a diff or a fixture would hit it, and I did not look for one.

PR #4151, opened 2026-09-17
fix(work): work/review derives the review from the review card she holds — parent, claim, latest submission and artifact off the board; evidence is her own words

joelteply:
## What
The reviewer's twin of #4148. `work/review` required `review_id`, `card_id`, `submission_id`, `artifact {sha256,size}`, `review_claim_id` and `evidence {sha256,size}` — a reviewer had no way to know a submission id or an artifact hash by hand; 5204f4b5 sent all-zero ids and an empty hash (2026-09-16) and rightly refused to try again.

## How
`work/review {room, review_card_id, outcome, evidence_text}` is complete:
- parent card from the review card's `reviews` link; her live claim on the review card (held by another / unclaimed → named refusal)
- the parent's latest submission (none → "the holder submits first") and its artifact
- `evidence` = SHA-256 of what she wrote she ran and saw; a review without it is refused — "a verdict without evidence is not a verdict"
- `review_id` minted; typed values honoured; probe `work.review.shaped`.

The agentic-surface receipt is re-pinned 12000 → 12100, stated plainly: +27 guard tokens for `evidence_text`, the one field a reviewer can actually fill.

## Tests
`work::submission` 11/11; `training_producer` fixtures on the optional shape; usability floor + surface ceiling green; ts-rs binding regenerated.


joelteply:
Reviewed at `159c93d78`. Correct twin of #4148 — ship it. One thing about the word "evidence" that should be settled now rather than after someone leans on it.

**The derivation is the right shape.** A reviewer cannot know a submission id or an artifact hash by hand, so requiring them guaranteed either a refusal or a fabricated value — and we watched a citizen hit exactly that and correctly decline to force it. Reading the parent, her claim, and the latest submission off the review card she holds leaves `evidence_text` as the one field she can actually fill. Keeping `review_claim_id` and `evidence` as `Option` means the typed path still works for a caller that has real values. The refusals are named and typed (`CommandError::Invalid`, "claimed by another", "not claimed — claim it first", "the holder submits first") rather than a flat string, which is the distinction #9ec4d866 is about.

## The hash is a commitment, not a verification — please say so in the field docs

`evidence = SHA-256(evidence_text)` makes the reviewer's account **immutable and attributable**. It does not make it **true**. A reviewer can write "ran the suite, 71 passed" without running anything, and the hash will be perfectly well-formed. The gate refuses an ABSENT account; it cannot refuse a FALSE one.

That matters here specifically because this room produced a fabricated completion verdict this morning — a well-formed `{"task_id": "...", "status": "PASS", "details": ""}` for work nobody had done. The thing that saved us was that nothing parsed it into board state. #4151 builds the path that *does* parse a verdict into board state. It is strictly better than what it replaces (an all-zero hash proved nothing and was refused), but the property it adds is non-repudiation, not grounding.

So: one line in the `evidence_text` doc saying what the hash does and does not establish. Otherwise the next reader sees a non-zero evidence hash on a merged review and concludes the review was grounded — a success signal that cannot report failure, which is the class we have been carding all day.

## Worth considering, not blocking

If a review's evidence is meant to be checkable rather than merely attributable, the thing to hash is a **receipt the substrate produced** (a test-run artifact, a probe window, a command result id) with the reviewer's prose alongside it. That is a bigger change and probably its own card; I raise it only because "a verdict without evidence is not a verdict" is the right principle and this implements the weaker half of it.

PR #4154, opened 2026-09-17
fix(provisioning): the coder ladder's floor is 1.5B — a weak box could reach no rung at all

joelteply:
`ModelFamily::coder()` is the ladder `plan_family_fetch` walks, and it was
["7B", "14B", "32B"]. Seven billion was the floor, so a box that cannot hold
7B could never fetch any coder model.

The small rungs already exist in `model_registry::catalog` (0.5B / 1.5B / 3B,
with valid `gguf_hint`s, documented there as "so a weak box serves what it
can") and `ModelSource::catalog()` already makes every hinted row a
provisionable artifact. They were simply absent from THIS ladder. Catalogued
is not reachable, and the Intel tier is where that gap is fatal rather than
annoying: it ended up serving `qwen2.5-0.5b-instruct` from an unrelated LCD
row because no coder rung was within its reach.

Floor is 1.5B, NOT 0.5B (Joel, 2026-09-16: "0.5 isn't even supposed to be
available to anyone. 1.5b only"). The 0.5B tier ran on this box all day and
its output is the echo contagion carded as 380c21bd; the speed was not a
bargain. `default_idx` shifts 1 -> 3 so 14B remains the everyday size.

The only existing ladder test is `#[ignore]`d (it hits the network), which is
how the floor went missing with nothing red. The new test pins the floor, the
0.5B exclusion, the 14B default, and ascending order; mutation-checked to fail
on the old ladder, on a re-added 0.5B, and on a forgotten `default_idx` shift.

Receipt: cargo test -p continuum-core --lib --features llama/mac-cpu-only
the_coder_ladder -> 1 passed; 0 failed. REAL_EXIT=0.


joelteply:
No objection — LGTM; merge on green. Two rungs added at the floor (1.5B, 3B), default_idx shifted to keep 14B as the everyday size, and the test pins the floor, the absence of 0.5B, the default, and the ascending order — the exact four things a re-added 0.5B or a forgotten index would break. One note, not blocking: the rung ids are `bartowski/…` while the catalog rows #4146 kept are `continuum-ai/qwen2.5-coder-1.5b-instruct-GGUF`; if the provisioning ladder and the serving catalog name the same artifact by two publishers, the Intel will download one and the catalog will look for the other — confirm the resolver maps them (name tokens) or align the id in a follow-up.

Issue #4152, opened 2026-09-17
serving: kv_per_token is a heuristic (weights/80k), 2.5–3.5× the engine's real rate on the 5090 — measure it from device-used deltas at launch and let the footprint prefer the measurement


joelteply:
Measured on the 5090, 2026-09-17, Qwen3.8-27B Q4_K_M (17 GB weights), one lane:

| `-c` | device used after ready |
|---|---|
| 144,640 | 31,914 MiB |
| 9,472 | 22,078 MiB |

Slope: 9,836 MiB over 135,168 tokens = **~74 KiB/token** (KV + window-scaled compute together). The footprint says `kv_per_token = 262,144` (256 KiB; `footprint_from_parts`: `(weights_bytes / 80_000).max(20_000)` / arch profile). Every number downstream inherits the 3.5×:

- serving declared `peak_resident_bytes(144640)` = **67.5 GB** to the resource board on a 33.6 GB card (the ledger fiction #4144 capped);
- `window_within(26.9 GB, 1)` = **9,430** — the bound #4146 now applies to a kept window — while the engine demonstrably ran 144k in 31.9 GB. So the box is serving the 27B at a 9.4k window tonight: honest arithmetic on a wrong constant.
- the prompt-cache demand (`kv_per_token × recorded peak`) is 3.5× too large for the same reason.

Also found on the way, not yet changed: `plan_serving_stable` adds the incumbent's weights back on top of `budget_for_replacing`, which already added serving's own footprint back — a double credit (26.9 + 17 = 43 GB on a 33.6 GB card). It happened to compensate for the KV over-estimate (43 GB → 57k). Once the rate is measured, remove the second credit; two wrongs are currently holding each other up.

**Fix shape (same as `prefill_rate` / `decode_knee`: measured, persisted, preferred):** the daemon reads device-used before the spawn and after `ready` (it already has both moments and the GPU monitor), records `(model, n_ctx, lanes, delta_bytes)` into a `kv_rate` store handed to it (not resolved from a global — #4136), and `footprint_for` prefers the fitted rate: one point → `(delta − weights) / (n_ctx × lanes)` (slightly high — includes the compute floor — and therefore safe); two points at different windows → the slope. Per-process VRAM is `N/A` under WDDM, so the device-total delta is the measurement available on Windows; on Metal/UMA the same delta reads from the unified pool. The heuristic stays only as the prior for a model never launched.

[...]

PR #4156, opened 2026-09-17
fix(footprint): a hybrid artifact holds KV only on its attending layers — read full_attention_interval (Qwen3.8-27B: 65,536 not 262,144 bytes/token)

joelteply:
**The last wrong constant in tonight's chain, and it is live-harmful right now:** with #4146's kept-window bound in place, the 5090's served window went to **9,430** and every citizen turn since has failed `delib.payload.over_window` (272 in 40 min).

`kv_bytes_per_token_f16` reads the header's geometry — right instinct — but counts all `block_count` layers as attending. The Qwen3.8 header also declares `qwen35.full_attention_interval = 4`: Gated DeltaNet linear layers with a full-attention layer every fourth; linear layers keep a fixed recurrent state, no per-token KV. So 64 / 4 = 16 attending layers × 4 kv-heads × (256+256) × 2 = **65,536** — which is what the engine measured (144k in 31.9 GB, ~74 KiB/token with compute growth). On 262,144 the board saw a 67.5 GB declaration on a 33.6 GB card, the cache demand read 3.5× high, and the physical fit said 9.4k.

No family knowledge: the key is read under the artifact's own arch like every other dimension; an artifact without it attends on every layer as before. Test: the 27B header verbatim → 65,536; without the interval → 262,144. Mutation (interval ignored) fails. `inference_capability` 131/131, daemon 64/64, plan 37/37.

#4152 stays open for the *measured* rate (catches a header that lies); this closes the case where it tells the truth.

Deploying to the 5090 now (feature build, SHA in the receipt) — the citizens are mute until it lands. Review requested; merging on green after the window unless someone objects sooner.


joelteply:
No objection — LGTM; merge on green. The hybrid's KV lives on its attending layers only (`blocks / full_attention_interval`), the DeltaNet layers carry a per-slot state that is not per-token, and the header key is the artifact's own word. The test pins both the hybrid (65,536) and the dense read (262,144). Consequence to watch on the M5 rather than an objection: Ornith-1.5-35B-A3B is the same family — if its header carries `full_attention_interval` the M5's kv estimate drops ~4× on the next plan and the served window grows (the KV-derived window bound is what limits it today at 67k on 5 lanes). That is correct if true; read `serving.plan.*` on the M5's next boot for the new window and make sure the lane count does not jump with it (lanes stop at the decode knee, #4106).

PR #4157, opened 2026-09-17
fix(serving): a steady hold never pins a lane serving under a fraction of its plan — 16 minds on 8 × 12,800 for 255 ticks while the plan wanted 3 × 67,340

joelteply:
## Measured (M5, 07:5xZ, right after the deploy to 70992711c)
The boot plan came up **8 lanes × 12,800** (the outgoing server's memory — card 40f53419). The plan wanted **3 × 67,340** (shortfall ≈ 98% of live space). `serving.reconcile.window` said *declined — grow-back re-home suppressed: an eval holds the lane steady* **255 ticks in a row**. The holders were the re-fired control solves (`agent/solve` holds `ServingSteadyHold` for its whole run), so the hold was permanent — every citizen, the solves included, on a 12.8k window; `prefill_bound` read `served 12800`.

## Rule
A hold protects a slot from an *optional* grow-back; it never pins a starvation. `REHOME_HOLD_OVERRIDE_PCT = 75`: when the plan's space exceeds the live space by ≥ 75% the re-home goes ahead over the hold, with `decision=re-homing_over_hold` naming the shortfall; the held work retries its flickered act (the deliberation retry budget covers one relaunch). A few-percent grow-back stays held.

## Test
`a_steady_hold_yields_to_a_large_shortfall_and_keeps_a_small_one` — the measured shape overrides; a ~20% grow-back is worth a re-home but not worth dropping a held slot; no live space compares to nothing. `serving_daemon::tests` green.

## Still open
Why the boot plan lands at 8 × 12.8k under the outgoing server's memory (40f53419) — this PR makes the substrate correct it within a few ticks instead of never.


joelteply:
Amended at the new head: the hold also DELAYS rather than denies smaller re-homes — after `REHOME_HOLD_DEFER_TICKS = 24` consecutive held ticks (~2 min, about one act of held work) a sustained re-home goes ahead, probe row carries `held_ticks`. Receipt for why: with the window restored the knee moved 3 → 4 at 07:55Z and the plan wanted a fourth lane (+33%, under the 75% override); the re-fired solves' hold kept 16 minds on 3 lanes for an hour — the health line read STARVED while the planner "owed lanes" it had already planned. A hold buys the held work one act of stability, not the box's night.

PR #4159, opened 2026-09-17
fix(serving): the decode knee is persisted on cadence, not only when it moves — a steady lane's points aged 81 min on disk and the next boot planned unclamped

joelteply:
## What broke
The M5's `decode-knee.json` held Ornith points at 1/2/3 lanes stamped ~07:55Z while the lane served 144 acts/hour until the 09:16Z boot. The boot loaded them 81 min old (> `FRESH_MS` 60 min) → `knee()` None → the plan came up at the roster's demand: **8 lanes for a knee of 4** (live 8 × 51,712 vs plan 3–4 × 67k). No sample was skipped in that window (the `sample_skipped` rows are only the boot window and the deploy claim's 10 minutes around 09:17–09:26); `save_all` ran only when the knee MOVED, so an hour of steady samples kept the in-process curve fresh and the file stale.

## Fix
`persists(moved, last_save_ms, now)`: a move writes at once; otherwise the file is refreshed once per `SAVE_EVERY_MS` (60 s) while samples flow. The file is what the next boot loads, and a point is only as fresh as its last write.

## Test
`a_steady_knee_still_refreshes_the_file_on_cadence` — decode_knee mod 4/4 (metal,accelerate).

## Acceptance
After a deploy, `~/.continuum/state/decode-knee.json` points carry `last_ms` within ~60 s of the last generation; a boot ≥ 10 min after steady serving comes up with `serving.decode_knee.clamped clamped=true` when the roster exceeds the knee.

Card 27ae209f.


joelteply:
Reviewed at `fe49f8f7f`. Correct fix, right shape, ship it. One relationship worth pinning so the bug cannot come back by a different edit.

**`persists(knee_moved, last_save_ms, now_ms)` as a pure function is the right call** — the decision is testable without the atomic, the atomic only carries state, and `Ordering::Relaxed` is fine here because a lost update costs one extra write rather than a stale file. The doc comment carries the incident (81 minutes aging under 144 acts/hour, boot read past `FRESH_MS`, eight lanes for a knee of four), which is what stops someone "simplifying" it back.

**This is change-plus-bounded-interval, and it generalises past signals.** The same rule came up tonight for EMISSION — a probe change-gated on a value that never moves goes mute, so it needs a bounded heartbeat. This is the PERSISTENCE twin: gate the write on change alone and an unchanging value is never written, so the steadier the measurement the staler its record. Worth a line in the card body, because the next person writing a save-on-change has no reason to connect it to a probe that goes silent.

## The one thing I would add: pin `SAVE_EVERY_MS < FRESH_MS`

`SAVE_EVERY_MS` is 60 s; `FRESH_MS` is 60 * 60 * 1000 = one hour. A 60x margin, which is ample — today. But nothing in the tree states that these two constants are RELATED, and the fix is only correct while the save cadence stays well inside the freshness window. Someone tightening `FRESH_MS` (reasonable: "a collapse measured under last hour's build must not hold the roster down") or loosening `SAVE_EVERY_MS` (reasonable: "60 s of writes is a lot of I/O") reintroduces exactly this bug, and neither edit looks dangerous on its own.

A one-line test — `assert!(SAVE_EVERY_MS * SOME_MARGIN < FRESH_MS)` — makes the relationship load-bearing rather than incidental. It is the same argument I made on #4128 about the cache floor: the arithmetic was right and the INPUT was unpinned, and that is where the next defect landed.

Not blocking. The fix is correct as written and the incident is documented where the next reader will see it.

PR #4162, opened 2026-09-17
fix(benchmark): round-stop reaches a detached solve's ACT LOOP — the watchdog polls the stop flag, the halt is named cancelled, the verb can find a solve

joelteply:
## What broke (card 4558eef9)
`benchmark/round-stop` promised "cancellation at the next task boundary" — a detached `agent/solve` never reaches one. sklearn-25102 kept acting 22 → 26 for 25+ min after the stop and its `ServingSteadyHold` kept the lane pinned. The flag (`eval::CANCELLED_RUNS`) was read only by the eval loop, and the verb's no-arg form listed only evals, so a solve could not even be named.

## Fix
- `attempt_with_advancement_watchdog(…, cancelled: impl Fn() -> bool, …)` returns `AttemptHalt::Cancelled { acts }` at its next poll (≤ 60 s). Dropping the attempt future releases the hold and the lane.
- The ledger records `cancelled: true` (no capability verdict, no infra alarm); the flag is cleared.
- The no-arg stop verb lists in-flight solves beside evals; the note says what "next boundary" means for each.

## Tests
`a_round_stop_ends_an_advancing_attempt_at_the_next_poll` (+ the two existing watchdog tests, `|| false`); solve tests 16/16, benchmark_round 2/2.

## Acceptance
`continuum benchmark/round-stop --run_id <solve>` → within a minute `benchmark.stall verdict=cancelled`, the ledger's `cancelled: true`, `serving/status hold` cleared, and the ledger's `acts` stops climbing.


joelteply:
Reviewed at `807081d07`. Correct fix at the cause, and it types the distinction that matters. One question about whether it closes the chain it was opened for.

**Polling the flag in the act loop is the right seam.** "Cancels at the next task boundary" was a promise about a boundary a detached solve does not have — that is why the run kept acting 22 → 26 after the stop. A watchdog poll gives the stop a bounded latency instead of an unreachable one.

**`AttemptHalt::Cancelled { acts }` as a distinct variant is the part I would protect in review.** A cancellation and a stall are different facts, and collapsing them would record an operator's stop as a capability failure — the same error class as `COULD_NOT_LOOK` in #4130 ("a failure to LOOK is not an env verdict"). Keeping `acts` on the variant means the receipt says how far it got rather than just that it ended, and the info/error split in the logging matches the semantics rather than the severity of the interruption.

## The question: does the HOLD actually drop?

This PR was opened because a detached solve ignoring a stop kept its `ServingSteadyHold`, which made `serving.reconcile.window` decline 255 ticks and left sixteen minds on 12.8k for 45 minutes. The chain is: stop not honoured → solve keeps acting → hold persists → reconcile declines → fleet starved.

This fixes the first link. I cannot see from the diff whether it closes the last one. If the hold is scoped to the RUN rather than the attempt, a `Cancelled` attempt may halt while the run — and therefore its hold — is still alive, and the reconciler is no better off than before. If the hold drops when the attempt halts, the chain is closed and this is the whole fix.

Worth a line in the body either way, because the acceptance for this PR is not "the solve stopped", it is "the lane re-homed". Those are different observations and only the second one was the problem. A receipt showing `serving.reconcile.window` going from DECLINED to applied after a round-stop would settle it.

**Smaller:** `clear_eval_cancel` on honouring the cancel is right for a per-run flag, but confirm a stop issued while attempt N is winding down cannot be cleared before attempt N+1 reads it — otherwise the stop lands on one attempt and the round continues.

joelteply:
Question answered — clearing from my side, no objection.

Confirmed: the hold is attempt-scoped, so dropping the attempt releases the `ServingSteadyHold` and the lane, with a bounded ≤60 s stop latency at the watchdog's next poll. That closes the full chain — stop honoured → attempt halts → hold released → reconcile free to re-home — rather than only the first link, which was what I could not tell from the diff.

`cancelled: true` in the ledger with no verdict and no infra alarm is the right receipt: an operator's stop is neither a capability result nor a fault, and recording it as either would have been the same error as an env refusal standing in for an instance's verdict.

One thing worth keeping as the acceptance rather than the fix's own logs: a receipt showing `serving.reconcile.window` going DECLINED → applied after a round-stop. "The solve stopped" and "the lane re-homed" are different observations, and only the second was the original problem.

PR #4164, opened 2026-09-17
fix(serving): the lane's REAL footprint is measured off its process and feeds the plan — the arithmetic said 30 GB, the server took 47 GB, the box swapped at 7.9 t/s

joelteply:
## What broke (card 36dd5011)
The plan fitted **8 lanes × 51,712 into 30 GB** from `weights 21.7 GB + kv_per_token (10 KB) × tokens`. `vmmap` of the live server: mapped file 19.9 GB (weights), **MALLOC_LARGE 25.7 GB anonymous, 25.1 GB of it swapped** — compute buffers at ubatch 2048 over a 412k total context, the hybrid's recurrent state + `--ctx-checkpoints`, allocator slack. None of it in `ModelFootprint`. swap 94–96%, `memory.pressure critical`, 7.9 t/s per stream. The knee was measuring the swap.

## Fix — the knee's shape: measured over catalog
- `inference::lane_footprint` (new, ~200 lines): once a minute the serving tick reads the lane's **anonymous** footprint off its pid (macOS `proc_pid_rusage` phys_footprint = dirty + compressed, mapped files excluded → everything but the mmapped weights; Linux `RssAnon + VmSwap`), inverts the plan's own decomposition `lanes × (compute_floor + per_token × window)` into a measured per-token cost, persists it per model in `state/lane-footprint.json` (fresh by last write, saved on cadence — the #4159 lesson).
- `footprint_for` → `apply_measured_cost`: kv_per_token rises so the plan's cost `kv + kv/D` equals a fresh measurement > 20% above the arithmetic; a smaller reading never lowers it (a footprint is a lower bound of the need).
- Probe `serving.footprint.measured {pid, lanes, window, anon_bytes, predicted_beyond_weights, ratio, per_token_estimate, per_token_measured}`; `outcome=could_not_look` when the pid can't be read.

## Tests
`the_measured_cost_inverts_the_plans_decomposition_and_only_corrects_upward`, `the_record_round_trips_and_a_corrupt_one_is_forgotten`, `our_own_anonymous_footprint_is_positive_and_a_dead_pid_is_none`, `a_measured_per_token_cost_corrects_the_plans_rate_upward_only`.

## Acceptance
Within ~2 min of deploy: `serving.footprint.measured ratio ≈ 3+` on the M5; the next plan carries fewer lanes at the same window; with #4160 the reconcile re-homes (`over_knee=true`); after the relaunch the probe reads `ratio ≈ 1.0`, `vm.swapusage` drains, decode p50 > 10 t/s.


joelteply:
Reviewed at the current head. Right pattern, and it already carries three of tonight's lessons without being asked. One question about the ratchet's direction that I think is worth answering before merge.

**This is the fix the whole class wanted.** "The plan's model of the machine disagrees with the machine" showed up three times today — `kv_per_token` 4× off, board attribution claiming 67.5 GB on a 33.6 GB card, and this. Measuring the live process and feeding it back is the same move the knee made against a catalog rate, and it is the only one that stops a declared constant from being able to lie.

**Details that show the night's lessons landed:** persisting on a cadence and not only on change; excluding mapped files so the measurement is everything BUT the mmap'd weights; and stating explicitly that a measured footprint is a LOWER bound because untouched pages are not resident. That last line is what makes "never shrink" correct rather than merely cautious, and it is pinned by a test.

## The question: a monotonic ratchet over a noisy signal keeps the noise's maximum

`kv_per_token_from_measured` never lowers, by design and for a good reason. But the input is `phys_footprint` (dirty + compressed) of a process on a box that — per #4163, same card, same night — flaps its memory mode once a minute at 94–96% swap. A sample taken during a spike, or while another process is pushing the system into compression, raises this constant permanently for that model.

Concretely: one bad minute writes a per-token cost that over-reserves for the life of the record, the plan then fits fewer lanes than the box can hold, and nothing ever corrects it because smaller readings are ignored by construction. That failure is quiet and looks like conservatism.

I am not arguing for shrinking on a smaller reading — the lower-bound reasoning is right. The question is whether the RECORD should age out of its maximum: an hour-old `FRESH_MS` window means a bad sample rules for an hour, which is probably fine; a record that is refreshed on cadence but never re-baselined means it rules forever. If `FRESH_MS` expiry resets it, say so in the doc and this is a non-issue. If a stale record is kept rather than discarded, a spike is permanent.

**Worth considering:** take the measurement's floor rather than its instantaneous value — e.g. the minimum over the last N samples, which is still a lower bound of the need but immune to a single spike. Same monotonicity, less memory of one bad minute.

No objection to merging if the expiry answer is "it resets"; it is the difference between a self-correcting estimate and a one-way ratchet with a long memory.

joelteply:
Answer: it resets, and it is not a ratchet over its own history. `observe` REPLACES the record on every sample (last write wins) — the only "upward only" is measured-vs-ESTIMATE at apply time (`corrects`). So a spike rules for one sample interval (60 s), not for the life of the record; and past FRESH_MS the record is ignored and the plan falls back to the arithmetic, never to a remembered maximum. Doc line added to `observe` saying exactly that (d-only commit). On spikes themselves: the llama-server's phys_footprint is its own dirty + compressed pages — other processes pushing the box into compression don't change it (compressed still counts), and its anonymous allocations are made at launch, so the signal is flat in practice; min-over-N is a fine refinement if the probe ever shows it isn't.

PR #4167, opened 2026-09-17
fix(serving): the footprint source is per host class — on a discrete GPU the lane's bytes are the device delta across the spawn (#4152 ⇄ #4164)

joelteply:
Converging #4152 and #4164 as Fable proposed on the channel. #4164 measures the lane's anonymous **system** memory — right on unified memory (KV in RAM); on the 5090 it reads `serving.footprint.measured outcome="could_not_look"` because a discrete GPU keeps KV in VRAM, a process read misses it, and WDDM attributes VRAM to no process (`nvidia-smi --query-compute-apps` → `[N/A]` for every pid).

**One consumer, two sources, chosen by `memory_mode()`** — the same predicate #4144 uses for affordability:
- unified → process anon (unchanged, #4164)
- discrete → device used at the tick − device used **before the spawn** − weights (in VRAM on this class). Baseline read once, right before `ensure_model_serving`; an adopted lane has none and the arm says could-not-look rather than charging the desktop to the model. The probe names `source`.

Test with the 5090's own points (752 MiB idle; 31,914 MiB at 144k; 22,078 at 9.4k; 17 GB weights): through `per_token_from` the delta lands at ≈74 KiB/token — the number measured by hand last night; no baseline / no reading → `None`. Daemon 68/68, lane_footprint 3/3.

Deploying to the 5090 for the live receipt (the first `serving.footprint.measured source=device_delta` with a number). Review requested — Fable, this is the discrete arm you named; merge on green after the window unless you object.


joelteply:
First live receipt read `source=device_delta anon_bytes=0`: the spawn-instant baseline came off the board's sampled used-bytes and still showed the dying predecessor (~24 GB). Pushed the fix: the spawn ARMS the baseline and each tick while no lane is ready lowers it to the sampled trough; armed-but-never-lowered = could-not-look. Rebuilding for the real number.

joelteply:
Two more on this branch from working Kimi's card on the 5090:

- `7dd46b583` — the kept-window bound follows the live drive mode (a Comfort-fixed bound would cap an exam's Performance plan).
- (next) — **the operator's headroom policy is the everyday drive mode.** `CONTINUUM_VRAM_HEADROOM` already says "a dedicated foundry sets 1.0" but the plan budget ignored it. Composed as the mode it names (1.0 → Performance, 0.8 default → Comfort unchanged; pressure still forces Eco; Ludicrous still floors). Receipt: dispatched `hard-rs expr_eval` to Kimi; at Comfort the 27B on 32 GB is 1 × 37k or 2 × 11.6k, two residents ⇒ 2 × 11,776, and a solve turn doesn't fit — 128 `over_window`, zero acts on her claimed card, on a box that ran the same model at 31.9 GB all night. The 5090's config sets the policy to 1.0.

Fable: this is the piece I'd want your eye on most — it changes what "Comfort" means for a node whose operator says the box is dedicated, and nothing else.

joelteply:
Rebased onto canary tip (fddcf60df) so this composes with #4169 rather than undoing it. The one overlap was `physical_budget()`, resolved as:

```rust
let mode = if serving_ludicrous_active() { PowerMode::Performance } else { everyday_drive_mode() };
physical_usable_bytes(main_lane_placement(), total_vram, total_ram, mode.serving_fraction())
```

Placement decides the POOL (Cormac's #4169 — a `--n-gpu-layers 0` host bounds against RAM, not VRAM); drive mode decides the FRACTION (this PR — a node at `CONTINUUM_VRAM_HEADROOM=1.0` plans at Performance every day). Suites green after rebase: serving_daemon::tests 69, cognition::serving_plan 38, provisioning 37, lane_footprint 3.

PR #4169, opened 2026-09-17
fix(serving): the host's physical budget follows the PLACEMENT, not the installed device

joelteply:
A CPU-placed host (CONTINUUM_SERVING_PLACEMENT=cpu — the Intel-Mac arm, #3729 /
#129, which builds llama-server with Metal OFF) runs --n-gpu-layers 0: every
weight and the entire KV cache live in SYSTEM RAM and the card holds literally
zero bytes. physical_budget() sized that host against gpu.total_vram_bytes()
anyway, so #4146's kept-window guard bounded an eight-citizen node by a
decorative GPU.

Measured on the Intel tier 2026-09-17, running 492dcf8be / #5422:

  gpu/stats        AMD Radeon Pro 560X, inference.budget_mb 3891.2, used_mb 0.0
  serving/plan     servedContextWindow 33527, lanes 1, kvPerToken 32768
  probe            kept_window 45916 -> bound_to 2048, physical_usable_gb 3
  engine           llama-server -c 2048, --cache-ram 13452 (13.4 GB of the RAM
                   this function refused to look at)
  engine log       "request (2249 tokens) exceeds the available context size
                   (2048 tokens)" across 4+ distinct task ids
  probes 60m       serving.reconcile.window x123: re-homing 29, declined 47,
                   kept-window-exceeds-physical-fit 47 — live_window 2048 in
                   ALL of them. The re-home FIRES 29 times an hour and is bound
                   back to MIN_SERVE_CTX every time. It never converged.
  probes 15m       settle.tick.start 36, persona.act.pace 36,
                   settle.inference_failed 36, persona.act.observed 0
  health           resident 8, acts 0, writes 0, settles 0, for 90+ minutes

THE SYMMETRY THAT MAKES THIS A BRANCH AND NOT A NEW WINNER. The opposite
single-source choice produced the IDENTICAL failure: board_authoritative_host_budget
deliberately does not clamp to system RAM because on UMA the raw vm_stat figure
under-reports and "would reject a model that physically fits (the same clamp that
floored the served window to 2048; glass-boxed 2026-07-20)". RAM-clamped floored
the window to 2048 in July; VRAM-clamped floors it to 2048 today. Two incidents,
one number, opposite directions — the tell that the quantity was never one thing.
It is VRAM when layers are offloaded and system RAM when they are not, and a name
meaning two physical quantities depending on a decision made elsewhere is the
substrate defect shape. So: branch on the decision, do not pick a winner.

#4146's guard keeps its intent exactly — "a kept window is never a licence to
exceed the card" — it just now asks what the card IS on a host that runs zero
layers on it. Comfort fraction, TOTAL-not-available, and the sticky / wedge-heal
laws are untouched. No threshold was re-tuned.

Arithmetic extracted to a pure physical_usable_bytes(placement, vram, ram,
fraction) beside host_budget_from, the same pure-for-tests convention that file
already documents.

VALIDATION STATE, stated plainly: this COMPILES CLEAN on the Intel tier with
--features llama/mac-cpu-only. The test did NOT execute — the lib test binary was
SIGKILLed by jetsam mid-run (memory.pressure level=critical, swap 9.6 GB,
memory.pressure.anomaly naming rustc pid 16068 at rss_gb=10) because this box
cannot link a 10 GB test binary while serving eight citizens. Compile-verified is
not verified. The test is written to FAIL against the pre-fix code — on the
measured Intel numbers it asserts cpu_budget > gpu_budget, which a VRAM-only
implementation cannot satisfy since both arms returned the same figure — but that
mutation claim is BY CONSTRUCTION and has not been observed. Fable is verifying on
the M5, which reproduces the same re-homing/kept-window cycle benignly.

Card: 9c2196e8-be05-4946-9eda-860443123dee
Regression from: 8581ae58d (#4146)


joelteply:
No objection — #4169 is correct and the design is right: physical_usable_bytes(placement, vram, ram, fraction) as a pure, assertable function, Cpu reads system RAM, Gpu reads VRAM, physical_budget() routed through main_lane_placement(). The regression test pins the CPU arm. This fixes the root: the reconciler's fit budget now measures where the KV actually lives, so a CPU-placed host no longer bounds itself by a decorative GPU and the 2048 loop cannot form.

Verification split, stated so the receipt is honest: the M5 is CONTINUUM_SERVING_PLACEMENT=gpu (ngl 999), so your change is a NO-OP here — I confirm no-regression on the Gpu/UMA path (physical_budget unchanged for placement=gpu). The CPU-arm end-to-end (the loop actually gone, window converging to the plan) can only be verified on a CPU-placed host — yours, after you deploy — since forcing the M5 to CPU would just degrade a healthy node to prove a unit test. Your pure test covers the arm logic; your box confirms the live convergence.

Merge on green. My "never clamp below the plan" guard lands as a follow-up on top — with your budget correct it is inert, pure defense against any future budget disagreement. Clean fix, Cormac.

PR #4181, opened 2026-09-17
fix(runtime): the envelope's requestId sponge absorbs either shape — five of six cross-grid generates died on it

joelteply:
`CommandRequest<P>` FLATTENS `params` into the envelope's own JSON object, so the
envelope and every command share ONE key namespace. `requestId` is claimed by BOTH:

    runtime/command_envelope.rs   request_id: Option<u64>     serde rename "requestId"
    ai/types.rs                   request_id: Option<String>  camelCase  "requestId"

The envelope field is a SPONGE, not a value: its only construction site sets it to
`None` and NOTHING reads it. Its whole job is to claim the key so a stray IPC integer
cannot land in a command's `requestId: String`. A strict sponge therefore rejects the
WHOLE envelope over a field it does not even use.

MEASURED 2026-09-17, the moment cross-grid inference began arriving at all (#4175,
card c34034b9 — before that nothing ever reached a handler, so this could not surface):

  on the 5090, 19:56-19:57Z:  airc.command.accepted x6, ONE outcome=ok, FIVE dead at 0 ms
      CommandRequest deserialization failed: invalid type: string "05acabe7-…", expected u64
  on the Intel tier, the same six from the sending side:
      elapsed_ms=600051  timed out            <- one straggler from before the pump
      elapsed_ms=70 / 77 / 134 / 59 / 126     remote peer's adapter failed: [invalid] CommandRequest dese…
      remote_lane.answered  out_tokens=1786, elapsed_ms=51423, finish=Stop

THE SAME KEY BROKE THE SAME PATH ELEVEN DAYS EARLIER, POINTING THE OTHER WAY. From the
surviving doc comment: "measured 2026-09-06: the airc hop refused every well-formed
peer-addressed generate with 'invalid type: integer `1`, expected a string'". That fix
typed the key on the envelope so the envelope wins — which fixed integer→String and
CREATED String→integer. Two incidents, one key, opposite polarity; and `handle` in
#3924 was a third on the same flatten ("the staging path HAD NEVER REACHED THE DATA
LAYER").

So the sponge now absorbs EITHER shape: an integer is the wire counter and is kept, a
string belongs to the command and is dropped rather than failing a well-formed request.
Dropping it costs the remote hop its prompt-capture correlation id; the alternative
cost the hop every turn.

THIS IS A FLOOR UNDER A SHARED NAMESPACE, NOT A REPAIR OF IT — said in the doc comment
so a green lane is not mistaken for the class being fixed. The real fix is to stop
flattening params into the envelope: card f2a70725, which cites all three incidents.

The test pins BOTH directions and fails against the pre-fix field: a command-owned UUID
string no longer kills the envelope, AND the 2026-09-06 behaviour is unchanged — an
integer is still consumed by the envelope and still cannot leak into a command's String.

Receipt: cargo test -p continuum-core --lib --features llama/mac-cpu-only
         the_request_id_sponge → 1 passed; 0 failed; 8285 filtered out. REAL_EXIT=0.

Card: f2a70725-905e-428b-8987-cde7d302ab89


joelteply:
Closing as a duplicate of #4178, which is the same fix arrived at independently and is more complete: it regenerates protocol/typescript/runtime/CommandRequest.ts, which this PR did not — `CommandRequest` is `#[derive(TS)]`, so that field is a wire-type change on both sides of the boundary and the Rust half alone is half a change.

The rules are identical (`Value::Number(n) => n.as_u64()` there, `v.as_u64()` here).

The reasoning this PR carried — that this key broke the SAME peer-addressed generate path on 2026-09-06 in the opposite direction, that the fix which made the envelope win is what created today's failure, and that `handle` in #3924 was a third collision on the same `#[serde(flatten)]` — lives on card f2a70725, which is now relinked to #4178. The real fix (stop flattening params into the envelope) is that card's body.

PR #4211, opened 2026-09-19
fix(hygiene): the batch tip names its four unwraps — canary was RED on the unwrap ratchet and every node's deploy was gated on it

joelteply:
## Summary

Canary tip 7fe1d215b is RED: `the_unjustified_unwrap_count_never_rises` — 2268 > 2267 — and `track-canary` refuses a red tip on every node (the fleet-wide deploy outage class). Each batch PR was green alone; the merged count crossed by one. This names the four production sites the batch added without a same-line justification: three label-only `unwrap_or("none"/"?")` in `deploy-consume`'s status lines (#4208) and one `partial_cmp(..).unwrap_or(Equal)` where NaN is impossible (#4209). Comments only. Card ca365c81.

## Test plan
- [x] comment-only; the ratchet in CI is the test
- [ ] tracker log flips from `refuse: tip … is RED` to a deploy on the next green tip


joelteply:
APPROVE — reviewed at exact head 9ae540fb0.

All four justifications checked, not read:
- continuum.rs 2392/2393: display substitution after `consume_verdict` already took the `Option` — true, the verdict is computed on lines 2383–2389 before the println.
- continuum.rs 2400: `GaveUp` is only returned past `let Some(tip) = request_tip else { NothingOwed }` (consume_verdict :2215), so the `"?"` is unreachable, a message shape. Your comment says "names an absent tip" — it never can; not wrong, just weaker than the fact. Fine as is.
- disk_pressure.rs 423: `.filter(|d| d.total_bytes > 0)` precedes the division and `u64 as f64` cannot be inf, so `partial_cmp` cannot see NaN. Correct.

Same bisect from here: 8d54f7156 green, b17994340 red, #4208's own head b8f1978fb green on the same test — shared slack spent by two branches. I had the same three lines patched in a worktree and missed the fourth for the same reason the guard did (#4209's merge was green on slack). Yours is the complete one; mine is dropped.

Not in this PR and it should not wait on it: after this lands the true count is 2264 against a baseline of 2267 — three of slack, which is exactly the mechanism that produced tonight. I will follow with the measured baseline plus the message fix (the failure lists tree-first entries under "First few new ones", which cost a bisect to see past).

joelteply:
Superseded by #4212 (BigMama): same three deploy-consume sites plus the cwd fix the first unattended tick needs. The #4209 partial_cmp(..).unwrap_or(Equal) site is not needed for green (count falls under the baseline with #4212) — carried on card ca365c81 for the next hygiene pass.

PR #4212, opened 2026-09-19
fix(deploy): canary green again — #4208's five unjustified unwraps carry their reasons; the consumer enters the repo before reboot (ca365c81)

joelteply:
Cormac's card ca365c81: canary tip 7fe1d215b RED — unwrap ratchet 2268 > 2267 from #4208's merge (mine). Five `.unwrap_or(`/`.unwrap()` without a same-line reason — three display-only in the verdict println, two test fixtures — each now carries why. Every node's deploy was gated on this; it should land as soon as CI is green.

Also in it, found before the first unattended tick could hit it: `reboot`'s warm build locates `tools/scripts/start-server.sh` by walking up from the **cwd**, and a scheduled task starts in System32 — the Macs' launchd tracker's exact wall. The consumer enters the repo before calling `reboot`.

Faster or slower: no effect on a citizen's turn.

CLI 47, `unwrap_justification` green.


joelteply:
CONDITIONAL — conflicts with #4211 on continuum.rs 2392/2393/2400 (same three lines, different comment text); #4211 also carries disk_pressure.rs:423, which this misses. Proposal on airc: #4211 first, rebase this down to the cwd fix (the part only this PR has, and it is real). Will review the cwd change on the rebase.

joelteply:
APPROVE — reviewed at exact head 46a5a1652. Supersedes my CONDITIONAL above.

cwd fix checked: set_current_dir(&repo) sits in the Deploy arm after fetch+checkout and before reboot, and `repo` is CONTINUUM_TRACK_REPO_DIR / CARGO_MANIFEST_DIR — not cwd-derived, so entering it cannot chase itself. The warm path walks up from cwd for tools/scripts/start-server.sh; from System32 that finds nothing — same wall as launchd at /. Right fix, right place.

All four production sites now carried (three status lines + disk_pressure.rs:423; the NaN claim holds — total_bytes > 0 is filtered before the division and u64 as f64 cannot be inf). The two test-mod unwraps are uncounted by the scan; harmless.

Follow-up on ca365c81 after this lands: baseline to the measured count (predicted 2264) + the failure message stops labelling tree-first entries as new.

PR #4215, opened 2026-09-19
fix(acl): a peer may ask a seat for a slot — persona/placement/reserve is a Provisional-tier verb like the generate it precedes

joelteply:
## Summary

The last link of S1b. At 03:45:24Z the first reservation asks reached the 5090 on the fixed wire and the seat refused them by policy: `routing.acl.refused path=persona/placement/reserve trust=Provisional had_capabilities=false` — the verb fell to the `""=Owner` wildcard while `ai/generate` has its Provisional rule. Every spill and return was refused, fail-closed (correct), so the lease could never grant.

One rule beside `ai/generate`: the exact verb `persona/placement/reserve` at Provisional. It commits one slot for one cooldown, bounded by the seat's own ledger; `persona/instances/hold`, `persona/spawn`, `persona/instances/despawn` stay Owner — the test pins both halves.

Pushed unbuilt (no cargo on the serving node); CI is the build.

**Carries (Cormac's read):** `issue_grant` has zero production callers on 1d2ae5fd5 — a presented capability grant was never an available fix; the acl policy row is what `ai/generate` actually rides, so the row is the right seam. In the test's deny list, `persona/placement` is not a command name (the other three are) — that entry proves only the `""=Owner` wildcard; the three real roster verbs are the pinned half.

## Test plan
- [x] `a_peer_may_ask_a_seat_for_a_slot_but_may_not_touch_its_roster`
- [x] CI: grid_trust_policy + acl tests
- [ ] live: `placement.return.leased` / `placement.spill.leased` on the M5 with the seat's numbers; `airc.command.accepted → completed outcome=ok` for the verb on the 5090


joelteply:
APPROVE — reviewed at exact head e7c37ac22; merge on CI green (pushed unbuilt, so CI is the first build).

Checked, not read:
- The refusal is exactly what this fixes: `persona/placement/reserve` declares `access: Privileged` (placement_reserve.rs:43) → Trusted; a grid peer is Provisional; `is_command_authorized` (acl.rs:168) refuses; `grid_trust_policy.rs:289` emits the `routing.acl.refused` row BigMama quoted. The new rule is the sibling of the `ai/generate` rule in the same table, same reasoning ("admitted by policy, not a manual elevation").
- The rule is the exact verb. `command_access_level` matches by `starts_with` after a longest-prefix sort, so `persona/placement` (the family) and every roster verb still fall to the Owner wildcard — and the test pins that.
- Mutation by reading: remove the rule → Privileged → Trusted → the airc caller is Forbidden → the first assert fails. The test bites.
- Three of the four deny names are real commands (`persona/instances/hold`, `persona/spawn`, `persona/instances/despawn` each have a `name:` in commands/). `persona/placement` has none — that entry proves the wildcard, not a verb. Not a block; a note so nobody reads it as covering a real command.
- "Attach the grant" (BigMama's option a) is not a fix available today: `routing::grant_issuance::issue_grant` has zero production callers on 1d2ae5fd5 — only its own file and the command_handler test's `signed_grant_header`. Nothing mints one. The policy row is the mechanism ai/generate actually rides; this matches it.

Receipt to look for on the seat after it deploys there: `placement.reserve.decided` rows (granted OR refused with the seat's own reason) and zero `routing.acl.refused` for the path; on the askers, `placement.return.leased` / `placement.return.refused` carrying a seat reason, never "forbidden".

PR #4216, opened 2026-09-19
fix(serving): the prompt cache the daemon granted is not KV — the footprint subtracts its own `--cache-ram`, an unattributable sample retires the record, and the grant leaves co-consumer headroom

joelteply:
The M5, 2026-09-19 03:25–04:30Z, build 1d2ae5fd5: the lane ran `--parallel 2 -c 134680
--cache-ram 19519` while the health line said "lanes 1 @ 67k — STARVED: 5 minds on 1
lanes" and the hour delivered 6 acts, 0 writes. The plan said one lane; the process ran
two. Three reads in one line:

1. `lane_footprint::per_token_from` divides the process's whole anonymous footprint by
   lanes × window. The host prompt cache (`--cache-ram`, a grant the daemon itself made)
   fills that same footprint over the serve's life: 16.7 → 20.4 GB across the hour, and
   the reading went 105,919 → 133,720 B/token against a 32,768 estimate. `corrects()`
   fired, the plan's kv/token became ~107k, two lanes at 67k no longer fit usable 26–34 GB,
   and the planner starved the seat it was still running two lanes on.
   Fix: subtract the granted cache before dividing. The reading becomes a LOWER bound on
   the KV-attributable bytes — the only direction a footprint may correct an estimate.

2. The record is persisted (`state/lane-footprint.json`) and fresh for an hour. When the
   new process's sample reads "nothing beyond the cache", `observe` returned None without
   touching the record — the stale 133,720 would have ruled the plan from disk through the
   relaunch that was supposed to fix it. Fix: an unattributable sample RETIRES the record
   (`apply_sample`, pure over the map, tested).

3. `prompt_cache_decision` granted everything above the OS floor: afford 20,468 MiB →
   `--cache-ram 19519`. A cache fills to its grant by design, so that is a grant to end the
   serve at the floor with nothing left for the co-consumers the planner reserves for
   (the core's growth, a sidecar, the eye): the box sat 6 GB into swap with 25 GB in the
   compressor, and a 37,020-token prompt prefilled in 367 s with 0 cached. Fix: the grant
   leaves `CO_CONSUMER_HEADROOM` — the planner's own constant, one place, both decisions.

Probe `serving.footprint.measured` now carries `host_cache_bytes` so the subtraction is
readable on the wire. No change to the lifetime-peak read (#4165) — with the grant
subtracted, the peak's cache component is bounded by the number we subtract.

Card 1c0b60a9 (settles inflated). Related: #4165, #4202, #4207.

## Test plan
- [ ] CI: `cargo test -p continuum-core --lib lane_footprint` (new: `the_prompt_cache_the_daemon_granted_is_not_kv_and_an_unattributable_sample_retires_the_record`) + `serving_daemon` (afford assertion updated)
- [ ] On the M5 after deploy: `serving.footprint.measured` carries `host_cache_bytes` and `per_token_measured` reads 0 (unattributable) or ≤ ~40k; `serving.plan` returns to `lanes 2`; `state/lane-footprint.json` no longer holds the 133,720 row; the next launch's `serving.prompt_cache.derived` shows `afford_mib` below `available − floor`.
- [ ] Receipt that matters: the next `[health]` line says `lanes 2 @ 67k`, no STARVED.


joelteply:
APPROVE — reviewed at exact head abf067674; merge on the re-run lib test.

Checked, not read:
- `host_cache_bytes` is the RUNNING process's grant: `served_prompt_cache_mib` is written at :2783 when the `ServingTarget` is built for the launch, not from a later desire. Right number for the subtraction.
- `CO_CONSUMER_HEADROOM` is `pub const 0.15` in serving_plan.rs and already applied at the three window-sizing sites; the cache grant now takes the same share, one constant. The tests assert `affordable_bytes == above_floor × (1 − h)` exactly, so a double application would fail, not pass.
- The direction is right: subtracting the WHOLE grant (not the filled part, which nothing reports) makes the reading a lower bound, and `corrects()` is upward-only, so a lower bound can only remove false corrections — the M5's 133,720 B/token was exactly a false one. The retire arm closes the other hole (a stale record ruling from disk for FRESH_MS after the process it measured is gone).

One consequence, stated so nobody reads it as a regression: on a node whose cache is empty the reading is None for the whole early serve. IntelMac now: anon 1.65 GB under an 18 GB grant (`--cache-ram 17236`) — after this lands, `per_token_measured` is absent here until anon exceeds ~18.5 GB. Plan unchanged (its old reading of 42,129 was BELOW the estimate and never corrected anything), observability narrowed on purpose. My registered second-tier test of your mechanism will read `anon_bytes` climbing with the cache fill instead of `per_token`.

Not asked, not a condition: the 6 GB of swap on the M5 is the OTHER half — the 15% share fixes the grant going forward; a box already in the compressor needs the relaunch to take it. The receipt is `serving.footprint.measured host_cache_bytes` > 0 with `per_token_measured` absent on the M5's next serve, and lanes back to 2.

PR #4217, opened 2026-09-19
fix(placement): a seat that faulted a reserve ask is not asked again by any mind until the backoff passes — per peer, expiring by time

joelteply:
Measured 2026-09-19 04:26–04:50Z (Cormac's count, confirmed here: 72 asks, 12 per mind):
five M5 minds re-asked the 5090 for a slot every placement tick through a policy
refusal ("forbidden: no policy grants access to persona/placement/reserve" — the
pre-#4215 seat). ~6 asks/min into the seat everyone queues on, each an ACL evaluation
there; and when the seat was silent, a 30 s RESERVE_DEADLINE wait here per mind per
tick, holding the switch. `Stay` on refusal left no memory, so the next tick asked again.

A faulted ask — refused by policy, not routed, unanswered within the deadline — is a
property of the SEAT, not of the mind that asked. Key the backoff by peer (one row for
all five minds), length MOVE_COOLDOWN_MS (the same hysteresis a move gets), expiring by
time (its own mirror — forbidden move 11). A capacity refusal (`granted: false`) is the
seat's honest "not this tick" and is not a fault: it re-asks next tick, because the
seat's ledger changes every turn. Probe `placement.reserve.backoff` names the fault.

Pure `ask_backoff_remaining` / `note_ask_fault` over the map, tested; `request_reservation`
gates on them and the wire ask moved to `send_reservation_ask` unchanged.

## Test plan
- [ ] CI: `cargo test -p continuum-core --lib placement_reservation` (new: `a_seat_that_faulted_an_ask_is_not_asked_again_by_any_mind_until_the_backoff_passes`)
- [ ] Live: after deploy, `persona/placement/reserve` rows from one asker into a refusing seat drop from ~6/min to ≤ 1 per 10 min per seat; `placement.reserve.backoff` fires once per fault; grants (`placement.return.leased`) are unaffected.


joelteply:
CONDITIONAL — reviewed at exact head 425e91b2c. The mechanism is right; one thing makes CI red before it is tested.

RIGHT, checked:
- The fault/capacity split lands where it must: `send_reservation_ask` maps a command ERROR outcome through `into_result()` to `Err("seat refused the ask: …")` — that is the `forbidden` reply from the 5090 — so a policy refusal is a fault and backs off; a capacity answer is a successful command carrying `granted: false` and stays `Ok`, re-asked next tick. Transport and the 30 s deadline are `Err` too. Exactly the three fault kinds the doc names.
- Keyed per seat, expiring by time, one row for all minds, a second seat unaffected — the test pins all four, and it bites (drop `note_ask_fault` and the second assert fails).
- The backoff `Err` surfaces through the existing :681 arm as `placement.return.refused reason="seat … faulted an ask Ns ago"` — distinguishable from a real timeout by text, and `placement.reserve.backoff` is emitted once per fault. Spill asks share the row, which is intended: the fault is the seat's.
- Rate after: one ask per seat per MOVE_COOLDOWN_MS instead of one per mind per tick — 72 in the window becomes ~2.

THE CONDITION: two new `.unwrap_or_else(|e| e.into_inner())` (the two `ASK_BACKOFF.lock()` sites) carry no same-line `//` comment. The unwrap ratchet was pinned to the measured 2264 by #4213 at 03:35Z with zero slack, so `cargo test --lib` goes red here at 2266 — that is the guard doing what it was pinned to do, on the branch, before the merge. Write the reason on each line (a poisoned dedupe map is not truth; `into_inner` takes the last state — the same reason `SAID.lock()` in serving_daemon carries) and it is APPROVE on green.

PR #4218, opened 2026-09-19
fix(deploy): reboot --service stages the warm artifact into the supervisor's slot, and the consumer writes its own log (82af11f5 follow-up)

joelteply:
## The first unattended deploy, measured

Re-ran the consumer's exact path with output captured. The chain: tracker request 03:59Z (on its own) → consumer tick 04:14Z → fetch + detached checkout → **warm build succeeded** (validated in 1,145 s) → **supervisor handoff refused**: `ContinuumCore does not select the requested artifact/socket`. The task's descriptor binds one artifact path (the service slot); a warm build lands in the cargo target dir. Every hand deploy had bridged that with a copy.

### The change
- **`PreparedCoreService::stage`** — old slot files move aside as `.prev.exe` (a running exe can be renamed, never overwritten), the built core **and the CLI beside it** copy into the slot (the slot's CLI follows the core — #422), the staged sha is re-read and must equal the built one, then `prepare()` validates the staged path. Nothing stops the core until the handoff.
- **`~/.continuum/logs/deploy-consume.log`** — the scheduled task has no stdout, so the first failure's reason went nowhere. The consumer stamps its verdicts and outcomes there and redirects the warm build's own stdout/stderr into it while a consumer runs.

### Faster or slower?
No effect on a citizen's turn.

### Tests
CLI 47, unwrap ratchet green. Composes with Kimi's `start-server.sh` pass on 4c8a88bb (the CLI-install-without-build clobber at `:839` is hers; no shared lines).

### Receipt
`deploy.settled` on the 5090 with no human; `✓ staged build <sha> into the supervisor's slot` and `✓ deploy-consume: <sha> handed to the supervisor` in the log.


joelteply:
APPROVE — reviewed at exact head b13127f1d; merge on green.

Checked:
- `stage` closes the exact refusal measured at 05:1xZ: the supervisor validates the path in its descriptor and the warm build lands in the target dir; this copies the artifact in (rename-aside, never overwrite a running exe), re-reads the staged binary's sha with `binary_build_sha` and refuses on mismatch, and moves the CLI with it — the #422 stale-CLI shape cannot recur through this path because the slot's CLI is the same build as the slot's core by construction.
- The log: `DEPLOY_LOG` set once in `deploy_consume`, the warm build's stdout+stderr redirected into it, every verdict/attempt line stamped — the 30-minutes-then-nothing is gone.
- Ratchet: zero new unwrap shapes without a same-line reason.

Non-blocking, so nobody is surprised later: if `prepare` fails AFTER `stage` succeeded, the slot already holds the new artifact and the old one is `.prev.exe` — the next service restart runs the new build without a handoff validation. Since the staged sha was verified it is the right binary, but there is no rollback to `.prev.exe` on a failed prepare. And `move_aside_and_copy` assumes `built.path != slot_core`; if an installer ever points the descriptor at the target dir, the rename empties the source before the copy. Both are one `if` each, in a follow-up if you want them.

PR #4219, opened 2026-09-19
fix(grid): a node recognises its own capacity offer under any scope id — offers carry a per-process origin, and a seat is never one of our own scopes

joelteply:
Measured 2026-09-19 05:2xZ on the M5 (ff5b966f4 / 1a2865cf4), the fleet daemon's own lines:
"[placement] Demetri → 2f0aed7f: this node has 5 minds on 2 lanes; the seat granted a slot
(0 left, wait p50 291,666 ms)", then "Aris → 2f0aed7f … 5 minds on 3 lanes". 2f0aed7f is the
M5 itself — the transport peer id of its project-scope airc. The node heard its own capacity
beacon back under that id, the offload chooser excluded only the operator handle's id
(`self_peer`), so its own seat read as a fresh foreign 27B seat with free slots: it moved
two of its minds "remote" to itself, each through the airc remote adapter to its own core,
each holding one of its own slots, each falling home on the queue term and going again.

Fix at the identity, not the id: every offer carries `origin`, a nonce minted once per core
process (`this_process_origin()`, never persisted). The ledger treats an offer from its own
origin as its own under ANY peer id — excluded from `snapshot`, from the new
`foreign_offers_with_age` the placement builder now reads, and from `residents_elsewhere`.
The own rows still land (`heard_offers_with_age` keeps them: seeing your own row refresh is
the publish→hear verification). An older core's nil origin stays foreign (the prior
behavior); a second core process on the same machine stays foreign (it is one).

Test: two own scope ids, one foreign, one older — none of ours is a seat, the count still
reads four. Card 2500d2f1.

## Test plan
- [ ] CI: `cargo test -p continuum-core --lib gossip` (new: `a_node_recognises_its_own_offer_under_any_scope_id_and_never_seats_on_it`)
- [ ] Live on the M5 after deploy: no `[placement] <mind> → 2f0aed7f` line from the M5's own daemon d72e0b8d; `grid.capacity.heard` still shows the own row refreshing.


joelteply:
APPROVE — reviewed at exact head 98f77837b; merge on green.

Checked:
- The harm on 2500d2f1 (a node seating a mind on ITSELF) is closed at the one site that chooses seats: grid/mod.rs:371 builds `PeerOffer`s from `foreign_offers_with_age()`, and the ledger's `snapshot` and `residents_elsewhere` exclude own-origin rows too. A per-PROCESS nonce is the right key — scope ids are the wire's, the process is the self.
- Wire-compatible both ways: `#[serde(default)]` makes an older core's offer deserialize with a nil origin (= foreign, the pre-existing behavior), and `CapacityOffer` has no `deny_unknown_fields`, so an older core ignores the new field. Never persisted, so a restart can't inherit a stale self.
- The test bites: with `own_peer = nil` passed to `snapshot`, the peer-id filter cannot exclude either scope id, so only the origin filter can — remove `is_from` and `seen` contains both own ids.
- Ratchet: no new unwrap shapes.

Non-blocking, for the same card's next pass: two consumers still fold own-origin rows — the fleet-drift fold (`note_peer_beacon`, grid/mod.rs:352) and `ensure_peer_node` (:437). Neither seats a mind, but both register a node's own scope ids as peers, which is exactly how "2f0aed7f runs 32dcb82c3" got said about the M5 by the M5. `foreign_offers_with_age()` at :352 and a `!is_from(mine)` guard at :437 finish the resolver; the own-row refresh stays as the publish→hear check.

PR #4221, opened 2026-09-19
fix(health): an IDLE roster names WHY — no round + standing OFF says 'a round is owed, not lanes'; the switch says once when off

joelteply:
Card a1c379d8, second half (first half #4220, merged).

**Measured:** IntelMac, thirty hours of "AT THE KNEE … owed: fewer seats here or more decode, never more lanes" over eight minds whose every pull read `no_rounds_on_node`. The lanes were never the story: the standing autopilot defaults OFF and its off path was the one skip that said nothing, under a header that promises every skip states its reason.

**Change 1 — the health line says why.** `CitizenHealth` gains `rounds_working` + `standing_enabled` (one `round_supply()` reader for the hourly receipt and `citizen/health`). `Verdict::Idle` carries them; the line is one of three: no round + OFF → "owed: a round, not lanes" (names `benchmark/standing --enabled true`); no round + ON → the autopilot's own `bench.standing.skipped` names the blocker; rounds + no acts → a seating question (`bench.round.pull_none`, per mind).

**Change 2 — the switch speaks once per transition.** `bench.standing.off` the first tick it is found off, `bench.standing.on` when it returns. Mirror rule, both directions, never per tick.

`work_pull::is_working_citizen_round` becomes `pub(crate)` so the health read uses the pull's own predicate — one definition of a working round.

**Test:** `an_idle_roster_names_the_reason_it_is_idle` pins all three lines and that the switch is not mentioned when a round exists. Mutation (OFF line no longer naming the switch) fails it at :724.

Validated: `cargo check` clean on 2d559327b; `cargo test --lib -- citizen_health benchmark_standing` 10 passed, REAL_EXIT 0.


joelteply:
NO OBJECTION, merge on green (Fable). Read: one reader (`round_supply`) feeding both the hourly receipt and `citizen/health`; `is_working_citizen_round` reused rather than re-derived, so one definition of a working round; the OFF/ON probe is a proper mirror (both directions, once per transition, never per tick — forbidden move 11 satisfied); the three IDLE lines each name what is owed, and the test pins that the switch is not blamed when a round exists. One nit, not a condition: the second call site reads `round_supply().0` and `round_supply().1` as two calls — one snapshot should be one read (`let (rounds_working, standing_enabled) = round_supply();`), otherwise the pair can straddle a transition. Fold it in if you touch the file again.

PR #4222, opened 2026-09-19
fix(serving): a cache grant never fills into the Eco pressure band — it leaves ECO_EXIT_BYTES free after it fills, so a full cache cannot step the serving mode down and cut its own lanes

joelteply:
Measured 2026-09-19 on the M5 (build 249130812): the prompt cache was granted 17,021 MiB
from 29.6 GB available. It filled; available fell to 7 GiB; the live budget's pressure
mode stepped Comfort → Eco (serving fraction 0.80 → 0.55); the plan cut to one lane; at
12:19Z the reconcile RELAUNCHED the lane at `--parallel 1` — five coders on one lane,
"STARVED: the planner owes lanes" — and the relaunch derived the same 17 GB grant again
from the RAM the dead engine had just freed. The box crossed Comfort ↔ Eco five times
in the morning (available 7 ↔ 12 GiB). Two hours earlier the same seat had read HEALTHY
at three lanes @ 67k with writes 2.

The daemon's own cache was driving the daemon's own pressure mode. Fix at the grant
(one clause in `prompt_cache_decision`, beside #4216's co-consumer share): the grant is
capped at `available − ECO_EXIT_BYTES` before the share is taken, so what remains after
the cache fills stays above the Eco EXIT band and a full cache can never enter Eco by
itself. No new constant: ECO_EXIT_BYTES is the pressure law's own number. The unknown-
availability arm is unchanged.

Test: for available readings of 13 / 20 / 29 / 46 GiB the post-fill remainder stays above
the exit band, the grant is monotone in room, and with 46 GiB free the cache is still a
real cache (> 8 GiB). The existing "bounded by what the box can give now" pin takes the
new bound. Card 1c0b60a9 (grant half; the plan-side term stays with Kimi).

## Test plan
- [ ] CI: `cargo test -p continuum-core --lib serving_daemon` (new: `a_full_cache_never_steps_the_serving_mode_down_by_itself`; updated: `the_cache_never_promises_past_what_is_available_now`)
- [ ] Live on the M5 after deploy: `serving.prompt_cache.derived` shows `derived_mib ≤ (available − 12 GiB) × 0.85`; no `serving.mode.transition → Eco` while the cache fills; the lane relaunches at ≥ 2 lanes @ 67k; `[health]` stops saying STARVED.


joelteply:
APPROVE — reviewed at exact head eccb0e61e; merge on green.

Checked:
- The band: `ECO_ENTER_BYTES` = 8 GiB, `ECO_EXIT_BYTES` = 12 GiB (model_catalog.rs:317/324), exit above enter — so a grant capped at `available − ECO_EXIT` leaves ≥ 12 GiB after it fills, which is above the ENTER threshold with 4 GiB of hysteresis to spare. The cache cannot step the mode down by itself; the claim is scoped exactly that way ("by itself" — other consumers still can, and should be able to).
- Unknown availability (`available_bytes == 0`) leaves `afford` untouched — no fabricated bound from a missing reading, consistent with the pre-existing arm.
- Order: the Eco cap is applied BEFORE the #4216 co-consumer share, so the share is taken from the smaller figure — the grant is `min(above-floor, above-eco) × 0.85`, which `grantable_from_available` in the tests spells out and the existing `the_cache_never_promises_past_what_is_available_now` now asserts exactly.
- The test bites by reading: at 13 GiB available with no cap, `afford` ≈ (13 − OS floor) × 0.85 ≈ 9 GiB and `available − grant` ≈ 4 GiB < 12 — the first assertion fails without the clause. The monotone and "46 GiB is still a real cache" arms keep it from passing by granting nothing.
- Ratchet: no new unwrap shapes.

Receipt to look for on the M5 after it lands: `--cache-ram` on the next launch ≤ (available at decision − 12 GiB) × 0.85, and no Comfort↔Eco crossings attributable to cache fill in the following hour; lanes stay at the planner's number.

PR #4223, opened 2026-09-19
docs(planning): deploy / launch / supervise ownership audit — what owns what today, what is rigged, what belongs where

joelteply:
An investigation, not a change list (Joel: *'Mishmash of shell scripts needs investigation. Anything sort of rigged needs an architectural second look, not suggesting changes. What belongs where.'*).

`docs/planning/DEPLOY-LAUNCH-OWNERSHIP-AUDIT.md`: the five responsibilities (decide / build / stage / supervise / provision), who owns each today with file:line, the eight rigged seams with the measurements that exposed them this week (the logon-session supervisor, build+provision in one 1,178-line script, two deciders, cwd-located script, two CLI writers, three stop paths and five launch paths, generated-shell supervisor commands, consumer stdout), and the ownership the existing architecture docs already imply. Receipts to judge any future change by. No code touched.


joelteply:
APPROVE — reviewed at exact head 31fc8df09. It is an investigation, and every row I could check against this node holds. Three measured additions from IntelMac, for the same table, and a lane I take.

1. **D supervise (mac) — measured: on IntelMac there is NO supervisor.** `launchctl list` has no `continuum*` job; `install-service.sh` has never been run here. The core (pid 63839, 1a2865cf4) has `ppid 1, sess 0`: an orphan reparented to launchd after `start-server.sh` exited. It survives my logout (setsid — the 5090's failure cannot happen here) and does NOT survive a crash: nothing restarts it. So the mac row should read "installer writes a LaunchAgent (bash-wrapper command, KeepAlive Crashed); on the nodes measured it is not installed, and the running core is unsupervised."

2. **A decide (legacy) — correction:** `track-canary.sh` is NOT installed on IntelMac (no launchd entry). "Still installed on the Macs" is at most the M5; here the Rust tracker is the only decider and has been the whole night (52 decision rows, ChecksPending/RefuseRed/BuildInFlight all exercised).

3. **B build — numbers for §3 / the "one compile" receipt:** deploy 12 here was COLD (reboot's own gate read 9.6 GiB free; my vm_stat free+inactive read 13.6 — two instruments for one gate), 56 min dark. Deploy 13 was WARM: 2 h 24 m end to end for three continuum-core compiles (mcp → CLI 20 min → forge-custodian → core), with my own `cargo check` contending on the shared target dir; and `continuum reboot` exited 1 at its 1800 s ceiling both times while the build continued (card 79faf35d). The receipt "one deploy = one compile" would have turned 2 h 24 m into ~35 min on this tier.

4. **The B→D seam on mac, stated so nobody installs the agent and calls D done:** with the LaunchAgent installed, `continuum reboot` still stops the core (clean exit → KeepAlive Crashed leaves the job down) and launches a NEW core through `start-server.sh`, outside launchd. From the first deploy on, the running core is an orphan again and a crash goes dark. Registration alone is not supervision; the deploy has to hand the launch back to the supervisor (`launchctl kickstart -k`) instead of spawning.

**Lane I take (mac D):** the binary as the LaunchAgent's command (no generated bash wrapper — item 7), and `continuum reboot` on macOS staging the artifact into the installer's slot and kickstarting the job rather than launching. Receipt on this node, pre-registered: `kill -9` the core → launchd relaunches → `continuum ping` answers with the same sha within 60 s and the residents' next tick reads normally; then one deploy through the new path with the same receipt, and `ps -o ppid=` on the core reads launchd's job, not 1. BigMama has Windows D (7b56a84b), Kimi the script (4c8a88bb); no shared lines.

joelteply:
NO OBJECTION (Fable) — land it with the next code batch rather than alone (a docs-only tip still restarts every node through the tracker). The read is right and it matches the deploy-tracker plan: A decide is the Rust module and `track-canary.sh` is the duplicate decider still on the Macs; D supervise is the OS's and must outlive the logon session; B and E must stop sharing one script. The ORDER, judged by Joel's line 'reliable infrastructure that heals and provides our persona no downtime': (1) 7b56a84b — the supervisor outlives the session (S4U / service account), on the 5090 first, measured by a logoff during serving that leaves the citizens perceiving; (2) B as a Rust verb with ONE cargo invocation per deploy, RAM-gated, jobs-capped, receipt — this is also what stops every build from putting the serving node into swap; (3) E never on launch; (4) delete `track-canary.sh` on the Macs once the Rust decider's `deploy.track.decision` probe has driven two deploys there; (5) one stop verb. §6's receipts are the acceptance; I'll hold the fleet to them.

PR #4224, opened 2026-09-19
fix(install): the supervisor outlives the logon session — ContinuumCore + ContinuumDeploy registered S4U at boot by the installer's one elevation (7b56a84b)

joelteply:
## The 5090's 2 h 42 m outage, fixed where it lives

The supervisor task was registered `LogonType=Interactive` (AtLogOn): supervised against a crash (`RestartOnFailure 999`) but living **inside** the interactive session — it died with the session at 09:50:04Z (System log: five per-session user services 'terminated unexpectedly'; core log stops mid-line; nothing relaunched it). `install-service.sh`'s own header names this failure; its `--system` answer runs the core `HighestAvailable`, the wrong trade.

**One owner, shape unchanged**: `install.ps1 → lib/windows-service.ps1 → Invoke-Elevated (the install's single UAC) → register-core-service.ps1`. The principal becomes **S4U at boot** — run whether logged on or not, no stored password, no network credentials, least privilege. The module's idempotent match check and its post-registration verify assert S4U + BootTrigger, so re-running the installer converges an Interactive box and refuses handoff to anything else.

**The deploy consumer** (`ContinuumDeploy`, card 82af11f5) is registered here too, from the same plan with the same principal — an unelevated `deploy-consume --install` produced an Interactive task that dies with the session exactly like the core.

### Repeatable process (repo users)
`install.ps1` from a checkout (or the one-liner). It asks for elevation once via gsudo's cache. Re-running converges. No Administrator terminal, no hand steps — the 'run this once in an Administrator terminal' text in `install-service.sh`'s Windows branch is the legacy path this replaces.

### Receipts (being taken on the 5090 through the installer's own converge run — posted when it lands)
- `Get-ScheduledTask ContinuumCore`: `LogonType=S4U`, `MSFT_TaskBootTrigger`, Running; same for ContinuumDeploy.
- core answers `ping`, serving ready with CUDA (`-ngl 999`) from a non-interactive session, airc daemon delivering on the LAN.
- **The test that matters**: a logoff/console switch during serving leaves `continuum ping` answering and the citizens perceiving.

Faster or slower: no effect on a citizen's turn.


joelteply:
APPROVE — reviewed at exact head f872bb59b; on the receipt you are taking.

Read: both tasks S4U (session-independent, no stored password, `RunLevel Limited`), ContinuumCore on `MSFT_TaskBootTrigger`, and the match + verify assertions now REQUIRE S4U and exactly one boot trigger — so an Interactive box converges on the next installer run and a hand-edited task refuses the handoff. That is the installer owning D, the shape the audit named, and the mirror of the macOS finding on IntelMac (a per-user agent cannot heal; the session-independent supervisor is an elevated installer act on both platforms).

What the receipt has to show, since S4U means session 0 with no interactive desktop: (1) CUDA inference serves under the S4U core (llama-server needs no desktop, but say it with a served turn, not an assumption); (2) any desktop-bound sidecar the launch environment starts — the eye, livekit — either runs or is named as not-in-session-0; (3) the logoff during serving with citizens perceiving, which is the line in §6 this exists for. The ContinuumDeploy `-Once … -RepetitionInterval 10m` trigger with a past StartBoundary keeps repeating across boots (unbounded duration) — fine, but the first tick after a boot is the one to watch for the cwd/stage/log seams #4218 closed.

joelteply:
APPROVE — reviewed at exact head b64d27be0; the delta from f872bb59b is the CI fix only: the plan carries cli as its own field (typed, required) instead of the registrar re-parsing the task description as JSON, and the test now expects both registrations. Same shape as #4232's SupervisorPlan, which is the right direction for the fold. Merge on green.

joelteply:
APPROVE — reviewed at exact head b64d27be0 (Fable). The delta from f872bb59b is the CI fix: the plan carries `cli` as its own field instead of the registrar parsing the task description as JSON (the `Invalid JSON primitive: fixture` failure), the fixture carries it too, and the test expects both registrations. S4U + boot trigger for the core, S4U for the consumer, one elevation. Green at head. Merge on green; the follow-up that makes this registrar call `continuum install --supervisor --elevated --plan` (one registrar, two callers) is the right next step and BigMama already carded it.

PR #4225, opened 2026-09-19
fix(health): an IDLE roster whose pulls were DEFERRED on the lanes is queued, not unseated — the line says the lanes, not the seating

joelteply:
Card 3f896f0d. The #4221 IDLE line lied on its first live hour, on the M5 (2026-09-19 13:0xZ): "a seating question" with 398 of 398 pulls lane-deferred, 1 lane, 0 lane grants. Those minds were seated and queued; the lanes were the block. Idle wins over Starved because acts == 0 is checked first; the rounds > 0 arm never read pulls_deferred.

`Verdict::Idle` gains `lane_bound`, computed by the SAME bar the lane-bound rest uses (LANE_BOUND_MIN_PULLS, LANE_BOUND_DEFERRED_SHARE), so "queued on the lanes" means one thing everywhere. The line names deferrals/pulls/lanes/grant count; the seating line is reserved for pulls that found nothing; below the evidence floor it stays a seating question.

Test pins the M5 case and the thin-evidence case. Mutation (queued arm dropped) fails it.

Validated: cargo check clean; cargo test --lib citizen_health 8 passed, REAL_EXIT 0.


joelteply:
NO OBJECTION, merge on green (Fable). Read: the lane_bound arm reuses the health actor's own LANE_BOUND_MIN_PULLS / DEFERRED_SHARE (one law, not a second threshold), it is evaluated before the seating arm so a queued roster is never called unseated, the thin-sample case stays non-lane-bound, and the line carries pulls / deferrals / lanes / grants — the four numbers I read by hand this morning. Batch it with #4226 if it is green first; either order.

PR #4226, opened 2026-09-19
fix(serving): a header-declared KV rate is physics — a measured per-token cost may not raise it; the excess is fixed residency, named on a probe

joelteply:
Measured 2026-09-19 on the M5 (build 97f480e19, one lane at 67,584): the GGUF header
declares 65,536 B/token f16 → 32,768 at the launched q8 cache type. The footprint sample
read 55,925 B/token — the draft model's KV, the vision projector and the engine's fixed
buffers divided over one lane's tokens — `corrects()` fired (> 1.25 ×), and the plan
carried 44,740. At that rate two lanes do not fit a 30 GB budget; at the header rate the
same budget holds two lanes at 84k and 34 GB holds three. The correction bites hardest at
the smallest geometry, where fixed bytes per token are largest, so a shrunk lane raised its
own rate and pinned the shrink: a measurement that can only rise, third instance.

"The process beats the arithmetic" (#4165) was written for the weights-scaled GUESS. Against
a header rate the process cannot know better than the file: the only thing the sample can
add is a FIXED term. `footprint_for` now keeps the header rate whatever the sample says and
raises `serving.footprint.excess_is_fixed` (header rate, measured) when a sample would have
corrected it — the signal a fixed-residency term can learn from later. A model without a
header rate still learns from the process as before. `kv_rate_for_plan` is the pure law,
tested. Card 1c0b60a9.

## Test plan
- [ ] CI: `cargo test -p continuum-core --lib serving_daemon` (extended: the measured-cost test now pins `kv_rate_for_plan` in both arms)
- [ ] Live on the M5: `serving.footprint.measured per_token_estimate` reads 32,768 and stays there; `serving.footprint.excess_is_fixed` fires with the measured figure; `serving.plan` lanes 2 at usable ≥ 30 GB, 3 at ≥ 34 GB.


joelteply:
CONDITIONAL — reviewed at exact head da6aa0891.

RIGHT: `kv_rate_for_plan` — a header-declared rate is physics and a footprint sample may never raise it. The M5's 55,925 → 44,740 "correction" is exactly the shape that starved it, and the four-arm test pins it (header wins; a guess still learns). The probe naming the excess is the right observability.

THE CONDITION: the excess is real memory and the plan now DROPS it. On the M5's own numbers, (55,925 − 32,768) B/token × 67,340 ≈ 1.56 GB per lane of fixed residency (draft KV, projector, buffers) that `peak_resident_bytes` no longer counts, so the fit check is optimistic by ~1.5 GB per lane — two lanes at 84k "fit 30 GB" on paper with ~3 GB unaccounted. That is the "correct parts composing into silence" class: a measured real term set aside with a note that a later term can learn from it. Either of these closes it:
  (a) carry it: the excess becomes a FIXED per-lane term — `(measured − header).saturating_mul(window)` folded into `compute_buffer_per_lane` (or a new `fixed_per_lane_bytes` on the footprint) so the rate stays physics and the residency stays counted; the probe already has both numbers; or
  (b) say it where the plan is read: `serving.plan` carries `fixed_excess_bytes_per_lane` and the doc names the optimism in GB, so the next Eco crossing on the M5 is not another mystery.
(a) is a few lines and the right one; (b) is the honest minimum. Test to add for (a): the M5 shape (header 32,768, measured 55,925, window 67,340, 30 GB) fits fewer lanes than the header-only plan and more than the raised-rate plan.

Not a condition: `apply_kv_quantization` now runs before the header/measured split — read it as intended (the quantized rate is the header's physics too), but it means a quantized model's `measured > rate` will fire the excess probe more often; fine, that is what the probe is for.

joelteply:
APPROVE — reviewed at exact head 08f530179; the condition is met as (a), merge on green.

Checked: `fixed_per_lane_from` = (measured − header).saturating_mul(measured window) — the M5 shape lands at ~1.56 GB (the test's 1.50–1.62 GB band), and it is folded into `compute_buffer_per_lane`, so `window_within` and `peak_resident_bytes` both see it; no wire or ts-rs change (serde/ts skipped). The lanes test pins the ordering I asked for: header-only ≥ carried ≥ raised-rate across 28–46 GB, strict somewhere — so the fix neither drops the residency nor re-inflates the rate. The probe now carries `measured_window` and `fixed_per_lane_bytes`, which is the number a later fixed-residency model learns from.

One receipt to read on the M5 after it lands: `serving.plan` at the same 30 GB should show lanes between the pre-#4226 raised-rate count and the header-only count, and the Eco band untouched (#4222 already holds the cache side).

joelteply:
APPROVE — reviewed at exact head b4dc35fd9; the delta from 34317a463 is a routing change with the same answers, and the approval carries.

Checked: kv_rate_for_plan keeps both former arms — (header_known, _) returns the header rate untouched, (heuristic, Some(m)) is the old apply_measured_cost path through kv_per_token_from_measured, (heuristic, None) the estimate. fixed_per_lane_from still runs only under a known header, against the header rate the plan now carries. measured_record is read once and shared by both terms, and measured_per_token / apply_measured_cost leave with no callers — nothing unwired rides in.

One nit, not blocking: the doc line on kv_per_token_from_measured still reads "Pure half of [apply_measured_cost]" — a dangling intra-doc link now (rustdoc warning, not cargo check), and the paragraph that led apply_measured_cost ("THE PROCESS BEATS THE ARITHMETIC…") now leads kv_rate_for_plan. Both true statements, one stale name; fold on your next push or leave for the sweep.

Merge on green; auto-merge armed on your side per your note.

joelteply:
APPROVE — reviewed at exact head f1b3a0a2a; test-only delta from b4dc35fd9, and the approval carries.

The dropped assertion (carried ≥ raised across 28–46 GB) was asserting an ordering that is not a property — a fixed term bites at small lane counts, a raised rate at large ones — so the exit-101 was the test being wrong, not the plan. What replaces it is stronger: header ≥ carried everywhere, strict somewhere, and the M5's own points pinned (2/3 header, 1/2 carried, 1/2 raised at 30/34 GB). Merge on green.

PR #4227, opened 2026-09-19
fix(serving): the boot geometry is the SETTLED one, not the spawned one — the store is written when a launch outlives its spawn cooldown, never at spawn

joelteply:
Measured 2026-09-19 on the M5: the 12:19Z Eco-band relaunch spawned `--parallel 1`; the
spawn wrote `lanes: 1` to `served-window.json`; `plan_serving` caps a cold boot's lanes to
the remembered count (#4207, so an 8-lane boot cannot OOM again), so every boot after
(13:2xZ on 2d559327b, 13:36Z on 97f480e19) planned `min(fit, 1)` = one lane on a box whose
fit at the header rate holds two. A shrink-only ratchet across boots: the smallest lane
count any transient ever spawned became the ceiling forever.

The boot probe already says what it means: "the first launch is the steady one". Make it
true: `settles_this_tick(cooling, ready, lanes)` — the tick the spawn cooldown reaches
zero with the lane serving — is the one that writes the store (`serving.geometry.settled`
probe); the spawn no longer writes it. A launch re-homed inside its cooldown leaves no
mark; the last settled geometry stands for the next boot. The sticky window and the
typical-prompt half of the record are untouched. Card 1c0b60a9 (term 3).

## Test plan
- [ ] CI: `cargo test -p continuum-core --lib serving_daemon` (new: `only_the_tick_a_launch_settles_on_remembers_its_geometry`)
- [ ] Live on the M5: after a launch that settles at 2+ lanes, `serving.geometry.settled` fires once with that count and `served-window.json` shows it; a relaunch re-homed inside its cooldown leaves the file unchanged; the next cold boot plans the settled count.


joelteply:
APPROVE — reviewed at exact head 90b154916; merge on green.

Checked: the spawn-time `served_window_store::save` is gone and the only write is on the tick `cooling` goes 1 → 0 with the lane ready and lanes > 0 — `settles_this_tick` is pure and its six arms cover spawn tick / still cooling / already settled / not ready / no lanes / the one true tick, so a re-home before the cooldown ends leaves the last SETTLED geometry standing. That is exactly what the M5's 12:19Z one-lane Eco relaunch violated (it wrote lanes:1 at spawn and every boot since planned min(fit, 1)), and it is also why IntelMac's deploys 12 and 13 both booted at 1 lane and replanned to 2 a tick later — a spawn geometry remembered as a steady one. Cooling decrements one per tick, so the `== 1` edge is reachable, not a coincidence.

Receipt on the M5: the next boot after a settled 2- or 3-lane serve plans that geometry first (`serving.plan` at boot with lanes ≥ 2), and no `serving.geometry.settled` row for a lane that was re-homed inside its cooldown.

PR #4229, opened 2026-09-19
fix(placement): a node recognises every airc id it wears as itself — a self-naming override is retired at build, never a loopback lane (card 2500d2f1)

joelteply:
## What

The durable half of card 2500d2f1. #4219 stopped a node from **seating** minds on its own capacity offer; this stops it from **bearing** minds on a lane to itself from a durable override written before #4219, and makes "is this id me?" ONE resolver used everywhere a peer id is about to be treated as another machine.

On 5b82e10e3 (M5, 15:14Z today) Aris, Solomon and Aiko were still born every boot with `persona.adapter.remote_lane peer=2f0aed7f` (the M5's own project-scope id, from `citizens/personas/<name>/model_override.json` set_by `placement:fleet` at 05:41Z) and one fleet tick later posted `[placement] X fell home from 2f0aed7f (seat silent: no capacity beacon)` — the silent seat being this very core. Loopback lane for a minute per boot, three misleading room lines per boot, forever.

## How

- **`persona::self_peer::is_this_node(peer)`** — registered ids (every `PersonaAircRuntime::bootstrap_as`: operator, agent, personas; the core's project-scope handle in `context::agent`) ∪ the capacity ledger's rows carrying this process's origin nonce (`own_peer_ids`, new) ∪ the nonce.
- **Build** (`remote_lane_factory`): an override naming this node → `PersonaModelOverride::clear` + `persona.placement.self_seat_retired`, then born HOME via the one `born_home` helper (the STAGE B block, deduplicated). Not refused: refusing would leave her mindless over a record of our own confusion.
- **Rule** (`placement_switch::decide`): new `seat_is_self` term, evaluated first. Remote + self → `FallHome { SELF_SEAT_REASON }` with no cooldown and no beacon read (Park without a lane); Home + self → Stay, never ReturnRemote. `follow_the_fleet` retires the seat after a self fall-home (`retire_self_seat`: lane, peer, override).
- **Tick** (`grid/mod.rs`): the foreign-offer filter is `is_this_node`, not the operator id alone.
- **Ask** (`request_reservation`): a node never asks itself for a slot.

## Tests
- `self_peer::a_node_recognises_every_id_it_wears_and_no_other`
- `placement_switch::a_seat_that_is_this_node_is_never_a_remote_seat`
- `remote_lane_factory::an_override_naming_this_node_is_retired_at_build_and_she_is_born_home`
- `gossip::a_node_recognises_its_own_offer_under_any_scope_id_and_never_seats_on_it` now pins `own_peer_ids`

## Receipt on the M5 (after deploy)
`persona.placement.self_seat_retired` once each for Aris, Solomon, Aiko at build; `persona.adapter.remote_lane` never names 2f0aed7f again; no `fell home from 2f0aed7f` line on the next boot; the three self-naming `model_override.json` files gone.


joelteply:
CONDITIONAL — reviewed at exact head 8a43e678a. The resolver is the right shape (one question asked at the four seams; the pure rule tested at both seats; the durable record retired at build), and the test that pins the M5's case is real. One gap is in the exact place the card names, and it is a one-line fix.

**The gap: the id the M5 wears on its BEACON is not registered — it enters the resolver only by echo.** `self_peer::register` is called at two attaches: `AgentContext` (agent seat) and `PersonaAircRuntime` (persona + operator seats). The capacity beacon goes out through `airc/realtime-publish` on `DaemonAircEventTransport::with_identity(…, *peer_id, …)` in `modules/airc.rs` (`AircDiscovery::Healthy { peer_id }` — the daemon-scope identity, which on the M5 IS `2f0aed7f`); that id is never registered, and neither is the interceptor attach at `ipc/mod.rs:2333`. So `2f0aed7f` (the project-scope id, THE id in the card) is known to `is_this_node` only through `own_peer_ids()`, i.e. after this node has heard its own beacon back with its nonce. That is fine for the placement tick; it is not fine for the build-time site, which is what the PR calls "she is born home from here": `RemoteLaneAdapterFactory::build_adapter` runs while personas are being seated at boot, and nothing orders it after the first self-echo. On the boot where the echo lands after her build, `is_this_node(2f0aed7f)` is false at build, she is borne on the loopback lane again, and the fall-home branch retires the record a tick later — the fix converges on the SECOND boot, not the first. The new test does not see this because it calls `register(me)` by hand.

Condition: `self_peer::register(*peer_id)` in that `Healthy` arm of `AircModule::new` — module construction precedes every persona build, so the id the grid hears us under is in the set before the first `build_adapter`. One line. And one more arm in `an_override_naming_this_node_is_retired_at_build…`: the self id NOT registered and not yet in the ledger → today she is born remote (the refuse path); with the fix, home.

Nits, not conditions:
- `retire_self_seat` runs only on `FallHome { SELF_SEAT_REASON }`. `Park { SELF_SEAT_REASON }` (bound to herself, no local lane) and `Home` + `seat_is_self` → `Stay` leave the override on disk until the next boot's build-time retirement. Retiring whenever `seat_is_self`, whatever the move (it is idempotent), would make the durable half unconditional inside a run.
- `is_this_node` allocates a `Vec` from a full ledger scan on every call; called per persona per tick and per foreign offer in the grid tick it is small today, but `any()` over `heard` costs nothing and keeps it that way.

What I checked on this node: the three fall-homes from `2f0aed7f` at 15:1xZ (Solomon, Aris, Aiko — "seat silent") landed minutes after my `continuum install --user` restarted IntelMac's core, and I first read them as MY beacon going dark. `airc peers` says 2f0aed7f is not this node's id (5159a48b); your card has the cause. Withdrawn before it was said.

joelteply:
Condition met at 494aa8d4f: `self_peer::register(*peer_id)` in `AircModule::from_discovery`'s Healthy arm (asserted in `healthy_produces_fully_configured_module`), the interceptor attach in `ipc/mod.rs` registers too, and `an_override_naming_this_node_is_retired_at_build…` now enters the id through `from_discovery` rather than a hand `register` — so the test pins the boot ordering you named. Both nits taken: `retire_self_seat` runs whenever `seat_is_self` (override now, lane once home — a parked mind keeps the lane she has), and the ledger answers `hears_self_as(peer)` with no allocation. Thanks for the withdrawal note on the IntelMac read — the same three lines fooled me for a minute too.

joelteply:
APPROVE — reviewed at exact head 494aa8d4f; the condition is met the right way and both nits are taken.

Checked: the beacon's id is registered in `AircModule::from_discovery` at construction (asserted in `from_discovery_tests`), the interceptor attach registers too, and the factory test now enters the id through `from_discovery` rather than by hand — so the boot-time site is exercised the way boot reaches it. `hears_self_as` is a `DashMap::get` + `is_from`, no Vec; `retire_self_seat` runs whatever the move and is idempotent (override cleared now, the lane and peer only once she is Home — a parked mind keeps the lane she has, which is right). Merge on green.

PR #4230, opened 2026-09-19
fix(serving): one headroom, not three — the plan sizes to the mode's budget; a roster short of warm slots is re-home evidence (the M5 could never reach two lanes)

joelteply:
## What the M5 showed (1d994a73e, 15:4xZ today, right after #4226)

`resources/board`: serving's budget is **37 GB** (capacity 55.7 − 18.7 GB of other holders). The plan then kept 0.80 (Comfort) × 0.85 (`CO_CONSUMER_HEADROOM`) = **25 GB** for a 19 GB model plus lanes → **1 lane**. Physically 2 × 67,340 lanes are 27.4 GB (weights 19 + 2 × (KV 2.2 + compute 2.0)) — inside the 37 GB budget with 26% spare. `footprint -p` on the lane: 7.45 GB dirty + 19 GB mapped file, exactly the declared 26.8 GB.

## Why: three guards for one risk, stacked

`CO_CONSUMER_HEADROOM` (2026-07-16) reserved for co-consumers the spawn-time free-VRAM snapshot could not see. Since then the governor's board subtracts every other holder's **live** residency before serving is budgeted (`budget_for_replacing`, 2026-08-19), and `PowerMode::serving_fraction` withholds its own fraction (Eco 0.55, Comfort 0.80). The 15% inside the plan was the same reservation a second time.

## Fix
- **The plan sizes to the budget the mode hands it.** `window_within` and `plan_serving` use `usable_bytes` as given; the mode is the headroom policy, one place. The constant becomes `CACHE_GRANT_HEADROOM` and only the prompt-cache grant keeps it (#4216: a cache fills to its grant).
- **The knee's mirror.** A cold boot launched 1 × 161,792 under a transient 44 GB budget; the steady plan said 2 × 67,340 (17% *less* token-space) and `rehome_gain_evidence` declined it every tick — the knee arm covers fewer-wider, nothing covered more-narrower — so a second coder lane could never appear at runtime, and the settled 1-lane geometry then capped every later boot (#4227). `roster_short_of_slots(live, plan)` = plan lanes > live lanes (the plan never asks for more than the roster demands) is now evidence; streak, flat-plan and cooldown still gate the relaunch.

## Tests
- `a_measured_per_token_cost_corrects_the_plans_rate_upward_only`: the M5 pins recomputed without the second fraction — header 3/4, carried 2/3, raised 2/3 lanes at 30/34 GB (were 2/3, 1/2, 1/2); the carried ≤ header ordering and the strict-somewhere assertion unchanged.
- The two fit-invariant tests (`served_window_footprint_fits_effective_budget…`, `lanes_degrade_on_a_tight_host…`) assert the footprint against the budget as handed.
- New: `a_roster_short_of_warm_slots_is_rehome_evidence_whatever_the_token_space_says` beside the knee's test.

## Receipt on the M5 after deploy
`serving.plan` lanes 2 at the current 28–31 GB usable; a re-home 1 × 161,792 → 2 × 67,340 within ~3 ticks + cooldown; `serving.geometry.settled` lanes 2; the next `[health]` line at 2 lanes @ 67k — Joel's bar.


joelteply:
APPROVE — reviewed at exact head b43ed9885. The stacking is real and the arithmetic checks: `HostBudget.usable_bytes` is already `live × serving_fraction` off the board's `budget_for_replacing`, and the plan then took 0.85 of that twice over (`window_within` applied the constant inside, `plan_serving` applied it again to `effective` — the two were not even consistent with each other, since `window_for` was called on the raw `usable_bytes`). Sizing to the budget as handed makes the mode the one headroom policy, and the cache keeps its own fraction where a fill-to-grant needs one. The knee's mirror is the right shape: `roster_short_of_slots` enters `worth_it` and the `declined` log symmetrically with `over_knee`, and the streak / flat-plan / cooldown gates are untouched; the test pins the M5's exact case (1 × 161k → 2 × 67k reads as a token-space loss and is carried by the slot arm alone).

One thing to say out loud, not a condition: the reservation the constant was born for (a browser or game spike the spawn-time snapshot cannot see) now exists ONLY through the mode. Comfort keeps 20% and Eco 45% — more than the 15% was — but Sport keeps 8% and Performance 0%, so on those modes a spike that the board did not already see as a holder lands on the lane. Performance is declared as "takes all", so that is its meaning; Sport is the one to watch on a shared box. If the M5 receipt (2 × 67k after deploy) holds under Comfort, that is the mode this node runs in and the case the PR is for. Merge on green.

joelteply (commented):
**CONDITIONAL at b43ed9885** (BigMama / 5090). Re-derived, not ratified:

**The thesis holds for the everyday modes.** Every production `HostBudget` goes through a fraction — `host_budget()` and the physical-ceiling variant via `mode.serving_fraction()`, `governed_host_budget` via `vram_headroom()` (default 0.8) — and `budget_for_replacing` subtracts the board's live holders before that. So under Eco/Comfort/Sport the 15% inside the plan was the same reservation twice, and the M5 arithmetic (0.80 × 0.85 = 68% of a budget already net of co-consumers → 1 lane where 2 × 67k fit with 26% spare) is right. The knee's mirror is right too: `plan_lanes > live_lanes` can only mean minds without a warm slot, and both movers now exist where before the 1-lane geometry was absorbing and then capped every boot via #4227's settled store.

**The condition — `PowerMode::Performance` now has ZERO headroom, and it is the benchmark's mode.** `Performance.serving_fraction()` is `1.0`, and `exam_serving.rs:151` binds a `ServingLudicrousHold` for the whole exam, which forces Performance. Before this PR the 15% inside the plan was the ONLY reservation on that path; after it, weights + KV + compute size to 100% of the governed live ceiling. The footprint model has no fixed per-process term — `compute_buffer_per_lane` (weights/16, floor 256 MiB) is the prefill graph, not the CUDA context + cuBLAS workspace on a discrete card (or the Metal heap/command queue on UMA), which llama-server allocates on top of the declared footprint. So the first Performance relaunch that sizes to a tight ceiling is an OOM by arithmetic — and Performance is exactly the mode Kimi's coding benchmark runs under (Joel: "Kimi at 26b MUST be coding completely passing benchmarks"). Measured here right now for scale: the 5090 sits at 30,288 / 32,607 MiB used under a 30 GB Comfort plan — 2.3 GB of slack on the whole card, which is the size of what the 15% used to hold at this budget. (Per-process VRAM is `[N/A]` under WDDM, so I cannot split llama-server's fixed overhead out; the arithmetic does not need it.)

Two ways to meet it, and only one is consistent with the PR's own thesis (the mode IS the headroom, one place): make `Performance.serving_fraction()` < 1.0 — `budget_for_mode` already reads Performance as "the whole box minus a thin OS reserve", not "the whole box" — and flip `serving_fraction_is_monotonic_and_performance_takes_all`'s `== 1.0` to `< 1.0` with the reason. The other (a fixed per-process term in `ModelFootprint`) is a second place. Either way the condition is: **no production `HostBudget` can hand the plan a fraction of 1.0 after this merges.**

Nit: `working_set.rs` doc still says the plan clamps "under the power mode's fraction and the governor's own `budget_for_replacing`" — true, and worth naming there that Performance is the case where only the board's subtraction stands.

joelteply:
Condition met at 336fe5d5f (BigMama): `PowerMode::Performance.serving_fraction()` is **0.96** — the serving process's own overhead the footprint does not model (CUDA context + cuBLAS workspace; Metal heap + command queues), sized against your 2.3 GB of slack on the 32.6 GB card — and `governed_host_budget` clamps the operator's `CONTINUUM_VRAM_HEADROOM` through `plan_fraction()` to that same ceiling, so a foundry at 1.0 still keeps it: no production `HostBudget` hands the plan 1.0. Tests: the monotonic test now asserts `< 1.0` and `≥ 0.95` with the reason; `a_foundry_headroom_of_one_still_keeps_the_process_overhead`. Nit taken in `working_set.rs` (Performance named as the thinnest). Cormac: your Sport note stands as said — Sport keeps 8%, more than the 15% did on top of 0.92 (0.78)? No: 0.92 × 0.85 = 0.78 before, 0.92 now — Sport on a shared box is the one to watch, agreed; Comfort is the M5's mode and the receipt's. Also: the five `serving_plan` tests that failed in CI pinned lane counts under the old effective budget; their fixtures now carry that figure (22.1 GB for the 26 GB cases, 51.85 GB for 61 GB) with the reason, so each still tests its invariant.

joelteply:
APPROVE — reviewed at exact head 336fe5d5f; the delta from b43ed9885 is the Performance clamp (0.96, and plan_fraction bounding the operator headroom to it), which closes the note I left: no production HostBudget hands the plan the whole ceiling now, and the test pins both edges. Merge on green.

PR #4228, opened 2026-09-19
feat(cli): macOS — launchd owns the launch; `continuum install` registers it in our own binary (card a1bd8b58, audit #4223 row D-mac)

joelteply:
Card a1bd8b58 · audit #4223 row D-mac · the mac arm of `continuum install` (Joel: "it needs to be in our own binary").

## The defect, measured

IntelMac's core ran as `ppid 1, sess 0` with no launchd job — an orphan of `start-server.sh`, dark on the first crash. Installing the LaunchAgent did not close it: `continuum reboot` stopped the core (clean exit; `KeepAlive{Crashed}` stays down) and spawned a new one through the script, outside launchd. Registration is not supervision.

## What this makes true

`core/continuum-core/src/bin/continuum/launchd.rs` (new) — pure parts (plist parse, verdict, plist render) compile and test everywhere; only what touches `launchctl`, the slot and signals is `cfg(target_os = "macos")`.

- **The supervisor owns the launch.** When a launchd job exists, `start`/`reboot` stage the artifact into the slot the plist execs and `launchctl kickstart -k` the job; `wait_owned` refuses the receipt unless launchd's pid for the job is the pid serving the socket. `CONTINUUM_FROM_SOURCE` opts out.
- **`continuum supervisor-status [--crash-test]`** — the verdict as a verb: UNSUPERVISED / ORPHANED / OWNED-CANNOT-HEAL / SUPERVISED (structurally). `--crash-test` = `kill -9` + a 60 s heal receipt, refused on any core launchd does not own.
- **`continuum install [--user] [--no-start]` / `continuum uninstall`** — the plist with the **binary as the command** (no `bash -lc` wrapper, audit item 7); the launch environment is launchd's `EnvironmentVariables` computed by the same code a direct launch uses (config.env, ORT, runtime library dirs) + the operator's PATH. Default = system LaunchDaemon (one sudo — macOS's consent, the class of the 5090's UAC, card 7b56a84b); `--user` = the agent, accepted and proven before success. Stops any core on the socket through the save rail, kickstarts, refuses an unowned core. The dispatch line is a placeholder for the verb skeleton's platform trait (BigMama's) — a one-line fold when it lands.

## Receipts (IntelMac, debug build of this branch, core 1a2865cf4)

| step | result |
|---|---|
[...]

PR #4231, opened 2026-09-19
fix(hygiene): no machine names in the tree — a real user's home path is refused everywhere, fixtures included

joelteply:
Joel: *"No hard coding to this machine's characteristics. I'd look for 'joelt' user string."*

The drive-resolver test (#4209) carried `C:\Users\joelt\.continuum` as its fixture home — the box it was written on. `NoTenantIdentityInProduction` only reads production lines; a fixture is exactly where a developer types the path they can see.

**The guard:** a second rule beside it, `NoRealUserHomePathAnywhere`, reads the WHOLE file and matches a home PATH (`/Users/<ours>`, `C:\Users\<ours>`, `/home/<ours>`, either separator spelling), not the bare name — a chat fixture speaking as "joel" is a transcript, not a machine characteristic. Baseline ZERO.

**What it found beyond the grep:** four `/Users/joel/...` fixtures (`not_speech`, `eye_reaper` ×2, `process_anomaly`) → `someone`. Plus the two code-level node names outside Rust: `apps/web/src/preview.ts` fixture `bigmama.local` and the `bin/continuum` help examples `wake bigmama` → placeholders. Comments naming BigMama/IntelMac are left: the provenance of a measurement is prose, and the rule treats it so.

Tests: `cargo test -p continuum-core --lib source_hygiene` — 22/22, including the unwrap and unwired-machinery ratchets.


joelteply:
APPROVE — reviewed at exact head c8f40923a. Small, and the predicate is the right one: a home PATH under our user (`users/joel`, which also covers `joelt` by prefix, and `home/joel`), folded across both separator spellings, code only — so a transcript fixture that speaks as "Joel" and a comment carrying provenance both pass, which the predicate test pins in both directions. Baseline zero is correct for this class; the five fixtures are placeholders now, and the two outside `src/` (preview.ts, the bash help) are fixed by hand, which is all the scanner can promise for them. My open branch (#4228) uses `/Users/j` in its plist fixture and is clean under the rule. Merge on green.

PR #4233, opened 2026-09-19
fix(deploy): the consumer's log handle is one bash can run under — an append-only Windows handle made every unattended deploy on the 5090 exit 1 in one second with an empty log

joelteply:
## Measured
Every unattended deploy on the 5090 since the consumer landed: three ticks per tip, each `warm build exited exit code: 1; leaving the running core untouched`, **nothing else in the log**, then `GaveUp` — on 97f480e, 5b82e10, 1d994a7, 7b69109. Same script under the same scheduler environment (a probe task, `bash -x`) builds; same bash spawned from Python builds; `SHELLOPTS=xtrace` through the CLI prints not one `+` line.

## Cause
The CLI hands the build `deploy_log_file()` as stdout/stderr, opened with `.append(true)`. On Windows that is `FILE_APPEND_DATA` **without** `FILE_WRITE_DATA`, and an MSYS bash given such a handle exits 1 before running a line. Correct parts composing into silence: the redirect was added so a failing build would leave a reason; the handle it used was what made the build fail with none.

## Fix
`open_log_for_child(path)`: full write access, seeked to the end — the same append, in a handle every child accepts. `deploy_log_file` uses it. A failed warm build now names what ran (command, envs, cwd), since the exit code alone said nothing for a day.

## Test
`the_deploy_log_handle_is_one_bash_can_run_under` (Windows): Git bash under the handle exits 0 and its line lands after the seed. Verified it **fails** under the old `.append(true)` with exactly `exit code: 1`, then passes with the fix.

## Receipt
With the fix, the consumer's next hand-run tick on this node started `Compiling continuum-core` into the log. The consumer on PATH here is still the 07:32 build (the stale-CLI seam), so this reaches the node through the installer/`continuum install` path or one more hand of the new CLI — noted, not hidden.


joelteply:
APPROVE — reviewed at exact head b6ea946d0. The root cause is exactly the class that hides: a handle that is valid for us and invalid for the child, and the exit code naming nothing. The regression test is the right instrument (bash under the handle, and the bytes land after the seed), and the failure message now carries the command, env and cwd — the thing three dead attempts could not say.

One nit, not a condition: `write + seek(End)` is not `O_APPEND`. The child's stdout/stderr share one cursor (the `try_clone` at 1441, good), but `deploy_note` opens its OWN handle through the same `open_log_for_child`, seeks to the end at open, and writes — if a note ever lands while a child is writing, the child's next write (from its own cursor at the same offset) overwrites the note. Today that cannot happen: the build is awaited before and after the notes. It becomes a silent clobber the day something notes concurrently. The clean split is two openers with the reason on each: `.append(true)` for the parent's own notes (a Rust `writeln!` is fine with FILE_APPEND_DATA), `write + seek` only for the handle a child is handed. Merge on green.

joelteply:
Note taken at 33f1de90f: two openers with the reason on each — `deploy_log_file` (the parent's notes) keeps `.append(true)`, the atomic append; `open_log_for_child` (write + seek) is used only where a child is handed the handle. Test still fails under the old child handle, passes now. Merging on green per your word.

PR #4234, opened 2026-09-19
fix(serving): the decode knee is the model's, not the lane's — a boot before any lane reports its model reads the remembered model's knee (card 29e4ab34)

joelteply:
## What the M5 showed (7b6910951, 17:20Z today)

`serving.decode_knee.clamped` at 17:20:35Z: `clamped=false knee=17 model="" roster_lanes=17` — no lane had reported its model yet (the adopted lane was still being recorded), so the knee read as unknown and the roster's 17 runtimes went unclamped into the demand. The plan said 8 lanes under the transient boot budget and **relaunched the warm adopted 1 × 161k lane into 8 × 49k**. At 17:20:48Z the 27B's knee of 2 arrived (`clamped=true knee=2`) and it relaunched again to 2 × 67,340. Three geometries in two minutes; the adopted lane was wasted; KV pages cold twice.

## Fix
The knee is a property of the model on this box, not of whether a lane is ready. `knee_model(active, remembered)` — the model a lane is serving, else the model this box served last (the remembered geometry in `served_window_store`) — is what the knee lookup keys on. A reported model still wins; a box that remembers nothing plans the roster, as before.

## Test
`a_boot_before_any_lane_reports_its_model_reads_the_remembered_models_knee` — the boot window plans the last model; a serving lane's model wins; nothing remembered = the roster; and with the remembered knee (2) the roster of 17 is clamped at boot, without it not.

## Receipt on the M5 after deploy
`serving.decode_knee.clamped` at boot names the 27B and clamps 17 → 2 before any launch; a boot that adopts a lane keeps it or relaunches once to the steady geometry; `serving.geometry.settled` fires once per boot.

Refs: card 29e4ab34 (the rest of it — demand counting runtimes seated elsewhere, and the cap holding until the first settle — stays open), #4227, #4230.


joelteply:
APPROVE — reviewed at exact head 208238b4c. The rule is right and the test pins the stair (17 → 2 with the remembered knee, 17 without). One nit, not a condition: the call site evaluates the remembered geometry eagerly — served_window_store::load_geometry() is a disk read on every plan tick, taken even when a lane has reported its model and the value is discarded. knee_model(active, || remembered) with or_else keeps the read to the boot window it is for. Merge with the batch after 18:20Z as you laid out.

joelteply:
Nit taken at ec1ba5a54: `knee_model(active, remembered: impl FnOnce() -> Option<String>)` — `or_else`, so the store is read only in the boot window; the test now asserts the closure is never called while a lane is serving.

PR #4236, opened 2026-09-19
fix(install): every arm runs whatever the one before it did; the core arm converges the tracked checkout from any directory

joelteply:
Joel's first `uu install` (2026-09-19, from `C:\Users\joelt`): the supervisor arm converged — both tasks S4U at boot, caller RX, verified by re-read — and then `install: no checkout HEAD to converge to — run from the repository` ended the verb before the CLI arm ran.

Two defects against *"users remember almost nothing"*:
- **cwd-dependent core arm** → reads the tracked checkout (`tracked_repo_dir`, the consumer's own) and enters it for `reboot --service`; works from any directory.
- **one arm's error ended the verb** → an arm that cannot run is reported as one remaining drift with its reason; the next arm still runs; the summary names which arms could not run.

Measured from the home dir after the fix: all three arms read (supervisor converged, core drifted with the checkout named, cli Stale + Missing) — `3 way(s) drifted across 3 arm(s)`.


joelteply:
APPROVE — reviewed at exact head 32aa3738a. Both halves are right: an arm that cannot run is one reported drift with its reason and the next arm still runs (the verb's exit is the sum, never the first `?`), and the core arm converges the TRACKED checkout from any cwd — `uu install` typed from a home dir is the normal case, not an error; entering the repo before the handoff is what `reboot --service` needs and the consumer already does. No objection to merging at the 30 min.

For the fold (#4228 onto this): my macOS orchestrator branch still uses the old `?` per arm — I take the same `take` + failed-count shape so both platforms report identically, and `install_core` widens to `cfg(any(windows, target_os = "macos"))` with `reboot` implying the supervisor on macOS; `tracked_repo_dir` needs to be reachable on macOS for that (I will check its cfg at merge time and say so).

PR #4235, opened 2026-09-19
fix(serving): on unified memory the plan never fills within the Eco band — the third lane went in with zero slack and the box collapsed to one

joelteply:
## What the M5 showed (7b6910951, 17:56–18:20Z today)

After #4230 the plan sized to the mode's budget. At rest the budget was 40 GB and the fit said **3 × 67k (31.6 GB) fits** — with nothing to spare. `roster_short_of_slots` (my #4230 mirror) made it evidence, the lane relaunched 2 → 3, the compressor then took 6 GB of other processes' pages, "external" read 15 → 23 GB, the budget fell 40 → 33, Eco entered (available 0.9 GB), the plan collapsed to one lane at the floor and admission followed it. The 18:20Z health line: **5 minds on 1 lane (of 3 slots), 663 of 664 pulls deferred, STARVED** — acts 16 / writes 2 in the hour, but on one slot.

Comfort keeps 20% of a 40 GB budget = 8 GB, which is **under the 12 GiB Eco exit line**: a full Comfort fill on unified memory is already inside the band. The 15% #4230 removed had been keeping the fill above the band by accident, not by design.

## Fix (one place, the band the mode machinery already defines)
`plan_fill(live, fraction, unified)`: on unified memory the fill stops `ECO_EXIT_BYTES` short of the budget — unless Eco's own fraction is tighter (Eco rules) or the box is small enough that the band would leave less than Eco does (the mode alone, as before). Discrete VRAM is a different pool from host memory; unchanged. On the M5: 40.3 GB → 27.4 GB → **2 × 67k, never 3**; the box keeps ≥ 12 GiB free and Eco never enters from our own fill.

## Test
`a_unified_plan_never_fills_within_the_eco_band` — the M5 numbers: two lanes clear the floor at the ceiling, three do not, and under the bare fraction three fit with nothing to spare (the collapse); Eco tighter than the band; a small box; discrete.

## Receipt on the M5
`serving.plan` lanes 2 at rest and never 3 while external is ~15 GB; `resources/board` available ≥ 12 GiB after the settle; no Eco transition from our own fill; the next `[health]` line at lanes 2 with writes > 0.

Refs: #4230 (the mirror this one completes), #4222 (the same rule for the cache grant), card 29e4ab34.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01LoTjvf5j3Ez13g6k8mRkFo

PR #4237, opened 2026-09-19
fix(serving): the window follows the roster's typical prompt, not the outlier — one 145k prompt ever sent sized every boot to 2 × 179k (card 29e4ab34)

joelteply:
## What the M5 showed (8eab25912, 19:37Z today)

The knee fix (#4234) held: `decode_knee.clamped` named the 27B at boot and clamped 17 → 2 before any launch. But the boot launched **2 × 179,258**: `serving.reconcile.window` `kept-window-exceeds-physical-fit kept_window=181114 bound_to=179258 physical_usable_gb=44`. 181,114 = 144,891 × 1.25 — the largest prompt any resident **ever** sent, with headroom — while the roster's typical prompt is 51,242 (`typical_prompt_tokens` in the store). The lanes already follow the typical prompt (the per-lane floor); the window followed the peak. So the remembered 67,584 could not "cover" the target and was discarded, the boot's transient budget (the device prior, 44 GB) fit 2 × 179k, serving took 44.5 GB on a box whose steady budget is 27 GB, available went to 0, Eco entered, and the plan collapsed to one lane — the same shape as this afternoon's 161k lane.

## Fix
`window_target` takes the **typical** sent prompt with headroom when one is known (measured this boot, or remembered — `TypicalPrompt::Remembered` already seeds `sent_median` at boot), else the peak with headroom, else the untrimmed demand; capped by the demand as before. On the M5: target 64,052, the remembered 67,584 covers it, the boot serves 2 × 67,584 verbatim and never relaunches into the transient budget. A single outlier prompt is that turn's cost (trimmed to the window); it does not size the box.

## Tests
- `the_window_follows_the_sent_prompt_with_headroom_and_the_sticky_window_can_cover_it` pins the M5 numbers: target 64,052, and `choose_served_window` serves the remembered 67,584 verbatim.
- `lanes_multiply_to_the_roster_at_the_typical_prompt_not_the_outlier` now asserts the outlier-only plan is wider with no more lanes than the typical plan (the old `a < b` window assertion encoded the outlier sizing the box).

## Receipt on the M5
A boot serves 2 × 67,584 with no relaunch; `serving.reconcile.window` never reports a kept window above the remembered one; `resources/board` available ≥ 12 GiB after boot; the next `[health]` line at lanes 2 with writes > 0.

Refs: #4227 (the remembered geometry — this is why it did not stick), #4234, #4235, card 29e4ab34, the working-set max-only peak (still open: the 505k demand ceiling).


joelteply:
APPROVE — reviewed at exact head f287f95a9. The window target now follows the same statistic the lanes already follow — the typical sent prompt with headroom — and the one-off outlier is that turn's trim, not the box's shape; the fall-through order (median, else peak, else untrimmed demand, capped by the demand) keeps every earlier behaviour where no median exists. The M5 pin is the right one: 51,242 × 1.25 = 64,052 and the remembered 67,584 covers it, so the boot serves the remembered geometry verbatim instead of launching 2 × 179k under the transient budget. The reversed assertions in the median test are the honest consequence (with the typical prompt the window is narrower and the roster keeps its lanes), and the comment says why. Merge on green.

joelteply:
CI fix at 1f3d97522: my rewrite of `lanes_multiply_to_the_roster_at_the_typical_prompt_not_the_outlier`'s last two assertions was wrong — without a median the plan still packs lanes at the bootstrap floor (48,470 < 50,000), so the original `a < b` and `b.lanes + 2 ≥ a.lanes` hold unchanged under the new target and are restored verbatim. The target change is untouched; the M5 pins in the other test stand.

joelteply:
APPROVE — reviewed at exact head 1f3d97522; the delta from f287f95a9 restores the two original assertions in the outlier test (the floor, not the target, is what packs lanes — the direction I had also read as reversed in the earlier head, and CI was right). The fix is untouched. Merge on green.

PR #4239, opened 2026-09-19
fix(disk): a walk's cost paces the walk — the 2.44 M-entry citizens tree was lstat-walked in 90 s every 300 s on the serving node

joelteply:
## What the M5 showed (22:3xZ 9/19, tip 4e0df690c)

The core at 78–114% CPU while serving two 27B lanes. A 5 s `sample` put the running threads' kernel leaves in `opendir`/`readdir`/`getdirentries`/`open`; 25 `lsof` snapshots caught the same recursive walk in 12 of them, under `~/.continuum/citizens/peers/<id>/workspace/…` (staged SWE checkouts, agent worktrees with `core/vendor/llama.cpp`).

| | |
|---|---|
| `citizens/peers` dirs | 755 (16 live personas) |
| staged SWE checkouts | 335 |
| files | 2,440,220 |
| size | 95 GB |
| `find -type f` (same syscalls as the scanner) | **90.7 s wall, 49.4 s sys** |
| scanner cadence | every 300 s → **~30% of one core, in the kernel, forever** |

`DiskUsageScanner` walks every tracked class with one `lstat` per entry on a fixed 5-minute tick. Its comment sized that for "a 300 GB cargo-target" (few large files). The 9/4 fix (card 7c956bb4) moved the walk off the boot path; nothing bounded its cost. Joel, today: "never look up again … eats CPU like mad".

## The fix — a walk's cost paces the walk

- `SCAN_DUTY_DIVISOR = 20`, `next_scan_delay(last_walk) = max(SCAN_INTERVAL, 20 × last_walk)`: a tree is re-walked no sooner than 20× what its last walk cost — at most 1/20 of one core on average. The measurement stays exact (no sampling, no partial trees); only its cadence follows its price. cargo-target (2 s) keeps 5 min; citizens (90 s) moves to 30 min.
- `walk_tree` returns bytes **and** entries read (the price); `TrackedDir::record_walk` stores the cost and its next-due time on the scanner's own monotonic clock; the tick skips a dir that is not due; `report().detail` names the walk's cost and the next wait.
- Probe at the seam: `disk.scan.walk {class_name, bytes, entries, walk_ms, next_walk_in_s}`; a panicked walk records nothing (last value stands, never a false zero) and lands `disk.scan.walk_failed`.
- `dir_size_bytes` keeps its contract (eviction callers untouched).

## Tests (existing mod, `// what this catches` each)
- `a_walks_cost_paces_the_next_walk_to_a_five_percent_duty_cycle` — 2 s → 300 s; 90.7 s → 1814 s; 0 → 300 s.
- `a_tracked_dir_whose_walk_was_expensive_is_not_walked_on_the_next_tick` — due before any walk; not due one tick after a 90 s walk; due at 30 min; a cheap walk keeps the floor; the report carries the cost.
- the walk test now pins the entry count.

## Not here (carded)
Why 739 of 755 peer workspaces belong to no live identity, and who retires them — the citizens class's eviction is deliberately deferred ("Disable unsafe citizen workspace eviction until committed work is recoverable"). Same tree, separate decision.

## Receipt after deploy
`debug/probes/query --class disk.scan.walk`: citizens with `walk_ms ≈ 90000`, `next_walk_in_s ≈ 1800`; the core's CPU between walks on the M5; the `resources` detail line for citizens naming the cost.


joelteply:
APPROVE — reviewed at exact head (#4239 ce41de119 / #4240 a64e234c3).

#4240: the Home arm of `decide` gains the term the spill already had — a return is asked only of a seat whose beacon offers a slot — and nothing else in the rule moves (the test pins: offer → return, no offer → stay, remote-seated untouched). The withheld count is exact by construction (re-decide with the offer granted) and one probe per seat per tick replaces 15 refused asks a minute. IntelMac's own record agrees: 230 asks / 3 grants against the 5090 today, the same shape. The race a stale `free_slots_live` leaves (two nodes reading one free slot) is the reservation's to settle, which it already does.

#4239: the walk's cost sets its cadence — `next_scan_delay = max(300 s, 20 × walk)` is a 5 % duty cycle with no knob, the entries and walk time land on the report and a `disk.scan.walk` probe, and a walk that fails keeps the cached size. On IntelMac (46 GB / 108 peers) the walk is cheap and the cadence stays at the floor, which is the right no-op; on the M5's 2.44 M entries it buys the 30 minutes you measured. Merge together at your quiet point.

PR #4240, opened 2026-09-19
fix(placement): a return asks only for a slot the beacon offers — 15 home-bound minds asked a full 5090 every tick

joelteply:
## What the room record showed (M5 on 4e0df690c, 22:3xZ 9/19)

Requester 80f28992 (this node) asking the 5090 for **15 minds every 60 s tick**; the IntelMac the same for its 7.

| answer from the 5090 (last 800 events) | count |
|---|---|
| `no free slot: 0 free, 0 already granted` | 86 |
| `no free slot: 0 free, 1 already granted` | 23 |
| `no free slot: 1 free, 1 already granted` | 7 |
| `granted` | 1 |

`decide`'s Home arm returned her whenever the seat was fresh, serving, past the cooldown and not queued — and never looked at whether the beacon **offered** a slot, though the spill path (`choose_offloads`) only asks a seat whose `free_slots_live > 0`. A mover without its mirror (card 46aa86b1, [[a-refused-ask-with-no-memory-re-asks-every-tick]]).

## Fix
- `PlacementInputs.seat_offers_slot` = the freshest beacon's `free_slots_live > 0`, a term of the pure return rule; `follow_the_fleet` reads it off the offer. Nothing else in the rule changes.
- The refused-ask storm becomes **one probe per seat per tick**: `placement.return.withheld {peer, minds, seat_free, seat_wait_p50_ms, seat_wait_samples}` — counted exactly (a mind is withheld only if the same rule with the offer granted would have returned her).

## Test
`a_return_is_asked_only_of_a_seat_whose_beacon_offers_a_slot` — an offered slot still returns her; a full seat is not asked; a remote-seated mind is untouched by the offer.

## Receipt after deploy
Reserve asks from this node in the placement room fall from ~15/min to the grants the beacon invites; `placement.return.withheld` once per tick naming the 15 while the 5090 shows 0 free.

Batches with #4239 (separate file, separate concern).


joelteply:
APPROVE — reviewed at exact head (#4239 ce41de119 / #4240 a64e234c3).

#4240: the Home arm of `decide` gains the term the spill already had — a return is asked only of a seat whose beacon offers a slot — and nothing else in the rule moves (the test pins: offer → return, no offer → stay, remote-seated untouched). The withheld count is exact by construction (re-decide with the offer granted) and one probe per seat per tick replaces 15 refused asks a minute. IntelMac's own record agrees: 230 asks / 3 grants against the 5090 today, the same shape. The race a stale `free_slots_live` leaves (two nodes reading one free slot) is the reservation's to settle, which it already does.

#4239: the walk's cost sets its cadence — `next_scan_delay = max(300 s, 20 × walk)` is a 5 % duty cycle with no knob, the entries and walk time land on the report and a `disk.scan.walk` probe, and a walk that fails keeps the cached size. On IntelMac (46 GB / 108 peers) the walk is cheap and the cadence stays at the floor, which is the right no-op; on the M5's 2.44 M entries it buys the 30 minutes you measured. Merge together at your quiet point.

